PR #1, opened 2019-10-04
Research object support

johanneskoester:
## Plan
* [ ] depend on https://github.com/ResearchObject/ro-python
* [ ] similar to snakemake --report, create research objects on request, from the info that is already captured via the persistence module under .snakemake


albangaignard:
Hi, I would be happy to contribute and provide source code on that topic (in particular provenance / PROV-O).

johanneskoester:
@albangaignard your help is greatly appreciated. I have added a skeleton. Basically, one just needs to feed the information that is retrieved in the skeleton with the python-ro API.

mdehollander:
This would be a nice addition to snakemake. How does this compare to [cwlprov](https://github.com/common-workflow-language/cwlprov) and [dataprov](https://github.com/fbartusch/dataprov/tree/development)?

albangaignard:
> This would be a nice addition to snakemake. How does this compare to [cwlprov](https://github.com/common-workflow-language/cwlprov) and [dataprov](https://github.com/fbartusch/dataprov/tree/development)?

Thanks very much for your feedback. This would be completely in line with the PROV profile of CWLprov (https://github.com/common-workflow-language/cwlprov/blob/main/prov.md). Regarding dataprov the approach is interesting but apparently it does leverage a standard for representing provenance metadata as proposed by the W3C (https://www.w3.org/TR/prov-primer/).

albangaignard:
@johanneskoester, I recently reviewed the code quality based on the automatic checks (SonarCloud), and code formatting best practicies (Black tool). Would you have time to review this pull request ? 

In summary : 
- `--provenance` option
- provenance capture in the AbstractExecutor, 
- two provenance serializations (RDF and JSON) in the working directory

snakemake/cwl.py
```diff
@@ -99,13 +99,13 @@ def job_to_cwl(job, dag, outputs, inputs):
...
     if job.conda_env_file:
-        files.add(os.path.relpath(job.conda_env_file))
+        files.append(os.path.relpath(job.conda_env_file))
 
     out = [get_output_id(job, i) for i, _ in enumerate(job.output)]
 
     def workdir_entry(i, f):
         location = "??inputs.input_files[{}].location??".format(i)
-        if f.is_directory:
+        if os.path.isdir(f):
```
johanneskoester, line 108: This I don't understand. We use the is_directory property of IOFile here, because the files may not yet be present. Then, isdir would not work.

snakemake/executors/__init__.py
```diff
@@ -33,6 +33,7 @@
 from snakemake.logging import logger
 from snakemake.stats import Stats
 from snakemake.utils import format, Unformattable, makedirs
+from snakemake.provenance_tracking.provenance import provenance_manager
```
johanneskoester, line 36: Since the provenance manager only has to be part of AbstractExecutor, I think we could avoid the singleton and just keep it in there instead.

snakemake/executors/__init__.py
```diff
@@ -178,7 +184,28 @@ def print_job_error(self, job, msg=None, **kwargs):
         job.log_error(msg, **kwargs)
 
     def handle_job_success(self, job):
-        pass
+        if provenance_manager.is_active():
+            prov_mgr = provenance_manager
+            # print(job.params['biotools_id'])
+            tool_name = ""
+            if "biotools_id" in job.params.keys():
+                tool_name = job.params["biotools_id"]
```
johanneskoester, line 192: Mhm, is it possible to extract the biotools_ids from the conda packages instead?

snakemake/executors/__init__.py
```diff
@@ -178,7 +184,28 @@ def print_job_error(self, job, msg=None, **kwargs):
...
+                cmd=job.shellcmd,
+            )
+            outputs = job.output
+            for o in outputs:
+                prov_mgr.add_output(
+                    output_id=o,
+                    input_id_list=job.input,
+                    tool_name=tool_name,
+                    job_uri=job.uri,
+                )
```
johanneskoester, line 208: As you know, Snakemake has its own metadata tracking. I wonder if it would make sense to export the research object from there instead of doing this while running in here. The advantage is that you will get information also for stuff that happened in a previous run. The flag would then rather be a post-hoc command, just like --report, instead of requiring the user to remember to add it while running. Also, multiple partial runs of the same workflow do not result in several separate provenance information files.

johanneskoester:
I am so sorry for the late response. This completely slipped my attention (I get so many Github notifications that I sometimes miss one). Nice work, please see my comments above.

cademirch:
going to close this since its quite old and main has diverged so far.

PR #17, opened 2019-10-10
Allow for passing 0 cores to mean all available

### LINKED ISSUE #11: Setting jobs in profiles to use all available cores
(Manually migrated from the original Bitbucket issue #1305)

I am testing out Snakemake profiles and would like to specify the number of jobs to be the same as the number of available cores. When using command-line arguments, I would be able to do this:

```snakemake -j```

and Snakemake will determine the right number of cores to allocate. If I instead of a `profile/config.yaml` file that looks like

```jobs:```

I get this error:

```snakemake: error: argument --cores/--jobs/-j: invalid int value: 'None'```

Is there a way to get the behavior of simply passing `-j` while using profiles?

mbhall88:
Addresses / closes #11 

When trying to pass `None` in a snakemake profile, `None` cannot be parsed as an `int`. This PR allows `0` to be passed to `--cores/--jobs/-j` and it will use all available cores. This way cores can be set to 0 in a snakemake profile.

Apologies if I have not setup the tests in the correct way. Obviously very happy to change things to fit into the project structure better.

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -791,14 +791,14 @@ def get_argument_parser(profile=None):
         "--cores",
         "--jobs",
         "-j",
-        action="store",
+        default=1,
         const=available_cpu_count(),
         nargs="?",
         metavar="N",
         type=int,
         help=(
             "Use at most N cores in parallel (default: 1). "
-            "If N is omitted, the limit is set to the number of "
+            "If N is 0 or omitted, the limit is set to the number of "
             "available cores."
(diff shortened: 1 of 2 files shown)


johanneskoester:
Thanks for looking into this! I have used your PR to come up with a slightly different solution in #33. Closing in favor of #33.

mbhall88:
No problems. However, I see you didn't bring across the tests to the CLI parser?

PR #18, opened 2019-10-10
Use follow_symlinks argument only on supported platforms

melund:
This PR fixes a bug where `follow_symlinks` arguments was used on non-supported platforms. 

The bug was probably a misinterpretation of how the system worked. It is not even allowed to set `follow_symlinks=False` on systems that doesn't support following symlinks. Doing so raises an `NotImplemented` exception.

Side note: 👍 🎉  on moving to github 😃


johanneskoester:
Thanks for pointing me to the issue! I have solved this in a bit different way now, see PR #32

PR #33, opened 2019-10-14
Allow to define cores as 'all'.

### LINKED ISSUE #11: Setting jobs in profiles to use all available cores

johanneskoester:
Replaces #17 and closes #11.

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -795,10 +795,9 @@ def get_argument_parser(profile=None):
         const=available_cpu_count(),
         nargs="?",
         metavar="N",
-        type=int,
         help=(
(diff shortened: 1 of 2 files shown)


mivade:
Thanks!

Issue #24, opened 2019-10-11
Subworkflow without a configfile gives error


SultanOrazbayev:
**Snakemake version**
5.7.0

**Describe the bug**
If a subworkflow does not have a `configfile` entry, an error is shown.

**Logs**

```
Traceback (most recent call last):
  File "/mypath/miniconda/envs/mydef/lib/python3.7/site-packages/snakemake/__init__.py", line 611, in snakemake
    export_cwl=export_cwl,
  File "/mypath/miniconda/envs/mydef/lib/python3.7/site-packages/snakemake/workflow.py", line 558, in execute
    updated_files=updated,
  File "/mypath/miniconda/envs/mydef/lib/python3.7/site-packages/snakemake/__init__.py", line 389, in snakemake
    overwrite_config.update(load_configfile(f))
  File "/mypath/miniconda/envs/mydef/lib/python3.7/site-packages/snakemake/io.py", line 1303, in load_configfile
    config = _load_configfile(configpath)
  File "/mypath/miniconda/envs/mydef/lib/python3.7/site-packages/snakemake/io.py", line 1265, in _load_configfile
    with open(configpath) as f:
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

**Minimal example**

Running this code gives error (if `configfile` entry is copied to the subworkflow, then the error will disappear):
```python
configfile: "config.yml"

subworkflow test:
    snakefile: "sub.snakefile"

rule all:
    input: test('test.txt')
```

The `sub.snakefile` can be:
```python
rule all:
    output: touch('test.txt')
```

wclose:
Plus one to this bug. Just found it myself.

wclose:
@johanneskoester does this have to do with the brackets on line 561 of workflow.py? 
```
configfiles=[subworkflow.configfile],
```
Based on the blame history, it looks like it's something that was changed in PR #3

SultanOrazbayev:
@wclose: thanks! removing the brackets resolves the issue. I'm running the tests to see if that introduces problems anywhere else, and if not will submit PR.

VukanJ:
In 5.10.0 this is still an issue

johanneskoester:
Fixed with PR #271.

PR #42, opened 2019-10-17
This commit fixes the issue with running subworkflows without defined configfile

SultanOrazbayev:
This commit fixes the issue with running subworkflows without defined configfile configfile, the original fix was suggested by @wclose (see https://github.com/snakemake/snakemake/issues/24). A test for this case has been added to ensure future versions take this into account.


SultanOrazbayev:
Running of an existing test "test_subworkflows" fails. However, running the latest version of snakemake on that test also fails. Looking at the test, it seems that the root workflow should also contain a `configfile` (or the workflow should be redefined to remove the reference to config dictionary).

johanneskoester:
Actually, that test runs fine on the master branch.

snakemake/workflow.py
```diff
@@ -558,7 +558,7 @@ def files(items):
                         subworkflow.snakefile,
                         workdir=subworkflow.workdir,
                         targets=subworkflow_targets,
-                        configfiles=[subworkflow.configfile],
+                        configfiles=subworkflow.configfile,
```
johanneskoester: Configfiles has to be a list. 
```suggestion
                        configfiles=[subworkflow.configfile] if subworkflow.configfile else None,
```

johanneskoester:
May I ask you to run black formatting with `black snakemake; black tests/tests.py` on this branch?

SultanOrazbayev:
I ran the commands above, but I'm not sure if the result is correct. After running the commands, the following change was made:
```
-                        configfiles=[subworkflow.configfile] if subworkflow.configfile else None,
+                        configfiles=[subworkflow.configfile]
+                        if subworkflow.configfile
+                        else None,
```
The suggested change doesn't seem to work (or am I missing something?).

mwort:
Any chance this could be merged soon? This is quite an urgent fix for workflows relying on subworkflows.

johanneskoester:
Ok, working on it. The formatting change suggested by black seems fine to me.

johanneskoester:
See PR #271

PR #72, opened 2019-10-30
Extend documentation.

johanneskoester:


docs/executable.rst
```diff
@@ -42,10 +42,41 @@ By specifying the number of available cores, i.e.
...
+.. code-block:: console
+
+    $ snakemake --cores 4 --batch myrule=1/3
+
+you instruct to only compute the first of three batches of the inputs of the rule `myrule`.
+To generate the second batch, run
+
+.. code-block:: console
+
+    $ snakemake --cores 4 --batch myrule=1/3
```
melund: Guess it should be ``myrule=2/3``

johanneskoester: Indeed :-)

melund:
Looks good. Where would you usually apply the batch feature? One of the places where I have that it is annoying that snakemake can take a long time to build the DAG is when I need to run "snakemake --unlock".  But the batch features doesn't really help with that right?

johanneskoester:
Indeed, unlock is not what batch is designed for. It is meant for normal execution and dry-run.

PR #79, opened 2019-11-01
Enable requester pays google storage remote objects

jaicher:
Issue: Current snakemake.remote.GS.remote() does not (to my knowledge) allow specifying a billable user project for accessing files in buckets with requester-pays set, e.g. GTEx v8.

This pull request updates `snakemake.remote.GS.RemoteObject` to allow specifying named parameter user_project. This allows remote downloads of files from google storage with requester pays enabled. Remote input files are specified using `GS.remote(url, user_project=example_project)`, where `example_project` is a string specifying the project that will be charged for accessing the data.


johanneskoester:
@vsoch, since you are also working on the specification of a google project (globally, outside of the workflow) it might be a good idea to coordinate. I think a project name should not be part of the workflow definition itself, because others would need to alter the workflow definition to make it run with their google account. Rather, this should be taken from @vsoch's global option.

johanneskoester:
Another question @jaicher: doesn't this automatically happen when configuring via gcloud init?

vsoch:
Taking a look! In case you are also having formatting errors with black (different results here vs locally) I opened an issue here https://github.com/psf/black/issues/1140.

jaicher:
@johanneskoester as @vsoch mentions, this doesn't happen when configuring using `gcloud init` (or `gcloud auth application-default login` either [as described in the docs][gs-snakemake-docs]).

I agree that we should not be hard-coding in project names in Snakefiles. I saw the `example_project` variable being derived from an input configuration file. I agree that it is unlikely that someone would want to bill different projects in a single workflow -- I think that using a boolean flag as @vsoch proposes is an excellent alternative if we are fine with restricting billing to a single project. Really like the ideas of how to report when requester pays is being used.

@vsoch if the bucket requires requester pays it raises `google.api_core.exceptions.BadRequest` with the message `google.api_core.exceptions.BadRequest: 400 GET {... url ...}: Bucket is requester pays bucket but no user project provided.`

@vsoch With the CI formatting issue -- I'm fairly new to contributing to these projects. I was looking at the [docs on contributing][contributing-snakemake-docs] and only saw information about the nosetests. I was trying to enable CircleCI/run locally but I thought it needed a `.circleci/config.yaml` file? Would appreciate any help setting up CI appropriately for testing future contributions to snakemake. I was trying to keep flake8 happy with my changes.

As for my original use-case, I was looking at processing cloud-only data from GTEx using a Snakemake pipeline -- the data are stored in a requester-pays only bucket. It turns out that this is a far from ideal way for downloading remote objects (downloads appeared to be single-threaded/blocking on the pipeline without compiled crcmod), so I've moved to an alternative approach.

   [contributing-snakemake-docs]: https://snakemake.readthedocs.io/en/stable/project_info/contributing.html
   [gs-snakemake-docs]: https://snakemake.readthedocs.io/en/stable/snakefiles/remote_files.html#google-cloud-storage-gs

vsoch:
@jaicher the `google.api_core.exceptions.BadRequest` would be perfect to catch (and then raise) and the error message provided is spot on to alert the user what is going on.

For formatting - let me see if I can help out! The repository uses [black](https://github.com/psf/black) (which you can install with pip) and I would recommend that you use the same version as I see in tests:

```bash
pip install black==19.3b0
```
I was having issues with consistency between local and remote, so I wound up installing from the GitHub master branch, and although the first time it failed, the second time seemed to go okay. The way you can check locally is to mirror the CI and do:

```bash
conda create --quiet --name black black
source activate black
black --check snakemake tests/*.py
```
The command says to check the snakemake module, and files that end in .py in the tests folder. The CI itself isn't CircleCI, but rather GitHub CI / workflows, so you can find the entire recipe under .github/workflows/main.yml. You should be able to click the check links at the bottom of the PR to see the results.

For the testing, it would be nuts to try and run all of the tests that Snakemake has, but we still should have tests for this GS.remote setting that you can write. What I did is to:

 1. Find an already existing folder that uses GS.remote commands
 2. Create a new folder (named by issue associated likely, make sure you include the prefix `test_github_issue*` as the issues in question are from bitbucket originally (numbers don't match up).
 3. Modify the Snakefile to include your variable, and possibly write additional Snakefiles that test other cases (e.g., one might have the flag for remotes, and the other not). You'd likely also need to use a remote bucket that you know has / doesn't have it. 

You'll probably end up with something like this:

```python
@pytest.mark.skip(reason="also need to figure out how to provide Google Cloud Project")
def test_github_issueX():
    workdir = dpath("test_github_issueX")
    tmpdir = run(
        workdir,
        use_conda=True,
        Snakefile="Snakefile.1"
        .... #other args
    )

    tmpdir = run(
        workdir,
        use_conda=True,
        Snakefile="Snakefile.1"
        .... #other args
    )
```
And note that you don't have to inspect the tmpdir after, that's only there if cleanup=False and you want to look at it for some reason.

Then to actually test, what I did is to open up ipython from the root of the repo, and basically import everything from tests

```python
from tests.tests import *
```
And then you got it - I copy pasted the various tests. And what I would do to interact / debug is insert this snippet into relevant sections where I wanted the code to stop and interact (breakpoints)

```python
import code
funcname = "GS.remote"
code.interact(local=locals())
```
You can set different funcnames as variables to give yourself hints about where you are, in the case there is more than one. Then you can at least run all your tests locally with your credentials to make sure things are working, and note that the function is marked for pytest to skip so it wouldn't run on CI.

Ah interesting! The test cases I've been doing thus far are using fairly small data files, and I suspect that's not going to be ideal for many projects. What would make sense to think about a little further down is (given an executor on Google Cloud) how to possibly mount a read only bucket as a drive (only works for read only) but yeah, for local runs where you need to download to your machine that's not even possible :P 

Hope that helps!

tests/tests.py
```diff
@@ -673,6 +673,40 @@ def test_remote_gs():
     run(dpath("test_remote_gs"))
 
 
+@pytest.mark.skip(reason="Need to choose how to provide billable project")
+@connected
+@not_ci
+def test_pr79(
```
vsoch: @johanneskoester is test_pr<number> an ok convention to use? If there aren't other test_pr* folders for GitLab we are probably good, and can assume that any test_pr folders are for GitHub. @jaicher another option would be to open an issue for this particular thing (if one isn't already) and reference it directly. If there isn't an issue, hats off to you for identifying an issue and going straight to a fix for it! :tophat:

tests/tests.py
```diff
@@ -676,11 +676,11 @@ def test_remote_gs():
 @pytest.mark.skip(reason="Need to choose how to provide billable project")
 @connected
 @not_ci
-def test_pr79(
+def test_issue96(
```
vsoch: ah one tiny tweak! The previous test_issueX I believe are referencing BitBucket (the two issue trackers don't line up) so here we would want a test_github_issueX instead. @johanneskoester is this correct?

tests/tests.py
jaicher: @vsoch I just created issue #96 and pushed a commit renaming the tests to reference this issue as suggested.

vsoch: Yes, saw it! I added a small comment about the naming convention (for GitHub vs Bitbucket) but otherwise we should be good. :)

tests/tests.py
jaicher: haha I was thinking about asking what happens once the github issues catch up -- I'll go ahead an push an update with the stated naming convention.

johanneskoester:
@jaicher, please just call the test such that we know what is tested: `test_gs_requester_pays`.

jaicher:
@johanneskoester Okay, made update to the name of the test -- please let me know if there are any other changes you need.

PR #90, opened 2019-11-05
[WIP] Adding Cloud Pipelines

vsoch:
**do not merge** this is a draft pull request!

This PR will encompass early work to add Google Cloud Pipelines to Snakemake. Specifically this means that:

 - the user selects --google-life-sciences provider (boolean) when running snakemake
 - GOOGLE_APPLICATION_CREDENTIALS and GOOGLE_PROJECT should be exported
 - a content hash named tarball of the working directory is uploaded to Google Storage
 - the worker uses a custom image to download and extract the working directory, and then run the script for the step in question.
 - the storage tarballs are deleted after the workflow completes, only if the user hasn't specified a flag to keep the cache.

We have several technical questions that still need answering, which I'll summarize here:
 - upload of environment variables vs. secrets
 - another method to discover application credentials and project (without exporting the envar)
 - how to provide an entrypoint + command with && (we've tried [almost all variations](https://gist.github.com/vsoch/076f2131826e3e55e822e51b5d9110e1) of this and the second command seems to run outside of the container, or the entire thing is parsed together and the result is command not found.
 - privileged containers OR custom addition of permissions (to allow running Singularity)
 - difference between action and resource environment
 - it would be cool if Google provided a function to choose a machine type based on these criteria, seems like it would be a common thing to need!

A few notes for some choices I made for the implementation:
 - As discussed in our meeting, I was able to use the [Google Discovery Build](https://developers.google.com/discovery/v1/building-a-client-library) function against the manifest endpoint to create a client.
 - regions is another variable that could be derived from some call to the API (to get a default region) but most likely the requester for storage objects would want to choose a region that is closest to the objects, which might not be where he or she is. The default regions are all zones in the US, which I chose based on guessing that the majority of Snakemake / GCP users are in the US. This might be entirely wrong, or a bad default.
 - the machine type is selected based on comparing the requested memory and cores to those available, and first eliminating the ones that are too small, and then choosing the smallest of those remaining. If the user provides a machine type prefix, we further filter down the choices to that.
 - Google Storage (for public objects) often has a 503 error (yesterday it popped up several times for me in testing) and if this happens again it might be a separate issue that the GS remote should look into handling - for most Google Cloud API stuffs retrying is essential.
 - speaking of retrying - the functions here are run by a client function I wrote to retry three times, and only then error. This is because BrokenPipeError (and other random failures) are common enough to warrant it.

I suspect this will fail linting and I'll push again with fixes :)


vsoch:
@johanneskoester this is a little confusing with black, because I installed the exact version that is run in the test suites, and there seems to be a difference in result. Here is what I'm doing locally:
```python
$ black --check snakemake tests/*.py
All done! ✨ 🍰 ✨
51 files would be left unchanged.
$ black --version
black, version 19.3b0
```
Note that the version is the same installed there. I also tried installing with conda in case it made a difference (it didn't). What am I missing?

vsoch:
That's so weird! The black formatting passed this time. I don't think I made any changes to that file. Must be ghosties.

johanneskoester:
@vsoch did you select a draft PR? I still have a merge button here and the icon is green instead of grey in the list.

vsoch:
oops I just totally forgot :/ Merging is still blocked so is it okay? I'll add WIP to the heading so others know.

johanneskoester:
Yeah, no problem ;-).

johanneskoester:
xref #91

vsoch:
Updates to the PR:

### Environment variables cannot be secrets
Andrew confirmed that since the underlying service is Google Compute Engine, we can't take advantage of secrets as we can with kubernetes. We should generally keep on top of this, and perhaps later brainstorm if other way could be used to allow for secrets. I'll add this to a little TODO list that I'll share in the project space after.

### Pipeline vs Action environment
I asked about the difference between Action / overall pipeline environment, and it's the case that every pipeline environment is passed to each action, and then an action can have it's own environment. Since we are defining actions per step for snakemake, it makes most sense to just add the environment variables that are needed there. I don't see a compelling use case here for using a pipeline environment instead. The one interesting detail is that an action cannot delete a pipeline environment, so I suspect pipeline environments would be appropriate if there is some service running a user workflow (and the user workflow shouldn't be allowed to mess with some parent environment). It doesn't seem hugely important here, but we can keep it in mind if the need arises.

### Machine Types

Andrew pointed me to the compute engine API to get [machine types](https://cloud.google.com/compute/docs/reference/rest/v1/machineTypes/get) that we can use instead of hard coding a list. The machineType list endpoint requires a specific zone, and the Life Sciences API allows the user to request a _region_, which is one or more zones (and would need to be derived from the region).

So to make this work, it comes down to:

 1. taking the user requested regions, and then using the compute engine api (zones endpoint) to derive zones available for the requested regions
 2. deriving the machineTypes for those zones 
 3. taking the set that is shared between all of them (otherwise we risk an error of getting an instance type not provided in a zone.) 

I added a user debug statement that says "Hey if you want more selection, then you need to select fewer regions." I think this approach is ok, and better than having the user select a zone, because the pipelines API expects a region, and if we derive a region from a user zone (e.g., user selects us-east1-a and we derive region us-east1 for the API) then we could hit an issue if the API allocates the instance to us-east1-b with some instance type only supported in us-east1-a. Anyhoo, I've deleted the custom function to derive the machineType, and it's now totally programmatic based on what the compute engine API says is available.

### Singularity

We probably won't get Singularity working any time soon:

> Privileged Containers: I don’t believe those are exposed in the API, and not sure if there’s a plan to.

I'm waiting on answers about the CMD/entrypoint but I'll update again when I hear!

vsoch:
@johanneskoester a test failed for test_remote which I don't think is related to the PR here, is this a one off error because of GitHub actions (and I should ignore and it can be re-run) or some issue that I can address in the PR here?

snakemake/__init__.py
```diff
@@ -1623,6 +1660,47 @@ def get_argument_parser(profile=None):
         " between S3 bucket and the run environment (container) is automatically"
         " handled by Tibanna.",
     )
+    group_google_life_science.add_argument(
+        "--google-life-sciences",
```
johanneskoester: @moschetti is there an official short name? --google-lsa?

snakemake/executors.py
```diff
@@ -450,6 +452,11 @@ def handle_job_error(self, job):
 
 
 class ClusterExecutor(RealExecutor):
+    """the cluster executor will start with the exec_job (the start of a
+       snakemake commant intended to run on some cluster) and then based
+       on input arguments from the workflow (e.g., containers? conda?)
+       build up the exec_job to be ready to send to the cluster or remote.
+    """
```
johanneskoester:
```suggestion
    """Backend for distributed execution. 
    
        The key idea is that a job is converted into a script that invokes Snakemake again, in whatever environment is targeted. The script is submitted to some job management platform (e.g. a cluster scheduler like slurm).
        This class can be specialized to generate more specific backends, also for the cloud.
    """
```

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        local_input=None,
+        restart_times=None,
+        exec_job=None,
+        max_status_checks_per_second=1,
+    ):
+
+        # Attach variables for easy access
+        self.workflow = workflow
+        self.quiet = quiet
+        self.workdir = os.path.dirname(self.workflow.persistence.path)
```
johanneskoester: This is the subfolder `.snakemake` inside of the working directory. Is that intended?

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        # IMPORTANT: using Compute Engine API and not k8s == no support secrets
+        self.envvars = envvars or []
+
+        # Quit early if we can't authenticate
+        self._get_services()
+        self._get_bucket()
+
+        # Akin to Kubernetes, create a run namespace, default container image
+        self.run_namespace = str(uuid.uuid4())
+        self.container_image = container_image or "vanessa/snakemake:dev"
```
johanneskoester: As previously discussed let's try everything in order to not need a special container image.

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        self.regions = regions or ["us-east1", "us-west1", "us-central1"]
+
+        # The project name is required, either from client or environment
+        self.project = (
+            os.environ.get("GOOGLE_CLOUD_PROJECT") or self._bucket_service.project
+        )
+        if not self.project:
+            raise WorkflowError(
+                "You must provide a --google-project or export "
+                "GOOGLE_CLOUD_PROJECT to use the Life Sciences API. "
```
johanneskoester: Let us just offer the env var. I think there is no need for a cli option for this (keeping the code simpler).

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+            assume_shared_fs=False,
+            max_status_checks_per_second=10,
+        )
+
+    def _get_services(self):
+        """use the Google Discovery Build to generate API clients
+           for Life Sciences, and use the google storage python client 
+           for storage.
+        """
+        from googleapiclient.discovery import build as discovery_build
```
johanneskoester: The first import of googleapiclient needs to be wrapped in a try statement with an error message instructing to install the missing dependency, similar to the kubernetes import in the kubernetes executor. Reason: for the snakemake-minimal package, I don't want to force people to install those dependencies if they do not need it.

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        self.shutdown()
+
+    def _generate_job_resources(self, job):
+        """given a particular job, generate the resources that it needs,
+           including default regions and the virtual machine configuration
+        """
+        # Right now, do a best effort mapping of resources to instance types
+        cores = job.resources.get("_cores", 1)
+        nodes = job.resources.get("_nodes", 1)
+        mem_mb = job.resources.get("mem_mb", 100)
```
johanneskoester: Also consider adding `disk_mb`, or is that not needed for the API?

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        # Update final list based on the remaining
+        to_remove = set()
+        for zone, types in lookup.items():
+            names = [x["name"] for x in types]
+            for machineType in list(machineTypes.keys()):
+                if machineType not in names:
+                    to_remove.add(machineType)
+
+        for machineType in to_remove:
+            del machineTypes[machineType]
```
johanneskoester: Without digging too much into the details here (I might be wrong), my feeling is that this could be a one-liner instead.

johanneskoester: Maybe not a one-liner but at least less dict editing? Rule of thumb, try to replace `del` statements with better initial filtering. Try to replace nested loops with comprehensions.

johanneskoester: One option to get this entire method cleaner would be to decompose it into smaller "filters" that are subsequently applied to the list or dict of machines?

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+                "Machine prefix {prefix} is too strict, and there are no options "
+                "available.".format(prefix=self._machine_type_prefix)
+            )
+
+        # Now find (quasi) minimal to satisfy constraints
+        machineTypes = keepers
+        keepers = dict()
+
+        minCores = 1
+        minMem = 3.75
```
johanneskoester: Use snake_case.

johanneskoester: Please add a comment explaining the rationale of min_mem.

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+                minMem = machineType["memoryMb"]
+
+        selected = machineTypes[smallest]
+        logger.debug(
+            "Selected machine type {}:{}".format(smallest, selected["description"])
+        )
+        virtualMachine = {
+            "machineType": smallest,
+            "labels": {"app": "snakemake"},
+            "bootDiskSizeGb": 100,  # default is likely 10
```
johanneskoester: Should this be disk_mb, maybe plus an offset?

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        }
+
+        resources = {"regions": self.regions, "virtualMachine": virtualMachine}
+        return resources
+
+    def _generate_build_package(self):
```
johanneskoester: Package is a very generic name. Maybe `source_package`?

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        logger.debug("bucket=%s" % self.bucket.name)
+        logger.debug("subdir=%s" % self.gs_subdir)
+
+    def shutdown(self):
+        """shutdown deletes build packages if the user didn't request to clean
+           up the cache. At this point we've already cancelled running jobs.
+        """
+        # Delete build packages only if user regooglquested no cache
+        if self._save_storage_cache:
+            logger.debug("Requested to save workflow cache, skipping cleanup.")
```
johanneskoester: Workflow cache is misleading (could mean results instead). Rather say workflow sources?

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        """find the SHA256 hash string of a file. We use this so that the
+           user can choose to cache working directories in storage.
+        """
+        if os.path.exists(filename):
+            hasher = hashlib.sha256()
+            with open(filename, "rb") as f:
+                for chunk in iter(lambda: f.read(4096), b""):
+                    hasher.update(chunk)
+            return hasher.hexdigest()
+        logger.warning("%s does not exist." % filename)
```
johanneskoester: This does not need to be a method. It can even go into `snakemake/common.py` for reuse in other places. Also, it should fail (with FileNotFoundError as coming from open) in case of non-existence I think.

snakemake/executors.py
```diff
@@ -2052,3 +2059,605 @@ def run_wrapper(
...
+        """The Google Python API client frequently has BrokenPipe errors. This
+           function takes a request, and executes it up to number of retry,
+           each time with a 2* increase in timeout.
+
+           Parameters
+           ==========
+           request: the Google Cloud request that needs to be executed
+           timeout: time to sleep (in seconds) before trying again
+           attempts: remaining attempts, throw error when hit 0
+        """
```
johanneskoester: @moschetti is there a chance to fix this upstream?

docs/executable.rst
```diff
@@ -178,6 +179,77 @@ a job intends to use, such that kubernetes can allocate it to the correct cloud
...
+credentials will be used.
+
+.. code-block:: console
+
+      $ export GOOGLE_CLOUD_PROJECT=my-project-name
+
+Using this executor typically requires you to start with large data files
+already in Google Storage, and then interact with them via the Google Storage
+remote executor. Here is an example script that shows you how you might
+want to upload files in advance using the Google API Python Client.
```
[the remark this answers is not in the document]
vsoch: Yes I think you can use `gsutil` https://cloud.google.com/storage/docs/uploading-objects - do you want a link added to it?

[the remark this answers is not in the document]
vsoch: okay added this locally.

snakemake/__init__.py
```diff
@@ -1623,6 +1660,47 @@ def get_argument_parser(profile=None):
...
+        "--google-life-sciences-env",
+        nargs="+",
+        metavar="ENVVAR",
+        default=[],
+        help="Specify environment variables to pass to the pipeline.",
+    )
+    group_google_life_science.add_argument(
+        "--google-life-sciences-regions",
+        nargs="+",
+        default=[],
```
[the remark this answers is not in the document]
vsoch: Done.

snakemake/__init__.py
```diff
@@ -1623,6 +1660,47 @@ def get_argument_parser(profile=None):
...
+        help="Specify one or more valid instance regions (defaults to US)",
+    )
+    group_google_life_science.add_argument(
+        "--google-life-sciences-cache",
```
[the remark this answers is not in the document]
vsoch: Done

snakemake/__init__.py
vsoch: Haha, whenever I read that I read "Google Lisa" and my mind wanders off to the Simpsons :) Lisa would definitely use Snakemake!

snakemake/executors.py
vsoch: Sure thing.

snakemake/executors.py
vsoch: Done.

snakemake/executors.py
vsoch: What do you mean upstream? @moschetti can comment, but I've worked with a lot of Google APIs, and the retry functionality (at least afaik) doesn't come with that, there is always a suggestion to implement some kind of [exponential backoff](https://cloud.google.com/iot/docs/how-tos/exponential-backoff). And when I haven't, it leads to a lot of random errors because there is the occasional error, even with Google.

snakemake/executors.py
vsoch: okay can do.

docs/executable.rst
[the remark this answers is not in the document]
vsoch: I'll leave this unresolved for @moschetti to comment.

snakemake/executors.py
vsoch: Is `disk_mb` an attribute that populates from Snakemake? Where does it come from?

docs/executable.rst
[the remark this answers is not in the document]
moschetti: yeah, `gsutil cp` (or `gsutil -m cp` which will enable multipart uploads for large files)

snakemake/__init__.py
moschetti: No, there's not an official short name. 

For reference, Nextflow used `google-pipelines` as the old API name was Pipelines API. So I wouldn't use this.

Maybe go with --google-lifesciences (as one word to match the API namespace though).

snakemake/executors.py
moschetti: as mentioned, the retry and backoff gets implemented here at the client level. that's above my pay grade, but is the general behavior of all the exposed APIs so that users can handle error behavior in a way that's appropriate for their use case.

docs/executable.rst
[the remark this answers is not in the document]
vsoch: okay I'll add this!

snakemake/__init__.py
vsoch: okay I am updating all references (variables and otherwise) of "google_life_sciences" to "google_lifesciences" so we can be consistent.

snakemake/executors.py
vsoch: @johanneskoester I think what I've implemented is a reasonable start (and perhaps we might want to adjust the timeout original value (2) or the retry count (3)? What we basically do is submit the operation of the bat (Google discovery requests have an .execute() function that allows us to pass the request and then execute). And then in the case of a BrokenPipe (the error I was seeing a few times) we run the function again, of course multiplying the timeout so we wait a little longer, and subtracting one from the retry count:

```python
        except BrokenPipeError as ex:
            if attempts > 0:
                sleep(timeout)
                return self._retry_request(request, timeout * 2, attempts - 1)
            raise ex
```
Also notice in the above that's where we wait (the sleep) and then in the case that we've exhausted the retry count, we raise the exception. I think this should work (I haven't hit issues with it yet) and maximally we might need to look for other exception types (haven't seen any yet).

snakemake/executors.py
vsoch: okay I'm going to assume `disk_mb` is a job resource, grab from there, default to 100 and add an offset of 50.

vsoch: or maybe just 10 is okay - I don't want to be single highhandedly responsible for people getting over-charged.

johanneskoester: So, disk_mb is a recommended resource (if needed), and people are able to set it via the command line with --default-resources (see the corresponding help test). The default is `max(2*input.size, 1000)`. You don't need to set any defaults in the executor yourself. Instead, require people to use --default-resources when they use the Google LS backend.

snakemake/executors.py
johanneskoester: Sounds good to me. I just wanted to check if Google wants us to do it like that.

docs/executable.rst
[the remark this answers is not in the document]
johanneskoester: Great, the gsutil cp approach should replace the custom script then.

snakemake/executors.py
vsoch: When you say require - do you mean I should exit with error if the flag is not used? Or the default setting will handle setting it, and I should add to the documentation to encourage users to use the flag?

vsoch: Are you looking at old code? Locally I see:
```python
        virtualMachine = {
            "machineType": smallest,
            "labels": {"app": "snakemake"},
            "bootDiskSizeGb": disk_gb,
        }
```
and it's derived from
```python
        disk_mb = job.resources.get("disk_mb", 102400)  # 100 GB default

        # Convert mb to gb, add buffer of 50
        disk_gb = math.ceil(disk_mb / 1024) + 10
```

vsoch:
I'm not sure about this Sonar addition - these failures didn't show up for the other pushes. :/

vsoch:
@johanneskoester I'm not sure how to address the remaining "code smells" and security checks, they don't seem like terrible issues but the tool seems to think so.

vsoch:
@johanneskoester test_issue635 is failing in testing, it's not associated with this PR.

vsoch:
woo my code isn't as smelly! :tada: I suspect the test will still fail, let me know if this is something related to the PR here.

vsoch:
@johanneskoester I accidentally ran black to format all the files in test - it's not necessary (and won't hurt) but just wanted to let you know when you see they have been changed.

snakemake/executors.py
```diff
@@ -2401,25 +2408,67 @@ def _generate_job_resources(self, job):
...
 
-        # Add all files to a package
-        for root, dirs, files in os.walk(self.workdir):
+        # Add all workflow_sources files
+        for filename in self.workflow_sources:
+            arcname = filename.replace(self.workflow.basedir + os.path.sep, "")
+            tar.add(filename, arcname=arcname)
+
+        # Add snakemake hidden directory
+        for root, dirs, files in os.walk(os.path.join(self.workdir, ".snakemake")):
```
johanneskoester: I don't yet understand what this block is needed for...

johanneskoester: Why do we need to upload the hidden directory?

snakemake/executors.py
```diff
@@ -2087,3 +2101,658 @@ def run_wrapper(
...
+        """generate a single action to execute the job.
+        """
+        # Write the jobscript, this will be the container entrypoint along
+        # with obtaining the working directory for the run
+        jobscript = self.get_jobscript(job)
+        self.write_jobscript(job, jobscript)
+
+        # Remote files (within Snakefile) should be referenced via remotes.GS
+        # For the workdir, we need to upload it, named based on hash of contents
+        targz = self._generate_build_source_package()
```
johanneskoester: Why don't you create the tarball once, upon construction of the executor? Then, we don't need to recreate it for each job.

snakemake/executors.py
vsoch: We have to add the entrypoint scripts, which are located in there.

snakemake/executors.py
vsoch: Definitely, let me fix that. I think I originally made this choice thinking that the build packages would be different for each step, but they aren't.

vsoch: Working on this now.

vsoch: okay I have an answer for why we cannot do that. The job file (the entrypoint) is generated on the level of a job step, so if we only copy the first we get an error:

```bash
/bin/bash: .snakemake/tmp.hzxks7k9/snakejob.copy.2.sh: No such file or directory
```
This is why it's done for each job.

vsoch:
@johanneskoester the merge from master broke the branch, I'm going to see if I can fix it.

snakemake/executors.py
johanneskoester: I don't think you need to create a script as an entrypoint. It should be possible to do what I do in the kubernetes executor (directly pass a shell command instead):
* https://github.com/snakemake/snakemake/blob/master/snakemake/executors.py#L1397-L1402
* https://github.com/snakemake/snakemake/blob/master/snakemake/executors.py#L1417

vsoch: It's generated by snakemake, and includes all the properties / configs, etc. as is done locally. Here is an example:

```bash
#!/bin/sh
# properties = {"type": "single", "rule": "copy", "local": false, "input": ["gcp-public-data-landsat/LC08/01/001/003/LC08_L1GT_001003_20170430_20170501_01_RT/LC08_L1GT_001003_20170430_20170501_01_RT_MTL.txt"], "output": ["snakemake-testing/1/landsat-data.txt"], "wildcards": {}, "params": {}, "log": [], "threads": 1, "resources": {"mem_mb": 100}, "jobid": 2, "cluster": {}}
snakemake snakemake-testing/1/landsat-data.txt --snakefile Snakefile --force -j --keep-target-files  --keep-remote --latency-wait 0 --attempt 1 --force-use-threads  --allowed-rules copy --nocolor --notemp --no-hooks --nolock  --use-conda  --default-remote-provider GS --default-remote-prefix snakemake-testing/1 
```
I don't think I understand how it would work without that.

vsoch: The entrypoint is already overloaded, I don't see how we can add more to it (e.g., see here https://github.com/snakemake/snakemake/pull/90/files#diff-c923148fbfa18c64d8a76ab68afd7b79R2520)

vsoch: Actually I see what you mean, let me give it a try.

vsoch: okay, all fixed, both the entrypoint update (not relying on a script) and consequently, also the generation of the build package (only done once!)

vsoch: okay this is all fixed.

snakemake/executors.py
vsoch: And this is also fixed (meaning that the .snakemake directory is not uploaded, and we generate the package once).

snakemake/executors.py
```diff
@@ -2093,3 +2107,641 @@ def run_wrapper(
...
+            self.exec_job, job, _quote_all=True, use_threads=use_threads
+        )
+
+        # The full command to download the archive, extract, and run
+        # For snakemake bases, we must activate the conda environment, but
+        # for custom images we must allow this to fail (hence || true)
+        commands = [
+            "/bin/bash",
+            "-c",
+            "mkdir -p /workdir && cd /workdir && wget -O /download.py https://gist.githubusercontent.com/vsoch/84886ef6469bedeeb9a79a4eb7aec0d1/raw/181499f8f17163dcb2f89822079938cbfbd258cc/download.py && chmod +x /download.py && source activate snakemake || true && python /download.py download %s %s /tmp/workdir.tar.gz && tar -xzvf /tmp/workdir.tar.gz && %s"
```
johanneskoester: The `|| true` looks suspicious. What is the reason?

vsoch: If they use a non traditional or custom container that doesn’t have an environment called snakemake it should still work.

vsoch:
@johanneskoester it's failing again with the merge with master, totally unrelated to my PR. I don't know how to address that so I'm going to leave it be. How are things being merged into master that don't pass?

snakemake/executors.py
johanneskoester: Actually, you shouldn't need the source activate snakemake at all. The official Snakemake docker container should have the envionment activated automatically.

vsoch: Indeed, I just tested this with the container and you are spot on! I'll test it with the pipelines first and then update the PR here upon success.

vsoch: hey @johanneskoester ! So I just tried this, and now I see why I added it (I think I must have started without it, because I couldn't have known that the container had the snakemake environment). If I don't do that source, then supporting libraries don't seem to be found. Here is the resulting error message:

```python
  File "/download.py", line 6, in <module>
    from google.cloud import storage
ModuleNotFoundError: No module named 'google.cloud'
Started running "snakejob-copy-2"
Stopped pulling "snakemake/snakemake:v5.7.2"
Started pulling "snakemake/snakemake:v5.7.2"
Worker "google-pipelines-worker-aa30ba7f3bbf98ad61c08b44d4d91fd4" assigned in "us-central1-a"
```
So I won't be updating the PR to remove that line, as it breaks function.

vsoch:
Note to @johanneskoester - reminder to add documentation section about local() (when you are back!)

vsoch:
The error is that AWS credentials are not provided in the environment - this is not related to this PR:

```
```

vsoch:
Note that error is still with respect to aws credentials:

```

Run aws-actions/configure-aws-credentials@v1
##[error]Input required and not supplied: aws-access-key-id
##[error]Node run failed with exit code 1
```

snakemake/executors.py
```diff
@@ -2115,3 +2128,741 @@ def run_wrapper(
...
+                len(machine_types), self.regions
+            )
+        )
+
+        # First pass - eliminate anything that too low in cpu/memory
+        keepers = dict()
+
+        # Also keep track of max cpus and memory, in case none available
+        maxCpu = 1
+        maxMem = 100
```
johanneskoester: please always use snake-case: max_cpu, max_mem, ...

snakemake/executors.py
johanneskoester: Can you test this again, now that you use the latest container 5.10.0?

snakemake/executors.py
vsoch: I must have been distracted by the Google keys that use camel! Fixed now.

snakemake/executors.py
```diff
@@ -2115,3 +2128,759 @@ def run_wrapper(
...
+           debug print the list of events, return True if all return codes 0
+           and False otherwise (indication of failure). In that a nonzero exit
+           status is found, we also debug print it for the user.
+        """
+        success = True
+
+        # https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/Event
+        for event in status["metadata"]["events"]:
+
+            logger.info(event["description"])
```
johanneskoester: Turn this into logger.debug(...)

vsoch: On it!

snakemake/executors.py
```diff
@@ -2115,3 +2128,759 @@ def run_wrapper(
...
+                logger.error("{}: {}".format(action["code"], action["cause"]))
+
+            elif "unexpectedExitStatus" in event:
+                action = event.get("unexpectedExitStatus")
+
+                if action["exitStatus"] != 0:
+                    success = False
+
+                    # Provide reason for the failure (desc includes exit code)
+                    msg = "%s" % event["description"]
```
johanneskoester: The above line seems to be superfluous to me.

johanneskoester: Ah, do you make a copy of the string in the event?

vsoch: I can definitely remove it - I was thinking we'd want to print to the console and to whatever error is defined. Would you like me to remove?

snakemake/executors.py
```diff
@@ -2115,3 +2128,759 @@ def run_wrapper(
...
+                    if "stderr" in action:
+                        msg += ": %s" % action["stderr"]
+                        logger.error(msg)
```
johanneskoester: This kind of duplicates the output, because it also prints the entire node-local snakemake output from the cloud job. The master process should usually already generate a sufficient error message. Hence, let us also make above line logger.debug().

snakemake/executors.py
vsoch: The goal of adding the event description to the error message was to give the user that reads the error log context about the action step that the error output is associated with.
```
                    msg = "%s" % event["description"]
                    if "stderr" in action:
                        msg += ": %s" % action["stderr"]
                        logger.error(msg)
```
I can remove it if you think redundant, but then the logger error won't have the description of the event that failed.

snakemake/executors.py
vsoch: gotcha.

snakemake/executors.py
```diff
@@ -2115,3 +2128,759 @@ def run_wrapper(
...
+        # https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/Event
+        for event in status["metadata"]["events"]:
+
+            logger.info(event["description"])
+
+            # Does it always result in fail for other failure reasons?
+            if "failed" in event:
+                success = False
+                action = event.get("failed")
+                logger.error("{}: {}".format(action["code"], action["cause"]))
```
johanneskoester: Please provide an example of what could be displayed here. Then, I'll let you know how we print it.

vsoch: Sure! See the first red line at the top - the FAILED_PRECONDITION is the content of `action["code"]` and then the cause follows. It's just that one little line.

image

Note that the rest of the message hasn't been updated for the changes just pushed to change info / error to debug.

snakemake/executors.py
johanneskoester: You can keep the description I think, but let us make this debug for now. Usually the message that snakemake prints in the master process (the `error in rule ...` thingy) is enough. Having a huge output on error just scares people and distracts them from the true information (which usually is contained in the log file of the rule and not in global stderr or stdout).

snakemake/executors.py
vsoch: What I can do is finish up changes as you suggest here, then run it again and share the updated output.

snakemake/executors.py
vsoch: Gotcha! This is now fixed, and pushed. I'm going to update the branches I have running the test pipeline and I'll show you the updated output in the comment below.

snakemake/executors.py
vsoch: okay here is updated terminal output - you are correct that it's a lot more cleaner!

image

I'm worried that there is sometimes meaningful information in the traceback that is printed, but this way is definitely cleaner looking, so I'm good to give it a try if you think this is best practice (notice that the job is still not exited so we might not see all of it yet).

vsoch: here is the entire thing (with the final error output)

image

vsoch:
Hmm, there seems to be a larger issue that extra arguments (e.g., --skip-cleanup-scripts) aren't passed to the executor. Looking at CPUExecutor and ClusterExecutor, there is repeated code to add these arguments to self.exec_job. I'm out of steam today but want to test this interactively tomorrow and I'll update the PR here with a solution (and hopefully resolve this new bug that just popped up!)

vsoch:
I rebased with master, and there is some newly introduced error that isn't finding a singularity container to use (which looks like it was defined for the workflow?)

```
  File "/home/vanessa/snakemake-work/snakemake/snakemake/__init__.py", line 680, in snakemake
    keepincomplete=keep_incomplete,
  File "/home/vanessa/snakemake-work/snakemake/snakemake/workflow.py", line 895, in execute
    success = scheduler.schedule()
  File "/home/vanessa/snakemake-work/snakemake/snakemake/scheduler.py", line 408, in schedule
    self.run(job)
  File "/home/vanessa/snakemake-work/snakemake/snakemake/scheduler.py", line 427, in run
    error_callback=self._error,
  File "/home/vanessa/snakemake-work/snakemake/snakemake/executors.py", line 2763, in run
    if job.needs_singularity and self.workflow.use_singularity:
  File "/home/vanessa/snakemake-work/snakemake/snakemake/jobs.py", line 307, in needs_singularity
    return self.container_img is not None
  File "/home/vanessa/snakemake-work/snakemake/snakemake/jobs.py", line 316, in container_img
    return self.dag.container_imgs[self.container_img_url]
KeyError: 'docker://continuumio/miniconda3'
```

vsoch:
The change was made here https://github.com/snakemake/snakemake/commit/f7041f963d0219eb07c558a6ebb708a1cf9ea7bb#diff-c923148fbfa18c64d8a76ab68afd7b79R390 @johanneskoester please advise.

johanneskoester:
Thanks for reporting. Issue found, fix is on the way.

vsoch:
Woot! This sounds good to me :)

vsoch:
phew, the SonarCloud errors were alarming - like missing urllib3 import, and a spelling error... *dinosaur screams*

vsoch:
okay, I've just added support for GPUs, along with documentation to show the user how to add `nvidia_gpu` or just `gpu` to get the requested GPUs. I used the following Snakefile (the one in the current test_google_lifesciences folder) modified to ask for GPU:

```
import os

# This test file is taken from Kubernetes - both use Google Cloud
from snakemake.remote.GS import RemoteProvider as GSRemoteProvider
GS = GSRemoteProvider()

rule all:
    input:
        "landsat-data.txt.bz2"

rule copy:
    input:
        GS.remote("gcp-public-data-landsat/LC08/01/001/003/LC08_L1GT_001003_20170430_20170501_01_RT/LC08_L1GT_001003_20170430_20170501_01_RT_MTL.txt")
    output:
        "landsat-data.txt"
    resources:
        mem_mb=100,nvidia_gpu=1
    run:
        # we could test volume size like this but it is currently unclear what f1-micro instances provide as boot disk size
        #stats = os.statvfs('.')
        #volume_gib = stats.f_bsize * stats.f_blocks / 1.074e9
        #assert volume_gib > 90
        shell("cp {input} {output}")

rule pack:
    input:
        "landsat-data.txt"
    output:
        "landsat-data.txt.bz2"
    conda:
        "env.yml"
    log:
        "logs/pack.log"
    shell:
        "bzip2 -c {input} > {output}; echo successful > {log}"
```

And then ran the test locally, here is the full output:

```
Building DAG of jobs...
Unable to retrieve additional files from git. This is not a git repository.
bucket=snakemake-testing
subdir=gpu
locations-available:
asia-northeast1
asia-southeast1
europe-west2
europe-west4
northamerica-northeast1
us
us-central1
us-west2
regions=['us-east1', 'us-west1', 'us-central1']
location=projects/snakemake-testing/locations/us-central1
container=snakemake/snakemake:v5.11.1
Using shell: /bin/bash
Provided cores: 9223372036854775807
Job counts:
	count	jobs
	1	all
	1	copy
	1	pack
	3
Resources before job selection: {'_cores': 9223372036854775807, '_nodes': 1}
Ready jobs (1):
	copy
Selected jobs (1):
	copy
Resources after job selection: {'_cores': 9223372036854775806, '_nodes': 0}

[Wed Apr  8 13:58:57 2020]
rule copy:
    input: gcp-public-data-landsat/LC08/01/001/003/LC08_L1GT_001003_20170430_20170501_01_RT/LC08_L1GT_001003_20170430_20170501_01_RT_MTL.txt
    output: snakemake-testing/gpu/landsat-data.txt
    jobid: 2
    resources: mem_mb=100, nvidia_gpu=1

build-package=source/cache/snakeworkdir-a560de7ca9adbabb53bdc4f2d878fe6fd1e3862bd7654e2b753af22d64952301.tar.gz
found resource request for 1 GPUs. This will limit to n1 instance types.
found 23 machine types across regions ['us-east1', 'us-west1', 'us-central1'] before filtering to increase selection, define fewer regions
Selected machine type n1-highcpu-2:2 vCPUs, 1.8 GB RAM
Get status with:
gcloud config set project snakemake-testing
gcloud beta lifesciences operations describe 3903726045302002262
gcloud beta lifesciences operations list
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Checking status for operation 3903726045302002262
Worker released
Stopped running "snakejob-copy-2": ving to: ‘/download.py’

     0K .                                                     100% 10.9M=0s

2020-04-08 20:06:22 (10.9 MB/s) - ‘/download.py’ saved [1246/1246]

Building DAG of jobs...
Using shell: /bin/bash
Provided cores: 2
Job counts:
	count	jobs
	1	copy
	1

[Wed Apr  8 20:06:33 2020]
rule copy:
    input: gcp-public-data-landsat/LC08/01/001/003/LC08_L1GT_001003_20170430_20170501_01_RT/LC08_L1GT_001003_20170430_20170501_01_RT_MTL.txt
    output: snakemake-testing/gpu/landsat-data.txt
    jobid: 0
    resources: mem_mb=100, disk_mb=128000, nvidia_gpu=1

Downloading from remote: gcp-public-data-landsat/LC08/01/001/003/LC08_L1GT_001003_20170430_20170501_01_RT/LC08_L1GT_001003_20170430_20170501_01_RT_MTL.txt
Finished download.
Uploading to remote: snakemake-testing/gpu/landsat-data.txt
Finished upload.
[Wed Apr  8 20:06:33 2020]
Finished job 0.
1 of 1 steps (100%) done
Complete log: /workdir/.snakemake/log/2020-04-08T200632.429349.snakemake.log
Started running "snakejob-copy-2"
Stopped pulling "snakemake/snakemake:v5.11.1"
Started pulling "snakemake/snakemake:v5.11.1"
Worker "google-pipelines-worker-86c1515f7609c9d3024589b1eb709769" assigned in "us-central1-c"
[Wed Apr  8 14:06:39 2020]
Finished job 2.
1 of 3 steps (33%) done
Resources before job selection: {'_cores': 9223372036854775807, '_nodes': 1}
Ready jobs (1):
	pack
Selected jobs (1):
	pack
Resources after job selection: {'_cores': 9223372036854775806, '_nodes': 0}

[Wed Apr  8 14:06:39 2020]
rule pack:
    input: snakemake-testing/gpu/landsat-data.txt
    output: snakemake-testing/gpu/landsat-data.txt.bz2
    log: snakemake-testing/gpu/logs/pack.log
    jobid: 1

build-package=source/cache/snakeworkdir-4df7b55bc6e4cc04b671c896185c6119c593d78e0a2638f29b81bb6c712905b1.tar.gz
found 23 machine types across regions ['us-east1', 'us-west1', 'us-central1'] before filtering to increase selection, define fewer regions
Selected machine type n1-highmem-4:4 vCPUs, 26 GB RAM
Get status with:
gcloud config set project snakemake-testing
gcloud beta lifesciences operations describe 5016011568141407592
gcloud beta lifesciences operations list
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Checking status for operation 5016011568141407592
Worker released
Stopped running "snakejob-pack-1": ad.py’ saved [1246/1246]

Building DAG of jobs...
Creating conda environment env.yml...
Environment for env.yml created (location: .snakemake/conda/0a062b8f)
Using shell: /bin/bash
Provided cores: 4
Job counts:
	count	jobs
	1	pack
	1

[Wed Apr  8 20:08:39 2020]
rule pack:
    input: snakemake-testing/gpu/landsat-data.txt
    output: snakemake-testing/gpu/landsat-data.txt.bz2
    log: snakemake-testing/gpu/logs/pack.log
    jobid: 0
    resources: mem_mb=15360, disk_mb=128000

Downloading from remote: snakemake-testing/gpu/landsat-data.txt
Finished download.
Activating conda environment: /workdir/.snakemake/conda/0a062b8f
Uploading to remote: snakemake-testing/gpu/logs/pack.log
Finished upload.
Uploading to remote: snakemake-testing/gpu/landsat-data.txt.bz2
Finished upload.
[Wed Apr  8 20:08:42 2020]
Finished job 0.
1 of 1 steps (100%) done
Complete log: /workdir/.snakemake/log/2020-04-08T200810.279853.snakemake.log
Started running "snakejob-pack-1"
Stopped pulling "snakemake/snakemake:v5.11.1"
Started pulling "snakemake/snakemake:v5.11.1"
Worker "google-pipelines-worker-586b436d4afc893eb5411d03e7b639bd" assigned in "us-west1-b"
[Wed Apr  8 14:08:50 2020]
Finished job 1.
2 of 3 steps (67%) done
Resources before job selection: {'_cores': 9223372036854775807, '_nodes': 1}
Ready jobs (1):
	all
Selected jobs (1):
	all
Resources after job selection: {'_cores': 9223372036854775806, '_nodes': 0}

[Wed Apr  8 14:08:50 2020]
localrule all:
    input: snakemake-testing/gpu/landsat-data.txt.bz2
    jobid: 0

Downloading from remote: snakemake-testing/gpu/landsat-data.txt.bz2
Finished download.
[Wed Apr  8 14:08:51 2020]
Finished job 0.
3 of 3 steps (100%) done
Requested to save workflow sources, skipping cleanup.
Complete log: /tmp/snakemake-rbeot8k5/.snakemake
```
While testing, I made sure that the workers generated had GPUs:

image

vsoch:
@johanneskoester I'm not sure about this error (for another test)
```
-- Docs: https://docs.pytest.org/en/latest/warnings.html
=========================== short test summary info ============================
FAILED tests/tests.py::test_issue916 - AssertionError: expected successful ex...
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
= 1 failed, 150 passed, 10 skipped, 2 xpassed, 8 warnings in 400.61s (0:06:40) =
##[error]Process completed with exit code 1.
```

johanneskoester:
Works fine in master though.

tests/common.py
```diff
@@ -99,8 +99,8 @@ def run(
         del os.environ["PYTHONPATH"]
 
     results_dir = join(path, "expected-results")
-    snakefile = join(path, snakefile)
-    assert os.path.exists(snakefile)
```
johanneskoester, line 103: Maybe the failure is a side effect of your changes here?

vsoch: My changes there honor using the Snakefile in the working directory - without that it uses the Snakefile in the original directory (and this breaks my previously working test). Isn’t the fixed approach I did more correct?

vsoch: @johanneskoester it shouldn't be the case that a job running from a working directory in /tmp is using a snakefile that is in the original folder. Or further, it shouldn't be the case that changing the path from the original working directory to the staged temp should break a test (as it did this one). From the test for google life sciences point of view, it breaks because we have to honor the path of the Snakefile (relative to the working directory) to build the .tar.gz package. So if we build with a Snakefile that is outside of the stated working directory, we get a long, convoluted system path inside the .tar.gz and the test fails. If the snakefile is inside the working directory as it is expected, the test works as it should. This is the reason that I changed those lines. I'm not sure what else to do.

vsoch: I can change it back and the other test might pass again, but when you enable google life sciences testing it will fail. Let me know what you'd like to do...

johanneskoester: The test relies on the location being unchaged, indeed. What you describe for GLS is an important point to consider though. Is there a reason to not cleanup the path in your GLS code if the snakefile/workflow sources are outside of the workdir? I think that should be possible. Maybe just always move them into a place like `sources/`, while keeping the folder structure, but without the system specific common prefix.

johanneskoester:
cc #4mmvtk.

tests/common.py
vsoch: I could definitely do that - is `sources` a convention?

vsoch: @johanneskoester this is a deeper bug in snakemake - I can provide a path to a Snakefile somewhere else, and the correct working directory, but the workflow.included adds the wrong files:
```
self.workflow.get_sources()
Unable to retrieve additional files from git. This is not a git repository.
Out[1]: 
{'../../home/vanessa/Desktop/Code/snakemake/snakemake/tests/test_google_lifesciences/Snakefile',
 '../../home/vanessa/Desktop/Code/snakemake/snakemake/tests/test_google_lifesciences/env.yml'}
```
We can assume that the Snakefile could be an outside path from the working directory, but the rest of the included files should be derived from the working directory, no?

vsoch: This gets into very dangerous territory to start copying files around a user working directory - in fact we can't always assume that the user is okay with this.

vsoch: I don't know how to fix this because all the files in self.included are wrong.

vsoch: I think the best approach is the following:

 - put a warning to the user if a file it outside of the working directory and exit on fail. It's not up to snakemake to change the organization / move files on behalf of the user, and it's not unreasonable to require that the Snakefile is in the same workdir. This is akin to a Dockerfile, you can't build a container with any context below the Dockerfile.
 - the bug with the other job run needs to be fixed. It should not rely on the snakefile being in a specific (and arguably wrong) location.

vsoch: okay I've fixed the issue with the test! It was basically hard coding the original path for the file (the issue name) into the output folder, which doesn't make sense. It's running again now and I think we should be ok... sorry for the unneeded noise!

vsoch:
okay we're good! :tada: There is still the issue if the user puts some path for the Snakefile outside of the working directory - which then seems to derive other (non relative) paths that will break for a build package. Since this use case is rare, and we have to preserve a relative context, I'm going to push a check that will ensure the user does not make this mistake:

```
WorkflowError:
All source files must be present in the working directory, /tmp/snakemake-90q8a1_i to be uploaded to a build package that respects relative paths, but /home/vanessa/Desktop/Code/snakemake/snakemake/tests/test_google_lifesciences/env.yml was found outside of this directory. Please set your working directory accordingly, and the path of your Snakefile to be relative to it.
unlocking
```
Testing locally now, will update PR soon.

johanneskoester:
PR #329 introduces string resources. Please see the example in the PR for how I would like people to specify the gpu_model.

vsoch:
Thanks @johanneskoester ! I'll give this an update when we have feedback if we want to use `machine` or `machine_type` (I used machine, but I think @joonjeong suggested `machine_type` and it seems like it is more specific and might be another option, in case we need to add other `machine_` prefix things in the future.

vsoch:
Just kidding! This is all set - the string variables are shown to be in quotes now: https://github.com/snakemake/snakemake/pull/90/commits/14148b5e02ba3690d44e02de95a789670adfc27d

snakemake/common.py
```diff
@@ -54,6 +55,43 @@ def get_uuid(name):
     return uuid.uuid5(UUID_NAMESPACE, name)
 
 
+def get_file_hash(filename, algorithm="sha256"):
+    """find the SHA256 hash string of a file. We use this so that the
+       user can choose to cache working directories in storage.
+    """
+    from snakemake.logging import logger
+
+    if os.path.exists(filename):
```
johanneskoester, line 64: The if is superfluous. `open()` already throws a proper exception.

snakemake/common.py
```diff
@@ -54,6 +55,43 @@ def get_uuid(name):
...
+        with open(filename, "rb") as f:
+            for chunk in iter(lambda: f.read(4096), b""):
+                hasher.update(chunk)
+        return hasher.hexdigest()
+
+    logger.warning("%s does not exist." % filename)
+    raise FileNotFoundError
+
+
+def bytesto(bytes, to, bsize=1024):
```
johanneskoester, line 82: I think this is not needed anymore?

snakemake/executors.py
johanneskoester: Ok, now I understand what you do here. You can instead just use `os.getcwd()`

snakemake/executors.py
```diff
@@ -2168,3 +2171,878 @@ def run_wrapper(
...
+        logger.debug("location=%s" % self.location)
+        logger.debug("container=%s" % self.container_image)
+
+        # The project name is required, either from client or environment
+        self.project = (
+            os.environ.get("GOOGLE_CLOUD_PROJECT") or self._bucket_service.project
+        )
+        if not self.project:
+            raise WorkflowError(
+                "You must provide a --google-project or export "
```
johanneskoester, line 2257: Adapt the error message, --google-project is not available anymore.

snakemake/executors.py
johanneskoester: Not fixed yet?

snakemake/executors.py
```diff
@@ -2168,3 +2171,878 @@ def run_wrapper(
...
+        """
+        if not gpu_count or gpu_count == 0:
+            return
+
+        logger.debug(
+            "found resource request for {} GPUs. This will limit to n1 "
+            "instance types.".format(gpu_count)
+        )
+        self.workflow.default_resources.parsed["nvidia_gpu"] = gpu_count
+        self.workflow.default_resources.args.append("nvidia_gpu=%s" % gpu_count)
```
johanneskoester, line 2503: Why should this become a default resource? It would mean that every job inherits it. It has to be set just in the job itself.

snakemake/executors.py
```diff
@@ -2168,3 +2171,878 @@ def run_wrapper(
...
+        # Look for if the user wants an nvidia gpu
+        gpu_count = job.resources.get("nvidia_gpu") or job.resources.get("gpu")
+        gpu_model = job.resources.get("gpu_model")
+
+        # If a gpu model is specified without a count, we assume 1
+        if gpu_model and not gpu_count:
+            gpu_count = 1
+
+        # Update default resources using decided memory and disk
+        self.workflow.default_resources = self.default_resources
```
johanneskoester, line 2529: Default resources should never be modified at this stage.

snakemake/executors.py
```diff
@@ -2168,3 +2171,878 @@ def run_wrapper(
...
+
+        logger.debug(
+            "found resource request for {} GPUs. This will limit to n1 "
+            "instance types.".format(gpu_count)
+        )
+        self.workflow.default_resources.parsed["nvidia_gpu"] = gpu_count
+        self.workflow.default_resources.args.append("nvidia_gpu=%s" % gpu_count)
+        self._machine_type_prefix = self._machine_type_prefix or ""
+        if not self._machine_type_prefix.startswith("n1"):
+            self._machine_type_prefix = "n1"
```
johanneskoester, line 2506: Why does this set an executor attribute? It is a method called while handling a single job, right?

snakemake/executors.py
```diff
@@ -2168,3 +2171,878 @@ def run_wrapper(
...
+        for snakefile in SNAKEFILE_CHOICES:
+            if os.path.exists(os.path.join(self.workdir, snakefile)):
+                self.snakefile = snakefile
+                break
+
+    def _set_workflow_sources(self, warning_size_gb=20):
+        """We only add files from the working directory that are config related
+           (e.g., the Snakefile or a config.yml equivalent), or checked into git. 
+           Additionally, given that we encourage these packages to be small,
+           we set a warning at 20GB.
```
johanneskoester, line 2711: Individual files in the sources should be much less. Set the limit to 200MB.

johanneskoester: Please move this into a `Workflow.check_source_sizes` function, and just call it from here. This is helpful as well for the kubernetes and the tibanna executor.

snakemake/executors.py
```diff
@@ -2168,3 +2171,878 @@ def run_wrapper(
...
+        # Add all workflow_sources files
+        for filename in self.workflow_sources:
+            arcname = filename.replace(self.workdir + os.path.sep, "")
+            tar.add(filename, arcname=arcname)
+
+        tar.close()
+
+        # Rename based on hash, in case user wants to save cache
+        sha256 = get_file_hash(targz)
+        hash_tar = os.path.join(self.workdir, "snakeworkdir-%s.tar.gz" % sha256)
```
johanneskoester, line 2771: Put this inside of `self.workflow.persistence.aux`, (not yet present, should be defined as `.snakemake/aux`.

snakemake/executors.py
```diff
@@ -2168,3 +2171,878 @@ def run_wrapper(
...
+
+        # Only copy if we don't have it yet, clean up if we do
+        if not os.path.exists(hash_tar):
+            shutil.move(targz, hash_tar)
+        else:
+            os.remove(targz)
+
+        # We will clean these all up at shutdown
+        self._build_packages.add(hash_tar)
```
johanneskoester, line 2780: There should only be one archive generated. It will be the same for all jobs. Hence, this does not need to be a set.

snakemake/executors.py
```diff
@@ -2168,3 +2171,878 @@ def run_wrapper(
...
+            tar.add(filename, arcname=arcname)
+
+        tar.close()
+
+        # Rename based on hash, in case user wants to save cache
+        sha256 = get_file_hash(targz)
+        hash_tar = os.path.join(self.workdir, "snakeworkdir-%s.tar.gz" % sha256)
+
+        # Only copy if we don't have it yet, clean up if we do
+        if not os.path.exists(hash_tar):
```
johanneskoester, line 2774: Move the generation of the source archive into the init method.

snakemake/executors.py
johanneskoester: So, this failed_precondition line still seems to be superflous to me. People can always get the full detail with describe, can't they? @moschetti is there any information in the returned error dict that we cannot get from the describe command?

docs/executing/cluster-cloud.rst
```diff
@@ -103,6 +104,201 @@ a job intends to use, such that kubernetes can allocate it to the correct cloud
...
+                filenames[name] = name.replace(here + os.path.sep, '')
+
+    # Upload files in script and base gooogle life sciences folder
+    for filename, remotename in filenames.items():
+        blob = bucket.blob(remotename)
+
+        if not blob.exists():
+            print("Uploading %s to %s" %(remotename, bucket_name))
+            blob.upload_from_filename(filename, content_type="text/plain")
+
```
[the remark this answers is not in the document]
vsoch: Removed locally.

docs/executing/cluster-cloud.rst
```diff
@@ -103,6 +104,201 @@ a job intends to use, such that kubernetes can allocate it to the correct cloud
...
+Note that you can also easily use the Google Cloud Console interface, if
+a graphical interface is preferable to you.
+
+Environment Variables
+:::::::::::::::::::::
+
+**Important:** Google Cloud Life Sciences uses Google Compute, and does
+**not** encrypt environment variables. If you specify environment
+variables with `--envars` they will **not** be secrets.
```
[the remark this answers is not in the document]
vsoch: Fixed locally.

docs/executing/cluster-cloud.rst
```diff
@@ -103,6 +104,201 @@ a job intends to use, such that kubernetes can allocate it to the correct cloud
...
+:::::::::::::::
+
+By default, Google Life Sciences uses the latest stable version of
+`snakemake/snakemake <https://hub.docker.com/r/snakemake/snakemake/tags>`_
+on Docker Hub. You can choose to modify the container base with
+the `--container-image` (or `container_image` from within Python), 
+however if you do so, your container must meet the following requirements:
+
+ - have an entrypoint that can execute a `/bin/bash` command
+ - have snakemake installed, either via `source conda snakemake` or already on the path
```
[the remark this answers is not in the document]
vsoch: Fixed locally.

docs/executing/cluster-cloud.rst
```diff
@@ -103,6 +104,201 @@ a job intends to use, such that kubernetes can allocate it to the correct cloud
...
+for this variable. If you don't specify a `gpu` or `nvidia_gpu` (with a count) but you do
+specify a `gpu_model`, the count will default to 1.
+
+
+Machine Types
+:::::::::::::
+
+To specify an exact `machine type <https://cloud.google.com/compute/docs/machine-types>_` 
+or a prefix to filter down to and then select based on other resource needs, 
+you can use `--machine-type-prefix` on the command line. If you want to specify
```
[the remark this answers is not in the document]
vsoch: fixed locally.

docs/executing/cluster-cloud.rst
```diff
@@ -103,6 +104,201 @@ a job intends to use, such that kubernetes can allocate it to the correct cloud
...
+rule a:
+    output:
+        "test.txt"
+    resources:
+        machine_type="n1-standard-8"
+    shell:
+        "somecommand ..."
+
+
+The command line argument currently takes preference, and will override
```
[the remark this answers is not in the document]
vsoch: Fixed locally.

snakemake/__init__.py
```diff
@@ -143,6 +143,12 @@ def snakemake(
     container_image=None,
     tibanna=False,
     tibanna_sfn=None,
+    google_lifesciences=False,
+    google_lifesciences_envvars=None,
```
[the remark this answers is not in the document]
vsoch: okay I've removed the google specific envars, and instead am using the set from the workflow (self.workflow.envars) transformed to a list.

snakemake/__init__.py
```diff
@@ -143,6 +143,12 @@ def snakemake(
...
+    google_lifesciences_regions=None,
+    google_lifesciences_location=None,
+    google_lifesciences_cache=False,
+    google_machine_type_prefix=None,
```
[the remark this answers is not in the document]
vsoch: Done.

snakemake/__init__.py
```diff
@@ -1731,6 +1766,55 @@ def get_argument_parser(profile=None):
...
+    group_google_life_science.add_argument(
+        "--google-lifesciences-regions",
+        nargs="+",
+        default=["us-east1", "us-west1", "us-central1"],
+        help="Specify one or more valid instance regions (defaults to US)",
+    )
+    group_google_life_science.add_argument(
+        "--google-lifesciences-location",
+        help="The Life Sciences API service used to schedule the jobs. "
+        " E.g., us-centra1 (Iowa) and europe-west-2 (London) "
```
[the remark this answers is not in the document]
vsoch: Yep I have an extra dash in europe-west-2 (should be europe-west2) https://cloud.google.com/compute/docs/regions-zones

snakemake/__init__.py
```diff
@@ -1856,7 +1940,14 @@ def adjust_path(f):
     try:
         resources = parse_resources(args.resources)
         config = parse_config(args)
-        if (args.default_resources is not None and not args.default_resources) or (
+
+        # Google Life Sciences has different defaults than others
+        if args.google_lifesciences and not args.default_resources:
+            # Maximum persistent disks size should be less than [3072] GB.
+            args.default_resources = ["mem_mb=15360", "disk_mb=128000"]  # n1-standard-4
```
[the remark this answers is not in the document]
vsoch: Will do, but I updated and tested this pretty critically, and had a lot of trouble when I used the tibanna defaults. Happy to change and then address issues if/when they come in.

snakemake/__init__.py
```diff
@@ -1998,6 +2090,24 @@ def adjust_path(f):
                 )
                 sys.exit(1)
 
+    if args.google_lifesciences:
+        if not os.environ.get("GOOGLE_APPLICATION_CREDENTIALS"):
+            print(
+                "Error: GOOGLE_APPLICATION_CREDENTIALS file path must "
```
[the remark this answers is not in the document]
vsoch: Fixed locally.

snakemake/__init__.py
```diff
@@ -1998,6 +2090,24 @@ def adjust_path(f):
...
+                "be exported with --google-lifesciences",
```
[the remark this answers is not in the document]
vsoch: Fixed locally.

snakemake/common.py
vsoch: Fixed.

snakemake/common.py
vsoch: I still use it under setting the workflow sources:

```python
        def check_size(f):
            """A helper function to check the filesize, and return the file
               to the calling function
            """
            gb = bytesto(os.stat(f).st_size, "g")
            if gb > warning_size_gb:
                logger.warning(
                    "File {} (size {} GB) is greater than the {} GB suggested size "
                    "Consider uploading larger files to storage first.".format(
                        f, gb, warning_size_gb
                    )
                )
            return f

```

snakemake/executors.py
vsoch: I think I can remove this entirely because `self._bucket_service.project` is always defined, granted that we successfully grab a bucket!

vsoch: I can't comment on the one above, but this comment:

> Ok, now I understand what you do here. You can instead just use os.getcwd()

That won't work because the workdir is definitely not always (usually not) the os.getcwd() Can you explain why this would work? (e.g., when we run a test and move to a temporary location, the working directory is the temporary location but os.getcwd() is still my present working directory where I'm running from).

snakemake/executors.py
vsoch: This was an issue that I fixed where the default resources weren't properly carried through. I don't have a link at the moment to show why, but I spent a long time debugging and if you remove it will lead to bugs.

For this comment:

> Not fixed yet?

I don't understand the comment? The discovery API is a good, stable way to use the raw Google APIs without worrying about bugs in their clients (at least I've had much more luck).

snakemake/executors.py
vsoch: This was done because something wasn't working, I really don't want to go through debugging it again because I know it's there for a reason. :(

snakemake/executors.py
vsoch: We need the machine type prefix in other places for the executor, and it's an easy / consistent way to do it instead of passing around variables. If a gpu is chosen it must be in the n1 family.

snakemake/executors.py
vsoch: All set.

snakemake/executors.py
vsoch: Could you clarify what you want/mean here?

snakemake/executors.py
vsoch: We've had this discussion before and I've tried it, and it doesn't work. See https://github.com/snakemake/snakemake/pull/90#pullrequestreview-338887510.

snakemake/executors.py
vsoch: I think this won't work until the job step for the same reason described in https://github.com/snakemake/snakemake/pull/90#pullrequestreview-338887510.

vsoch:
> So, this failed_precondition line still seems to be superflous to me. People can always get the full detail with describe, can't they? @moschetti is there any information in the returned error dict that we cannot get from the describe command?

@johanneskoester I agree with you - this message is typically that it exited with error code 1 and I haven't found it useful for debugging. I'm going to change to debug.

vsoch:
@johanneskoester once we rename the files and put them under an executors folder, it will almost be impossible to do any updates from master. So I think we should only do this at the very end, absolutely when everything else is good and we are ready for merge otherwise.

snakemake/common.py
vsoch: And now this is a function provided by the workflow class.

vsoch:
@johanneskoester I'll want to ask for your suggestions for how to reduce "cognitive complexity" of functions - I made them modular at a level I thought made sense, and most of the length for the ones flagged is due to really long WorkflowError messages. I'm happy to refactor in any way you think is better.

snakemake/executors.py
johanneskoester: See commit a8945fe6. 
```suggestion
        hash_tar = os.path.join(self.workflow.persistence.aux_path, "workdir-%s.tar.gz" % sha256)
```

snakemake/executors.py
johanneskoester: Mhm, in that comment you said that it is fixed. The job step (or jobscript) does not need to be part of the archive. It does not even need to be a script. See the kubernetes executor, where we just specify a shell command to be run. The job properties line you refer to in your previous [comment](https://github.com/snakemake/snakemake/pull/90#pullrequestreview-338887510) is meaningless here, it is for cluster environments. Just directly specify the exec statement when submitting the job.

snakemake/executors.py
johanneskoester: See my comment above.

snakemake/executors.py
vsoch: @johanneskoester I think we already switched it so that the exec_job statement is the command given to the container, e.g.,
```python
        exec_job = self.format_job(
            self.exec_job, job, _quote_all=True, use_threads=use_threads
        )

        # Now that we've parsed the job resource requirements, add to exec
        exec_job += self.get_default_resources_args()

        # The full command to download the archive, extract, and run
        # For snakemake bases, we must activate the conda environment, but
        # for custom images we must allow this to fail (hence || true)
        commands = [
            "/bin/bash",
            "-c",
            "mkdir -p /workdir && cd /workdir && wget -O /download.py https://gist.githubusercontent.com/vsoch/84886ef6469bedeeb9a79a4eb7aec0d1/raw/181499f8f17163dcb2f89822079938cbfbd258cc/download.py && chmod +x /download.py && source activate snakemake || true && pip install crc32c && python /download.py download %s %s /tmp/workdir.tar.gz && tar -xzvf /tmp/workdir.tar.gz && %s"
            % (self.bucket.name, self.pipeline_package, exec_job),
        ]
```
and if we look at logs [from the tutorial](https://github.com/snakemake/snakemake/pull/331/files#diff-eb71c39d80e149671e6f4b862fda5612R730) we can verify that is what is happening. So I'll test moving the archive generation to the beginning again, it will either work or there is something else I'm forgetting :)

vsoch: Testing now!

vsoch: okay that worked! Sorry for not trying it before remembering that we made all these changes.

vsoch:
Not sure what is going on with the build container image step, but it's been going for over 2 hours.

image

johanneskoester:
That issue is resolved now.

snakemake/executors.py
johanneskoester: Good work, thanks!

johanneskoester:
> @johanneskoester I'll want to ask for your suggestions for how to reduce "cognitive complexity" of functions - I made them modular at a level I thought made sense, and most of the length for the ones flagged is due to really long WorkflowError messages. I'm happy to refactor in any way you think is better.

I have looked at them again, and I think they are indeed ok to remain like this. Sorry for the noise.

johanneskoester:
GCloud credentials are now in place. I think you should be able to activate your test case here. In line with the kubernetes and tibanna tests, please move your test case in a separate file `tests/test_google_life_sciences.py` which shall be called in a separate job step.

vsoch:
@johanneskoester what is the name of the envar secret - did you just call it `GOOGLE_APPLICATION_CREDENTIALS`? I'm adding the step in main.yml now, should be able to push to test as soon as I know that name!

vsoch:
E.g., this is what I'd guess, let me know if I'm off
```yaml
    env:
      AWS_AVAILABLE: ${{ secrets.AWS_ACCESS_KEY_ID }}
      GCP_AVAILABLE: ${{ secrets.GCP_SA_KEY }}
      GOOGLE_APPLICATION_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
```

vsoch:
Dearest sonarcloud - you've had a change of heart! Because before you said my code was smelly. Maybe it's good smelly now :)

.github/workflows/main.yml
```diff
@@ -69,6 +70,14 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-1
 
+      - name: Test Google Life Sciences Executor
+        if: env.GOOGLE_APPLICATION_CREDENTIALS
```
johanneskoester, line 74:
```suggestion
        if: env.GCP_AVAILABLE
```

The setup gcloud step above enables the env var GOOGLE_APPLICATION_CREDENTIALS for all following steps. That should be all.

.github/workflows/main.yml
```diff
@@ -35,6 +35,7 @@ jobs:
     env:
       AWS_AVAILABLE: ${{ secrets.AWS_ACCESS_KEY_ID }}
       GCP_AVAILABLE: ${{ secrets.GCP_SA_KEY }}
+      GOOGLE_APPLICATION_CREDENTIALS: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
```
johanneskoester, line 38:
```suggestion
```
Not needed.

vsoch:
@johanneskoester the test did not run still - I can't see settings/secrets so I'll need your guidance about what needs to be checked.

johanneskoester:
You are on a fork. github does not expose secrets to forks.

johanneskoester:
Let us go on in PR #384

PR #91, opened 2019-11-05
Define required environment variables via the Snakefile syntax.

johanneskoester:
This introduces a new keyword `envvars:` that allows to request the
definition of certain environment variables before a workflow is executed.
If those a not present, a helpful error message is displayed.
The given environment variables are automatically passed to cloud jobs (kubernetes for now).

# TODO before merge
* [x] Implement envvar passing for tibanna.
* [x] Write a section in the docs.


johanneskoester:
@SooLee may I ask you to ensure that the defined environment variables (available as a list in `workflow.envvars`) are passed to tibanna jobs?

snakemake/workflow.py
```diff
@@ -837,6 +839,19 @@ def current_basedir(self):
...
+        Register environment variables that shall be passed to jobs.
+        """
+        undefined = [var for var in envvars if var not in os.environ]
+        if undefined:
+            raise WorkflowError(
+                "The following environment variables are requested by the workflow but undefined. "
+                "Please make sure that they are correctly defined before running Snakemake:\n"
+                "{}".format("\n".join(undefined))
+            )
+        self.envvars = envvars
```
vsoch: okay, so it looks like we are just keeping a list of the keys here - for local this makes sense because the job is grabbing from the environment directly. For Kubernetes (and also Google Life Sciences / GLS that created a similar variable) we will need to actually grab and then pass these variables on to the cloud executors. My one concern is with respect to secrets - it looks like the kubernetes client apis can handle secrets well (in which case it would make sense to just pass all as secrets, why wouldn't we want secrets?) but for Google Life Sciences, it isn't clear to me if the environment passed gets encrypted automatically (we need to ask). Either way, @johanneskoester once this is integrated I'll update the GLS to also use the global envvars.

johanneskoester: What you describe is what the kubernetes executor already does. So far, it took the envvar names from a special CLI arg. Now, it uses `workflow.envvars`. That is why you do not see it in the diff. However, you see the encoding of the envvar values as a secret [here](https://github.com/snakemake/snakemake/pull/91/files#diff-c923148fbfa18c64d8a76ab68afd7b79R1329).

johanneskoester: Yes, it is kind of mandatory that the variables are passed on in an encrypted way (at least via HTTPS). This is the case for kubernetes, and it has to be ensured for Google LS as well.

johanneskoester:
Indeed, it would be best to wait with the merge until tibanna is there and the docs have been updated. I'll add todo items to the top.

snakemake/workflow.py
vsoch: okay understood - I hope that LS / pipelines can do similar!

SooLee:
It works now with local testing with tibanna >= 0.11

PR #97, opened 2019-11-08
Use posixpath.relpath to preserve '/' in target files (fixes #46)

### LINKED ISSUE #46: Rules with outputfiles in subfolders on Windows (possible path separator issue)
**Snakemake version**
5.7.1
**Describe the bug**
Unable to specify output files in subfolders on windows. I have tried both 
This works on linux and WSL.

**Logs**
Powershell
```
PS > snakemake -np tmp/entries.feather
Building DAG of jobs...
MissingRuleException:
No rule to produce tmp\entries.feather (if you use input functions make sure that they don't raise unexpected exceptions).
```

Git-bash (MINGW)
```
$ snakemake -np tmp/entries.feather
Building DAG of jobs...
MissingRuleException:
No rule to produce tmp\entries.feather (if you use input functions make sure that they don't raise unexpected exceptions).
```

**Minimal example**
Snakefile
```
rule convert_xlsx:
    input:
        "input/pi/{filename}.xlsx"
    output:
        "tmp/{filename}.feather"
    shell:
        "echo {input} > {output}"
```

**Additional context**
I believe this is caused by the path separator. Are there any recommended best practices for writing windows and linux compatible workflows?

coroa:
It's important to preserve '/' before resolving target files in the DAG.

Alternatively, `Dag.file2jobs` could replace them back to `/`, or one could improve `OutputIndex` to match with different os separators.

### DIFF
--- a/snakemake/workflow.py
+++ b/snakemake/workflow.py
@@ -9,6 +9,7 @@ import sys
 import signal
 import json
 import urllib
+import posixpath
 from collections import OrderedDict
 from itertools import filterfalse, chain
 from functools import partial
@@ -425,7 +426,7 @@ class Workflow:
         else:
 
             def files(items):
-                relpath = lambda f: f if os.path.isabs(f) else os.path.relpath(f)
+                relpath = lambda f: f if os.path.isabs(f) else posixpath.relpath(f)
                 return map(relpath, filterfalse(self.is_rule, items))
 
         if not targets:


melund:
Just a suggestion. Could you add (or point to an existing test) which triggers this on Windows?

melund:
I am not sure this will do the right thing on windows:

image

coroa:
I just confirmed in a virtual machine, that this is thought to simplistically. Will close.

What exactly is the reasoning behind allowing non-relative paths in the first place?

coroa:
> Just a suggestion. Could you add (or point to an existing test) which triggers this on Windows?

Since you were able to make the tests work on windows in your PR and @fneum confirmed that even on your branch, the target file path is affected, it means there is no test atm covering this case.

The simple example in #46 would be a good addition, but I won't take it, sorry.

Issue #70, opened 2019-10-30
Support execution of Jupyter Notebooks using script directive


kpj:
When writing pipelines which involve many visualizations and light data processing, I tend to implement the individual steps using Jupyter Notebooks and glue them together using Snakemake.
At the moment, I execute them manually using `nbconvert`. This is inconvenient and makes passing parameters difficult/impossible.

An elegant solution to this situation would be to allow Jupyter notebooks (`*.ipynb`) to be used via the script directive:
```
rule foo:
  input: 'data.csv'
  output: 'figure.pdf'
  script: 'CreateFigure.ipynb'
```

This was already discussed in a now stale [bitbucket issue](https://bitbucket.org/snakemake/snakemake/issues/529/support-running-from-ipython-notebooks), but no solution was proposed.
Other [efforts](https://github.com/nick-youngblut/snakemakeNB) exist, but seem outdated and not well integrated into Snakemake itself.
Taking a quick look at the [current Snakemake source](https://github.com/snakemake/snakemake/blob/cd0be446d310a3a6aa9a65bec563ef53d548d311/snakemake/script.py#L266) suggests that no implementation efforts have been started.

If this is a feature others would be interested in as well, I'd be very happy to contribute a PR.

edit: I started an implementation [here](https://github.com/kpj/snakemake/tree/jupyter_notebook_support).

SultanOrazbayev:
Looks good, but maybe a `notebook` would be better instead of `script`.

kpj:
@SultanOrazbayev: I feel like using `script` would fit the existing functionality better.

sappjw:
I've implemented a wrapper for calling [papermill](https://papermill.readthedocs.io) from Snakemake.  This may be a cleaner way to do it out of the box.

kpj:
@sappjw: Very cool! How do you handle passing parameters (more specifically, the `snakemake` object) to the notebook?

sappjw:
The wrapper handles putting together the shell command, so it would suffer the same limitations as `papermill`.  No objects can get passed ([even using the Python API](https://github.com/jupyter/nbconvert/issues/1070#issuecomment-515798400)), just parameters.  I didn't think about the advantage of integrating notebook-running functionality into Snakemake is that you could pass objects to the notebook.

kpj:
[...]

PR #101, opened 2019-11-12
Support Jupyter Notebook execution in script directive

kpj:
This PR implements the enhancement suggested in #70. In short, it allows Jupyter notebooks to be passed to the script directive in order to be executed.

Additionally, it provides a test case, extra install requirements for `jupyter`, as well as a short section in the documentation.

It might be necessary to discuss conda integration as selecting a kernel from such an environment seems non-trivial (https://github.com/jupyter/nbconvert/issues/515).


docs/snakefiles/rules.rst
```diff
@@ -433,6 +433,27 @@ Apart from Python scripts, this mechanism also allows you to integrate R_ and R
...
+
+    rule NAME:
+        input:
+            "path/to/inputfile",
+            "path/to/other/inputfile"
+        output:
+            "path/to/outputfile",
+            "path/to/another/outputfile",
+            notebook_output = "path/to/processed_notebook.ipynb"
+        script:
```
johanneskoester: I would like to have a separate directive for this, called `notebook`:
```suggestion
        notebook:
```

This provides more clarity with the other formats that are allowed below.

setup.py
```diff
@@ -58,7 +58,10 @@
         "gitpython",
         "psutil",
     ],
-    extras_require={"reports": ["jinja2", "networkx", "pygments", "pygraphviz"]},
+    extras_require={
+        "reports": ["jinja2", "networkx", "pygments", "pygraphviz"],
+        "jupyter": ["jupyter", "jupyter_contrib_nbextensions"]
```
johanneskoester: I think it should be enough to depend on nbformat. For running the notebook rule, one simply has to define a conda or singularity directive which includes jupyter and whatever other packages are needed for running the notebook.

snakemake/script.py
```diff
@@ -309,7 +313,7 @@ def script(
     f = None
     try:
         path, source, language = get_source(path, basedir)
-        if language == "python":
+        if language == "python" or language == "jupyter":
```
johanneskoester: This is suboptimal, because jupyter is in principle language agnostic. Suggestion:
instead of keeping the script function as it is, convert it into a class, thereby splitting up the central parts into methods (get_preamble1, get_preamble2, write_script, execute_script, etc.). Then, inherit from that class in a new module `snakemake/jupyter.py`, where you redefine the write_script method into writing the notebook, and redefine the get_preamble2 method to add the jupyter specific code where you set the working directory. Finally, you redefine the execute_script method to perfom the jupyter specific execution.

I know, this is a lot to ask for. Hence, if you feel uncomfortable with that, I would also be fine to do the refactoring myself first, and then you can simply adapt to those changes.

snakemake/script.py
```diff
@@ -484,10 +500,14 @@ def script(
         with tempfile.NamedTemporaryFile(
             suffix="." + os.path.basename(path), dir=dir, delete=False
         ) as f:
-            if not language == "rmarkdown":
-                f.write(preamble.encode())
-                f.write(source)
-            else:
+            if language == "jupyter":
+                nb = nbformat.reads(source, as_version=4)  # nbformat.NO_CONVERT
```
johanneskoester: I wonder how you envision drafting such notebooks? Since the notebook code has to depend on those variables being defined, it is a bit cumbersome to initially develop it, in particular because one somehow looses the biggest advantage of jupyter, which is its interactiveness. Maybe we can offer the following:
When running Snakemake with

```
snakemake --export-prototype-notebook <rulename>=<notebookfile>
```
It only executes the workflow up to the necessary input of the given rule. Then, it even runs this code if the notebook is not yet present, but it does not execute the notebook. Instead, it just writes a draft file including the preamble. The user can then open that notebook with jupyter, edit it under consideration of the real data coming from the rule, and save it.

Finally, with
```
snakemake --import-prototype-notebook <rulename>=<notebookfile>
```
Snakemake copies the notebook into whatever is defined in the rule, while removing the preamble block from the prototype.

This is not necessary for this PR (it would definitely be a separate PR), but I wanted to hear your thoughts.

snakemake/script.py
```diff
@@ -535,6 +558,61 @@ def script(
...
+                    raise WorkflowError(
+                        "Invalid Jupyter Notebook output format: '{ext}'".format(
+                            ext=ext
+                        )
+                    )
+
+                # remove preamble from output
+                nb = nbformat.read(tmp_output, as_version=nbformat.NO_CONVERT)
+
+                nb["cells"].pop(0)
```
johanneskoester: I think for provenance reasons it would actually be good if the preamble was kept. This way, people will be able to see all the parameters used inside of the rendered notebook. Also, they would be able to simply reexecute or modify that particular step. This can be super powerful, e.g. if a core facility passes workflows as reports including the corresponding notebooks for some plots. Power users can just open the notebooks and interactively modify e.g. the styling or add annotations for their publication. Then, (ideally) they can pass them back to the core facility for re-import into the original workflow, in order to get an updated analysis.

johanneskoester:
Regarding this: https://github.com/jupyter/nbconvert/issues/515
I think we can overcome this if we simply execute the notebook with `jupyter` itself, instead of doing it via nbconvert.

setup.py
Emiller88: I agree, I think this should be mentioned in the documentation though for new users.

Emiller88:
Would this support other kernels besides python?

johanneskoester:
@Emiller88 not yet in the current PR, but this is what I have in mind with the proposed refactoring.

docs/snakefiles/rules.rst
kpj: Is the idea for the `script` and `notebook` directives to be mutually exclusive, or would they be executed in some order?

setup.py
kpj: That seems sensible. Should `nbformat` be added to `install_requires`?

snakemake/script.py
kpj: Ah yes, very good point! Since I mostly use the Python kernel I forgot about other options.
The refactoring sounds like a great idea. If you don't mind I can commit a rough draft, and only continue if it fits your vision.

To clarify: there would be a `ScriptExecutor` base class with methods:
* `get_preamble` (handling https://github.com/snakemake/snakemake/blob/0c1e56c163083034a6502e30b07dd7e5ef5ff67e/snakemake/script.py#L312-L479)
  * how would your suggested `get_preamble1` and `get_preamble2` differ?
* `write_script`(handling https://github.com/snakemake/snakemake/blob/0c1e56c163083034a6502e30b07dd7e5ef5ff67e/snakemake/script.py#L484-L504)
* `execute_script` (handling https://github.com/snakemake/snakemake/blob/0c1e56c163083034a6502e30b07dd7e5ef5ff67e/snakemake/script.py#L506-L562)

Would the end goal be for every script type (Python, R, Julia, ...) to be added to this framework, or only Jupyter?

snakemake/script.py
kpj: Good point. Do you think this is also necessary if the notebook is converted to a non-interactive format (e.g. pdf)?
Knowing about the executed parameters is probably always a good thing, but if the rendered pdf is distributed to non-pipeline users, they might be confusing.

snakemake/script.py
kpj: This point definitely needs some thought. At the moment I develop the notebook outside of Snakemake and use hard-coded paths to existing Snakemake outputs. Once the notebook is "finished", I add it to the pipeline and adjust the paths (using `snakemake.input`, etc). This is certainly far from optimal.

I think to answer this we need to figure out what the central goal of using Notebooks with Snakemake is. For me (so far, I am happy to change my mind 😄) Jupyter notebooks are just another piece of code which computes results (and as a side effect produces a nicely readable PDF with text and figures). Thus my usage of the `script` directive.
If we think of them as a way of building Snakemake pipelines "step-by-step", your proposal might be the way to go.

docs/snakefiles/rules.rst
johanneskoester: Mutually exclusive. Like shell and script.

johanneskoester: It requires some knowledge of the Snakemake parser. I can do that part if you want.

setup.py
johanneskoester: Yes, I think so. Once this is available, many workflows will make use of it, because it is such a cool feature :-).

snakemake/script.py
johanneskoester: Every script type should be added to this framework.

snakemake/script.py
johanneskoester: I think it is both. To me, it would kind of optionally replace the normal python or R scripts, allowing for a more interactive way of drafting them. But also in your case, it would simplify the inclusion in the pipeline.

snakemake/script.py
johanneskoester: I think this could be handled by some kind of explanatory comment at the beginning. Another possibility would be to print the values in a structured format instead of having the cell with the pickled object in such cases.

docs/snakefiles/rules.rst
kpj: Yes, that would be nice :)

setup.py
kpj: Great, I amended the requirements!

snakemake/script.py
kpj: Perfect, I'll start with an implementation for Jupyter notebooks only and try to keep it general.

snakemake/script.py
kpj: I like the idea of printing the values in a structured format.
I'll see how it fits the class-based implementation discussed above.

snakemake/script.py
kpj: Very fascinating use case :)
Maybe a good way of approaching this becomes clearer after the other proposed changes are implemented.

kpj:
I refactored the script execution into an abstract class `ScriptBase` and a child class `JupyterNotebook` which implements `get_preamble`, `write_script`, and `execute_script`.

Let me know whether this is heading in the right direction :)

johanneskoester:
Btw. this will be one of the key features in the new Snakemake paper I will start writing soon. I would like you to be one of the co-authors if that is ok for you?

SultanOrazbayev:
If there's no objection to external packages, then `papermill` (https://github.com/nteract/papermill) could simplify the code quite a bit.

The example below uses papermill and wrapping it into a `notebook` directive might be simpler.

```
from papermill import execute_notebook as nb

rule all:
  input: 'test.txt'

rule run_notebook:
  output: '{sample}.txt'
  log: 'nb.{sample}.ipynb'
  run: nb(input_path='testnotebook.ipynb', output_path=log[0], parameters = {'output':output, 'wildcards':wildcards})
```

So with a `notebook` directive implemented the above might look like:
```
rule run_notebook:
   notebook: 'testnotebook.ipynb'
   output: '{sample}.txt'
   log: 'nb.{sample}.ipynb'
```

kpj:
@johanneskoester: Yes, that sounds great. I'll implement those classes.
I'd be very happy to be a co-author.

kpj:
Do you have a suggestion for how to share the Python preamble between the `PythonScript` and `PythonJupyterNotebook` classes?
They are mostly the same, but require minor changes (e.g. Jupyter Notebooks do not define `__file__`)

One way of solving this would be to have a `get_preamble_skeleton` function which returns a preamble with placeholder, which is then filled with appropriate values for either Python scripts or notebooks in the respective class. The same could also be done for R and R notebooks.

johanneskoester:
> Do you have a suggestion for how to share the Python preamble between the `PythonScript` and `PythonJupyterNotebook` classes?
> They are mostly the same, but require minor changes (e.g. Jupyter Notebooks with Python kernel do not define `__file__`)
> 
> One way of solving this would be to have a `get_preamble_skeleton` function (probably as a member of `PythonScript`) which returns a preamble with placeholder, which is then filled with appropriate values for either Python scripts or notebooks in the respective class. The same could also be done for R and R notebooks.

Yes, that would have been my suggestion as well.

johanneskoester:
> If there's no objection to external packages, then `papermill` (https://github.com/nteract/papermill) could simplify the code quite a bit.
> 
> The example below uses papermill and wrapping it into a `notebook` directive might be simpler.
> 
> ```
> from papermill import execute_notebook as nb
> 
> rule all:
>   input: 'test.txt'
> 
> rule run_notebook:
>   output: '{sample}.txt'
>   log: 'nb.{sample}.ipynb'
>   run: nb(input_path='testnotebook.ipynb', output_path=log[0], parameters = {'output':output, 'wildcards':wildcards})
> ```
> 
> So with a `notebook` directive implemented the above might look like:
> 
> ```
> rule run_notebook:
>    notebook: 'testnotebook.ipynb'
>    output: '{sample}.txt'
>    log: 'nb.{sample}.ipynb'
> ```

This is indeed a good suggestion that would substantially simplify the code. @kpj any objections?

kpj:
> This is indeed a good suggestion that would substantially simplify the code. @kpj any objections?

I also think it's a great idea. Using `papermill`'s parameter system should make prototyping notebooks and interacting with `snakemake` easier.

What would be the preferred way of integrating it? One could either use its commandline interface with `shell(..)` or the Python API. In the latter case, how would one handle `self.bench_record`?

johanneskoester:
> I also think it's a great idea. Using `papermill`'s parameter system should make prototyping notebooks and interacting with `snakemake` easier.
> 
> A few question one needs to ponder about:
> 
>     * What would be the preferred way of integrating it? One could either use its commandline interface with `shell(..)` or the Python API. In the latter case, how would one handle `self.bench_record`?
The shell way makes sense, because even when using the python API, we would still need to spawn a new process in order to allow for local parallelization (like with do with the run directive). Hence, it makes sense to just use the shell.

>     * Assuming that we still prepend the Snakemake preamble, which variables should we make available as parameters? Even if it's only `input` and `output` one would have to deal with duplicates.
I think it should be the same as with normal scripts. Hence, a snakemake object as the only parameter.

>     * Should we use the `log` directive to store the converted notebook (as suggested before)?
I don't think that is necessary. The converted notebook can always be recreated from the rule via a special command line flag. In fact we should have the same for scripts. Something like `snakemake <targetfile> --export-script <scriptpath>` and `snakemake <targetfile> --export-notebook <notebookpath>`.

kpj:
> The shell way makes sense, because even when using the python API, we would still need to spawn a new process in order to allow for local parallelization (like with do with the run directive). Hence, it makes sense to just use the shell.
> [..]
> I think it should be the same as with normal scripts. Hence, a snakemake object as the only parameter.

Yes, this consistency makes sense.

However, since we will use `shell(..)` to call papermill, this would mean that the snakemake object needs to be passed as a parameter on the commandline ([or via YAML](https://papermill.readthedocs.io/en/latest/usage-execute.html#using-a-yaml-string-for-parameters)).
As the snakemake object is currently being pickled, this conversion might be non-trivial/lead to inconsistency issues with other script types.

Do you think it makes more sense to stick with the embedded preamble, or should we try to make the snakemake object a papermill parameter?

> I don't think that is necessary. The converted notebook can always be recreated from the rule via a special command line flag. In fact we should have the same for scripts. Something like `snakemake <targetfile> --export-script <scriptpath>` and `snakemake <targetfile> --export-notebook <notebookpath>`.

That makes sense. We would however need a way of specifying a path to store the *executed* notebook (i.e. still in ipynb format).
Currently, this is handled by looking for `snakemake.output.notebook_output` and discarding the output if the variable is not set.

johanneskoester:
> > I don't think that is necessary. The converted notebook can always be recreated from the rule via a special command line flag. In fact we should have the same for scripts. Something like `snakemake <targetfile> --export-script <scriptpath>` and `snakemake <targetfile> --export-notebook <notebookpath>`.
> 
> That makes sense. We would however need a way of specifying a path to store the _executed_ notebook (i.e. still in ipynb format).
> Currently, this is handled by looking for `snakemake.output.notebook_output` and discarding the output if the variable is not set.

Not sure I understand. Does this mean that the executed notebook is stored as well, containing additional information like generated plots?


> > The shell way makes sense, because even when using the python API, we would still need to spawn a new process in order to allow for local parallelization (like with do with the run directive). Hence, it makes sense to just use the shell.
> > [..]
> > I think it should be the same as with normal scripts. Hence, a snakemake object as the only parameter.
> 
> Yes, this consistency makes sense.
> 
> However, since we will use `shell(..)` to call papermill, this would mean that the snakemake object needs to be passed as a parameter on the commandline ([or via YAML](https://papermill.readthedocs.io/en/latest/usage-execute.html#using-a-yaml-string-for-parameters)).
> As the snakemake object is currently being pickled, this conversion might be non-trivial/lead to inconsistency issues with other script types.
> 
> Do you think it makes more sense to stick with the embedded preamble, or should we try to make the snakemake object a papermill parameter?

Ok, you are right. Let us for now keep papermill out for now, and stick to your implementation. When using the snakemake preamble, the benefit is indeed minimal. If we have a good idea for a special integration in the future, we can add an additional `papermill` directive.

snakemake/workflow.py
```diff
@@ -54,6 +54,7 @@
 from snakemake.persistence import Persistence
 from snakemake.utils import update_config
 from snakemake.script import script
+from snakemake.notebook import notebook
```
johanneskoester: May I ask you to now move the Notebook specific implementations of the ScriptBase class into a new module `snakemake.notebook` and add a function `notebook(...)` there, similar to the `script(...)` function in `snakemake.script`. Also, could you please move the script specific implementations of ScriptBase into the `snakemake.script` module?

kpj:
> Not sure I understand. Does this mean that the executed notebook is stored as well, containing additional information like generated plots?

Yes, exactly! I found this very useful as a way of investigating what exactly happened during execution.

> Ok, you are right. Let us for now keep papermill out for now, and stick to your implementation. When using the snakemake preamble, the benefit is indeed minimal. If we have a good idea for a special integration in the future, we can add an additional papermill directive.

Sounds good.

kpj:
A few (possibly) last things to consider:

* Storing the executed notebook is a good idea, I think. Currently, the output file `notebook_output ` is used to specify the destination path. Is this the best solution?

* Adding the option to automatically convert the executed notebook to another format (pdf, html, ...) is not needed (so I should remove the according code).

* The preamble is currently being removed before saving the notebook output. For now, it would probably be the best to simply leave it in.

* `jupyter nbconvert` instead of `jupyter-nbconvert` (note the dash) is currently used to execute the notebooks. The former is only provided by [jupyter_contrib_nbextensions](https://github.com/ipython-contrib/jupyter_contrib_nbextensions#2-install-javascript-and-css-files) (I believe), so we should probably use the latter.

* One should probably test how conda/... integration works on machines which are not my own :)

kpj:
A note regarding conda environments:

I believe there is currently a [bug](https://github.com/kpj/snakemake/blob/f8a2f0b20c9726489f2b7ca7db791560696ee835/snakemake/shell.py#L99), because `shell(..)` makes use of `context.get("conda_env", None)` which does currently always return `None`.
It might be enough to simply replace it with `context["self"].conda_env` (seems to work for me), but there might be more side-effects which I am currently missing.

johanneskoester:
> A note regarding conda environments:
> 
> I believe there is currently a [bug](https://github.com/kpj/snakemake/blob/f8a2f0b20c9726489f2b7ca7db791560696ee835/snakemake/shell.py#L99), because `shell(..)` makes use of `context.get("conda_env", None)` which does currently always return `None`.
> It should be enough to simply replace it with `context["self"].conda_env` (seems to work for me), but there might be more side-effects which I am currently missing.

Are you sure you ran snakemake with --use-conda? Conda support is rock solid and used all the time by me and others.

johanneskoester:
> A few (possibly) last things to consider:
> 
>     * Storing the executed notebook is a good idea, I think. Currently, the output file `notebook_output` is used to specify the destination path. Is this the best solution?

To me, this is rather something like a log file. Hence, I suggest to request it like this
```
    log:
        notebook="logs/theexecutednotebook.ipynb"
```
of course with any more reasonable filename.

>     * Adding the option to automatically convert the executed notebook to another format (pdf, html, ...) is not needed (so I should remove the according code).

Agreed.

>     * The preamble is currently being removed before saving the notebook output. For now, it would probably be the best to simply leave it in.

Yes, definitely. This way, people can simply re-execute or modify the notebook, e.g., in order to make cosmetic changes to a plot.

>     * `jupyter nbconvert` instead of `jupyter-nbconvert` (note the dash) is currently used to execute the notebooks. The former is only provided by [jupyter_contrib_nbextensions](https://github.com/ipython-contrib/jupyter_contrib_nbextensions#2-install-javascript-and-css-files) (I believe), so we should probably use the latter.

Agreed.

>     * One should probably test how conda/... integration works on machines which are not my own :)

I think it would be best to remove jupyter from the global test environment, and instead have a specific conda env containing jupyter in the test case. Can you do that?

kpj:
> Are you sure you ran snakemake with --use-conda? Conda support is rock solid and used all the time by me and others.

I have the feeling that the script related changes might have broken something.

Consider the following files:
```python
#Snakefile
rule all:
    input:
        'foo.txt'

rule script_rule:
    output:
        'foo.txt'
    conda:
        'env.yaml'
    script:
        'script.py'
```
```yaml
#env.yaml
channels:
  - default
dependencies:
  - python>=3.6
```
```python
#script.py
with open(snakemake.output[0], 'w') as fd:
    fd.write('foo')
```

Running `snakemake --use-conda` will not activate the conda environment, because `context.get("conda_env", None)` returns `None`.

I am not entirely sure how to debug this, since I don't completely understand `inspect.currentframe().f_back.f_locals` ([ref](https://github.com/snakemake/snakemake/blob/97b01802b7afd508dc7b924eb54e1d753d71f6c0/snakemake/shell.py#L88)). Maybe jumping to the next frame does not work anymore, since we are now using classes?

snakemake/script.py
```diff
@@ -176,71 +259,510 @@ def __init__(
...
+                        "script. Note that this cannot be avoided, "
+                        "because the script uses data structures from "
+                        "Snakemake which are Python >={0}.{1} "
+                        "only.".format(*MIN_PY_VERSION)
+                    )
+        if self.singularity_img is not None:
+            # use python from image
+            py_exec = "python"
+        # use the same Python as the running process or the one from the environment
+        shell("{py_exec} {fname:q}", bench_record=self.bench_record)
```
johanneskoester: Ok, I see the problem that manifests with conda_env for you. Here, `conda_env` and `singularity_img` must be available as local variables (this is where shell searches them).

johanneskoester: I guess it makes sense to have a separate method in the ABC for executing the command, which handles this for all inheriting classes.

johanneskoester: The following will work once you have merged in the latest master.
```
    def execute_cmd(self, cmd):
        shell("{cmd}, bench_record=self.bench_record, conda_env=self.conda_env, singularity_img=self.singularity_img, shadow_dir=self.shadow_dir, env_modules=self.env_modules)

Comment (johanneskoester) [reply snakemake/script.py]: Ok, master is merged in. Let's see.

Comment (kpj) [reply snakemake/script.py]: Yes, this generally makes sense.
However, if we introduce the function like this:
```python
    def _execute_cmd(self, cmd, **kwargs):
        shell(
            "{cmd}",
            bench_record=self.bench_record,
            conda_env=self.conda_env,
            singularity_img=self.singularity_img,
            shadow_dir=self.shadow_dir,
            env_modules=self.env_modules,
            **kwargs)
```
then this method will no longer have e.g. `py_exec` as its local variable (since we are one function call deeper).

kpj: I now implemented it like this:
```python
    def _execute_cmd(self, cmd, **kwargs):
        context = inspect.currentframe().f_back.f_locals
        context.update(kwargs)

        shell(
            cmd,
            bench_record=self.bench_record,
            conda_env=self.conda_env,
            singularity_img=self.singularity_img,
            shadow_dir=self.shadow_dir,
            env_modules=self.env_modules,
            **context)
```
(see 9579090ea8f62b06679b8cbb3cdaf9738bc3accc)

It seems to work but is a bit convoluted.

johanneskoester: I don't think we need context here. You should be able to simply pass py_exec to _execute_cmd as a keyword argument, shouldn't you?

johanneskoester:
Can you do a `black snakemake` on your branch?

snakemake/script.py
kpj: That will certainly also work.

It would then be necessary to pass all used variables to `self._execute_cmd(..)`. See e.g.:
```python
self._execute_cmd("{py_exec} {fname:q}", py_exec=py_exec, fname=fname)
# instead of
self._execute_cmd("{py_exec} {fname:q}")
```

This would be fine for me (I think I even prefer it, because it is more explicit).
So far I was mainly trying to replicate the way `shell(..)` works.

kpj:
The formatting checks now pass.

The code tests fail because the `nbformat` module is not found.
Is it necessary to add it to `test-environment.yml` even though it is contained in `setup.py`'s `install_requires`?

snakemake/script.py
johanneskoester: Exactly, I prefer it for the same reason. Shell just works like that for convenience when writing a workflow. For internal use, it should always be explicit over implcit.

johanneskoester:
> The formatting checks now pass.
> 
> The code tests fail because the `nbformat` module is not found.
> Is it necessary to add it to `test-environment.yml` even though it is contained in `setup.py`'s `install_requires`?

I think it should be part of a conda environment in the test case rule. It does not need to be a snakemake dependency.

johanneskoester:
> The formatting checks now pass.
> 
> The code tests fail because the `nbformat` module is not found.
> Is it necessary to add it to `test-environment.yml` even though it is contained in `setup.py`'s `install_requires`?

You need to add it to `test-environment.yml` and `environment.yml`.

PR #104, opened 2019-11-12
Update snakemake vim

tinyheero:
This PR adds the `wildcard_constraints` directive to the snakemake.vim syntax file so that it gets highlighted in vim.


johanneskoester:
Thank you!

Issue #115, opened 2019-11-18
Printing parameter values


SultanOrazbayev:
**Is your feature request related to a problem? Please describe.**
The rule details print out wildcards, inputs and outputs, but not parameters.

**Describe the solution you'd like**
It would be good if the parameter values are also printed.

cademirch:
Closing due to extended inactivity. If this is still something you'd like to see, please reopen!

PR #121, opened 2019-11-23
printing parameter values

SultanOrazbayev:
This allows printing parameter values in the job information: https://github.com/snakemake/snakemake/issues/115


johanneskoester:
Note that params can also contain complex objects, which cannot be easily printed. Hence, I think this should always be an optional feature. Maybe add a flag --printparams?

SultanOrazbayev:
Will be referring to the approach here: https://github.com/snakemake/snakemake/pull/81

PR #125, opened 2019-11-26
Zenodo remote provider

tpall:
This PR adds Zenodo remote provider to upload and download files from Zenodo <https://zenodo.org>.

- Access token with write scope is required. 
- Uploads files to pre-existing depositions based on deposition id.
- When deposition id is not provided, creates a new deposition.
- Publishing of a deposition + adding all relevant metadata must be done 'manually'.
- Files can be downloaded with this remote provider only from user's own depositions (HTTP remote provider can be used to download files from other public Zenodo uploads). 
- Zenodo sandbox <https://sandbox.zenodo.org> can be used for testing purposes, needs separate sandbox account and access token.
- Uploads are limited to 100MB with current API (Zenodo says that new API with 50G limit is in the works).

Have been using this remote provider for my project for some months now.


snakemake/remote/zenodo.py
```diff
@@ -0,0 +1,212 @@
...
+from snakemake.remote import AbstractRemoteObject, AbstractRemoteProvider
+from snakemake.exceptions import ZenodoFileException, WorkflowError
+from snakemake.common import lazy_property
+from snakemake.utils import makedirs
+
+try:
+    # Third-party modules
+    import requests
+    from requests.exceptions import HTTPError
+except ImportError as e:
```
johanneskoester: Exception handling is not needed here, requests is already a default dependency of snakemake.

snakemake/remote/zenodo.py
```diff
@@ -0,0 +1,212 @@
...
+    @property
+    def list(self):
+        return [i for i in self._zen.get_files()]
+
+    @property
+    def name(self):
+        return self.local_file()
+
+
+class ZENHelper(object):
```
johanneskoester: Wouldn't it make sense to use one of the zenodo helpers that are already there? For example pyzenodo?

tpall: I considered using zenodo libraries, like Tommos0/pyzenodo and space-physics/pyzenodo3. I am not aware if there are any others. 
The first one didn't had API upload or download implemented, only searching for records which uses another api.
Pyzenodo3 looks very good, has also upload, but no download. 
So it seemed more straightforward to start from scratch and write another wrapper.
Plus, here we have zenodo sandbox for testing.

snakemake/remote/zenodo.py
tpall: Removed exception. Whereas, I just added it based on other remote provider examples:-)

snakemake/remote/zenodo.py
```diff
@@ -99,11 +99,12 @@ def download(self):
...
-            self._zen.upload(self.local_file(), self.remote_file())
-        else:
-            raise ZenodoFileException(
-                "Current Zenodo stable API supports <=100MB per file."
+        with open(self.local_file(), "rb") as lf:
+            self._zen._api_request(
+                self._zen.bucket
+                + "/{}".format(os.path.basename(self.remote_file())),
+                method="PUT",
+                data=lf
```
tpall: No 100Mb limit anymore: updated upload method to new Zenodo files API with a PUT request to a 'bucket' link.

snakemake/remote/zenodo.py
johanneskoester: Thanks! Those might be historical.

tpall:
Fixed missing definition (7167696). Test fails now because of missing access token.

johanneskoester:
The test case fails, can you have a look?

tpall:
> The test case fails, can you have a look?

401 Client Error: UNAUTHORIZED for url: https://sandbox.zenodo.org/api/deposit/depositions

Zenodo sandbox account (https://sandbox.zenodo.org) and access token with write scope is necessary to run this test.

Test run creates a new deposition (as no deposition=id is provided) in sandbox.zenodo and then uploads and downloads test file. Whereas each test run creates one new deposition in zenodo sandbox. The sandbox environment can be cleaned by zenodo at anytime, so it's not a big problem.

tpall:
Zenodo says that their UI and REST API responses were designed with having in mind uploads of a total of 20-30 files. Should this be mentioned in documentation?

johanneskoester:
> Zenodo says that their UI and REST API responses were designed with having in mind uploads of a total of 20-30 files. Should this be mentioned in documentation?

Yes, please.

johanneskoester:
Finalizing this inside of an upstream branch, since I don't know whether the tests will properly work in a fork.

PR #131, opened 2019-11-30
Stata support

SultanOrazbayev:
This introduces integration with Stata (https://www.stata.com/). Documentation and tests are to be added.

### DIFF
--- a/snakemake/script.py
+++ b/snakemake/script.py
@@ -136,7 +136,7 @@ class StataEncoder:
     def encode_items(cls, items):
         def encode_item(item):
             name, value = item
-            return 'string {} = {}'.format(name, cls.encode_value(value))
+            return 'string {} = {}'.format(name, '"'+cls.encode_value(value).strip("'")+'"')
         return "\n".join(map(encode_item, items))
 
     @classmethod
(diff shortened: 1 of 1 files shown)


johanneskoester:
Thank you! This is indeed a good idea. How would we test it though? As far as I know, stata is not available as FOSS?
Apart from that, this PR has to wait on PR #101 because in there, we are currently refactoring and cleaning up script support.

SultanOrazbayev:
Would the following work? Running a python-based test that verifies Stata syntax based on pre-defined rules rather than actually calling Stata (which is not available for free). 

Another alternative is to write a Stata-specific program, but also not sure how well the Python integration will work (in the latest version).

Still need to work on the documentation and testing, but will wait for PR#101.

johanneskoester:
The PR is merged, but I still think that it is not a good idea to integrate closed-source software like this. We will never be able to automatically test if it still works in the future. Unless stata provides us with some way to really execute a test in CI, I don't see a way here.

johanneskoester:
Closing because of above reason. Please feel free to reopen if the situation changes.

SultanOrazbayev:
Perhaps the best way forward is to make a separate repo that adds the stata functionality as a separate 'plug-in', not unlike the https://github.com/mbhall88/snakefmt.

PR #138, opened 2019-12-04
Move format of shell prefix and suffix to cmd context

SamStudio8:
Moves the formatting of shell prefix and suffixes to the context in which the command is submitted by the shell. This change permits generic expansion of snakemake variables that pertain to the current job (such as {jobid}), rather than only variables that are in frame when the shell is first initialised.


snakemake/shell.py
```diff
@@ -101,7 +101,9 @@ def __new__(
         shadow_dir = context.get("shadow_dir", None)
 
         cmd = "{} {} {}".format(
-            cls._process_prefix, cmd.strip(), cls._process_suffix
+            format(cls._process_prefix, stepout=2),
```
SamStudio8: I should probably add the args and kwargs here?
```
format(cls._process_prefix, *args, stepout=2, **kwargs),
```

johanneskoester: Yes, please define a lambda above, which you then use for cmd, prefix and suffix.

johanneskoester:
Please also fix formatting by running black on your branch: `black snakemake`

snakemake/shell.py
SamStudio8: 53472922a5664ff8748843ff52c5a8f6ff841f09

SamStudio8:
I'm not sure why the formatting failed on my latest commit as `black snakemake` gives me a slice of cake and gold stars on my local copy now.

SamStudio8:
@johanneskoester Sorry, I commented on the wrong half of the diff. However, I've noticed an existing issue where calling `shell.executable` and `shell.prefix` leads to different behaviour based on the order in which they are called if the executable is `bash`.

Calling `shell.executable('bash')` sets the prefix to `set -euo pipefail`, which is then overwritten when a user calls `shell.prefix`. Setting the prefix before setting the shell would replace the user provided prefix with `set -euo pipefail`. I don't think these should lead to different behaviours so I propose we either:

* always ignore the default `set -euo pipefail` prefix when a user provides any prefix, which can be done by only writing the prefix when calling `shell.executable` if the prefix is already empty (*i.e.* has not already been set by a prior call to `shell.prefix`)
* always ensure the default `set -euo pipefail` prefix is included by changing the `shell.prefix` command to use `+=` instead of `=`

I think I would lean towards the former as I would expect that a user who is overriding the shell prefix may anticipate that they would also be overriding any defaults. This is potentially outside the scope of this PR but let me know what you think.

SamStudio8:
@johanneskoester Just checking in on this

SamStudio8:
> @johanneskoester However, I've noticed an existing issue where calling `shell.executable` and `shell.prefix` leads to different behaviour based on the order in which they are called if the executable is `bash`.

Any thoughts on this @johanneskoester?

johanneskoester:
> @johanneskoester Sorry, I commented on the wrong half of the diff. However, I've noticed an existing issue where calling `shell.executable` and `shell.prefix` leads to different behaviour based on the order in which they are called if the executable is `bash`.
> 
> Calling `shell.executable('bash')` sets the prefix to `set -euo pipefail`, which is then overwritten when a user calls `shell.prefix`. Setting the prefix before setting the shell would replace the user provided prefix with `set -euo pipefail`. I don't think these should lead to different behaviours so I propose we either:
> 
>     * always ignore the default `set -euo pipefail` prefix when a user provides any prefix, which can be done by only writing the prefix when calling `shell.executable` if the prefix is already empty (_i.e._ has not already been set by a prior call to `shell.prefix`)
> 
>     * always ensure the default `set -euo pipefail` prefix is included by changing the `shell.prefix` command to use `+=` instead of `=`
> 
> 
> I think I would lean towards the former as I would expect that a user who is overriding the shell prefix may anticipate that they would also be overriding any defaults. This is potentially outside the scope of this PR but let me know what you think.

I prefer the latter. bash should always imply pipefail. Can you modify this PR to change the behavior to `+=`?

SamStudio8:
@johanneskoester Thanks, I've done that now. This should ensure that pipefail will be set, regardless of whether you set the prefix, or the shell first.

SamStudio8:
@johanneskoester I've swapped the lambda back out, as I couldn't seem to pass the kwargs from the parent scope correctly... Perhaps your python-fu can help me here! This has fixed the errors that were occurring earlier, but now it seems to bug out on one of the singularity tests.

SamStudio8:
> @johanneskoester I've swapped the lambda back out, as I couldn't seem to pass the kwargs from the parent scope correctly... Perhaps your python-fu can help me here! This has fixed the errors that were occurring earlier, but now it seems to bug out on one of the singularity tests.

I've realised my error: the lambda was fine, I needed to increment the `stepout` value to allow `util.format` to access the right frame!

johanneskoester:
Don't forget to format with black :-)

SamStudio8:
> Don't forget to format with black :-)

I'm not sure why that's still failing, as I did run `black`:

```reformatted /home/sam/git/snakemake/snakemake/shell.py
All done! ✨ 🍰 ✨
1 file reformatted, 52 files left unchanged.```

**Comment by SamStudio8** [comment]
Aha, it's not me:
```
Run export PATH="/usr/share/miniconda/bin:$PATH"
would reformat /home/runner/work/snakemake/snakemake/snakemake/exceptions.py
Oh no! 💥 💔 💥
1 file would be reformatted, 60 files would be left unchanged.
##[error]Process completed with exit code 1.
```

**Comment by SamStudio8** [comment]
```
subprocess.CalledProcessError: Command ' singularity exec --home /tmp/snakemake-sr9e1cgh  --bind /home/runner/work/snakemake/snakemake:/mnt/snakemake /tmp/snakemake-sr9e1cgh/.snakemake/singularity/36b22e49e8a03fd08160e9345dd1034e.simg bash -c 'set -euo pipefail; source ~/.bashrc;  ls test.txt; echo "test" > test.out '' returned non-zero exit status 1.
```
This error would not have been caught before, because setting the prefix to `source ~/.bashrc` would have previously overwritten setting pipefail.

It seems `.bashrc` does not exist in the container:

```
docker run bash "source" "~/.bashrc"
/usr/local/bin/docker-entrypoint.sh: line 11: exec: source: not found
```

The prefix is set in `test01/Snakefile`, and I guess this carries through the rest of the tests:
```
test01/Snakefile:6:shell.prefix("source ~/.bashrc; ")
```
Should we set the `prefix` in `test_singularity/Snakefile` to an empty string to override?

**Comment by SamStudio8** [comment]
The prefix is set in `test01/Snakefile`, and I guess this carries through the rest of the tests:
```
test01/Snakefile:6:shell.prefix("source ~/.bashrc; ")
```
@johanneskoester Should we set the `prefix` in `test_singularity/Snakefile` to an empty string to override? (How would we do this without carrying this to the other tests?)

Issue #12, opened 2019-10-08
Customizing snakemake console output


ekiefl:
Hello!

Over at anvi'o we are using snakemake extensively to automate our workflows and it works gorgeously. Imo anvi'o's ease of use can really be demarcated by "before snakemake" and "after snakemake". So thank you so much for all of your commitment to this project!

One thing I was interested in doing is capturing all of the logging info and processing it in a customized manner, so that I can tailor exactly what is sent to the console. I'm probably thinking about this the wrong way, but I was imagining perhaps handling this by creating a custom `Logger.text_handler` method found in `logging.py`. Anyways, I'm very interested to see your perspective one whether snakemake is set up to handle client-side customization of the console output.

Thanks so much for your time.

Cheers,
Evan

johanneskoester:
Hi,
glad to hear that. Indeed, settting up custom loggers is exactly meant for things like that. The [main Snakemake entry point](https://snakemake.readthedocs.io/en/stable/api_reference/snakemake.html) has a parameter `log_handler`, where you can pass a custom logger. And indeed, a custom terminal logger would inherit from [snakemake.logging.Logger](https://snakemake.readthedocs.io/en/stable/api_reference/internal/snakemake.html#snakemake.logging.Logger) and overwrite the method `text_handler`.

ekiefl:
@johanneskoester, thanks for the quick response.

Our codebase calls snakemake with the following code:

```
        sys.argv = ['snakemake',
                    '--snakefile',
                    get_workflow_snake_file_path(self.args.workflow),
                    '--configfile',
                    self.args.config_file]

        if self.additional_params:
            sys.argv.extend(self.additional_params)

        if self.list_dependencies:
            sys.argv.extend(['--dryrun', '--printshellcmds'])
            snakemake.main()
            sys.exit(0)
        else:
            sys.argv.extend(['-p'])
            snakemake.main()
```

I notice `snakemake.main()` calls `snakemake.snakemake(...)` without a `log_handler` parameter. Am I correct that running the snakefile through `main()` limits us from providing a `log_handler` parameter?

ekiefl:
@johanneskoester, As a workaround, I have done something very ugly. I redefined snakemake's `snakemake.main()` function in our own code base, adding the following change to the call of `snakemake.snakemake()`:

```
-cluster_status=args.cluster_status)
+cluster_status=args.cluster_status,
+log_handler=AnvioSnakeMakeLogger.anvio_logger)
```

[...]

PR #141, opened 2019-12-05
add python_args to main(...) so that software can use all  arguments in snakemake(...)

ekiefl:
This very small change provides access to all arguments in `snakemake` when `snakemake` is called through `main(...)`.

A detailed example of why this is useful as provided in this issue: https://github.com/snakemake/snakemake/issues/12. To summarize, the user calls `main` programmatically and wants to supply a custom logger function, but cannot because the `snakemake` call inside `main` has no exposed `log_handler` argument. The user calls `main` instead of `snakemake` directly because of the extensive argument parsing provided in `main`.


ekiefl:
I actually think probably the best example of a log function is the default, called `text_handler`: https://snakemake.readthedocs.io/en/stable/_modules/snakemake/logging.html#Logger.text_handler. I don't propose using the logger function I wrote, because it's very specific to the codebase I wrote it for: https://github.com/merenlab/anvio/blob/snakemake_logger/anvio/workflows/__init__.py

And the best information I found to learn about the message dictionary passed to that function was in the snakemake api file: https://snakemake.readthedocs.io/en/stable/api_reference/snakemake.html

---

So how can this pull request be improved?

I think it would be helpful in the docstring of [text_handler](https://snakemake.readthedocs.io/en/stable/_modules/snakemake/logging.html#Logger.text_handler) to have a full description of the formatting of the input dictionary (just the way that it is described in the parameter specifications of the [snakemake api file](https://snakemake.readthedocs.io/en/stable/api_reference/snakemake.html)).

In a similar fashion, I think it would be useful in the snakemake api parameter specifications if it was explained that the log handler parameter overwrites `Logger.text_handler`,  which will make it clear to users where to go looking for an example logger function.

I'll add these documentation changes tomorrow if I don't hear back from you. Cheers

ekiefl:
Okay I made the changes. Let me know what you think :)

johanneskoester:
Why would you call main(), if you could just call snakemake() instead?

ekiefl:
main() has a lot of checks and argument parsing

johanneskoester:
And what is the use case? I can understand that one calls `snakemake` if one wants to run it e.g. from a web server. But this seems to be different.

ekiefl:
Well, maybe we are doing things wrong on our end. Let me explain the use case.

anvi'o has a suite of command line programs that are all prefixed with `anvi-`. One of them is `anvi-run-workflow`, which eventually ends up running `snakemake.main`. `anvi-run-workflow` takes non-snakemake parameters first, and then if the user wants to pass additional flags to be processed by snakemake, they are put after `--additional-params`. For example, here is a potential command:

```
anvi-run-workflow --workflow metagenomics --additional-params --jobs 20 --cluster "clusterize -j={rule}"
```

When run, this boils down to running snakemake in this method:

```python
    def go(self, skip_dry_run=False):
        """Do the actual running"""

        self.sanity_checks()

        if self.save_workflow_graph or (not skip_dry_run):
            self.dry_run()

        if self.dry_run_only:
            return

        # snakemake.main() accepts an `argv` parameter, but then the code has mixed responses to
        # that, and at places continues to read from sys.argv in a hardcoded manner. so we have to
        # overwrite our argv here.
        original_sys_argv = copy.deepcopy(sys.argv)

        sys.argv = ['snakemake',
                    '--snakefile',
                    get_workflow_snake_file_path(self.args.workflow),
                    '--configfile',
                    self.args.config_file]

        if self.additional_params:
            sys.argv.extend(self.additional_params)

        if self.list_dependencies:
            sys.argv.extend(['--dryrun', '--printshellcmds'])
            snakemake.main()
            sys.exit(0)
        else:
            sys.argv.extend(['-p'])
            snakemake.main()

        # restore the `sys.argv` to the original for the sake of sakity (totally made up word,
        # but you already know what it means. you're welcome.)
        sys.argv = original_sys_argv
```
 
(For the record, I do not know what the first comment is talking about but maybe @ShaiberAlon knows).

It seems like we could call `snakemake.snakemake` from here, but then we would have to replicate the argument processing and sanity check for each of the parameters that are already beautifully processed by `snakemake.main`. We would also have to keep our codebase up to date with snakemake whenever you guys change `snakemake.main`'s behavior.

This is essentially why this behavior was requested. Please let me know if you have questions or suggestions.

ShaiberAlon:
>(For the record, I do not know what the first comment is talking about but maybe @ShaiberAlon knows).

That comment was actually written by @meren :-)

johanneskoester:
Ok I see. Thanks for explaining. In that case, a true fix would be to make the CLI itself extensible. The best solution IMHO is to turn the main function into a class, with a separate method for obtaining the argparser. Then, you can simply inherit that class and add args to the argparser.

ekiefl:
I'm not sure I completely understand, but is this kind of what you mean?

```python
def main(argv=None):
    cli = CLI(argv)
    cli.parse_args()
    cli.run()


class CLI(object):
    def __init__(self, argv):
        self.argv = argv

    def parse_args(self):
        parser = get_argument_parser()
        args = parser.parse_args(self.argv)

        if args.profile:
            # reparse args while inferring config file from profile
            parser = get_argument_parser(args.profile)
            args = parser.parse_args(self.argv)

        ...
        ...
        ...
        ...

        if args.snakefile is None:
            for p in SNAKEFILE_CHOICES:
                if os.path.exists(p):
                    args.snakefile = p
                    break
            if args.snakefile is None:
                print(
                    "Error: no Snakefile found, tried {}.".format(
                        ", ".join(SNAKEFILE_CHOICES), file=sys.stderr
                    )
                )
                sys.exit(1)


    def run(self):
        if args.gui is not None:
            ...
            ...
            ...

        else:
            success = snakemake(
                args.snakefile,
                batch=batch,
                cache=args.cache,
                ...
                ...
                ...
            )

        sys.exit(0 if success else 1)
```

johanneskoester:
Yes, exactly. Feel free to make it as finegrained as needed in order to extend it.

johanneskoester:
Closing this in favor of a future Cli class as outlined above. Looking forward to that PR!

Issue #143, opened 2019-12-06
Namedlist set_name overrides class functions ( for example --report  crashes if a params: variabel is called 'items')


Gutorm:
**Snakemake version**
5.7.4 
**Describe the bug**
`snakemake --report` will crash if one of the `params:` is called "items" (see example below)
(snakemake workflow execution has no problem with this, only `--report`)

This is probably due to `items()` being a method of python 

**Logs**
(Corresponds to example below)
```
snakemake  --report
Building DAG of jobs...
Creating report...
Adding foo.rst (0 MB).
Could not detect mimetype for foo.rst, assuming text/plain.
Traceback (most recent call last):
  File "/mnt/work/Qc.py375.venv/lib/python3.7/site-packages/snakemake/__init__.py", line 618, in snakemake
    batch=batch,
  File "/mnt/work/Qc.py375.venv/lib/python3.7/site-packages/snakemake/workflow.py", line 617, in execute
    auto_report(dag, report)
  File "/mnt/work/Qc.py375.venv/lib/python3.7/site-packages/snakemake/report/__init__.py", line 552, in auto_report
    FileRecord(f, job, report_obj.caption, env, category)
  File "/mnt/work/Qc.py375.venv/lib/python3.7/site-packages/snakemake/report/__init__.py", line 326, in __init__
    self.params = logging.format_dict(job.params)
  File "/mnt/work/Qc.py375.venv/lib/python3.7/site-packages/snakemake/logging.py", line 384, in format_dict
    for name, value in dict.items()
TypeError: 'str' object is not callable
```

**Minimal example**

```python
rule test:
  output: report(touch("foo.rst"))
  params: items = "hello darkness"
  run:
      print (params[0])
```
 

**Additional context**
Note that `  params: foo = "hello darkness"` will work.

)

troycomi:
At first I was thinking, "why would you override a builtin function?" but the actual cause is more of a problem.  The params, along with most directives, are stored as a Namedlist, defined in [io.py](https://github.com/snakemake/snakemake/blob/master/snakemake/io.py#L1120).  When a value is [set](https://github.com/snakemake/snakemake/blob/master/snakemake/io.py#L1170), it sets the attribute for the name provided ('items') overriding any functions.  You can also break this with params like 'set_name' or resources called 'items', which causes crashes on execution.  Oddly setting an output file as items="foo.txt" doesn't cause the same crashes.

[...]

PR #149, opened 2019-12-11
change Namedlist attributes into private naming schema (_*).

johanneskoester:
Provides a fix to issue #143 that does not break existing workflows
and leaves more flexibility when choosing attribute names.


johanneskoester:
Hi @troycomi I know. However, we need the ability to set attributes named as e.g. sort, for backwards compatibility (see e.g. the bwa mem wrapper). Also, it would be really inconvenient if all those names are suddenly unavailable. I plan to carefully watch this, such that nothing allowed is used internally.

PR #159, opened 2019-12-18
Environment modules support

johanneskoester:
This PR provides an initial draft for environment module support. This should help with Snakemake adoption in HPC environments. 
Rules can be annotated with modules, which will be loaded in the given order, e.g.:

```
rule bwa:
    input:
        "genome.fa"
        "reads.fq"
    output:
        "mapped.bam"
    envmodules:
        "bio/bwa/0.7.9",
        "bio/samtools/1.9"
    shell:
        "bwa mem {input} | samtools view -Sbh - > {output}"
```

The downside is that modules may be not available outside a specific HPC environment, or are named the same but contain something else. Hence, using `envmodules:` does not guarantee reproducibility like using `conda:` or `singularity:`.

### Todo

* [ ] add docs


johanneskoester:
cc @cmeesters

johanneskoester:
@cmeesters while implementing, I came to the conclusion that trying to find suitable modules from a given conda env will never reliably work in general (see our scipy discussion). Rather than having a half-working solution, I figured that it is better to provide full env module support, with the notice in the docs that one should always complement it with an optional conda env to ensure full reproducibility.
So, people could write now

```
rule bwa:
    input:
        "genome.fa"
        "reads.fq"
    output:
        "mapped.bam"
    conda:
        "envs/bwa.yaml"
    envmodules:
        "bio/bwa/0.7.9",
        "bio/samtools/1.9"
    shell:
        "bwa mem {input} | samtools view -Sbh - > {output}"
```

In their HPC environment, they can issue `snakemake --use-envmodules`. For reproduction outside of their HPC environmnent, they can issue `snakemake --use-conda`, to get the same result.

PR #161, opened 2019-12-19
add option for log handler script

christopher-schroeder:
Add a log handler script argument for custom log handling


GuillaumeDerval:
Here is an example of script that pushes the progress to slack: https://gist.github.com/GuillaumeDerval/ee6d20c94448236dcca2e6bd8d47ca0f

PR #162, opened 2019-12-19
Add --create-envs-only example to deployment.rst

colinbrislawn:
One of colleagues had trouble finding the `--create-envs-only` flag, so I've added an example to the Distribution and Reproducibility page.

Feedback welcome!


johanneskoester:
Awesome, thank you!

Issue #156, opened 2019-12-15
Access the rule name within a rule


a-slide:
**Is your feature request related to a problem? Please describe.**

At the moment it is only possible to access the rule name inside a shell command using `{rule}`.

**Describe the solution you'd like**

It would be great to be able to access it from any keywords inside a rule, in particular input, output and log. For example it could be stored in a variable named `rule.name`.

**Additional context**

This has already been raised for the log keyword when Snakenake was on Bit-Bucket https://bitbucket.org/snakemake/snakemake/issues/199/rule-name-cant-be-accessed-by-rule-in

Gutorm:
I think this would be a good thing, but it might be that the parsing algorithms make it difficult (since `rule` might is not be established yet). 
But this request reminded me of a workaround I use, see https://stackoverflow.com/questions/40975993/current-rules-name-in-snakemake/59172511 .

a-slide:
I have use the same hack as well, but it feels a bit "wrong".

m-jahn:
Currently, the rule name can be accessed in shell but not in other rule directives. Nevertheless closing this issue now as there seems to be no urgent need or activity for this request. Feel free to ask for re-opening if this becomes urgent again.

PR #175, opened 2020-01-05
Enhancement: Allowing to write rule decorator

smatsumt:
# What's this PR?

This PR allows to write **decorator wrapping a rule** in Snakefile.

This is example of decorator in Snakefile. We can enable remote file access just putting `@s3_remote_wrapper` decorator.
```
@s3_remote_wrapper
rule rule1:
    input: 'some-bucket-name/input.txt'
    output: 'some-bucket-name/output.txt'
    shell: 'grep foo {input} > {output}'
```

Decorator implementation is like this.

```
from snakemake.remote.S3 import RemoteProvider as S3RemoteProvider
S3 = S3RemoteProvider()

def s3_remote_wrapper(ruleinfo):
    # wrap S3.remote() with all args and kwargs of input/output
    ruleinfo.input = (tuple(map(S3.remote, ruleinfo.input[0])), {k: S3.remote(v) for k, v in ruleinfo.input[1]})
    ruleinfo.output = (tuple(map(S3.remote, ruleinfo.output[0])), {k: S3.remote(v) for k, v in ruleinfo.output[1]})
    return ruleinfo
```

# What's the benefit?

- **Keep rule description simple.** Just putting one decorator is simpler than writing `S3.remote()` call for all inputs and outputs.
- **Control wrap functions easily.** The decorator can be easily disabled with just uncomment it. In the example above, we can easily change using remote file or not.
- **Enables to write ruleinfo based behavior.** For example, we can write output filename based on rule name using the ruleinfo argument. This might solve the Issue #156, "Access the rule name within a rule". 

# (Current) Limitation

1. `shellcmd`, `wrapper` and `script` cannot be overwritten. `ruleinfo.shellcmd` is just information, not actual executed command. Assigning `ruleinfo.shellcmd` with new value does not change actual executed command.
2. Getting rulename from ruleinfo is not straight forward. `ruleinfo.func.__name__[7:]`will return rule name.
3. Have to know ruleinfo structure details when writing wrapper.

About limitation 1. and 2., I'm planning to write PR for them if this decorator PR is accepted.

### DIFF
--- a/snakemake/parser.py
+++ b/snakemake/parser.py
@@ -804,7 +804,7 @@ class Snakefile:
 def format_tokens(tokens):
     t_ = None
     for t in tokens:
-        if t_ and not t.isspace() and not t_.isspace():
+        if t_ and (not t.isspace() and not t_.isspace()) and (not t_ == '@'):
             yield " "
         yield t
         t_ = t
--- a/snakemake/parser.py
+++ b/snakemake/parser.py
@@ -804,7 +804,7 @@ class Snakefile:
 def format_tokens(tokens):
     t_ = None
     for t in tokens:
-        if t_ and (not t.isspace() and not t_.isspace()) and (not t_ == '@'):
+        if t_ and (not t.isspace() and not t_.isspace()) and (not t_ == "@"):
             yield " "
         yield t
         t_ = t
--- a/snakemake/workflow.py
+++ b/snakemake/workflow.py
@@ -918,7 +918,7 @@ class Workflow:
                 continue
             # find @workflow.rule decorator, and pull it to the first decorator
             try:
-                rule_decorator_t = next(filter(lambda x: getattr(x[1].func, 'attr', '') == 'rule', enumerate(ast_elm.decorator_list)))
+                rule_decorator_t = next(filter(lambda x: getattr(getattr(x[1], 'func', ''), 'attr', '') == 'rule', enumerate(ast_elm.decorator_list)))
             except Exception as e:
                 logger.error(str(e))
                 continue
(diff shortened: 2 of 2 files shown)


snakemake/parser.py
```diff
@@ -804,7 +804,7 @@ def __exit__(self, *args):
 def format_tokens(tokens):
     t_ = None
     for t in tokens:
-        if t_ and not t.isspace() and not t_.isspace():
+        if t_ and (not t.isspace() and not t_.isspace()) and (not t_ == '@'):
```
smatsumt: This change is needed to treat decorator correctly.
Before this change, parser changes "@decorator" to "@ decorator", and Python interpreter does not treat it as decorator.

johanneskoester:
Thanks for the suggestion! At the moment, I don't see a use case yet. To shift all IO to e.g. S3, you can simply use --default-remote-provider at the command line. This is preferable, because it keeps platform specific choices out of the workflow definition making it more reproducible for others. For partially having S3 IO, the current way of expressing it is more readable and explicit to non-python experts. Since readability is more important than brevity, I would like to stick with what we have now.

I am always feeling bad about rejecting work, so please accept my deepest apologies. But I also feel responsible to ensure that the snakemake language does not become too heterogeneous and stays readable for everybody.

johanneskoester:
If you disagree or have further arguments, please feel free to open an issue and we can go on with the discussion over there.

smatsumt:
Thank you very much for kind and thoughtful reply.

I agree with that decorator might damage workflow readability. Especially it is true for unfamiliar people. And I understand that "readability is more important than brevity" is important policy for snakemake.

I didn't know that the option `--default-remote-provider` can use for it. Absolutely, it will solve the example usage situation 😃

For me, snakemake is the best workflow engine (that's why I sent the PR 😁 ), I feel the command line options are a little bit too much. I sometimes feel it is hard to find useful options. I guess this comes from that command line options includes both workflow behavior changing options (--force-all, etc) and each task behavior changing option (--default-remote-provider, etc)

I'm thinking second group options might be separated somehow, and making decorator might mitigate it. Actually, I'm not sure decorator is the best solution for it...

Anyway, thanks again for your reply!

PR #180, opened 2020-01-07
Try enabling tibanna tests again

johanneskoester:


johanneskoester:
@SooLee hooray! It works again! There was an interference with the moto based mocked S3 tests.

Issue #144, opened 2019-12-06
`conda activate` instead of `source activate` in documentation


sappjw:
In the [Testing Guidelines](https://snakemake.readthedocs.io/en/stable/project_info/contributing.html#testing-guidelines) documentation section it suggests

    source activate snakemake-testing

This should now be

    conda activate snakemake-testing

as of conda 4.4.0 ([for reference](https://docs.conda.io/projects/conda/en/latest/release-notes.html#id189)).

PR #189, opened 2020-01-10
Use repr() to quote strings for embedding into Python

### LINKED ISSUE #85: "SyntaxError" for the Snakefile located in the directory which name cntains an apostrophe symbol
Environment: Snakemake 5.4.5, Python 3.6.5, Windows 10, MinGW.

Way to reproduce: take a fully valid project and put into a directory that has an apostrophe in it’s name, e.g. “I’ll not work“, try to run. The Snakemake raises an SyntaxError “invalid syntax“ on the first rule found. Rename the directory back and remove the apostrophe, e.g. “I will work“, the project works again.

tbooth:
Fixes #85 

While looking at the above issue I spotted several bits of manual quoting logic including an `escape_backslash()` utility function that seemed to be better served with `repr()`, or equivalently `{!r}.format()`. I also made an edit to the docs for how to run the Nose tests as they seemed to be out of date. Hopefully this code is both clearer and more correcter.

I didn't run the full test suite yet but I'll let the CI system do that.


tbooth:
As a bonus, fixes #144

PR #203, opened 2020-01-23
Fix envmod

cmeesters:
module files come in different flavours: with or without system dependencies and particularly build with different compilers. Hence, to avoid conflicts the need to purge all previously loaded environment modules for local rules or to clean polluted environments for cluster jobscripts arises. 

This fix should solve it.


cmeesters:
not sure how to interpret this CI failure - merely a timeout issue? Or else?

johanneskoester:
@cmeesters the {cmd} was missing, so it did not execute anything. I have fixed this in a slightly different syntax now in commit 07cc089.

PR #205, opened 2020-01-24
Added support for Azure storage authentication via SAS

andreas-wilm:
This is a safer and more flexible alternative to using an account key (which is still supported). See official [SAS documentation](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview) and [Sebastian's original PR](https://bitbucket.org/snakemake/snakemake/pull-requests/378/azure-integration/diff). Also refer to email exchange with Sebastian from Thu, Oct 10, 2019.

Happy to provide temporary read/write authentication tokens for testing (the one in the email won't work anymore) and refer to tests/test_remote_azure/README.md

### DIFF
--- a/docs/snakefiles/remote_files.rst
+++ b/docs/snakefiles/remote_files.rst
@@ -144,7 +144,8 @@ Microsoft Azure Storage
 
 Usage of the Azure Storage provider is similar to the S3 provider.
 For authentication, one needs to provide an account name and a key
-or SAS token, which can for example be read from environment variables.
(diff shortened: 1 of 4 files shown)


johanneskoester:
I trust that you have tested this. In the long run, I think it might make sense that I enable the free tier and we try to get an infinite extension.

andreas-wilm:
Yes I did. Happy to share a SAS and account name for testing via PM

PR #19, opened 2019-10-10
Fix windows test suite and enable Windows test CI

melund:
I reopening the PR from Bitbucket on fixing the test suite on windows. 

I have enabled Windows CI tests on Github actions.  

Orginal discription from bitbucket: 
---------------------------------------

This PR fixes the test suite for Windows. From a a handful of passed tests to 97 passed and 53 skipped.

This is done by fixing the some actual Windows issues and skipping tests for features which can’t work on Windows. (Many windows fixes where done i separate PRs on bitbucket which have already been merged) 

Most of the tests rely on bash to work, which is a problem on Windows were "cmd.exe" is the default shell. So this PR implements a new option in the shell class to use bash as external executable on Windows. It adds a simi hidden ``shell.use_bash_on_win(bashcmd)`` convenience function to easily configure bash. If `bashcmd` is `None` it will try to locate a bash from either a conda installed git or from system installed "git for windows".

In some tests the expected files on Windows will be different. To handle this, a new option is added which allows for Windows specific expected files in the test setup. If files in the "expected-results" folder are postfix'ed with "_WIN" they will be used instead of the non postfixed files.

It also adds an option to the test-framework to ignore line endings when comparing output files on Windows. Controlling line endings on Windows is a pain.

PR #212, opened 2020-01-31
split tests and make AWS tests optional

johanneskoester:
Such that the test suite also works on forks.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,7 +60,6 @@ jobs:
           pytest -v -x tests/test_expand.py tests/test_io.py tests/test_schema.py tests/tests.py
 
       - name: Configure AWS Credentials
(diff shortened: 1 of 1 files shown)


melund:
@johanneskoester Correct me if am mistaken, but this still shows tests as failing for forks, since the step that configures the aws credentials fails.

See:
https://github.com/snakemake/snakemake/pull/19
and the test failures [here](https://github.com/snakemake/snakemake/pull/19/checks?check_run_id=431913958)

Maybe you need to add an [`continue-on-error`](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) to the step,so the whole job doesn't fail. I don't know what it does to your `if: succes()` statement in the subsequent step.

PR #222, opened 2020-02-10
Enable test suite for windows

melund:
> I recreated this PR from #19 to use a local branch in an attempt to avoid AWS test failures for forks. 

This PR fixes the test suite for Windows. This is done mainly by enabling bash as an alternative to cmd.exe on windows. 

Most tests rely on bash to work, which is a problem on Windows were "cmd.exe" is the default shell. So this PR implements a new option in the shell class to use bash as external executable on Windows. It enables the `shell.executable('bash')` to easily configure bash. A usable bash must exist on PATH for this to work. Note: this not the default case on Windows 10 where the WSL bash in "System32" will be found first. Pytest handles this by prepending bash from conda git client to PATH. 

These fixes has the effect that most test can also run and pass on Windows.  So we can verify that snakemake works as expected. 

In some tests the expected files on Windows will be different. To handle this, a new option for the test framework is added which allows for Windows specific expected files. If files in the "expected-results" subfolder are postfix'ed with "_WIN" they will be used instead of the non postfixed files.

It also adds an option to the test-framework to ignore line endings when comparing output files on Windows. Controlling line endings on Windows is a pain.

FInally it enables CI testing with Github Actions.


melund:
I would have thought the AWS failures would go away now that the PR doesn't come from a fork. But apparently not ...

melund:
@johanneskoester I have fixed the problem with tests failing when the AWS secret is not available. 

I added an environment variable to indicate if AWS is available. 

```yaml
env:
   AWS_AVAILABLE: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
The environment variable is empty when the key is not available. 

The two steps involving AWS are then skipped with:
```yaml
if: env.AWS_AVAILABLE
```

melund:
@johanneskoester: I am glad @samesense could my fixes for the AWS tests.

I have updated this branch with master to remove merge conflicts on the pipeline file.

snakemake/shell.py
```diff
@@ -79,6 +103,20 @@ def cleanup(cls):
         with cls._lock:
             cls._processes.clear()
 
+    @classmethod
+    def use_bash_on_win(cls, bashcmd=None):
```
johanneskoester: I am not yet convinced that it is a good idea to expose this via the public API. Why not just installing bash via conda for the windows tests?

snakemake/shell.py
```diff
@@ -33,16 +33,25 @@ class shell:
...
         return cls._process_args.get("executable", None)
 
     @classmethod
     def check_output(cls, cmd, **kwargs):
-        return sp.check_output(
-            cmd, shell=True, executable=cls.get_executable(), **kwargs
-        )
+        if ON_WINDOWS and cls.get_executable():
+            return sp.check_output(
+                cls._process_prefix + " " + argvquote(cmd),
```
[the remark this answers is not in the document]
melund: No. It is for quoting windows commands. It is a python for  implementation of the c algorithm here: https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

snakemake/io.py
```diff
@@ -13,6 +13,7 @@
 import datetime
 import json
 import copy
+import platform
```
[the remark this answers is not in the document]
melund: True. I will remove it.

snakemake/io.py
```diff
@@ -416,7 +419,7 @@ def prepare(self):
                 if e.errno != 17:
                     raise e
 
-        if is_flagged(self._file, "pipe"):
+        if is_flagged(self._file, "pipe") and not ON_WINDOWS:
```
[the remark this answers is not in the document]
melund: Well. I couldn't make it work, so this was the easy fix. How do I make the pipe flag a no-op on windows?

snakemake/shell.py
```diff
@@ -57,6 +66,12 @@ def executable(cls, cmd):
                 )
         if os.path.split(cmd)[-1] == "bash":
             cls._process_prefix = "set -euo pipefail; "
+        if ON_WINDOWS and cmd.lower().endswith("bash.exe"):
+            # If bash is selected as the executable on Windows
+            # configure _preocess_prefix and win_quote_cmd for
+            # this to work correctly
+            cls._process_prefix = '"{}" -c'.format(cmd)
```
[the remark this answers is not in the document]
melund: The `cls._process_args["executable"]` is passed to Popen as the `executable` parameter. That doesn't do anything on Windows. So if we need to use bash instead of cmd.exe we need to pass use `bash.exe -c ` as part of the command. The `cls._process_prefix` is perfect for that. I still let code set bash.exe into the executable argument so it doesn't differ for from posix.

snakemake/utils.py
```diff
@@ -527,4 +527,31 @@ def argvquote(arg, force=True):
         return cmdline
 
 
+def os_sync():
+    """Ensure flush to disk"""
+    if not ON_WINDOWS:
+        os.sync()
+
+
+def find_bash_on_windows():
```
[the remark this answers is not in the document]
melund: Not sure I follow.  This is just a convenience function for locating a useable bash installation on windows. If we didn't have this every one who wants to run the test suite on windows needs to configure which bash to use depending on their setup. 

We can't just rely on finding `bash.exe` on the path since that defaults to the WSL bash (ubuntu) on new windows 10 system.

snakemake/workflow.py
```diff
@@ -213,7 +220,14 @@ def norm_rule_relpath(f, rule):
...
             f = local_path(f)
             if f:
-                files.add(os.path.relpath(f))
+                if (
+                    ON_WINDOWS
+                    and os.path.splitdrive(f)[0] != os.path.splitdrive(os.getcwd())[0]
+                ):
+                    pass  # relpath doesn't work on win if files are on different drive
+                else:
+                    f = os.path.relpath(f)
```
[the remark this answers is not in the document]
melund: I will fix that.

snakemake/shell.py
melund: I agree that this shouldn't really be in the public API. It is mostly something we need to easily make the tests run on windows. I.e. so snakemake runs all the shell arguments with bash instead of cmd.exe which is default. 

 We could make the method private and then only use it in snakemakes internal tests. An other option I could think of is an argument to snakemake it self and the test system could launch test with that argument set so all the test were run with bash as the shell instead of cmd.exe.  That would also remove the `shell.use_bash_on_win()` from all the test `snakefile`s

melund: > Why not just installing bash via conda for the windows tests?

We still need a way to tell snakemake to bash instead of cmd.exe to interpret shell commands. 

I think I like the idea with a argument to snakemake.

either:
```
snakemake --use-bash-on-windows
```
or a general command line argument which could work for both windows/posix
```
snakemake --shell-executable=<path-to-bash.exe>
```
The test framework could then just set this argument when running test on windows.

snakemake/io.py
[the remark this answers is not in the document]
melund: @johanneskoester I have tried to make it a no-op as you suggested. See if it makes sense.

snakemake/shell.py
johanneskoester: Brainstorming:

We add a directive to the language:
```
shell: "bash"
```
We could restrict possible values to "bash" and "cmd.exe"?

Modify the linter to advise to always add this to a workflow.

And if that is set, [this](https://github.com/snakemake/snakemake/blob/master/snakemake/shell.py#L176) code path will be also taken on windows, thereby failing with a proper error if bash is not in PATH.

melund: @johanneskoester, I like the idea. It is much cleaner. 

The only downside is that "bash.exe" is on path by default in Windows. 

image

But it is a version that can't be used. It is just a link for starting the windows subsystem for linux. It will just give some weird errors for all users, until they realize that the need to set the PATH manually. But maybe that is OK. The main point of all this work is to get as much of the test suite running on windows. 

Could we also allow `shell:` to take a full path? So you do something like this:

```
shell: "path/to/a/good/bash.exe"
```
or 
```
shell: locate_a_good_bash()
```

How do I add a new shell directive to the language? Or could you do that, and then I could rebase and then use the directive?

melund: @johanneskoester I you could add the "shell:" directive in this branch then I will look at getting the whole thing merged with master again.

johanneskoester: Again thinking about this. Snakemake already offers:
```
shell.executable("bash")
```
Which could be done in all the tests, or even in the run function of tests/common.py. Hence, on windows you can simply put your bash finding code into tests/common.py, and set `shell.executable` to whatever full path is determined there. Does that work? This way, we don't expose the method in the regular Snakemake API. 
So, either put something like 
```
if ON_WINDOWS:
    shell.executable(find_bash_on_windows())
```
into the run function of tests/common.py. And also move the function `find_bash_on_windows()` there. This way, we don't need to touch all the tests.

snakemake/workflow.py
```diff
@@ -243,7 +250,13 @@ def norm_rule_relpath(f, rule):
         for f in self.included:
             f = local_path(f)
             if f:
-                files.add(os.path.relpath(f))
+                if not (
+                    ON_WINDOWS
+                    and os.path.splitdrive(f)[0] != os.path.splitdrive(os.getcwd())[0]
+                ):
+                    # relpath doesn't work on win if files are on different drive
+                    f = os.path.relpath(f)
```
johanneskoester, line 258: What does doesn't work mean in that case? Is it returning a wrong path or failing with an exception? In the latter case, I think it would be more pythonic to catch the exception and add special windows handling for it.

snakemake/shell.py
```diff
@@ -58,6 +68,12 @@ def executable(cls, cmd):
...
+            cls.win_quote_cmd(True)
```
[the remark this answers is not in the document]
melund: Nice surprise :) 

But I think we still need the `bash.exe -c` process prefix because `shell=True` only [works for "cmd.exe" on windows](https://github.com/python/cpython/blob/3.8/Lib/subprocess.py#L1294) and can't be used when setting an external executable.

Reading up on how `CreateProcessW` api works on windows I think we also need to start the command line with the process we are starting, e.i `bash.exe -c "<bash command>"`. I think c programs on windows expect argv[0] to still it self. 

From microsoft's [`CreateProcessW`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) API:

> If both lpApplicationName and lpCommandLine are non-NULL, the null-terminated string pointed to by lpApplicationName specifies the module to execute, and the null-terminated string pointed to by lpCommandLine specifies the command line. The new process can use GetCommandLine to retrieve the entire command line. Console processes written in C can use the argc and argv arguments to parse the command line. Because argv[0] is the module name, C programmers generally repeat the module name as the first token in the command line.

`lpApplication` is the `executable` from python and `lpCommandLine` is the command line (`cmd` in python ))

snakemake/shell.py
melund: Yes. I can see how that would work for tests that are started directly by pytest. 

But setting stuff in `common.py` would have no effect on the sub work flows i.e. the inner work flows where snakemake starts it self.

melund: Maybe we could get pytest to modify the Windows `PATH` so the good version of `bash` is found first?

snakemake/workflow.py
melund: True. I will add that.

snakemake/workflow.py
```diff
@@ -243,7 +250,13 @@ def norm_rule_relpath(f, rule):
...
+                files.add(f)
```
melund, line 259: @johanneskoester Will this work instead? 
```suggestion
                try:
                    f = os.path.relpath(f)
                except ValueError:
                    # relpath raise ValueError on Win if files are on different drive, we ignore it
                    if not ON_WINDOWS:
                        raise
                files.add(f)
```

snakemake/shell.py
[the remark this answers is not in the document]
johanneskoester: ok!

snakemake/workflow.py
johanneskoester: Nice!

snakemake/shell.py
johanneskoester: Yes, that's a good idea, can you do that please?

melund: I will. But give me some days. I also need to fix a lot of test regressions.

melund:
@johanneskoester, I have rewritten the way the bash executable is handled on windows in `shell.py`. It now follows the posix code more closely.

All the snakefiles that need bash now has the  `shell.executable("bash")` line and pytest alters the PATH variable on windows so the correct bash is found. 

Thanks for the suggestions. I like this implementation much more. 

The only test I couldn't get working was the jupyter-notebook tests. I couldn't get it working. I think it fails due to some issue with conda. Maybe I can figure out the problem if I start to use the snakemake-conda combo in my daily projects.

I hope this is ready to go in. But let me know if you need more changes :)

.github/workflows/main.yml
```diff
@@ -140,13 +134,20 @@ jobs:
...
+          python-version: 3.8
+          channels: conda-forge, bioconda
+
       - name: Setup Snakemake environment
         run: |
-          conda install -c conda-forge -q mamba
-          mamba env create -q --name snakemake --file test-environment.yml
-        
+          conda install -q --name base mamba
+          mamba env update -q --file test-environment.yml
```
johanneskoester, line 147: Mhm, why putting everything into base instead of creating a separate environment?

melund: It does goes into the snakemake environment. But I agree that it is a little hard to spot. 

I had a lot trouble getting the conda working properly. I ended up using the ["setup-miniconda" gihtub action](https://github.com/marketplace/actions/setup-miniconda). It creates and activates the snakemake environment for all subsequent steps. So it is only mamba that goes into the base environment. The `mamba env update ..` line works on the snakemake environment. 

I hope the action will soon support mamba and then this should become nicer.

melund:
Thanks for merging this @johanneskoester. This was 1.5 year and countless debug sessions to keep snakemake running on Windows. Hopefully, this will make my life easier in the future 🥳 

If I may come with one small suggestion. I don't think you should [squash PR from your contributors](https://thanpol.as/javascript/Dont-Squash-its-rude-and-a-lie) in general. I know the number of commit in my branch got a little out of hand in the end. I should have cleaned it up a bit. But it feels a bit disheartening to see 1.5 year of work get reduced to a single commit. Especially since that is all GitHub uses to measure open source contributions. 

Any way, thanks for an awesome library 👍

PR #227, opened 2020-02-12
nvidia/gpu support in kubernetes executor

joonjeong:
It would be great if some tasks running kubernetes cluster can uses GPU resources.
This commit is just a workaround in this time, but if there's someone who want to use snakemake jobs with GPU in kubernetes cluster, then this feature could be useful for them.

Actually, there's no standard way to allocate GPU to each pod, but many cloud vendors are supporting nvidia GPU to their VM services and nvidia also supports in-container gpu usage scenario.


snakemake/executors.py
```diff
@@ -1492,6 +1492,10 @@ def run(self, job, callback=None, submit_callback=None, error_callback=None):
             container.resources.requests["memory"] = "{}M".format(
                 job.resources["mem_mb"]
             )
+        if "nvidia_gpu" in job.resources.keys():
+            container.resources.limits["nvidia/gpu"] = "%d".format(
```
johanneskoester: I like this, but it needs to be generic. Is there a fixed format for resources names in kubernetes? I.e. always separated by a slash?

joonjeong: Maybe you already pick up on this feature, about naming gpu resources, it depends on GPU vendor. I'm not certain but just saying, this rule started with nvidia. They just named on resource with `nvidia.com/gpu`. And many cloud vendor is now supporting nvidia-gpu-enabled VM. After time passed, recently AMD is also named their gpu resource as `amd.com/gpu`

You can check kubernetes pod manifests with below links.

* https://github.com/NVIDIA/k8s-device-plugin#running-gpu-jobs
* https://github.com/RadeonOpenCompute/k8s-device-plugin/blob/master/example/pod/alexnet-gpu.yaml

Actually, my previous commit has problem because of missing `.com`, I'll fix it.

joonjeong:
@johanneskoester could you review once more?

snakemake/executors.py
johanneskoester: Ok. That side is good then. So you could also add handling for `amd_gpu`. Does a generic `gpu` make sense? I guess only if it would be possible to query available gpus beforehand via the kubernetes api?

johanneskoester: @joonjeong can you add amd_gpu handling as well?
@SooLee is there a way to pass gpu requirements via tibanna?
@vsoch  is there is a way to pass gpu requirements via GLS?

vsoch: On it! For a brief look over, it looks like the user can select an N1 instance type and then an accelerator type and count. So we could expose that to the user by way of a --gpu N flag (or similar) that would then select an instance from the N1 family and configure an accelerator. How is the current UI (command line or snakefile) setup to receive a gpu request? Is it part of the job metadata?

@johanneskoester I'll have an email going out to you and Andrew shortly with some questions, and once I'm familiar with how the input is done with snakemake, I can add to the Google Life Sciences PR. Stay tuned!

johanneskoester: > On it! For a brief look over, it looks like the user can select an N1 instance type and then an accelerator type and count. So we could expose that to the user by way of a --gpu N flag (or similar) that would then select an instance from the N1 family and configure an accelerator. How is the current UI (command line or snakefile) setup to receive a gpu request? Is it part of the job metadata?
> 
> @johanneskoester I'll have an email going out to you and Andrew shortly with some questions, and once I'm familiar with how the input is done with snakemake, I can add to the Google Life Sciences PR. Stay tuned!

People will define the needed gpu resource in the resources definition of the rule:
```
rule a:
    output:
        "test.txt"
    resources:
        nvidia_gpu=1
    shell:
        "somecommand ..."
```

We might expose more control in the future, but that's it for now.

vsoch: okay, on it! I think GCP might just have support for nvidia, so we wouldn't have the middle option. I'll dig more into it when I code this up.

vsoch: Just so we have it later, here are the acceleratorTypes that are available for an n1 (reasonably sized) instance (it's based on zone actually):

```
{'id': 'projects/snakemake-testing/zones/us-west1-a/acceleratorTypes',
 'items': [{'id': '10004',
   'creationTimestamp': '1969-12-31T16:00:00.000-08:00',
   'name': 'nvidia-tesla-p100',
   'description': 'NVIDIA Tesla P100',
   'zone': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a',
   'selfLink': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a/acceleratorTypes/nvidia-tesla-p100',
   'maximumCardsPerInstance': 4,
   'kind': 'compute#acceleratorType'},
  {'id': '10007',
   'creationTimestamp': '1969-12-31T16:00:00.000-08:00',
   'name': 'nvidia-tesla-p100-vws',
   'description': 'NVIDIA Tesla P100 Virtual Workstation',
   'zone': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a',
   'selfLink': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a/acceleratorTypes/nvidia-tesla-p100-vws',
   'maximumCardsPerInstance': 4,
   'kind': 'compute#acceleratorType'},
  {'id': '10019',
   'creationTimestamp': '1969-12-31T16:00:00.000-08:00',
   'name': 'nvidia-tesla-t4',
   'description': 'NVIDIA Tesla T4',
   'zone': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a',
   'selfLink': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a/acceleratorTypes/nvidia-tesla-t4',
   'maximumCardsPerInstance': 4,
   'kind': 'compute#acceleratorType'},
  {'id': '10020',
   'creationTimestamp': '1969-12-31T16:00:00.000-08:00',
   'name': 'nvidia-tesla-t4-vws',
   'description': 'NVIDIA Tesla T4 Virtual Workstation',
   'zone': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a',
   'selfLink': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a/acceleratorTypes/nvidia-tesla-t4-vws',
   'maximumCardsPerInstance': 4,
   'kind': 'compute#acceleratorType'},
  {'id': '10008',
   'creationTimestamp': '1969-12-31T16:00:00.000-08:00',
   'name': 'nvidia-tesla-v100',
   'description': 'NVIDIA Tesla V100',
   'zone': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a',
   'selfLink': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a/acceleratorTypes/nvidia-tesla-v100',
   'maximumCardsPerInstance': 8,
   'kind': 'compute#acceleratorType'}],
 'selfLink': 'https://www.googleapis.com/compute/v1/projects/snakemake-testing/zones/us-west1-a/acceleratorTypes',
 'kind': 'compute#acceleratorTypeList'}
```
I'm thinking we want the smallest number that is greater than or equal to the number of gpus requested.

vsoch: I'm not familiar with the subtle differences between these, and if we would need to add some ability to specify, for example, the T4 vs P100. For now I'll chose the T4 since it seems slightly better, and we can add some kind of variable if a user requests it in the future. https://www.microway.com/hpc-tech-tips/nvidia-turing-tesla-t4-hpc-performance-benchmarks/ like, damn! GPUs...!

vsoch: okay all set! https://github.com/snakemake/snakemake/pull/90#issuecomment-611173810

joonjeong: Thank you guys for reviewing my opinion!

@johanneskoester Ok, I'll attach more commit on this branch.

joonjeong: One thing I'd like to tell more, there's no way to select GPU type in kubernetes manifest because it's hard to define generic way, so nvidia-docker solve this problem with [OCI(Open Container Initiative) Runtime Spec](https://github.com/opencontainers/runtime-spec) and you guys can find the way how to configure GPU options on container in [this article](https://devblogs.nvidia.com/gpu-containers-runtime/) and you can check how it's implemented in  [nvidia-docker](https://github.com/NVIDIA/nvidia-docker) repository.

I tested this method on AWS and Google Cloud but to describe how it works, there should be more examples. I'll prepare more about this example soon.

vsoch:
hey @joonjeong ! I want to coordinate with you to discuss two job resources that would be available for both kubernetes and google life sciences API for gpus (just finished a call with @johanneskoester discussing this):

 - gpu_model: would be optional, for the user to select a (cloud/service specific) name identifier for a gpu. For example, in my [list above](https://github.com/snakemake/snakemake/pull/227#discussion_r405732172) they could define a resource `gpu_model` to be `nvidia-tesla-p100`
 - machine: would mimic the current `--machine-type-prefix` (and we would decide at some future point if keeping this command line argument makes sense) but so the user could specify the machine type (exactly or a prefix) also via a job resource.

Let me know your thoughts! I would like to add this to the Google Life Sciences executor PR, and I think @johanneskoester would like us to coordinate so it's added here too. :)

joonjeong:
@vsoch Thank you for your opinion! Actually, now I'm mainly working this job on GKE(Google Kubernetes Engine) in Google Cloud, but never experiences with Google Life Sciences, so I'd like to see their API and track your works.

Basically, here's just my guess but, it looks that GKE and Google Life Science both are working on Google Cloud, so selecting and restricting resource like gpu_model, `nvidia-tesla-t4` could be possible with `node selector` or `node affinity` features in kubernetes. Also `machine type` could be possible because kubernetes node manifest has generic label like `beta.kubernetes.io/instance-type`, here's example.

```yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    container.googleapis.com/instance_id: "1448287777796925774"
    node.alpha.kubernetes.io/ttl: "0"
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: "2020-04-08T09:22:09Z"
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/fluentd-ds-ready: "true"
    beta.kubernetes.io/instance-type: e2-standard-2
    beta.kubernetes.io/masq-agent-ds-ready: "true"
    beta.kubernetes.io/os: linux
    cloud.google.com/gke-nodepool: default-pool
    cloud.google.com/gke-os-distribution: cos
    failure-domain.beta.kubernetes.io/region: asia-northeast1
    failure-domain.beta.kubernetes.io/zone: asia-northeast1-b
    kubernetes.io/arch: amd64
    kubernetes.io/hostname: gke-batch-default-pool-00fbeb44-zk3d
    kubernetes.io/os: linux
    node.kubernetes.io/masq-agent-ds-ready: "true"
    purpose: default
  name: gke-batch-default-pool-00fbeb44-zk3d
  resourceVersion: "2834114"
  selfLink: /api/v1/nodes/gke-batch-default-pool-00fbeb44-zk3d
  uid: 6bef7c0b-797a-11ea-a69c-42010a920180
```

Anyway, I'll take a look your works and find a way to integrate this.

vsoch:
Great! To be clear, even between Google Life Sciences (which uses compute engine) and K8 the namespaces don't necessarily have to be the same - I suspect there will be some different set of machine types and/or gpu models for K8. But generally, the basic idea is that the parameter names, gpu_model and machine, could be shared (akin to mem_mb but with different possible values).

joonjeong:
> Great! To be clear, even between Google Life Sciences (which uses compute engine) and K8 the namespaces don't necessarily have to be the same - I suspect there will be some different set of machine types and/or gpu models for K8. But generally, the basic idea is that the parameter names, gpu_model and machine, could be shared (akin to mem_mb but with different possible values).

I agree about that. I think sharing some keywords for CLI parameters is enough.

So,  how do we work? If there's already keywords are decided on your side then, I'll follow.


---
Here's more about Kubernetes

Currently, KubernetesExecutor now using just one identifier, `mem_mb`.

```python
        container.resources = kubernetes.client.V1ResourceRequirements()
        container.resources.requests = {}
        container.resources.requests["cpu"] = job.resources["_cores"]
        if "mem_mb" in job.resources.keys():
            container.resources.requests["memory"] = "{}M".format(
                job.resources["mem_mb"]
            )
```

and with this keywords, Snakemake create Pod with `core` and `mem_mb` value like this

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: snakemake-task
spec:
  containers:
  - name: snakemake
    image: snakemake/snakemake
    resources:
      requests:
        cpu: "250m"
        memory: "64Mi"
```
This is because Kubernetes is just like 'general resource manager' like Hadoop YARN, so Pod is allocated any Worker Node in k8s cluster.

What I firstly wanted is here. with `--resource nvidia_gpu=1`, then manifest is generated like below and requested to k8s cluster.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: snakemake-task
spec:
  containers:
  - name: snakemake
    image: snakemake/snakemake
    resources:
      requests:
        cpu: "250m"
        memory: "64Mi"
      limits:
        nvidia.com/gpu: 1 # leaving requests in this line is not permitted now...
```

If machine is added like `--resource machine_type=n1-standard-4` then, manifest should be generated like this.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: snakemake-task
spec:
  nodeSelector:
    beta.kubernetes.io/instance-type: n1-standard-4
  containers:
  - name: snakemake
    image: snakemake/snakemake
    resources:
      requests:
        cpu: "250m"
        memory: "64Mi"
```

If machine and gpu model both are added like `--resource machine_type=n1-standard-4 gpu_model=nvidia-tesla-t4` then, manifest should be generated like this.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: snakemake-task
spec:
  nodeSelector:
    beta.kubernetes.io/instance-type: n1-standard-4
    cloud.google.com/gke-accelerator: nvidia-tesla-t4 # Here's ambiguous point. AWS doesn't support this node label.
  containers:
  - name: snakemake
    image: snakemake/snakemake
    resources:
      requests:
        cpu: "250m"
        memory: "64Mi"
```

I'll think more about node selector cases in gpu.

Anyway, about keywords, `machine_type` and `gpu_model`, I'll follow what you're working on.

snakemake/executors.py
johanneskoester: PR #329 introduces string resources. In short, people could specify `gpu_model="nvidia-tesla-p100`. Would that be feasible for kubernetes as well @joonjeong?

joonjeong: @johanneskoester That could be good for KubernetesExecutor to enforce snakemake pod to select specific gpu model. but `nvidia.com/gpu` still needed because kubernetes decide to allocate actual resource(gpu) with specified under `resources` section.

So, if `gpu_model` keyword is added, I guess command line will be written like this.

```bash
$ snakemake --kubernetes --resource gpu_model=nvidia-tesla-p100 nvidia_gpu=1
```

Anyway, I think it's good to add `machine_type` and `gpu_model`. I'll also take care about that in this commit.

To @johanneskoester @vsoch, resource keywords are decided? I'll follow that keywords

vsoch:
Ah that's a good suggestion - I was using `machine` but not `machine_type` (it's what we discussed in our call) but I'd be happy to use `machine_type` instead - it's more clear. @johanneskoester what do you think?

snakemake/executors.py
vsoch: @joonjeong we had discussed `gpu_model` and `machine` but I like your suggestion of `machine_type` to be more specific, and I can change my PR if that is @johanneskoester preference.

johanneskoester: > @johanneskoester That could be good for KubernetesExecutor to enforce snakemake pod to select specific gpu model. but `nvidia.com/gpu` still needed because kubernetes decide to allocate actual resource(gpu) with specified under `resources` section.
> 
> So, if `gpu_model` keyword is added, I guess command line will be written like this.
> 
> ```shell
> $ snakemake --kubernetes --resource gpu_model=nvidia-tesla-p100 nvidia_gpu=1
> ```

I think nvidia_gpu can be omitted, because the KubernetesExecutor can simply infer that from the gpu_model definition, right?

@vsoch @joonjeong `machine_type` sounds good.

vsoch: @johanneskoester all set, updated machine to machine_type https://github.com/snakemake/snakemake/pull/90/commits/81a5b2507227c969e6d68473370a6a0c77afc4df

johanneskoester:
My main concern is that the kubernetes executor should remain generic, not tied to one vendor. So if there is no platform independent way to map machine_type and gpu_model to a pod, it should probably not be implemented.

johanneskoester:
Wait, machine_type seems fine, right? Just gpu_model is currently not properly standardized?

vsoch:
Yes I am using machine_type too :)

joonjeong:
@johanneskoester @vsoch Oh, I'm sorry to see this issue late, I was a little busy.

`machine_type` and `gpu_model` is enough, I think. It would be great if `gpu_count` is added(is that already exist?)

> My main concern is that the kubernetes executor should remain generic, not tied to one vendor. So if there is no platform independent way to map machine_type and gpu_model to a pod, it should probably not be implemented.

More about Kubernetes, I just look some cloud vendors' interfaces(labels of node) and it is hard to make them standardized, so I think we need to provide another option for kubernetes pod creation, `Pod Template`

Basically, concept `Pod Template` is already applied as a [Jenkins Plugin](https://github.com/jenkinsci/kubernetes-plugin) and some other open source solutions and it is working like this, I'll describe with some command if this is applied to snakemake.

1. Suppose that this below pod template exists for GPU.

```yaml
# pod-template_nvidia-tesla-p100.yaml
# I copied this from https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/ and fix some
apiVersion: v1
kind: Pod
metadata:
  name: snakemakejob-xxxxxxx
spec:
  containers:
    - name: snakemake
      image: "snakemake/snakemake:latest"
      resources:
        limits:
          nvidia.com/gpu: 4
  nodeSelector:
    accelerator: nvidia-tesla-p100 # this means custom label applied in cluster nodes
    # if it's working on GKE(Google Kubernetes Engine), below label could be possible.
    # cloud.google.com/gke-accelerator: nvidia-tesla-k80
```

2. When submit `snakemake` jobs to kubernetes cluster, run command like this
```bash
$ snakemake --kubernetes --kubernetes-pod-template=pod-template_nvidia-tesla-p100.yaml --use-conda --default-remote-provider $REMOTE --default-remote-prefix $PREFIX
```

I think `Pod Template` is just general way in Kubernetes Executor. How about you guys? @johanneskoester @vsoch 

Below option  is also possible. Suppose that we discussed parameters are applied and we just submit snakemake jobs to kubernetes.

```bash
$ snakemake --kubernetes --use-conda --default-remote-provider $REMOTE --default-remote-prefix $PREFIX --resource machine_type=n1-standard-1 gpu_model=nvidia-tesla-p100 gpu_count=1
```

And, above command generates Pod on kubernetes like this.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: snakemakejob-xxxxxxx
spec:
  containers:
    - name: snakemake
      image: "snakemake/snakemake:latest"
  nodeSelector:
    machine_type: "n1-standard-1"
    gpu_model: "nvidia-tesla-p100"
    gpu_count: "1"
```

This is just simple solution for just who want to use kubernetes with CLI parameters.
Actually, I'm not certain with above yaml pod template can use GPU directly, but if there's kubernetes cluster has node with label `{"machine_type": "n1-standard-1", "gpu_model": "nvidia-tesla-p100", "gpu_count": "1"}`, then Pod is allocated to that node.(Unfortunately, in this case, I'm not certain Kubernetes will provide CUDA to container or not.)

Long story, to summarize my ideas,

1. In Kubernetes Executor, if `machine_type`, `gpu_model`, `gpu_count` is declared, then just add them to labels.(Just Allocate Pod to Specific Node Strategy)
2. If there's someone want to use Kubernetes more advanced way, then just provide parameter `--kubernetes-pod-template` to make customized pod template on their own.(Pod Template)

Actually, what I really wanted is second way. I'm not hesitated with discarding this PR.
If it would be great with you guys, then I'll just close this PR and rework with above two ways.

vsoch:
> machine_type and gpu_model is enough, I think. It would be great if gpu_count is added(is that already exist?)

@joonjeong I've added all three of these resource specifications to the Google Life Sciences Executor. They default to unset (None) of course, and then for Google Life Sciences we only can choose between different nvidia types (no other families exposed yet!) Given asking for a GPU, the machine_type is also filtered down to those in the N1 family. I suspect K8 also has some subset of allowed machine types or GPUs, so although the resource variables would be shared between executors, we can assume the allowed values will be somewhat different. For gpu_count, given that a GPU is asked for, if the user doesn't set this variable we default to 1.

> Below option is also possible. Suppose that we discussed parameters are applied and we just submit snakemake jobs to kubernetes.

I think it's reasonable to have some default template that is packaged with snakemake, and that the user doesn't need to know about, and that gets filled in with resource specifications. I think the bulk of users won't be familiar with these template, but a small subset will want to specify it directly (and this could be done with `--kubernetes-pod-template`. Here are a few things to think about:

- if a user provides a pod template _and_ specifies resource arguments, how are they consolidated? Is it the case that some subset (e.g., machine type) are ignored?
- how would it work to request a pod template for one part of a workflow but not another (or would this not be reasonable to support?)
 - I think if the pod template variable is added, we should also provide a selection of examples for different use cases. If we can't come up with use cases easily and create, test, and share these templates, then I would question if it would be highly used or if we are over-engineering it. Do you have a sense for the popularity of this executor, and how it's being used?

PR #230, opened 2020-02-14
Make previous output files available as prevoutput

johanneskoester:
This way, it becomes possible for jobs to perform smart updates of existing files, saving computation time.

### DIFF
--- a/snakemake/jobs.py
+++ b/snakemake/jobs.py
@@ -669,6 +669,9 @@ class Job(AbstractJob):
                 )
             )
 
+        # TODO: instead of removing existing output, move it into a temp dir
+        # and make the files available to shell/run/script as prevoutput.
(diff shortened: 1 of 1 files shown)


SamStudio8:
Will this be optional? I wouldn't want to wait for very large files to be moved to a temporary location *in case* I wanted to use `prevoutput`.

johanneskoester:
mv on the same filesystem should be as fast as rm, basically instantaneous. We will not move files to /tmp. Perhaps even just renaming them instead of moving. The true delete will then happen after the job is finished.

johanneskoester:
Already implemented via update() and before_update() in Snakemake 8.7

johanneskoester:
Yes, this is fully optional.

PR #233, opened 2020-02-14
Add a mechanism to define services

johanneskoester:
I.e. a job that has to run along with all consuming jobs, before being terminated. This can be used to integrate the invocation of databases, or setting up shared memory allocation (e.g. for loading a STAR index before mapping). A service is automatically submitted together with its consuming jobs. Moreover, there can be multiple instances of the same service for multiple groups


johanneskoester:
cc @cmeesters

johanneskoester:
Has been added in another PR-.

PR #239, opened 2020-02-21
added an option to the sftp remote to create output path

cclienti:
Hello

I added an option to the remote sftp to create the path in the remote
server. This is to be consistent with the standard posix file behavior
of SnakeMake which creates automatically output directories.

The added option is RemoteProvider(mkdir_remote=True)


cclienti:
Hello,

It seems that the formatting of the SFTP.py causes some issue during the "check formatting" pass. If needed I can try to change the formatting in another commit.

cclienti:
I close this merge request in order to propose it via the github organization of my company.

Issue #204, opened 2020-01-23
Unable to collect output logs from cluster jobs


tylergannon:
**Is your feature request related to a problem? Please describe.**
It would be very handy if Snakemake would play nicely with GridEngine output logs.  This would dry up `shell` sections of rule definitions by letting the cluster handle redirecting output in the same way it would without Snakemake.

When developing snakemake rules, it's nice to be able to simply copy the shell command that's printed by Snakemake, after the rule fails.  This is one step more tricky if that shell command includes output redirections, which must be edited out before I can paste the command into a terminal to run it directly.

Better than this is if the qsub submission script handles configuring output redirection, so that developers don't have to write it directly into the shell script.

**Describe the solution you'd like**
All that really needs to change in order to allow this, is for Snakemake to clear log files for jobs _before cluster submission_ rather than _at invocation time_.   I believe that's all.

I already wrote a qsub wrapper (invoked with `--cluster "my_script {dependencies}"`), which more or less calls `qsub -o {log[0]} -e {log[1]} ... {snakemake_job_script}`.

This is convenient because I can simply specify two files under `log` and then I don't have to specifically configure stream redirection in the `shell` script.  Even better yet if I just specify `log: my_log_fn`, which would return something like:

```
return multiext("logs/" + rule_name + "." + ".".join(val for val in wildcards), ".log", ".err"
```

**Describe alternatives you've considered**
The alternative is to write stdout and stderr redirections in all the `shell` scripts inside the workflow, which works okay.  It's less DRY, but the biggest drawback is this:

running stuff through the cluster is just a little more difficult for troubleshooting.  It's very nice to be able to paste the same command Snakemake ran, into a terminal, so that I can directly see where it failed.  As it stands, I have to edit the command, removing shell redirection instructions, before I can re-run it manually.

tylergannon:
Another reason this would be useful is that it's kind of annoying to get the automatically-generated log files created by gridengine, which are empty (in the case of stdout) except for a tiny bit of snakemake output (stderr file) because they will lead to issues if I write them to the same paths configured in the Snakemake rule.

The other option is to write to files that Snakemake doesn't know about, but this somewhat belies the reason for having a log section on the rule definition.

tylergannon:
Addressed by #289

PR #243, opened 2020-02-25
Add "preserve" flag to allow authors to skip deleting a log.

tylergannon:
I very simply added a flag called "preserve", and altered
job initialization to skip deleting log files marked with this flag.

I addressed this wish in snakemake/snakemake#204.  Without it
I see no way to use the log files listed on my job definition,
as the paths to which my cluster will save the output streams
for submitted jobs.

Thank you!

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -672,7 +672,7 @@ def gcloud_cluster():
                     default_remote_provider="GS",
                     default_remote_prefix=self.bucket_name,
                     no_tmpdir=True,
-                    **kwargs
+                    **kwargs,
                 )
             except Exception as e:
                 shell(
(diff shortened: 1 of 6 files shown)


tests/tests.py
```diff
@@ -672,7 +672,7 @@ def run(self, test="test_kubernetes", **kwargs):
                     default_remote_provider="GS",
                     default_remote_prefix=self.bucket_name,
                     no_tmpdir=True,
-                    **kwargs
+                    **kwargs,
```
tylergannon: Comma added by Black formatter.

tylergannon:
How about a different flag-oriented solution, coming at my problem from a totally different angle.

The following will actually accomplish what I want to do much more explicitly, and in a way that might be interesting to others.

```
rule myrule:
log: stdstreams("logs/myrule.{molecule}", ".out", ".err")
```
which would be equivalent to:
```
rule myrule:
log: [stdout("logs/myrule.{molecule}.out"), stderr("logs/myrule.{molecule}.err")]
```

Then I can just make a small change to `shell.py`, so that it will scan the `log` list for the flags `stdout` and `stderr`, and...

1. Do nothing different if not found
2. Open file objects for the specified files, if found, and pass them to Popen for the output streams
3. Raise an error unless each of those flags is specified either zero or one times.

This is such an easy solution and I think it will enable many workflow developers to DRY up their shell scripts inside `Snakefile`s.


If you don't like the `stdstreams` function, another suggestion would be `multiext("logs/myrule", stdout(".out"), stderr(".err"))`.  We'd just have to then make sure `multiext` will pass flags from the given extensions over to the resulting expanded paths.

I might code this up tomorrow anyways and propose it, but I'm curious what you think.

tylergannon:
Superseded by #289

Issue #244, opened 2020-02-26
to_remove does not honor --keep-remotes


vsoch:
**Snakemake version**
5.10.x (development)

**Describe the bug**
When the user adds the flag `--keep-remotes` the docstring says that files downloaded from remotes will not be removed. In practice, the files are removed.

**Minimal example**

If I'm running something with Google Life sciences, I might have this full command generated for my worker:

```
snakemake snakemake-testing/kim-wxs-varlociraptor/igv-report/23.present.html --snakefile workflow/Snakefile --force -j --keep-target-files  --keep-remote --latency-wait 0 --attempt 1 --force-use-threads  --allowed-rules igv_report --nocolor --notemp --no-hooks --nolock  --use-conda  --default-remote-provider GS --default-remote-prefix snakemake-testing/kim-wxs-varlociraptor  --default-resources
```

I think the bug [is here](https://github.com/snakemake/snakemake/blob/e064a076ebb01310447ef9d09b229951cf52a04a/snakemake/jobs.py#L761) because we add remote files (that are local) to the list of to_remove without checking if the flag is set. Note that the flag goes from the client -> DAG and then the dag variable is held by the scheduler, both of which interact with jobs (the function linked above is part of a job). I'm not sure the best way to pass forward this variable/setting to the job, since it's a child of the instances where it's set. Likely if either the dag or scheduler creates the job, this would be where to pass it on!

I'm currently testing just running the same command (and commenting out that line) to see if that resolves the error.

PR #245, opened 2020-02-26
Adding check for keep remotes to job

vsoch:
This is a suggested fix (we can call WIP) to address #244. Basically, the variable gets passed from the client to the DAG object, and then when the dag creates a new job, it hands off the variable (and then it's available in cleanup).

Note that for testing I just commented out the line (the one that will be triggered by the variable) and the files are no longer removed! There they are! :clinking_glasses: 

```
$ ls snakemake-testing/kim-wxs-varlociraptor/
igv-report  merged-calls  recal  refs
```

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -1208,6 +1208,7 @@ class DAG:
             wildcards_dict=wildcards_dict,
             format_wildcards=format_wildcards,
             targetfile=targetfile,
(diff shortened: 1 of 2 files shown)


johanneskoester:
Thanks for pointing me to the issue. Actually, remote inputs should never be cleaned up upon job failure. This is now implemented in PR #739.

PR #249, opened 2020-02-28
Set linear algbra library thread usage environment variables for all jobs

dlaehnemann:
For any process that snakemake runs, avoid implicit parallelism beyond the requested number of threads for the respective rule. Default to 1 thread for implicit parallelism if the rule does not specify a number of threads.

When e.g. [R is compiled with a parallelized linear algebra library](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Linear-algebra), calling any linear algebra (like handling large matrices) can lead to implicit parallelism that will grab all cores/threads available on the host system. On large servers, this will lead to a degree of parallelization that does more harm than good, slowing everything down with cross-talk between threads.

For [R parallel computing](https://cran.r-project.org/web/views/HighPerformanceComputing.html), there's a package that can control this variable within scripts: [RhpcBLASctl](https://cran.r-project.org/web/packages/RhpcBLASctl/index.html). However, the same issue exists in other languages, e.g. for [numpy in python](https://stackoverflow.com/a/53224849/2352071) or for [Julia](https://stackoverflow.com/q/37501181/2352071). Thus, I think setting all known environment variables for every created process to the number of threads that process has specified via its rule definition is the cleanest solution that is easiest to maintain and document.


dlaehnemann:
It seems the error currently thrown is still related to yesterday's commits. But the failing tests also reveal, that the environment variables are added three times each, so the current way of adding them is somehow executed three times -- not sure what this is a side-effect of?

snakemake/shell.py
```diff
@@ -103,6 +110,16 @@ def __new__(
         env_modules = context.get("env_modules", None)
         shadow_dir = context.get("shadow_dir", None)
 
+        # environment variable list for linear algebra libraries taken from:
+        # https://stackoverflow.com/a/53224849/2352071
+        cls._process_prefix += (
```
johanneskoester: This will always extend process_prefix, making it longer and longer with each shell command. Instead, rather create a new variable `envvars` and add it to `cmd` below (line 123), after `cls._process_prefix`.

dlaehnemann:
This took days to pinpoint...

Have a quick look, if that is how you meant your suggestions. Now, the variables are only added once, as intended, and should contain the right threads number applicable to the current process.

The failing test seems to be due to the new requirement to specify `--cores`...

SamStudio8:
Once #138 is accepted, you could add these lines to the [part of `excecutable` where pipefail is set](https://github.com/snakemake/snakemake/blob/5dca619abe66bc2d5c25c11538448d053a715774/snakemake/shell.py#L58), that overrides the shell prefix when `bash` is the shell of choice. This might be a bit cleaner than adding them inside of the command context each time.

snakemake/shell.py
```diff
@@ -134,7 +147,7 @@ def __new__(
             stdout=stdout,
             universal_newlines=iterable or None,
             close_fds=close_fds,
-            **cls._process_args
+            **cls._process_args,
```
johanneskoester: Please do the following for setting the variables in a portable way:
```
envvars = dict(os.environ)
envvars["OMP_NUM_THREADS"] = threads
...
```
and then pass that as `env=` argument to `sp.Popen`.

dlaehnemann: @johanneskoester This is finally passing. Spent an over an hour figuring out that Popen insists on all values in the `env=` `dict` being of type `str`, so `threads has to be `str`...

PR #252, opened 2020-02-28
Spot

SooLee:
This PR adds a spot instance support for snakemake-tibanna


snakemake/__init__.py
```diff
@@ -144,6 +144,7 @@ def snakemake(
     precommand="",
     default_remote_provider=None,
     default_remote_prefix="",
+    spot_instance=False,
```
johanneskoester: I think we should call this `tibanna_spot_instance` and the flag accordingly.

johanneskoester:
And also please format with black. Thank you!

SooLee:
@johanneskoester Thanks for the review! I think we will need to expose more tibanna config parameters, so I'm thinking we could generalize it to something like `--tibanna-configfile` and have a json that defines spot instance, security group, subnet, etc (I had a use case for having to use these parameters). What do you think?

johanneskoester:
> @johanneskoester Thanks for the review! I think we will need to expose more tibanna config parameters, so I'm thinking we could generalize it to something like `--tibanna-configfile` and have a json that defines spot instance, security group, subnet, etc (I had a use case for having to use these parameters). What do you think?

In order to make tibanna straightforwardly configurable via [profiles](https://snakemake.readthedocs.io/en/stable/executing/cli.html#profiles), I'd like to stick with command line flags, all prefixed with --tibanna, and in its own group.

johanneskoester:
Maybe something like `--tibanna-args subnet=xxx spot=true security-group=foo`

SooLee:
@johanneskoester Sounds good! I can use `--tibanna-config` (we use args for a different purpose, so config would be a better term).

SooLee:
@johanneskoester I had to comment out the envvar lines in the Snakefile - apparently the latest Snakemake Docker image doesn't recognize this syntax. It may fail some tests if they're using envvar. 
 I also turned off default resource because the latest image didn't recognize input size_mb.

PR #258, opened 2020-03-03
added an option to the sftp remote to create output path

cclienti:
An option to the remote sftp is added to create the path in the remote
server. This is to be consistent with the standard posix file behavior
of SnakeMake which creates automatically output directories.

The added option is RemoteProvider(mkdir_remote=True)

### DIFF
--- a/snakemake/remote/SFTP.py
+++ b/snakemake/remote/SFTP.py
@@ -26,7 +26,13 @@ class RemoteProvider(AbstractRemoteProvider):
     allows_directories = True
 
     def __init__(
-        self, *args, keep_local=False, stay_on_remote=False, is_default=False, **kwargs
+        self,
+        *args,
+        keep_local=False,
+        stay_on_remote=False,
+        is_default=False,
+        mkdir_remote=True,
+        **kwargs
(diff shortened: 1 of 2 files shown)


snakemake/remote/SFTP.py
```diff
@@ -26,7 +26,13 @@ class RemoteProvider(AbstractRemoteProvider):
     allows_directories = True
 
     def __init__(
-        self, *args, keep_local=False, stay_on_remote=False, is_default=False, **kwargs
+        self,
+        *args,
+        keep_local=False,
+        stay_on_remote=False,
+        is_default=False,
+        mkdir_remote=False,
```
johanneskoester: I think it should be True by default then, right?

johanneskoester:
Please update to the latest master branch from upstream. The test failures are fixed now.

snakemake/remote/SFTP.py
cclienti: I did not dare to set it to true in order not to break the existing code behaviour. But I agree, it would be better to set the default value to true. I'll change that.

PR #263, opened 2020-03-05
bug fix - resource unit

SooLee:


SooLee:
@johanneskoester input.size is in bytes so disk_mb should be divided by 1024*1024. Not doing this caused tibanna ebs size estimates to be in thousands of terabytes.

johanneskoester:
Good catch! I have fixed this slightly differently in commit 43775a05.

PR #273, opened 2020-03-12
Adding crc32c checksum for Google cloud storage

vsoch:
The Google Storage api does not expose a function to stream to file, however the blob.write_to_file accepts a file object, so we can create a class to wrap this object, exposing the same functions, but also generating a hasher that we can update for each write action. This allows us to calculate the crc32c while the file is being downloaded and written to file as opposed to needing to download and then read again, which will make a difference for large files. See the docstrings in the GS.py remote
file for links to further documentation. I also ran black on setup.py so it's pretty, and added an extras install for google-cloud since the dependency list just grew by one.

Please edit as you see fit! I'm off to bed.


snakemake/remote/GS.py
```diff
@@ -13,13 +15,42 @@
...
 except ImportError as e:
     raise WorkflowError(
-        "The Python 3 package 'google-cloud-sdk' "
-        "needs to be installed to use GS remote() file functionality. %s" % e.msg
+        "The Python 3 packages 'google-cloud-sdk' and `crc32c` "
+        "need to be installed to use GS remote() file functionality. %s" % e.msg
     )
 
 
+class Crc32cCalculator(object):
```
johanneskoester: `(object)` is superfluous with Python 3.

snakemake/remote/GS.py
```diff
@@ -95,12 +126,27 @@ def size(self):
...
-            os.makedirs(os.path.dirname(self.local_file()), exist_ok=True)
-            self.blob.download_to_filename(self.local_file())
+        if not self.exists():
+            return None
+
+        # Create the directory for the intended file
+        os.makedirs(os.path.dirname(self.local_file()), exist_ok=True)
+
+        # Continue trying to download until hash matches
+        while not os.path.exists(self.local_file()):
```
johanneskoester: Instead of checking existence on disk, have a boolean variable that becomes true once the file is there or so. THis way, we save the additional existence checks, which can play a role when downloading many files in parallel.

snakemake/remote/GS.py
vsoch: Fixed.

snakemake/remote/GS.py
vsoch: I almost implemented it this way! I changed at the last minute because I didn't like the is, else in the loop. I didn't think of this. Now fixed!

snakemake/remote/GS.py
```diff
@@ -95,12 +126,30 @@ def size(self):
...
+            with open(self.local_file(), "wb") as blob_file:
+                parser = Crc32cCalculator(blob_file)
+                self.blob.download_to_file(parser)
             os.sync()
-            return self.local_file()
-        return None
+
+            # Compute local hash and verify correct
+            if parser.hexdigest() != self.blob.crc32c:
+                os.remove(self.local_file())
```
johanneskoester: A final thought: in theory, this could go on forever. I think we should give it 3 tries, and then raise a `WorkflowError("Failed to download {} from remote: checksums did not match 3 times.".format(self.remote_file()))`.

vsoch: okay, added! I made `retry_count` a variable so (at some future point) it might be exposed. If some kind of issue comes up with requests being too quickly one after the other, we could add a delay (e.g., exponential or some kind of backoff) but I don't see a strong need for this until there is evidence for it.

PR #288, opened 2020-03-22
clarify availability of pythonic names

goi42:
The shell command can only access `input` and `output` in the python format mini language. This updates the docs to make that clear.


tbooth:
Hi,
I'm not sure how much this clarifies things. The key piece of info here is that the string formatting with `{input}` `{output}` etc. occurs when the rule is made into a job, whereas any direct manipulation of the shell string by using, for eg., `.format(...)` is going to happen earlier, as the rule is being parsed, when those values are simply not available.
Your added explanation implies (to me) that this is a syntactic thing, but in fact it's a semantic one.

Cheers,
TIM

goi42:
> I'm not sure how much this clarifies things.
> ...
> Your added explanation implies (to me) that this is a syntactic thing, but in fact it's a semantic one.

@tbooth I have added clarifying text that hopefully makes this clear.

goi42:
I do not understand why this test failed.

PR #289, opened 2020-03-23
Add output redirect flags.

tylergannon:
Workflows involving many shell scripts can become quite cluttered
with bash redirections if every line in the script has them, especially
if stdout and stderr are redirected to separate files.

This allows the developer to instruct Snakemake to redirect the entirety
of a given script over to files, by flagging log files with stdout and
stderr.

This PR replaces #243, and it's a little closer to the spirit of what I was hoping to
accomplish with that one.

There are a couple of choices I've made here, which I anticipate you might disagree
with.  I hope that you'd be open to the spirit of this change, which would really mean a lot to me,
and work with me to come up with an implementation that matches your design decisions.

thanks!

### DIFF
--- a/snakemake/executors.py
+++ b/snakemake/executors.py
@@ -392,7 +392,7 @@ class CPUExecutor(RealExecutor):
             job.wildcards,
             job.threads,
             job.resources,
-            job.log._plainstrings(),
+            job.log,
             benchmark,
             benchmark_repeats,
             conda_env,
--- a/snakemake/io.py
+++ b/snakemake/io.py
@@ -1327,8 +1327,27 @@ class Resources(Namedlist):
     pass
 
 
-class Log(Namedlist):
+class LogStreamError(AttributeError):
     pass
+
+
+class Log(Namedlist):
+    def streams(self):
+        _streams = {
+            name: str(value)
+            for name, value in self.items()
+            if name in ("std", "stderr", "stdout")
+        }
+        if "std" in _streams and ("stderr" in _streams or "stdout" in _streams):
+            raise LogStreamError("Can't combine std stream with stdout/stderr.")
+        if any(path for path in _streams.values() if " " in path):
+            raise LogStreamError("Std output stream has more than one entry.")
+
+        for path in _streams.values():
+            dirname = os.path.dirname(path)
+            if any(dirname) and not os.path.exists(dirname):
+                os.makedirs(dirname)
+        return {name: open(value, "w") for name, value in _streams.items()}
 
 
 def _load_configfile(configpath, filetype="Config"):
--- a/snakemake/shell.py
+++ b/snakemake/shell.py
@@ -16,6 +16,7 @@ from snakemake.logging import logger
 from snakemake.deployment import singularity
 from snakemake.deployment.conda import Conda
 from snakemake.exceptions import WorkflowError
+from snakemake.io import Log
 import snakemake
 
 
@@ -80,6 +81,18 @@ class shell:
         with cls._lock:
             cls._processes.clear()
 
+    @classmethod
+    def __get_output_streams(cls, log: Log, capture_stdout):
+        log = log if isinstance(log, Log) else Log()
+
+        streams = log.streams()
+
+        if "std" in streams:
+            return {"stdout": streams["std"], "stderr": sp.STDOUT}
+
+        streams.setdefault("stdout", sp.PIPE if capture_stdout else STDOUT)
+        return streams
+
     def __new__(
         cls, cmd, *args, iterable=False, read=False, bench_record=None, **kwargs
     ):
@@ -90,15 +103,12 @@ class shell:
         # add kwargs to context (overwriting the locals of the caller)
         context.update(kwargs)
 
-        stdout = sp.PIPE if iterable or read else STDOUT
-
         close_fds = sys.platform != "win32"
 
         jobid = context.get("jobid")
         if not context.get("is_shell"):
             logger.shellcmd(cmd)
 
-        env_prefix = ""
         conda_env = context.get("conda_env", None)
         container_img = context.get("container_img", None)
         env_modules = context.get("env_modules", None)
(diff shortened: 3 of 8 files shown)


snakemake/executors.py
```diff
@@ -387,7 +388,7 @@ def job_args_and_prepare(self, job):
             job.wildcards,
             job.threads,
             job.resources,
-            job.log._plainstrings(),
+            job.log,
```
tylergannon: I guess that by mapping the file collections to their `_plainstrings` equivalents, you're trying to obscure the AnnotatedString protocol from  the executors, but I left it this way in hopes at least of getting the functionality ready, and work out the object model with you.

Is there an alternative way to pass on awareness of the stdout/stderr flags to the shell class?

Also ... should this functionality extend to the python `script` blocks?  I can follow up with that, if you'd give me some instruction.

snakemake/io.py
```diff
@@ -965,7 +981,16 @@ def multiext(prefix, *extensions):
             "Extensions for multiext may not contain path delimiters "
             "(/,\) and must start with '.' (e.g. .txt)."
         )
-    return [flag(prefix + ext, "multiext", flag_value=prefix) for ext in extensions]
+
+    def copy_flags(source, dest):
```
tylergannon: I think it's rather helpful to be able to propagate specific flags to individual members of a `multiext` list.

snakemake/shell.py
```diff
@@ -79,6 +80,25 @@ def cleanup(cls):
...
+            match = [f for f in log if isinstance(f, _IOFile) and prop.fget(f)]
+            if len(match) == 0:
+                return None
+            if len(match) > 1:
+                raise SyntaxError(name + " allowed on just one log per rule.")
+            makedirs(os.path.dirname(os.path.abspath(match[0])))
+            return open(match[0], "w")
+
+        stdout = find_log(_IOFile.is_stdout, "stdout")
+        stderr = find_log(_IOFile.is_stderr, "stderr")
```
tylergannon: It's kind of just right here where I feel like I'm losing separation of concerns, with the `io` objects bleeding into your `shell` class.  It might be justifiable, given that the `shell` class already has a somewhat hazy interface boundary through its use of `inspect.currentframe().f_back.f_locals`.

tylergannon:
@johanneskoester okay so instead of flags, we would have "keyword" log names, that would signify capturing output streams.  That's fine as long as you don't mind the risk of someone naming their logs with those names, without knowing that they are keywords.

This is nice though, I was trying to think up a simple way to express capturing both streams.  I'll get back to you this evening with those changes.

tylergannon:
Okay the changes you suggested are implemented.  I've also simplified the solution by moving the logic that validates opens output streams over to the `io.Log` class.

snakemake/shell.py
```diff
@@ -132,10 +142,10 @@ def __new__(
             cmd,
             bufsize=-1,
             shell=True,
-            stdout=stdout,
             universal_newlines=iterable or None,
             close_fds=close_fds,
-            **cls._process_args
+            **cls.__get_output_streams(context.get("log"), iterable or read),
```
johanneskoester: Very nice. What happens if a user uses e.g. stderr="foo.log" and in addition also manually pipes into `{log.stderr}`. I guess since nothing gets captured from here and the manual pipe gets precedence, we are fine, right?

snakemake/shell.py
```diff
@@ -132,10 +142,10 @@ def __new__(
...
+            **cls._process_args,
```
johanneskoester: The open files that you get with `Log.streams()` should be closed below.

snakemake/io.py
```diff
@@ -1327,10 +1327,29 @@ class Resources(Namedlist):
...
+        if "std" in _streams and ("stderr" in _streams or "stdout" in _streams):
+            raise LogStreamError("Can't combine std stream with stdout/stderr.")
+        if any(path for path in _streams.values() if " " in path):
+            raise LogStreamError("Std output stream has more than one entry.")
+
+        for path in _streams.values():
+            dirname = os.path.dirname(path)
+            if any(dirname) and not os.path.exists(dirname):
+                os.makedirs(dirname)
+        return {name: open(value, "w") for name, value in _streams.items()}
```
johanneskoester: The log files should be opened in append mode, such that multiple shell commands below each other don't overwrite their output.

snakemake/shell.py
tylergannon: That's an extremely good question.  I'll make a test case for this.

snakemake/shell.py
tylergannon: They are closed by Popen, which has close_fds=True, though I guess if I close them manually then this feature might even work for Windows platform.

snakemake/io.py
tylergannon: Sure thing

johanneskoester:
Hi @tylergannon any chance to finalize this soon?

dfm:
@tylergannon, @johanneskoester — I'd be very keen to have access to a feature like this which, as far as I can tell, still isn't supported. Would you be open to me taking a stab at bringing this PR over the finish line, or do we feel like a different approach would be better?

tylergannon:
Hey all, apologies for a very tardy update from me on this issue.  I'm no longer at the company that had me developing tools using Snakemake.

@dfm please feel free to take this over.  Been a while since looking back here and I still feel that the syntax proposed is pretty decent.  I don't remember exactly what was the difference between my proposal an @johanneskoester's reply.

But there are the two minor issues to resolve, being 

- [ ]  a test case that asks whether there will be any adverse result if someone specifies `log: stdout=foo.log` and also redirects stdout to that file inside their script.  It is possible that this could lead to data corruption, if Python buffers the captured output and doesn't write it straight to disk. 
- [ ] If the above does have an adverse result, then what?  Would a note in the docs be sufficient?  
- [ ] Open logs in append mode per https://github.com/snakemake/snakemake/pull/289#discussion_r399346400
- [ ] Add explicit close to opened log files, per https://github.com/snakemake/snakemake/pull/289#discussion_r399345645

Let me know if you have any questions!
Tyler

dfm:
@tylergannon — Thanks!! I've opened a new PR over in #2100 with these changes and my initial attempts to address the outstanding issues.

PR #293, opened 2020-03-25
only assert statinfo is identical for matches

goi42:
XRootD will occasionally return multiple matches for a file. This behavior is transient and probably harmless, but it causes the assertion in _get_statinfo to fail. This updates the assertion to check only the relevant properties. Note that, even with all relevant properties identical, match[0].statinfo != match[1].statinfo, hence the wordiness of the assertion.


goi42:
@johanneskoester I am afraid I'm not sure what's wrong with my formatting. Perhaps you can provide some insight?

goi42:
@johanneskoester No change :(

goi42:
Thanks for your attention and for fixing the formatting issues.

PR #298, opened 2020-03-30
fix `patterns=` example to working keyword, corresponding to keyword in text below

dlaehnemann:
with the previous example of `pattern=""`, snakemake complains about `pattern` being an unexpected keyword argument for `result()`


johanneskoester:
Thanks!

Issue #312, opened 2020-04-04
snakemake does not have default --cores value


albertas:
I want to print "Hello world" using snakemake, so I have created Snakefile and executed snakemake in the directory. However got the error:

```Error: you need to specify the maximum number of CPU cores to be used at the same time with --cores.```

I just want to use snakemake as replacement for makefile, it really doesn't matter how many cores it uses for workflow. Would it be possible to set a default value for --cores?

Freshly installed snakemake 5.13.0 version, without any configuration. This problem does not occur with version 4.

Snakefile I used:
```
rule hello_world:
    shell:
        "echo Hello world"
```

**Comment by johanneskoester** [comment]
This change was intentional. People not coming from make frequently didn't anticipate the fact that it uses 1 core by default, wondering why it does not parallelize. Hence, we decided to make the specification of cores explicit. I know that it causes additional noise for simple cases, but not for the majority. There is always the possibility to define your own default via a [profile](https://snakemake.readthedocs.io/en/stable/executing/cli.html#profiles).

**Comment by nick-ulle** [comment]
> There is always the possibility to define your own default via a profile.

And is there a way to set a default profile, so that `snakemake --cores ...` doesn't just become `snakemake --profile ...`? This question has been [asked on StackOverflow](https://stackoverflow.com/questions/67531322/can-i-define-a-default-profile-for-snakemake), but never got a clear answer.

It seems like the goal here is to just be able to type `snakemake` to build.

**Comment by mschilli87** [comment]
@albertas: My workaround for cases where I want to 'hide' Snakemake (and its complexity) from the end user is the following:

```snakemake
#!/usr/bin/env -S snakemake --cores 1 --quiet --snakefile
rule hello_world:
    shell:
        "echo Hello world"
```

If you place this in a file called `hello_world` and make  that executable (`chmod +x hello_world`) you can execute it without any additional flags (`./hello_world`) and get

```
Job counts:
        count   jobs
        1       hello_world
        1
Hello world
```

Note: This requires `env` from coreutils >= 8.30.

Also be aware that command line arguments passed to `hello_world` will be passed down to `snakemake`:

```sh
./hello_world --verbose
```

```
Job counts:
        count   jobs
        1       hello_world
        1
Resources before job selection: {'_cores': 1, '_nodes': 9223372036854775807}
Ready jobs (1):
        hello_world
Selected jobs (1):
        hello_world
Resources after job selection: {'_cores': 0, '_nodes': 9223372036854775806}
Hello world
[...]

PR #313, opened 2020-04-04
Use 1 core by default

albertas:
Provided default=1 value for cores option in order to solve this issue: https://github.com/snakemake/snakemake/issues/312


johanneskoester:
Thanks a lot, but it was an intentional change. For a detailed explanation, see my comment in issue #312. Sorry for closing!

PR #338, opened 2020-04-18
change to white-list of overridable attributes in Namedlist._set_name

### LINKED ISSUE #337: Namedlist set_name still overrides essential class functions
**Snakemake version**
5.14.0

**Describe the bug**
This is related to previous issue #143, which is marked as closed. The final fix involved a blacklisting the attributes `items()`, `keys()` and `get()` as attributes that cannot overridden by `Namedlist._set_name()` by raising an exception in these cases. This fix is incomplete as it is still possible to break snakemake, such as by accidentally overriding member functions such as `extend()` and `append()`, which are used internally by `Namedlist`. 

**Minimal example**
The following `Snakefile`:
```
rule all:
    input:
        'append.log'

rule override_extend:
    output:
        append = 'append.log',
        test = 'test.log'
    shell:
       """
       echo "append" > {output[0]}
       echo "test" > {output[1]}
       """

```
Result:
```
TypeError in line 10 of ~/Snakefile:
'_IOFile' object is not callable
  File "~/Snakefile", line 10, in <module>
```

**Additional context**
This occurs because in the example because `append()` is overridden in by the `Namedlist` `output` when specifying the first output file. Therefore, appending the second tries to call the overridden attribute `append`, which is no longer a function, leading to the error.

In the commit history, the original fix #145 (commit `ba4cc28`) prevented any Namedlist override, but this was updated in #149 (commit `9d917f3`) because the original fix broke some workflows that relied using `index` and `sort` as attributes. Fix #149, however, is too permissive.

I have a suggested fix that I think better solves both issues, and will submit a pull request.

twlee79:
Fixes #337. 

**Background**
Currently `Namedlist._set_name()` only black-lists `items()`, `keys()` and `get()` as attributes that cannot be overridden by a workflow, but this is demonstrably too permissive (#337). 

**Details**
In my view, blacklisting additional functions (e.g. `extend()` and `append()`) is not a suitable fix. This is because future code that uses `Namedlist` might be written to depend on other (non-blacklisted) existing functions without users being aware of this, leading to errors cropping up where these functions are overridden by workflows. As it is, a user may be unaware that use of `index()` and `sort()` are unsafe and that these are overridden by some existing workflows and could potentially cause unexpected results. Note that any such code will only fail under certain circumstances (e.g. `index` being used as an attribute), so these issues will not be easily detected.

In this fix, all class attributes present in `Namedlist` are disallowed from being overridden (similar to  #145), but there is also a white-list of attributes that _can_ be overridden (`self._allowed_overrides`). This should address the problems mentioned in #149, where `index` and `sort` are used by some existing workflows, and were broken by #145. As a precaution the `__init__` function sets the default values of these white-listed attributes to a function that raises an `AttributeError`, so that users are aware that these functions cannot be called. This acts as a fail-fast (with a clear error message) for any unsafe code that tries to use these functions, e.g. if someone tries to use `Namedlist.sort()`, they will be immediately aware that this is not allowed.

The white-list can be expanded by altering the `self._allowed_overrides` definition. Currently I have set this to only contain `index` and `sort` as these are the only attributes I'm aware of that need white-listing.


johanneskoester:
We now get a pickling error: https://github.com/snakemake/snakemake/pull/338/checks?check_run_id=652626915#step:6:191

Maybe get rid of the lambda and refer to a class method instead (not looked into it in detail though)?

twlee79:
Apologies, I thought I'd run the relevant unit tests but it appears my system was skipping a whole bunch. Fixed by replacing the lambda with `functools.partial` and a static method.

Issue #213, opened 2020-01-31
[question] cores reported as sys.maxsize


vsoch:
This is more of an observation that I had, and I'm just walking through the code to see how it generates. Specifically, when we use a flavor of cloud cluster executor, the number of cores is reported to be an unusually large number. It turns out it's the sys.maxsize:

```python
    if cluster or cluster_sync or drmaa or tibanna or google_lifesciences:
        cores = sys.maxsize
    else:
        nodes = sys.maxsize
```
I'm not sure if this is done intentionally, but if it is, since we print it out as cores to the user, there might be some better way to go about this. I'm opening the issue for now for discussion, I know @moschetti had asked about it. I'll update the issue with an example of the printing if/when I stumble on one (very likely will!) during testing stuffs.

vsoch:
As promised, here is what it looks like to the user running the pipeline:

```
Using shell: /bin/bash
Provided cores: 9223372036854775807
Rules claiming more threads will be scaled down.
Job counts:
	count	jobs
	1	all
	1	build_oncoprint_table
	1	plot_oncoprint
	3
```
so many cores!!

moschetti:
TL;DR - sys.maxsize doesn't count size of system, but reflects the max pointer size, use os.cpu_count() instead

That number is 2**63 - 1
Python's [sys.maxsize](https://docs.python.org/3/library/sys.html#sys.maxsize) lists the maximum value that can be stored in the underlying C compiler's ssize_t variable, so it should report that value on all 64-bit systems (and likewise report 2**31 - 1 on a 32-bit system. It shouldn't be used to count cores.

[os.cpu_count()](https://docs.python.org/3/library/os.html#os.cpu_count) will report the number of cores on the system (as of Python >= 3.4), or len(os.sched_getaffinity(0)) to report the number of usable cores.

So [the code that is setting that cores/nodes](https://github.com/snakemake/snakemake/blob/0607695047290effb44367cd004523e5e3398171/snakemake/__init__.py#L321) should be updated.

[...]

PR #346, opened 2020-04-23
updating sys.maxsize to os.cpu_count() for number cores

vsoch:
this commit will fix issue #213, namely that we should use os.cpu_count() to get the max number of cores and not sys.maxsize. The reason, as stated in the issue, is that sys.maxsize reflects the maximum pointer size and not the maximum number of cores. Locally on my machine I can verify this:

```python
import os
import sys

os.cpu_count()                                                                                                                                                           
# 8
sys.maxsize                                                                                                                                                              
# 9223372036854775807
```

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -335,7 +335,7 @@ def snakemake(
         updated_files = list()
 
     if cluster or cluster_sync or drmaa or tibanna:
-        cores = sys.maxsize
+        cores = os.cpu_count()
     else:
         nodes = sys.maxsize
 


vsoch:
A few points to discuss:

 - there is also a `get_available_cores` function that seems to be used later
 - should we be doing args.cores or os.cpu_count()
 - is this first conditional even necessary if we do parsing later?

It seems like the logic could be cleaned up a bit, but I need to better understand the original reason it was like this.

vsoch:
I don't think this is a desired fix so closing.

PR #359, opened 2020-04-30
setup gcloud

johanneskoester:


dcroote:
The latest build error in the Kubernetes test:
```
The Python 3 packages 'google-cloud-sdk' and `crc32c` need to be installed to use GS remote() file functionality. No module named 'google'
```
looks to be related to the recent `--only-deps` change to the Dockerfile:

Works in `snakemake/snakemake:v5.17.0`:
```bash
$ docker run -it snakemake/snakemake:v5.17.0 python -c "import snakemake"
$ docker run -it snakemake/snakemake:v5.17.0 python -c "import google.cloud"
$
```
But in `snakemake/snakemake:latest`, which is built on 7951ec2:
```bash
$ docker run -it snakemake/snakemake:latest python -c "import snakemake"
$ docker run -it snakemake/snakemake:latest python -c "import google.cloud"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'google'
$
```

Importantly, both are also missing the `crc32c` dependency:
```bash
$ docker run -it snakemake/snakemake:latest python -c "import crc32c"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'crc32c'
$ docker run -it snakemake/snakemake:v5.17.0 python -c "import crc32c"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'crc32c'
$
```

johanneskoester:
Yes, indeed. I am in the process of fixing that.

PR #362, opened 2020-04-30
adding basic functionality for retry to select functions in remotes.GS

vsoch:
This is a very basic, vanilla first shot at adding [google retry](https://googleapis.dev/python/google-api-core/latest/retry.html) to the GS remote. Specifically, I chose the absolute minimum setup (no params) that is supposed to work for exceptions raised by the calling functions, and I chose those functions that are making API calls. In the case that a function is mostly subfunctions making calls (e.g., mtime) I only added retry to those subfunctions. Note that by default:

 - it will handle transient api errors
 - implements exponential backoff
 - default params are

```
 class google.api_core.retry.Retry(predicate=<function if_exception_type.<locals>.if_exception_type_predicate>, initial=1.0, maximum=60.0, multiplier=2.0, deadline=120.0, on_error=None)[source]
```

I added these without modifying tests, thinking that likely the current tests will be able to test. We can update the PR further with customization if necessary.


snakemake/remote/GS.py
```diff
@@ -125,6 +127,7 @@ def size(self):
         else:
             return self._iofile.size_local
 
+    @retry.Retry()
```
johanneskoester, line 130: Do you think we need the internal retry here as well?

vsoch: They are different purposes - the internal one is for the case that the download "works" but the hash is wrong. The external one is for connectivity. If the API is working but there is some blip so my hash is wrong, I would want the internal retry to exist, because otherwise I'd just see an error that the hashes don't match (and the user would likely say "Why can't we retry?") If the API is haven't issues, that would raise the exception caught by the external decorator. So I think we are better having both.

johanneskoester: Yes, but you can simply pass a custom [predicate](https://googleapis.dev/python/google-api-core/latest/retry.html#google.api_core.retry.Retry), can't you? Pass a function that also checks for the exception caused by the hash mismatch (we should add a special exception for that in snakemake/exceptions.py), in addition to what is [default](https://googleapis.dev/python/google-api-core/latest/retry.html#google.api_core.retry.if_transient_error) now.

vsoch: @johanneskoester here are some implementation details:

## Adding CheckSumMismatchException
I first modified the download function to instead of having a loop with retry, to just raise a new exception (defined in snakemake/exceptions.py) for CheckSumMismatchException). Then that would be caught by the decorator. Having the download function raise a special exception type for this case of strings mismatching seems like a logical way to have the predicate know exactly that it happened.

## Writing a Google Storage decorator
I had first started writing the decorator function in snakemake/executors.py, but since we require the retry.is_transient_error function, this would mean trying to do an import in executors.py that might not be relevant for the user (e.g. a Tibanna executor user doesn't care about this Google decorator but it would trigger an error and tell them they need the library). We could move the import into the function, but then we run a similar issue of trying to run the function for some reason without the library, and having the error trigger internal to the check. I thought since we already do this check and then potential raise of import error for the retry library, it would be more straight forward to define the decorator directly in remotes/GS.py, and have it use the CheckSumMismatchException from exceptions.py. The decorator is very simple - it just checks for transient error or this particular kind of new exception. Take a look and let me know if this is what you had in mind!

vsoch:
@johanneskoester many of these code smells are for the original executors file (now moved) is this something I should be concerned about?

vsoch:
I'm going to retry updating with current master to see if that resolves the weirdness.

snakemake/exceptions.py
```diff
@@ -469,6 +468,14 @@ class SpawnedJobError(Exception):
     pass
 
 
+class CheckSumMismatchException(Exception):
```
johanneskoester, line 471: Please inherit from WorkflowError. This way, Snakemake will by default hide the stack trace and just display the message.

snakemake/remote/GS.py
```diff
@@ -23,6 +24,27 @@
     )
 
 
+def google_cloud_retry_decorator(ex):
```
johanneskoester, line 27:
```suggestion
def google_cloud_retry_predicate(ex):
```

snakemake/remote/GS.py
```diff
@@ -125,41 +148,29 @@ def size(self):
...
+        # https://github.com/googleapis/python-storage/issues/29
+        with open(self.local_file(), "wb") as blob_file:
+            parser = Crc32cCalculator(blob_file)
+            self.blob.download_to_file(parser)
+        os.sync()
+
+        # Compute local hash and verify correct
+        if parser.hexdigest() != self.blob.crc32c:
+            os.remove(self.local_file())
+            raise CheckSumMismatchException
```
johanneskoester, line 169: Please add a proper error message here, saying that checksum of file ... was wrong.

snakemake/exceptions.py
vsoch: Ah, good idea!

snakemake/remote/GS.py
vsoch: Gotcha!

PR #371, opened 2020-05-04
Fix security hole in Dockerfile

rebecca-palmer:
(I attempted to report this privately, to no response.  Maybe a spam filter blocked it??)

snakemake's Dockerfile uses `wget -O- http://neuro.debian.net/...` to obtain the Neurodebian repository key and address, then uses them to install the singularity-container package.  As http:// is insecure against malicious modification, this is a remote code execution (inside the container) security hole.

The first commit fixes this by using https instead; the second changes an (unrelated and not such a problem, but might as well while we're thinking about it) link to https.

Not addressed here but maybe something to think about: you seem to be using Neurodebian xenial on top of base stretch, which is both mismatched (Ubuntu vs Debian) and old (2016-7).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,8 @@ ENV PATH /opt/conda/bin:${PATH}
 ENV LANG C.UTF-8
 ENV SHELL /bin/bash
 RUN /bin/bash -c "install_packages wget bzip2 ca-certificates gnupg2 squashfs-tools git && \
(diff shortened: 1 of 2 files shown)


johanneskoester:
Awesome, thanks a lot! Sorry for not seeing your email. I have a huge pile of mails to cope with these days: homeoffice + two children and no daycare :-).

johanneskoester:
Yes, the docker container needs an update. I always wanted to postpone this until a switch to podman, as an alternative to singularity, that will come later this year.

PR #382, opened 2020-05-08
Refactor executors

johanneskoester:
* [x] Directly pass all read jobs to the executor (which still by default handles them one by one).
* [x] Move executors into a full module that allows for submodules.


johanneskoester:
@cmeesters here we go. For now, I left all old executors in one place, but you can already start with a new submodule.

vsoch:
@johanneskoester what is the plan for maintaining some sense of previous contributions? If you just copy paste chunks into new files, all that history and credit will be lost. Perhaps the different contributions should be offered the chance to do it (PR to this Th is branch?) I could give this a test with the GLS executor - even if it’s not fully working just to create the new file.

vsoch:
I think you could try copying the same executors file multiple times so it would maintain the fit history, and then deleting chunks from each? I also found this https://devblogs.microsoft.com/oldnewthing/20190919-00/?p=102904. Maybe it would make sense to ping contributors to executors.py and see what would be best to do. If something needs work in the future we want to be able to use got blame to know who to ping too. :)

johanneskoester:
Very good point! I did not anticipate it. That blog post seems to solve it, so I think it should be fine to just do it, right?

vsoch:
Possibly? I think it would need testing. Having contributors help (and copy paste the executors they wrote) would also be a simple solution.

johanneskoester:
* @SooLee may I ask you to move your tibanna executor into a separate file snakemake/executors/tibanna.py, in a follow-up PR?

* @vsoch may I ask you to move your google executor into a separate file snakemake/executors/google_lifesciences.py?

vsoch:
Definitely! Will get in a PR in a little bit.

PR #400, opened 2020-05-18
moving GoogleLifeSciencesExecutor to be executors/google_lifesciences.py

vsoch:
This is an extension to #382 to move the Google Life Sciences executor out of the current __init__.py in the executors folder and into it's own file. Since we only import the executor in scheduler.py we just need to update the one import there. I also did my best to remove un-used imports in both files.


vsoch:
PR number 400! How cool is that!

PR #401, opened 2020-05-18
adding support for number types to tibanna-config option

nhartwic:
I modified how "tibanna-config" parameters are parsed so that number-like values are converted to floats/ints as appropriate.

Potential concern, its now impossible to pass the string "32" as a value to "tibanna_config". I don't think this is a thing anyone actually wants or needs to do though anymore than anyone would have wanted to pass the string "false" as a value.

Fixes 395#


nhartwic:
Another note, this current implimentation doesn't support negative ints, they will instead by interpretted as floats. Here is an alternative implementation that would support negative ints...

```
elif v == "false":
    v = False
else:
    funcs = [int, float]
    for f in funcs:
        try:
            v = f(v)
            break
        except ValueError:
            pass
tibanna_config_dict.update({k: v})
```

johanneskoester:
I don't think negative ints are needed currently. Thank you!

Issue #404, opened 2020-05-22
request to not ignore exceptions in S3Helper


arjanvandervelde:
Hi!

I'm troubleshooting some intermittent failures with rules involving S3 downloads. It took me a while to figure out that these were S3 download errors, which are silently ignored by snakemake. It'd be great if you could raise the exception or at least print a warning when a S3 download fails....

see: https://github.com/snakemake/snakemake/blob/master/snakemake/remote/S3.py#L280-L292 

Thanks!

johanneskoester:
Fixed with commit 889f2630.

arjanvandervelde:
awesome. thanks!

PR #405, opened 2020-05-22
Fix race condition.

arjanvandervelde:
Solves the issue where a directory would be removed prematurely
if multiple files are downloaded from the same key hierarchy on S3,
in multiple concurrent jobs.  This commit also fixes
https://github.com/snakemake/snakemake/issues/404 where S3 exceptions
were silently ignored.

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -664,8 +664,6 @@ class DAG:
                     logger.info("Removing local output file: {}".format(f))
                     f.remove()
 
-            job.rmdir_empty_remote_dirs()
-
     def jobid(self, job):
         """Return job id of given job."""
         if job.is_group():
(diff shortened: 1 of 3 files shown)


johanneskoester:
Thanks for spotting! This is fixed now in a slightly more generic way in commit 84c545856613f358ed2e8cca8c7cb11bfd755897.

johanneskoester:
And the exception issue is fixed in commit 889f2630.

arjanvandervelde:
> Thanks for spotting! This is fixed now in a slightly more generic way in commit [84c5458](https://github.com/snakemake/snakemake/commit/84c545856613f358ed2e8cca8c7cb11bfd755897).

awesome. thanks!

PR #406, opened 2020-05-24
changed --edit-notebook to use default_url instead

akhanf:
Passing the ipynb notebook as a positional argument forces the browser to open, even if `--no-browser` is used. By using` --NotebookApp.default_url` this does not happen. However, there does not seem to be a way to quit the notebook from the browser if this alternative is used. Any other suggestions?


johanneskoester:
It seems like one can pass the null command to --browser, in which case nothing is opened:
```jupyter notebook --browser ":"```

johanneskoester:
Thanks to drawing my attention to this and explaining why --no-browser does not work (which sounds like a bug to me). I have fixed this now as mentioned above in commit 497ffc3f.

Issue #251, opened 2020-02-28
Execution order and priority


jprnz:
I would like to be able to prioritize the execution order of certain jobs such that their dependencies are not run until they are needed. 

For example, if I have a a rule that makes parts and another that aggregates those parts, I typically see all parts processed first, followed by the aggregation for each target. 

I would like to be able to use either of the methods described in [this post](https://groups.google.com/forum/#!topic/snakemake/9l7TR4Nysoo) to have Snakemake only work on the dependent parts when a new aggregated target is needed.

Here is a minimal example:
```python
targets = ["A", "B", "C"]
parts = ["1", "2", "3"]

rule all:
    input: [v + ".final" for v in targets]

rule make_target:
    input: lambda wc: [wc.target + v for v in parts]
    output: "{target}.final"
    priority: 3
    resources: batch=1
    shell: "echo {input:q} > {output}"

rule make_part:
    output: "{target}{part}"
    shell: "echo > {output}"
```
I would expect the order of the jobs to be:
A1, A2, A3, A.final, B1, B2, B3, B.final, etc.

Regardless if I use `--resources batch=1` or not, I get the following order:
A1, A2, A3, B1, B2, B3 ... A.final, B.final, C.final

I am using Snakemake v5.10.0

Thank you for your help!

jprnz:
@johanneskoester with a little guidance, I would be happy to help contribute to this feature.

PR #409, opened 2020-05-25
Ilp scheduler

FelixMoelder:
Introducing an alternative job scheduler defined as an ilp.
The scheduler selects jobs by priority, number of used threads and used temp files in order to maximize used threads and remove temporary files as fast as possible.
As a result workflows should be finished faster due to the more effective use of available resources.

Open questions:

- The libraries `math` and `pulp` are imported inside the `job_scheduler_ilp` function. This does not look like the right place to me.
- Should the old scheduler be removed or do we keep both and set one of them as optional?


snakemake/scheduler.py
```diff
@@ -522,11 +523,103 @@ def exit_gracefully(self, *args):
...
+    def job_selector_ilp(self, jobs):
+        """
+        Job scheduling by optimization of resource usage by solving ILP using pulp 
+        """
+        import pulp
+        import math
+        from pulp import lpSum
+
+        # assert self.resources["_cores"] > 0
+        scheduled_jobs = [
```
johanneskoester, line 535: If you change this into a dict that has the job object as a key and the LpVariable as a value, you don't have to do the split on the name below, right?

FelixMoelder: Done

jprnz:
This sounds like it could address #251. I am curious to know if user defined resources could also be used in the scheduling process? Thank you for your work on this!

FelixMoelder:
> This sounds like it could address #251. I am curious to know if user defined resources could also be used in the scheduling process? Thank you for your work on this!

Currently, the scheduler optimizes by priority, core load and temp size as we would like to reduce runtime and memory consumption of the workflow.
But if I'm not mistaken adding a custom resource would not solve the issue of your example.
Right now the solver is only looking at the currently executable jobs without optimizing for future executable jobs.
To solve your issue the scheduler needs to look at all existing jobs and prefer those jobs that lead to the job with the highes priority (or custom resource).
As default behavior this would strongly slow down the complete workflow in most cases, waiting for jobs to finish being necessary for executing your prioritized rule.

But maybe I'm just missing something here, so I would discuss that with @johanneskoester when he is back from holidays.

johanneskoester:
@jprnz indeed, this PR will likely solve your issue. We do not hardcode the behavior you intend, but the behavior of greedily running all jobs kind-of level wise is given up here. Instead, we truly optimize by priority and saturation of given cores, while minimizing temp file residence on disk.

johanneskoester:
Custom resources are considered as well, like previously by making sure that they are not exceeded.

jprnz:
Great news--looking forward to seeing it in action!

snakemake/scheduler.py
```diff
@@ -391,7 +392,7 @@ def schedule(self):
                         "Ready jobs ({}):\n\t".format(len(needrun))
                         + "\n\t".join(map(str, needrun))
                     )
-                    run = self.job_selector(needrun)
+                    run = self.job_selector_ilp(needrun)
```
johanneskoester, line 395: Please make this configurable via the command line. `snakemake --scheduler [ilp|greedy]`, with ilp being the default.

johanneskoester (changes requested):
Awesome, tests pass now.

PR #410, opened 2020-05-25
generalization of multiext output

jafors:
Generalization of the `multiext()` command. Allow extensions to start with different characters, not only with `.` .

Example (to catch multiple output files from `macs2 callpeak`):
`output:
    multiext("peaks/{sample}", "_peaks.xls", "_narrowPeaks.peak", "_model.R")`


johanneskoester:
Thanks a lot!

PR #416, opened 2020-05-28
unlock before building DAG

goi42:
There is no reason to build a DAG if the `--unlock` argument is given.

### DIFF
--- a/snakemake/workflow.py
+++ b/snakemake/workflow.py
@@ -648,6 +648,14 @@ class Workflow:
                     "you don't have the permissions?"
                 )
                 return False
+
+        logger.info("Building DAG of jobs...")
+        dag.init()
+        dag.update_checkpoint_dependencies()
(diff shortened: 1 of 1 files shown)


snakemake/workflow.py
```diff
@@ -670,6 +663,13 @@ def files(items):
             )
             return False
 
+        logger.info("Building DAG of jobs...")
+        dag.init()
+        dag.update_checkpoint_dependencies()
+        # check incomplete has to run BEFORE any call to postprocess
+        dag.check_incomplete()
+        dag.check_dynamic()
+
```
johanneskoester, line 672: This has been moved to the wrong place. `self.persistence.lock()` above needs the DAG to be fully computed.

goi42: Thanks for pointing this out. Corrected.

Issue #220, opened 2020-02-06
Allow include directive to accept a Path object


mbhall88:
**Is your feature request related to a problem? Please describe.**
I love the introduction of the `input` and `output` directives taking `Path` objects. One other place I would really like to see this same functionality is in the `include` directive - when including other snakefiles.

**Describe the solution you'd like**
Allow `include` to accept/handle a `pathlib.Path` object.

Something like this would be an example use case:
```py
from pathlib import Path
rules_dir = Path("rules")
include: rules_dir / "map.smk"
```

**Describe alternatives you've considered**
Currently, I just turn the `Path` object into a `str` when passing it to the `include` directive:
```py
from pathlib import Path
rules_dir = Path("rules")
include: str(rules_dir / "map.smk")
```

**Additional context**
It looks as though the code that would need to be altered to allow this is the `snakefile` parameter in this function: https://github.com/snakemake/snakemake/blob/af722a1b827160a5bc5b81f46b2990f6c327028b/snakemake/workflow.py#L892-L931

A very simple solution (hack) would be to just turn the `snakefile` into a `str` in the first line of the function. However, this probably isn't the best long-term solution?  
Outside of that, this function looks to be doing a lot of auxiliary things and so  I don't know if I should be going in and fiddling.  

One comment I would make is having a path object would allow you to easily confirm the existence of the snakefile the user is asking to include and then provide a helpful error message. In the past, I have found that when this include directive fails for that exact reason it usually takes me a while to track down where it is coming from.

engenegr:
It could be connected to https://github.com/snakemake/snakemake/issues/206

I'd like to see snakemake taking care of absolute paths for the rules inside included Snakefiles.

PR #418, opened 2020-05-29
Allow include directive to take a pathlib.Path object

mbhall88:
This addresses #220 .

I have implemented the "easiest" solution to the problem. I tried following the issue upstream, but allowing `snakefile` to be "PathLike" would not be a simple fix I think. However, if you think there is a nicer way of implementing the ability to use `pathlib.Path` objects in `include` directives point me at the relevant code and I can make the changes.

On a side note, I noticed quite a lot of unused `import` statements. Would you like me to clean those up? (Will take one second - thank god for JetBrains/PyCharm)

### DIFF
--- a/snakemake/workflow.py
+++ b/snakemake/workflow.py
@@ -991,6 +991,9 @@ class Workflow:
         """
         Include a snakefile.
         """
+        if isinstance(snakefile, Path):
+            snakefile = str(snakefile)
+
         # check if snakefile is a path to the filesystem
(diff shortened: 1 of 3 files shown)


johanneskoester:
The import hints could be false positives though, since Snakemake parses Snakefiles in the workflow.py namespace. Hence, those imports also define what is availaible to a workflow. If they are cleaned up, one needs to find a good way to pass them [here](https://github.com/snakemake/snakemake/blob/master/snakemake/workflow.py#L1030) instead.

PR #425, opened 2020-06-02
Add the rule name to finished logging statement

nh13:
I find it difficult to see what jobs finished just based on job id, so I added the name of the rule.  More info could be added, for example when running with `-p`.

### DIFF
--- a/snakemake/logging.py
+++ b/snakemake/logging.py
@@ -393,7 +393,7 @@ class Logger:
                     self.logger.warning(indent(msg["msg"]))
             elif level == "job_finished" and not self.quiet:
                 timestamp()
-                self.logger.info("Finished job {}.".format(msg["jobid"]))
+                self.logger.info("Finished job {}: {}.".format(msg["jobid"], msg["name"]))
                 pass
             elif level == "rule_info":
                 self.logger.info(msg["name"])


snakemake/logging.py
```diff
@@ -393,7 +393,7 @@ def timestamp():
                     self.logger.warning(indent(msg["msg"]))
             elif level == "job_finished" and not self.quiet:
                 timestamp()
-                self.logger.info("Finished job {}.".format(msg["jobid"]))
+                self.logger.info("Finished job {}: {}.".format(msg["jobid"], msg["name"]))
```
nh13, line 396: any reason not to use f-strings?

johanneskoester: Yes, Snakemake still support python 3.5. Once that is gone, I'd be happy to change all that. But I am still hesitating.

johanneskoester:
Can you run black on your changes?

nh13:
Reformatted with black but it looks like the checks timed out.

nh13:
@johanneskoester bump

nh13:
@johanneskoester bump

nh13:
@johanneskoester I think this is ready for review (sorry the many year delay)

nh13:
Closing as this feature is now implemented on main - finished job messages already include the rule name (e.g., "Finished jobid: 0 (Rule: all)").

PR #430, opened 2020-06-09
encode script before writing out

odenas:


johanneskoester:
I have fixed it slightly differently in PR #431.

johanneskoester:
But thanks for pointing me to the issue!

PR #442, opened 2020-06-12
fixing bug with hardcoded snakemake-testing

vsoch:
There is a bug that the snakemake-testing bucket is hard coded into the executor - I don't know how we both missed this @johanneskoester ! This is a quick fix for that.


johanneskoester:
Good catch :-)

Issue #413, opened 2020-05-28
ChildIOException when re-running a pipeline that create links to files in a directory output


blaiseli:
**Snakemake version**
5.18.1

**Describe the bug**
I have a workflow with two rules:

Rule `make_test_dir_and_test_file` creates a directory, a subdirectory, and a file within that subdirectory:

```python
rule make_test_dir_and_test_file:
    output:
        directory("test_dir/subdir"),
    shell:
        """
        mkdir -p {output[0]}/
        echo "test" > {output[0]}/test_file
        """
```

The other rule, `make_other_file`, takes as input the directory created by `make_test_dir_and_test_file` and creates in the "main" directory (the parent of its input) a link to the contained file:

```python
rule make_other_file:
    input:
        "test_dir/subdir"
    output:
        "test_dir/other_file"
    shell:
        """
        ln -s subdir/test_file {output[0]}
        """
```

This link is my final target:

```python
rule all:
    input:
        "test_dir/other_file"
```

I can run the workflow once successfully, but when I try to re-run it, I get a `ChildIOException` that tells me that the linked file is a child of it's containing grand-parent directory:
```
ChildIOException:
File/directory is a child to another output:
/home/bli/src/10825_Cecere/Genomes/test_dir/subdir
/home/bli/src/10825_Cecere/Genomes/test_dir/subdir/test_file
```

I do not expect this to happen, because I'm not creating this file in a separate rule: this file was created in the same rule as its containing directory. What I create in the second rule is a link, which resides in another directory (the main `test_dir`).

With another workflow where I remove one level in the file hierarchy (the `subdir` part on the paths), I get a `ChildIOException` right away, but that seems expected, because the link created in the second rule (`test_dir/other_file`) is now indeed a child of the directory created by the first rule (`test_dir`).

**Logs**
```
# Clean output
$ rm -rf test_dir/
$ snakemake --cores 1 --snakefile test4.snakefile 2> log
# Did it work?
$ echo $?
0
$ snakemake --cores 1 --snakefile test4.snakefile 2> log
# Did it work again?
$ echo $?
1
$ cat log 
Building DAG of jobs...
# Re-works after cleaning the output:
$ rm -rf test_dir/
$ snakemake --cores 1 --snakefile test4.snakefile 2> log
$ echo $?
0
```

**Minimal example**

Here is `test4.snakefile`:

```python
rule all:
    input:
        "test_dir/other_file"

rule make_test_dir_and_test_file:
    output:
[...]

PR #447, opened 2020-06-17
bugfix for issue #413: stop using realpath when looking for nested output paths

jmeppley:
I think the use of os.path.realpath does nothing for us when checking for outputs that are children of each other. It has no effect before the outputs exist, and I don't think we want different behavior here after the outputs exist. I also don't see the benefit of resolving links for this, but maybe I'm missing something.

Assuming it can be safely removed, then its removal solves issue #413.


tbooth:
I've also been affected by this issue and the fix here looks reasonable to me. It may still be possible to induce unexpected behaviour by making symlinks to directories, but I think anyone who's doing that gets what they deserve.

alienzj:
Very useful, merge please.

twrightsman:
This issue has also been affecting my workflows. A typical usecase is having additional processing steps for one type of files but skipping those in another. If a file skips this processing step I'll just symlink it to the output of the processing step. Patching the changes in this PR fixes my DAG resolution issues doing this.

tbooth:
Hi, Just another plea to consider merging this fix or equivalent ASAP. I'm having to apply this patch every time after updating Snakemake or else my workflows break.

VGalata:
Another plea to review this patch.

jmeppley:
My first test wouldn't work without altering tests.common.run to fix the no_tmpdir flag. (At least I thought it was a fix). But that seems to have broken at least one windows test.

I've created a new PR (#825) with this fix that should pass all the tests.

PR #452, opened 2020-06-18
Not following links when looking at child outputs.

blaiseli:
This is an attempt to solve issue 413.


blaiseli:
Note: I'm not at all confident this approach is correct. https://github.com/snakemake/snakemake/pull/447 might be a better solution.

PR #459, opened 2020-06-25
Changed py_version detection

jafors:
Changed the detection of the python version installed in the current environment to `sys.version_info` instead of `--version`.

### DIFF
--- a/snakemake/script.py
+++ b/snakemake/script.py
@@ -26,7 +26,6 @@ from snakemake.io import git_content, split_git_path
 from snakemake.deployment import singularity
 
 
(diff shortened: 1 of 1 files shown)


vineetbansal:
I realized that `conda` + `scripts` directives in combination don't work in snakemake for Python 2 environments, because `python --version` writes to `stderr` instead of `stdout` for Python < 3.4 (see https://docs.python.org/dev/whatsnew/3.4.html#changes-in-python-command-behavior). The old implementation thus doesn't get anything from `stdout` and crashes. You beat me to it!

PR #462, opened 2020-06-26
Rearrange conditional statement in debug check

twrightsman:
If cluster mode is enabled then cores is set to `None` on line 363, which results in a `TypeError` in the following comparison statement.


johanneskoester:
Thank you!

PR #467, opened 2020-06-30
use "default" profile if present and nothing else is specified

johanneskoester:
THis behavior can be deactivated with "--profile none".

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -816,6 +816,14 @@ def get_appdirs():
     return APPDIRS
 
 
+def get_default_profile():
+    """Return "default" if default profile exists in one of the search dirs."""
+    dirs = get_appdirs()
+    if get_profile_file("default", "config.yaml"):
+        return "default"
+    return None
+
+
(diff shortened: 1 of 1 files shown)


Hugovdberg:
What is currently holding back this PR? It seems a simple enough fix to publish and keep all the `--cores` naggers happy :-)

Hugovdberg (approved):
The only problem I could see is people with a profile named none will not be able to select that profile. I expect that to be very slim chance, but we could add a warning that that profilename will no longer be accepted in a future version.

deto:
Just wanted to comment here to bump this PR - I think this would be a great addition.

When I run workflows, I need to set certain config params.  Currently there is a risk that I run things and forget to set the profile.   This is often frustrating for me when I wake up the next day and find that jobs failed overnight :(

johanneskoester:
THanks for the bump. I had decided to implement another approach (allowing a default profile to be set via an environment variable). However, I forgot a part of the documentation. This is fixed now: 2b32bba6efa53a9359bc67d8eb8ce06f1106e8ef

PR #473, opened 2020-07-02
Speed improvement to cleanup_workflow

jmeppley:
When working with lots of files in one directory, cleanup_workdir can take minutes because it's running os.listdir() on a large directory once for every time any file is an input or output of any job. So a thousand input files in a folder and a couple rules run on each one would lead to 5 thousand calls to listdir on the same directory. 

This PR make two changes to the logic:

First, it only runs listdir if the input or output file doesn't exist. (If it exists, we know the dir is not empty).

Second, it only runs listdir once per directory. To do this, I removed the empty_dirs property of Job and put all the logic in dag.py


jmeppley:
I don't see the problem with the failed test. It got to "6 of 6 done (100%)"

Issue #480, opened 2020-07-07
Snakemake deletes input directory in dry-run


baraaorabi:
**Snakemake version**
v5.19.3 (installed via conda)

**Describe the bug**
One of the rules has a directory as input.
When running Snakemake in dry mode (`-n` flag), Snakemake deletes the input directory.

**Logs**
```bash
(conda) baraa@machine:~$ mkdir test
(conda) baraa@machine:~$ ll test
total 0
drwxrwxrwx 1 baraa baraa 4096 Jul  7 12:08 ./
drwxr-xr-x 1 baraa baraa 4096 Jul  7 12:08 ../
(conda) baraa@machine:~$ snakemake -np
Building DAG of jobs...
Job counts:
        count   jobs
        1       all
        1       mkdir
        2

[Tue Jul  7 12:09:05 2020]
rule mkdir:
    input: test
    output: test/testing
    jobid: 1

mkdir test/testing

[Tue Jul  7 12:09:05 2020]
localrule all:
    input: test/testing
    jobid: 0

Job counts:
        count   jobs
        1       all
        1       mkdir
        2
This was a dry-run (flag -n). The order of jobs does not reflect the order of execution.
(conda) baraa@machine:~$ snakemake -np
Building DAG of jobs...
MissingInputException in line 5 of /home/baraa/Snakefile:
Missing input files for rule mkdir:
test
(conda) baraa@machine:~$ ll test
ls: cannot access 'test': No such file or directory
```

**Minimal example**

```python
rule all:
    input:
        'test/testing'

rule mkdir:
    input:
        'test'
    output:
        directory('test/testing')
    shell:
        'mkdir {output}'
```
**Additional context**

baraaorabi:
Note: if the input folder is not empty, it seems that Snakemake does not delete it

jmeppley:
This bug may have been added with commit 84c545856613f358ed2e8cca8c7cb11bfd755897 by @johanneskoester. 

It looks like job.empty_remote_dirs() became job.empty_dirs() and it carried over the behavior of finding empty input dirs but now it's being applied to all inputs, not just remotes. I could see how it would make sense to clean up empty remote input dirs, but not other inputs? 

We may want to qualify the cleanup_workdkir method to only delete empty input dirs if they are for remote jobs?

johanneskoester:
Fixed in commit becd36ca.

johanneskoester:
Thanks for reporting!

PR #481, opened 2020-07-07
Fix issue 480: cleanup_workdir should remove input dirs if they are from remotes

jmeppley:
#### #480: Snakemake deletes input directory in dry run.

This changes DAG.cleanup_workdir from deleting all empty input dirs to only those from remote inputs.

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -230,7 +230,7 @@ class DAG:
         for io_dir in set(
             os.path.dirname(io_file)
             for job in self.jobs
-            for io_file in chain(job.output, job.input)
+            for io_file in chain(job.output, job.remote_input)
             if not os.path.exists(io_file)
         ):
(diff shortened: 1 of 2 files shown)


jmeppley:
@johanneskoester I just restored the check from line 776 [here](https://github.com/snakemake/snakemake/blob/eb9b647dfec1ce6b5c5fd0da115dc3215f286220/snakemake/jobs.py#L776).

There are two things that were just style choices that can easily be changed:

I created a new property: Job.remote_input, but it just as easily could be one line added directly to DAG.cleanup_workdir().

The property could cache the result in self._remote_input, but I think this only gets called once at then end of the run. Caching won't save anything, unless I'm wrong about that.

johanneskoester:
Thanks! This is now fixed in a different way in commit becd36ca. I did not mean to clean up anything in dryrun at all. Thanks for pointing me to the issue.

PR #483, opened 2020-07-08
Connection Pool for SFTP and FTP

jmeppley:
This is proof of concept that pooling connections will speed up snakemake for workflows with hundreds of remote inputs.

It allows snakemake to re-use connections when checking the metadata of multiple files without having too many connections open at once.

#### DomainObject
I added a connection_pool property to DomainObject and an abstract connect() method that subclasses need to implement. There are also a couple methods for getting connection parameters from args and kwargs that can be over-ridden as needed. 

#### SFTP and FTP
The STFP.sftpc() and FTP.ftpc() methods have been replaced with the connect() method. And calls to self.ftpc() and self.sftpc() have been replaced with self.connection_pool.item().

I dropped the immediate_close parameter from FTP because I think it's no longer necessary with connection pooling. 

#### ConnectionPool
This is imported from https://github.com/jmeppley/ConnectionPool (also available as `conda install -c jmeppley connectionpool`). There may be better libraries out there.

#### 2000+ RefSeq files over FTP
I was able to use FTP.remote to build a DAG to download the ~2400 protein files that make up the RefSeq 200 release in about 30 minutes. Without connection pooling, it was still working after 4 1/2 days. 

#### 250 to 1000 files over SFTP
I tested with this Snakefile that downloads files over SSH:
```
from snakemake.remote.SFTP import RemoteProvider
SFTP = RemoteProvider(private_key='/home/jmeppley/.ssh/id_rsa')                            
            
N = config.get('count', 250)                                                               
                                                                                           
file_template = 'tyr.soest.hawaii.edu/mnt/delong/dataproducts/mesoscope/mapping/qc/stats/{file}.stats'  
files, = SFTP.glob_wildcards(file_template)                                                
files = files[:N]
            
rule total: 
    input: expand("read_counts/{file}.txt", file=files)                                    
    output: 'report.txt'                                                                   
    shell:
        "cat {input} >> {output}"
            
rule read_count:
    input: SFTP.remote(file_template)
    output: 'read_counts/{file}.txt'                                                       
    shell: """echo -ne "{wildcards.file}\t" > {output}                                     
              grep reads {input} | cut -f 3 >> {output}"""                                 
```

I tested it using dry-runs with a few different values for count:

 * 100 files: 1m 44s with the current snakemake and 3.0s with pooled connections
 * 250 files: 4m 14s with the current snakemake and 3.5s with pooled connections
 * 1000 files: 16m 50s with the current snakemake and 4.5s with pooled connections


johanneskoester (approved):
Awesome, this is a very good idea! Does it only work for FTP?

johanneskoester:
Can you please format your PR with black?

jmeppley:
I'll get the reformatted code up soon. 

It's been a minute since I've looked at this, but I think only a few of the connection types lent them selves to pooling. It looks like I did FTP and SSH/SFTP. HTTP is supposed to disconnect between actions. I think other remotes, eg S3, already have some sort of background object managing the connection.

The question that should be answered before we make this live is what to do about the underlying connection pool code. I had to add a .close() method to the version that's available in pip.

I've been using my personal conda channel for testing, but that's not going to cut it for a real release. I can try to get it into conda-forge. I'm not sure about the process there. I've put things into bioconda, but this is more general.

It's a single file under the MIT License, so we could easily fold it into the snakemake/remote code with credit, but I'm not sure of the legality of that.

I think conda-forge is probably the way to go, but I'm not wedded to anything.

johanneskoester:
Yes, conda-forge sounds good.

jmeppley:
conda-forge will take a little bit because I want to give the original developer a chance to weigh in. The project is only a couple years old.

jmeppley:
I updated the PR code to do a few things:

 * fall back to creating new connections for every action if connection_pool is not installed
 * adds connection_pool as a python dependency in setup.py and test.environment.yml

There is no conda package yet for connection_pool, but the pip version works. I've used the pip version in the environment yaml file.

johanneskoester:
Once the package is on conda-forge, please ping me for merging! Thanks!

jmeppley:
It's in pip. Can we pull it from there?

johanneskoester:
Technically yes, but for consistency, I'd like to have it in conda-forge.

johanneskoester:
Otherwise, the conda snakemake package cannot depend on it.

enryH:
apart from performance, [multiplexing](https://blog.scottlowe.org/2015/12/11/using-ssh-multiplexing/) would be nice in a server setup where only a certain amount of new connections (although successful) are allowed in a certain time-window (20 in 5mins).  Snakemake seems to check access to all files on DAG creations which let's me run into a `paramiko.ssh_exception.SSHException: Unable to connect to server.com: [Errno 111] Connection refused`.

@jmeppley Do you want help packaging [ConnectionPool](https://github.com/jmeppley/ConnectionPool) into a conda package? Pur Python projects should be relatively easy to setup I heard (and I would be interested to do it).

snakemake/remote/__init__.py
```diff
@@ -7,8 +7,17 @@
...
+from functools import partial
 from abc import ABCMeta, abstractmethod
 from wrapt import ObjectProxy
+from contextlib import contextmanager
+
+try:
+    from connection_pool import ConnectionPool
+except ImportError:
+    # we just won't pool connections if it's not installed
+    #  Should there be a warning? Should there be a runtime flag?
```
enryH, line 19: To replace ` "connection_pool" in sys.modules` in L349?

snakemake/remote/__init__.py
```diff
@@ -290,3 +299,101 @@ def local_path(self):
...
+            kwargs_to_use[k] = v
+        for k, v in self.kwargs.items():
+            kwargs_to_use[k] = v
+
+        return args_to_use, kwargs_to_use
+
+    @contextmanager
+    def get_connection(self):
+        """ get a connection from a pool or create a new one """
+        if not self.immediate_close and "connection_pool" in sys.modules:
```
enryH, line 349: could be replace by a flag (see comment above)

jmeppley:
@enryH  I'll gladly accept help. I was able to set up a conda package, but getting it into conda-forge was harder. I'm more familiar with bioconda, but -- since it's more general -- this belongs in conda-forge. 

I got as far as creating a [conda-forge PR](https://github.com/conda-forge/staged-recipes/pull/12683) and going through one round of code-review, but it stalled. 

What exactly are you picturing for a flag in this code? Just a simple boolean when the import fails that we can check later (EG at __init__.py:L347)? Or something more sophisticated?

enryH:
Finally I read the conda-forge docs:) So it seems all [5 checks passed](https://github.com/conda-forge/staged-recipes/pull/12683#event-4137432669) this time?! 

I would simply add a boolean variable specifying it. `CONNECTION_POOL_INSTALLED` or `connection_pool_installed` (I would prefer the upper case as it is a environment variable, although it is set dynamically?!).  

[Most docstrings](https://github.com/jmeppley/ConnectionPool/blob/master/connection_pool/__init__.py) in the original package are Non-English. Do you auto-translate them somehow? (The code is quite clear, but I think English comments are often preferred?) Ah I just see you also use google translate already on some of the non so clear ones:)

jmeppley:
It looks like I finally got all my ducks in a row with conda-forge and they
merged my PR. I think it will take some time to get from the staging
repository to actually being available in conda-forge, but I think all I
can do at this point is wait.

enryH:
upvote:)

johanneskoester:
Sorry for being so slow. The last months were crazy. I try to go through all PRs soon.

PR #484, opened 2020-07-08
Use py-tes client and add support for task resources

uniqueg:
* Replace manual requests to TES instance with [`py-tes`](https://github.com/ohsu-comp-bio/py-tes) client
* Add dependency `py-tes` to `test-environment.yml` (requires installation via `pip` as it is unavailable via `conda`)
* Add support for the following TES task resources:
  * `cpu_cores`: use Snakemake's `resources["_cores"]`
  * `ram_gb`: use Snakemake's `resources["mem_mb"]` divided by 1000
  * `disk_gb`: use Snakemake's `resources["disk_mb"]` divided by 1000
* Add resource requirements to some jobs in `tests/test_tes`

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -18,6 +18,7 @@ import concurrent.futures
 import subprocess
 import signal
 import tempfile
+import tes
 from functools import partial
 from itertools import chain
 from collections import namedtuple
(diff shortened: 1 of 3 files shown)


uniqueg:
@johanneskoester: this is done in collaboration with Sven Twardziok, to be merged into branch `integrate_tes_v2`, so that he can continue based on that.

Issue #448, opened 2020-06-17
Support alternate signal handling for killing shell processes


nh13:
Snakemake currently [kills shell sub-processes](https://github.com/snakemake/snakemake/blob/73627b6ed042fa598ff2a4d48f9d2f71d8a80f9f/snakemake/shell.py#L87-L92) vai the [`subprocess.kill()`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.kill) method, which sends a `SIGKILL` signal on POSIX OSs.  The underlying process has no chance to clean up or exit gracefullly.

While the default `SIGKILL` works well in most cases for snakemake, this does not allow the underlying sub-process to handle any error.  While snakemake may clean up intermediate output files and the like, the sub-process may be want to close a http connection or do some other final cleanup with some remote API (ex. aws batch).  Consider using snakemake to run snakemake with the `--keep-going` option for a silly example.

It would be nice if one of the following would be implemented:
- either send `SIGTERM` via [`subprocess.terminate`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate)
 or `SIGNINT` via [`subprocess.send_signal`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.send_signal), with the former preferred, but consider support configurability for the signal.
- have a timeout such that if the process doesn't terminate using the above signal within the period of time, a `SIGKILL` is called.  This can default to zero to just use `SIGKILL` right away.

Currently, the work-around is to use the `onerror` to parse the logs to find if the given process you care about terminated and do the clean up manually.

nh13:
I should mention, I'd be happy to submit a PR if @johanneskoester think it would be merged.

millerh1:
Hello -- I don't want to write a duplicate of this by mistake so I will add on here. I think it would be a great improvement if this issue with snakemake execution termination could be addressed (text quoted below): https://groups.google.com/forum/#!topic/snakemake/HbKd1_QsMNs

It poses a fairly serious problem when running snakemake in a non-interactive environment such as within a flask application -- where ctrl+C is not possible. It seems like killing snakemake requires grabbing the whole process tree and supplying it to SIGINT. Is that correct? I am fairly novice when it comes to messaging so it would be great to hear your take. 

> Hello, I haven't found anything on the docs about this
> so I made a few tests and it raised some questions.
> 
> all tests were made with snakemake running this rule:
> rule a:
>   output: touch('flag')
>   shell: 'sleep 500'
> 
[...]

PR #492, opened 2020-07-12
Support alternate signal handling for killing shell processes 

nh13:
See: https://github.com/snakemake/snakemake/issues/448

This also changes the behavior of shutdown for local execution.  Previously, if a running job fails, it may not terminate them even when `-k/--keep-going` is not being used.  This is fixed.

I think a test could be added, something like below, but I couldn't get the trapping to work

<details>

<summary>Snakefile</summary>

```
rule all:
    input:
        ['out1.txt', 'out2.txt']


rule test1:
    output: 'out1.txt'
    shell:
        """
        echo out1 > out1.txt
        sleep 1
        exit 1
        """

rule test2:
    output: 'out2.txt'
    shell:
        """
        _handle_sigterm() {{
            echo "trapped: $1"
            echo "handled" > result.txt
            exit 1
        }}
        trap "_handle_sigterm SIGTERM" SIGTERM
        trap -p
        trap -l
        sleep 1000
        echo "unhandled" > result.txt
        echo out2 > out2.txt
        exit 0
        """
```

</details>

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -160,6 +160,7 @@ def snakemake(
     messaging=None,
     edit_notebook=None,
     envvars=None,
+    local_timeout=0
 ):
     """Run snakemake on a given snakefile.
 
@@ -275,6 +276,7 @@ def snakemake(
         log_handler (function):     redirect snakemake output to this custom log handler, a function that takes a log message dictionary (see below) as its only argument (default None). The log message dictionary for the log handler has to following entries:
         keep_incomplete (bool):     keep incomplete output files of failed jobs
         edit_notebook (object):     "notebook.Listen" object to configuring notebook server for interactive editing of a rule notebook. If None, do not edit.
+        local_timeout (int):        The number of seconds to wait until immediately killing a shell job upon workflow failure.  If 0, kill immediately with SIGKILL, otherwise use SIGTERM and upon timeout use SIGKILL.
         log_handler (list):         redirect snakemake output to this list of custom log handler, each a function that takes a log message dictionary (see below) as its only argument (default []). The log message dictionary for the log handler has to following entries:
 
             :level:
@@ -622,6 +624,7 @@ def snakemake(
                     cluster_status=cluster_status,
                     max_jobs_per_second=max_jobs_per_second,
                     max_status_checks_per_second=max_status_checks_per_second,
+                    local_timeout=local_timeout,
                 )
                 success = workflow.execute(
                     targets=targets,
@@ -701,6 +704,7 @@ def snakemake(
                     export_cwl=export_cwl,
                     batch=batch,
                     keepincomplete=keep_incomplete,
+                    local_timeout=local_timeout,
                 )
 
     except BrokenPipeError:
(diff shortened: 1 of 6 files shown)


snakemake/scheduler.py
```diff
@@ -356,18 +362,20 @@ def schedule(self):
 
                 # handle errors
                 if user_kill or (not self.keepgoing and errors):
-                    if user_kill == "graceful":
+                    if (user_kill == "graceful" or self.keepgoing) and running:
                         logger.info(
                             "Will exit after finishing " "currently running jobs."
                         )
-
-                    if not running:
```
nh13, line 364: previously, it would wait for running jobs, even when `keepgoing` is false, as `shutdown()` waits for all futures to complete

snakemake/shell.py
```diff
@@ -87,10 +88,36 @@ def win_command_prefix(cls, cmd):
         cls._win_command_prefix = cmd
 
     @classmethod
-    def kill(cls, jobid):
+    def kill(cls, jobid, timeout=0):
         with cls._lock:
             if jobid in cls._processes:
-                cls._processes[jobid].kill()
+                proc = cls._processes[jobid]
```
nh13, line 94: this section needs careful review

tests/tests.py
```diff
@@ -45,6 +45,9 @@ def test_issue956():
     run(dpath("test_issue956"))
 
 
+def test_issue448():
+    run(dpath("test_issue448"), shouldfail=True)
```
nh13, line 49: this should be added when we have a test that works

nh13:
@johanneskoester bump

snakemake/executors/__init__.py
```diff
@@ -516,11 +518,21 @@ def cached_or_run(self, job, run_func, *args):
...
+    def shutdown(self, jobids=None):
+        if jobids is None:
+            return
+        logger.debug("CPUExecutor shutting down, will kill {} jobs".format(len(jobids)))
+        # kill all jobs!
+        for jobid in jobids:
+            logger.debug("Killing job {}".format(jobid))
+            shell.kill(jobid=jobid, timeout=self.timeout)
+            logger.debug("Killed job {}".format(jobid))
+        logger.debug("Shutting down the thread pool")
```
johanneskoester, line 530: So, actually not killing jobs upon local failure (but instead finish all currently running without scheduling new ones) was intentional, and is also in line with what happens in cluster mode. However, I can understand that it is sometimes interesting to have a different behavior (which one can already get by hitting ctrl+c after the first failure is reported). So, how about adding this behavior as an optional flag `--kill-upon-failure`?

johanneskoester (changes requested):

snakemake/__init__.py
```diff
@@ -1642,6 +1645,14 @@ def get_argument_parser(profile=None):
         "Snakemake will notify the service on errors and completed execution."
         "Currently only slack is supported.",
     )
+    group_behavior.add_argument(
+        "--local-timeout",
+        default=0,
+        type=int,
+        help="The number of seconds to wait until immediately killing a shell job upon "
+        "workflow failure.  If 0, kill immediately with SIGKILL, otherwise use SIGTERM "
+        "and upon timeout use SIGKILL (default: 0)."
```
johanneskoester, line 1654: The timeout can then be an argument of --kill-upon-failure.

nh13:
Closing as stale - branch is 5+ years behind main with major codebase restructure; issue #448 remains open for a fresh implementation.

Issue #352, opened 2020-04-24
Allow conda environment activatoin by name on a per-rule basis


sstadick:
**Problem**
I currently have several snakemake workflows that require different environments on a per-rule basis when running within a docker container. I want the creation of the docker container to build all my conda environments, not snakemake, but I want snakemake to be able to use different environments per-rule.

**Ideal Solution**
Ideally I would like to be able to set a conda environment by name on a per-rule basis. Something like

```snakemake
rule bwa_map:
    input:
        "data/genome.fa",
        "data/samples/A.fastq"
    output:
        "mapped_reads/A.bam"
     condaname:
        "my-alignment-env"  # This is the name of the conda env in the environment
    shell:
        "bwa mem {input} | samtools view -Sb - > {output}"
```

**Current solution**
Currently this is 'solved' by activating the conda env in the `shell` portion of the rule. However this requires a good bit of boilerplate shell code and is finicky. 

EX:

```snakemake
rule bwa_map:
    input:
        "data/genome.fa",
        "data/samples/A.fastq"
    output:
        "mapped_reads/A.bam"
    params:
        conda_env = "my-bwa-env"
    shell:
         'set +eu '
        ' && PS1=dummy ' 
        ' && . $(conda info --base)/etc/profile.d/conda.sh '
        ' && conda activate {params.conda_env} '
        ' && echo $CONDA_PREFIX; ' 
        ' bwa mem {input} | samtools view -Sb - > {output} '
```

dcroote:
The "ideal solution" is possible if **snakemake** is used to create the conda environments in the correct working directory when the Docker image is built. The key phrase in the preceding sentence is "correct working directory," which, for Kubernetes execution, is `/workdir`:
 https://github.com/snakemake/snakemake/blob/2d94d4e217b2b088e1dd873d77ca6a62c8d3d5eb/snakemake/executors.py#L1490

Consequently, running snakemake in `/workdir` with "empty" rules can generate all of the necessary `conda` environments within the Docker image at build time:

`Dockerfile`:
```Dockerfile
FROM snakemake/snakemake:latest

WORKDIR /workdir

COPY envs envs/
COPY create-envs.smk .

RUN conda config --add channels defaults && \
    conda config --add channels bioconda && \
    conda config --add channels conda-forge && \
    snakemake -r --cores 1 --use-conda -s create-envs.smk
```

[...]

PR #494, opened 2020-07-12
add support for named conda environments

nh13:
Use the conda_env directive

Implements: https://github.com/snakemake/snakemake/issues/352

### DIFF
--- a/snakemake/utils.py
+++ b/snakemake/utils.py
@@ -16,6 +16,7 @@ import string
 import shlex
 import sys
 from urllib.parse import urljoin
+import urllib
 
 from snakemake.io import regex, Namedlist, Wildcards, _load_configfile
 from snakemake.logging import logger
@@ -198,6 +199,11 @@ def makedirs(dirnames):
         dirnames = [dirnames]
     for dirname in dirnames:
         os.makedirs(dirname, exist_ok=True)
+
+
+def is_file_or_url(path):
+    """Returns true if the path is a file or a URL"""
+    return urllib.parse.urlparse(path).scheme or os.path.exists(path)
 
 
 def report(
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -290,7 +290,7 @@ class DAG:
 
         if not init_only:
             for env in self.conda_envs.values():
-                if not dryrun or not quiet:
+                if os.path.isfile(env.file) and (not dryrun or not quiet):
                     env.create(dryrun)
 
     def pull_container_imgs(self, dryrun=False, forceall=False, quiet=False):
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -30,7 +30,7 @@ from snakemake.common import DYNAMIC_FILL, group_into_chunks
 from snakemake.deployment import conda, singularity
 from snakemake.output_index import OutputIndex
 from snakemake import workflow
-
+from snakemake.utils import is_file_or_url
 
 PotentialDependency = namedtuple("PotentialDependency", ["file", "jobs", "known"])
 
@@ -290,7 +290,7 @@ class DAG:
 
         if not init_only:
             for env in self.conda_envs.values():
-                if os.path.isfile(env.file) and (not dryrun or not quiet):
+                if is_file_or_url(env.file) and (not dryrun or not quiet):
                     env.create(dryrun)
 
     def pull_container_imgs(self, dryrun=False, forceall=False, quiet=False):
--- a/snakemake/deployment/conda.py
+++ b/snakemake/deployment/conda.py
@@ -214,6 +214,10 @@ class Env:
                 env_file = tmp.name
                 tmp_file = tmp.name
 
+        # If the environment file is really just the name of a conda environment
+        if not os.path.isfile(env_file):
+            return None
+
         env_hash = self.hash
         env_path = self.path
 
--- a/snakemake/jobs.py
+++ b/snakemake/jobs.py
@@ -953,7 +953,11 @@ class Job(AbstractJob):
 
         if self.shadow_dir:
             wait_for_files.append(self.shadow_dir)
-        if self.dag.workflow.use_conda and self.conda_env:
+        if (
+            self.dag.workflow.use_conda
+            and self.conda_env
+            and os.path.isfile(self.conda_env.file)
+        ):
             wait_for_files.append(self.conda_env_path)
         return wait_for_files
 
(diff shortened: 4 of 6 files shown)


snakemake/parser.py
```diff
@@ -427,6 +427,12 @@ class Conda(RuleKeywordState):
     pass
 
 
+class CondaEnv(RuleKeywordState):
+    @property
+    def keyword(self):
+        return "conda_env"
```
nh13, line 433: this can be `condaenv` but using snakecase to make it clearer

snakemake/utils.py
```diff
@@ -555,3 +556,9 @@ def _find_bash_on_windows():
         except FileNotFoundError:
             bashcmd = ""
     return bashcmd if os.path.exists(bashcmd) else None
+
+
+CondaEnvTuple = namedtuple(
+    "CondaEnvTuple",
+    "is_named env"
+)
```
nh13, line 564: This could also be removed, and instead of using `is_named`, we just using `isinstance(conda_env, str)` in its stead.

johanneskoester:
Thank you! I'll try to review this ASAP. Just give me some additional time as there is a deadline tomorrow and I need to release some urgent fixes in Varlociraptor.

johanneskoester:
Further, may I ask you to format the PR with black?

nh13:
@johanneskoester I'll try to disentangle the reliance on an environment file existing, as that's what I was trying to make work.  Likely I'll get to this in early September when I have some time off for side-projects :)

conchoecia:
I take it that this still isn't resolved? Just was trying to figure out calling up a conda env for a single rule today!

nh13:
@conchoecia I haven't had time to get around to this, and have stuck with boiler plate code for now.

I like to store all my conda environments in one place using [`conda-env-builder`](https://github.com/nh13/conda-env-builder) (which I wrote).  It's super handy if I have a single git repo where I'm building multiple docker images (each a snakemake pipeline), each with one or more conda environments.  

In the snakefile I use the following shell boilerplate:
```smk
rule:    
    params:
        conda_env = "some_prebuilt_and_named_env"
    shell:
        """
        set +eu
        && PS1=dummy
        && . $(conda info --base)/etc/profile.d/conda.sh
        && conda activate {params.conda_env};
        set -eu 
        ...
        """

nh13:
@johanneskoester I really tried to reason about how `conda_env` was really used in `snakemake`. So much of the code is predicated on having a valid and present environments YAML file, I had a lot of trouble finding every place it is used (jobs, persistence, cleanup), and coming up with a good abstraction.  So many objects (workflows, rules, persistence) keep copies of things, and it's really hard to tell exactly what (path, `Env` class, string) without types.  Sorry for the rant, I do think that it could really benefit from some refactoring like you suggested, but that's beyond the time I have right now.

I think the latest set of changes are good compromise for a minimal set of changes to get it working.  Thank-you for your consideration!

nh13:
@johanneskoester tests pass, this is ready to go!

mw55309:
Just adding a +1 for this feature - is there a timeline for implementation?

LiterallyUniqueLogin:
I'm also interested in this. I also don't need snakemake to manage my conda environments for me. Given that the code has already been helpfully written, can we get traction on this?

colinbrislawn:
Any updates?

I'm looking forward to using Snakemake with complex (but standardized!) environments like this one:
https://data.qiime2.org/distro/core/qiime2-2021.11-py38-linux-conda.yml

snystrom:
Chiming in to say this would be very useful in a couple of my more esoteric workflows with external non-conda deps.

Kdreval:
This would be useful enhancement of snakemake functionality! Any updates on the timeline for this feature?

johanneskoester (commented):
Thanks a lot for pointing me again to this. Sorry for the delay. This slipped my attention, because, to be honest, I personally rather favor to have the yaml in the workflow definition, because I think this is much more transparent and adaptable. Nevertheless, Snakemake best practices are not a religion and if we can make a change that fits your habits without totally rewriting the way Snakemake works, I am fine with giving it a try.

Nevertheless, I don't want to merge this in the current form, because of the hack using something considered to be a file path as a name instead. I really think the OO solution is worth giving a try. But I know, this is quite some refactoring work, and I can understand that you probably also don't have the time for this. Given the fact that I forgot this PR for such a long time and you patiently waited for it, I will have a look into this today. Maybe I can come up with a solution quickly.

nh13:
Thank you @johanneskoester for giving it a shot. I commend you for making reproducibility easy and the first choice when running Snakemake, but also understand the pragmatic choices that other people (including myself) make that require alternatives. Like you said, we shouldn’t become dogmatic about the right way to ensure reproducibility. In some cases, I want Snakemake just to the excellent workflow engine that it is, and not force me down a specific path to build my environment. I’d argue that it’s great both are included in Snakemake, but should be allowed to be decoupled if wanted. 

I did try to figure out how to implement the OO approach, but like I said, there’s a lot of coupling in the code and as a newcomer to the code it was quite daunting. I’d love to learn and contribute more (along with our team) as we really love Snakemake and use it in our day to day across a large set of clients. 

Seeing how you’d approach this feature could help us understand the code, and later we can add more functionality and/or code docs to help our future selves.  Thank you for your consideration!

colinbrislawn:
Thanks for returning to this PR, @johanneskoester

I'm glad we waited to revisit this issue. Much has changed, so my use case for named environments is different.

My user story: most rules I write call a single program (bwa) or some unix tools (grep + sed). I find it easy to
1) write env.yaml files and --use-conda or
2) call snakemake from within whatever conda environment I am building

But sometimes I use Qiime2. Each version has +400 dependencies, and is installed with conda into an environment with a [fixed name](https://docs.qiime2.org/2021.11/install/native/#install-qiime-2-within-a-conda-environment). Because the conda environment for each release of is consistent, calling `conda activate qiime2-2021.11` should work for anyone already using Qiime2.

Before mamba, installing more copies of qiime2 was very slow. But times have changed! I'll try putting this into an snakemake conda env and see how it goes!

colinbrislawn:
Well, [qiime2-2021.11-py38](https://data.qiime2.org/distro/core/qiime2-2021.11-py38-linux-conda.yml) is causing it's own issues

<details>
  <summary>Error with qiime2-2021.11-py38</summary>
  
Creating conda environment workflow/envs/qiime2-2021.11-py38-linux-conda.yml...
CreateCondaEnvironmentException:
Could not create conda environment from /net/fs01.cluster.com/home/cbrislawn/bin/github-repos/unite-train/workflow/envs/qiime2-2021.11-py38-linux-conda.yml:
Encountered problems while solving.
Problem: nothing provides __glibc >=2.17,<3.0.a0 needed by nodejs-16.12.0-h73e1104_0
Problem: nothing provides __glibc >=2.17,<3.0.a0 needed by nodejs-16.12.0-h73e1104_0 

</details>

For some reason `conda` is able to solve this environment, mamba can not... 🧐

I suppose this shows another use case for named environments.

PR #511, opened 2020-07-23
Fix GS inventory cache by removing blob prefix

dcroote:
I think caching remote files is an excellent addition to snakemake, so first of all thank you @johanneskoester and @vsoch . However, 01cf67fd54f8c86e61a26da1d9b44518414ef20a in v5.19.3 broke kubernetes on GCP for me. This PR fixes the `MissingInputException` errors I'm seeing in nested buckets containing wildcards by removing `prefix=` from the `client.list_blobs()` function.

## Example 

Snakefile:
```python
rule all:
    input:
        expand("project1/results/{sample}/output.txt", sample=["sample1"])

rule do_something:
    input:
        "project1/reads/{sample}/empty-input.txt"
    output:
        "project1/results/{sample}/output.txt"
```

Demonstration:
```bash
# empty-input.txt exists in the sm-test-gs bucket with a nested structure
$ gsutil ls gs://sm-test-gs/project1/reads/sample1/empty-input.txt
gs://sm-test-gs/project1/reads/sample1/empty-input.txt

# v5.19.3 failes (and v5.20.1 as well, not shown)
$ snakemake --version && snakemake --kubernetes --default-remote-provider GS --default-remote-prefix sm-test-gs -n -r
5.19.3
Building DAG of jobs...
MissingInputException in line 6 of /home/derek/gs-sm-test/Snakefile:
Missing input files for rule do_something:
sm-test-gs/project1/reads/sample1/empty-input.txt

# now with v5.19.2
# (also works with this PR branch)
$ snakemake --version && snakemake --kubernetes --default-remote-provider GS --default-remote-prefix sm-test-gs -n -r
5.19.2
Building DAG of jobs...
Job counts:
        count   jobs
        1       all
        1       do_something
        2

[Wed Jul 22 21:32:38 2020]
localrule do_something:
    input: sm-test-gs/project1/reads/sample1/empty-input.txt
    output: sm-test-gs/project1/results/sample1/output.txt
    jobid: 1
    reason: Missing output files: sm-test-gs/project1/results/sample1/output.txt
    wildcards: sample=sample1

[Wed Jul 22 21:32:38 2020]
localrule all:
    input: sm-test-gs/project1/results/sample1/output.txt
    jobid: 0
    reason: Input files updated by another job: sm-test-gs/project1/results/sample1/output.txt

Job counts:
        count   jobs
        1       all
        1       do_something
        2
This was a dry-run (flag -n). The order of jobs does not reflect the order of execution.
```

### DIFF
--- a/snakemake/remote/GS.py
+++ b/snakemake/remote/GS.py
@@ -138,9 +138,7 @@ class RemoteObject(AbstractRemoteObject):
             - cache_mtime
             - cache.size
         """
-        for blob in self.client.list_blobs(
-            self.bucket_name, prefix=os.path.dirname(self.blob.name)
-        ):
+        for blob in self.client.list_blobs(self.bucket_name):
             # By way of being listed, it exists. mtime is a datetime object
             name = "{}/{}".format(blob.bucket.name, blob.name)
             cache.exists_remote[name] = True


vsoch:
AFAIK this wasn’t tested with Kubernetes, which might explain the issue being undetected. I think we would still need the prefix, but have it be an argument passed from where it’s called (and default set to none meaning just the bucket). The issue is that if we are querying a very large public bucket that has a subset of files in some “folder” path, removing the prefix would mean trying to cache the entire bucket, which we often don’t want to do. The strategy implemented takes the Is akin to how Docker sets the context - we use the directory path of a referenced file as that context and don’t go above that. So I agree that we should fix this but perhaps explore a different solution?

johanneskoester:
Mhm, I don't really understand why the prefix should hurt in your case @dcroote. Can you add some debug statements that highlight where it causes problems?

dcroote:
Hi @johanneskoester, thanks for reviewing. The issue is due to the directory hierarchy and how `prefix` uses `os.path.dirname(self.blob.name)`. As @vsoch mentioned, this limits the search to the parent directory but none higher. In the example case, the directories are structured as follows, with the output separated by more than the parent (note `reads/` vs `results/`):
```
    input:
        "project1/reads/{sample}/empty-input.txt"
    output:
        "project1/results/{sample}/output.txt"
```

Adding a few print statements above  https://github.com/snakemake/snakemake/pull/511/files#diff-a7a981381d42d76677652a60c6ad02b2R141 to compare the presence / absence of `prefix`:
```python
print('Prefix: ', [x.name for x in self.client.list_blobs(self.bucket_name, prefix=os.path.dirname(self.blob.name))])
print('No prefix: ', [x.name for x in self.client.list_blobs(self.bucket_name)], '\n')
```
Results in:
```bash
Building DAG of jobs...
Prefix:  []
No prefix:  ['project1/', 'project1/reads/', 'project1/reads/sample1/', 'project1/reads/sample1/empty-input.txt']
```

Here removing `prefix` allows the search to find the `input` file `project1/reads/sample1/empty-input.txt`

vsoch:
I think this is a bug with the google library, and likely has something to do with a permission or the fact that list_blobs returns an iterator without making any API calls. I found a few issues that look similar:

 -  https://github.com/googleapis/google-cloud-python/issues/9079
 -  https://github.com/googleapis/google-cloud-python/issues/920#issuecomment-110376383

Logically it doesn't make sense given the path `project1/results/sample1/output.txt` that the prefix `project1/results/sample1` wouldn't find the file there. Both above report getting empty lists when they should not, and that appears to be what is happening here. I think we should try to debug the issue with the storage client and not remove it from the code - I still don't think indexing above the context of where the user is working, especially for huge buckets, would be an okay thing to do. How about we try something like:

```python
iterator = self.client.list_blobs(delimiter='/', prefix=prefix)
response = iterator._get_next_page_response()
for prefix in response.get("prefixes", []):
    print("gs://%s/%s" %(bucket_name, prefix))
```
And we should also ensure that you are using an up to date version of the client.

dcroote:
Hi @vsoch. Thank you for investigating, but I believe this the empty list output is correct. `project1/results/` is and should be empty because the input file is in `project1/reads/`. Note `reads/` for the input vs. `results/` for the output.

vsoch:
That was just a mistake on my part, I meant to copy paste `project1/reads/sample1/empty-input.txt` but grabbed the other one. If I understand correctly, we should be looking for the input.

dcroote:
Ah, I think I've found the point of confusion. snakemake **should** be looking for the _input,_ but caching is based on the _output_ blob name and its parent directory (via `os.path.dirname(self.blob.name)`). This demonstrates it (print statements added above [here](https://github.com/snakemake/snakemake/pull/511/files#diff-a7a981381d42d76677652a60c6ad02b2R141)) :
```python
print(f'Blob name: {self.blob.name}')
print(f'Directory: {os.path.dirname(self.blob.name)}')
```
Output:
```bash
Blob name: project1/results/sample1/output.txt
Directory: project1/results/sample1
```
The input is in `project1/reads/`, not `project1/results` hence  the `MissingInputException`

johanneskoester:
> Ah, I think I've found the point of confusion. snakemake **should** be looking for the _input,_ but caching is based on the _output_ blob name and its parent directory (via `os.path.dirname(self.blob.name)`). This demonstrates it (print statements added above [here](https://github.com/snakemake/snakemake/pull/511/files#diff-a7a981381d42d76677652a60c6ad02b2R141)) :
> 
> ```python
> print(f'Blob name: {self.blob.name}')
> print(f'Directory: {os.path.dirname(self.blob.name)}')
> ```
> 
> Output:
> 
> ```shell
> Blob name: project1/results/sample1/output.txt
> Directory: project1/results/sample1
> ```
> 
> The input is in `project1/reads/`, not `project1/results` hence the `MissingInputException`

That would at least not be intentional. Inventory caching happens on file level, without knowledge of whether it is input or output.

vsoch:
hey @dcroote ! Please see https://github.com/snakemake/snakemake/issues/546#issuecomment-671527329 for the full debugging of this issue - it was correct that caching at the root would fix it, but not for the reasons we originally thought. I've opened PR https://github.com/snakemake/snakemake/pull/548 to fix the issue so the iocache should work as expected. There is an unrelated windows install issue that hopefully @johanneskoester has seen before.

dcroote:
@vsoch - Success! Thank you! And your PR is a much better solution than I proposed. Closing this (which probably should have been an issue rather than a PR in the first place) in favor of #548.

johanneskoester:
Thanks a lot you two!

PR #527, opened 2020-07-30
highlight as code

enryH:
In the [webversion](https://snakemake.readthedocs.io/en/v5.20.1/snakefiles/deployment.html#using-environment-modules) the \`something\` was shown as italic, missing one dash  before use-envmodules. (-use-envmodules instead of --use-envmodules)


johanneskoester:
Thank you!

PR #529, opened 2020-07-30
Added info about unparseable argument to parse_key_value_arg

terrycojones:
This is completely minimal and could be done in other ways, but I just want to see what you think. I haven't run any tests as I'm not sure how one does that.

I suggest adding this because I just did this:

```sh
$ snakemake --config xxx=33 target
```

and got the "Invalid config definition: Config entries have to be defined as name=value pairs." error message. Turned out that because `--config` can accept many key/value pairs, Python's argparse (I guess) was adding `target` to the list of values assigned to `config`. But the error message didn't show me what was going wrong until I went in and changed the code to print the troublesome value.

This doesn't happen if the user uses `snakemake --config xxx=33 -- target` but not everyone will be in the habit of doing (or even know about) this.

I'm happy to write or fix tests, etc., if someone tells me how to run them.

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -759,7 +759,7 @@ def parse_key_value_arg(arg, errmsg):
     try:
         key, val = arg.split("=", 1)
     except ValueError:
-        raise ValueError(errmsg)
+        raise ValueError(errmsg + ' Unparseable value: %r.' % arg)
     return key, val
 
 


terrycojones:
I just spent a bunch of time trying to figure out something else that would have been easier if the unparsable / invalid values were added to output. In this case I was trying to use `--profile` and ran into the problems described in https://stackoverflow.com/questions/63198867/errors-when-trying-to-set-command-line-values-via-snakemake-profile-file

It took some digging into to find out what was wrong. I printed the `args` coming back from argparse and saw my config values were being added as targets. After rearranging my command line, I got an error as above. I rearranged my command line again and got a different error, this time telling me "Invalid threads definition: entries have to be defined as RULE=THREADS pairs (with THREADS being a positive integer)." but with no value reported. So I added printing of the value and it showed me it was trying to parse the value `trim=7 diamond_dna=5`. Only then did I go back to my `config.yaml` and change

```yaml
set-threads: "trim=7", "diamond_dna=5"
```
to
```yaml
set-threads: ["trim=7", "diamond_dna=5"]
```

and that worked!   So `parse_set_threads` (and `parse_batch`) could be trivially changed to also print the values that are causing the error, and this would save people (people like me, at least) some time trying to debug the awkwardness that comes with some use cases of Python's `argparse`.

While I'm making suggestions, somewhere in the docs it mentions that in a YAML config file you drop the `--` of the option name. It would be good if it also mentioned that in the case of an `nargs=*` option (like `--set-threads` as in my case) that the value on the right hand side must be a list (`["trim=7", "diamond_dna=5"]` in my case), not the string that would normally be put on the command line.

Thanks! :-)

johanneskoester:
Good point, I have added more explanations to the docs (in master).

johanneskoester (approved):

terrycojones:
Thanks @johanneskoester   I'm happy to submit better prs in the future, I pushed this one just to make the suggestion. Anyway, thanks for merging!

PR #531, opened 2020-07-31
ENH: use multi-stage builds + install singularity from source + use nonroot user

kaczmarj:
The Dockerfile prior to this commit install Singularity version 2.x. This commit installs Singularity 3.x from source. This commit also adds multi-stage builds, so unnecessary programs are not present in the final image. The separate stages can be built in parallel with `DOCKER_BUILDKIT=1 docker build .` on systems that have BuildKit.

This commit is a result of the comment https://github.com/snakemake/snakemake/pull/371#issuecomment-625167668. Podman can replace the Singularity installation in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,72 @@
-FROM bitnami/minideb:stretch
-MAINTAINER Johannes Köster <johannes.koester@tu-dortmund.de>
-ADD . /tmp/repo
-WORKDIR /tmp/repo
-ENV PATH /opt/conda/bin:${PATH}
-ENV LANG C.UTF-8
-ENV SHELL /bin/bash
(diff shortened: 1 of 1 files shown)


kaczmarj:
And out of my own interest, I started working on a Dockerfile for podman. It requires user-namespaces to be enabled in the kernel, so I don't know how feasible using podman in Docker will be.

https://gist.github.com/kaczmarj/5e15a449269209a732f602431022f7bd

johanneskoester (approved):
Awesome, thanks a lot!

johanneskoester:
> And out of my own interest, I started working on a Dockerfile for podman. It requires user-namespaces to be enabled in the kernel, so I don't know how feasible using podman in Docker will be.
> 
> https://gist.github.com/kaczmarj/5e15a449269209a732f602431022f7bd

Will user namespaces ever become default?

PR #532, opened 2020-08-01
start of work for wrapper to support --use-singularity if a container…

vsoch:
This is start of work to support defining a container image for a snakemake wrapper. I'll do my best to walk through the logic I've done thus far, and we can discuss questions along the way!

### 1. A Wrapper has a Docker container

The basic idea is that a wrapper can have defined a container, for example in this branch of snakemake-wrappers I've added a  [Dockerfile to bwa/mem](https://github.com/snakemake/snakemake-wrappers/blob/adding-containers/bio/bwa/mem/Dockerfile).  It's actually quite simple - it bootstraps the latest snakemake container, and then adds the installs from the wrapper's environment.yaml.
```
FROM snakemake/snakemake
# docker build -t snakemake-wrappers/bwa/mem
# How to test from test folder:
# snakemake --cores 1 mapped/a.bam --use-conda -F -s Snakefile_samtools
MAINTAINER Johannes Köster <johannes.koester@tu-dortmund.de>
ADD . /tmp/repo
WORKDIR /tmp/repo
RUN /bin/bash -c "source activate snakemake && \
    conda env update --file /tmp/repo/environment.yaml --prune"
```
This way, we can (to some extent) reproduce the conda install in the container. The structure of this particular file should be discussed when there is some PR to the wrappers repository, and actually there are several things we need to discuss:

 - how we could do automated builds given a mono-repo and not modular wrapper repos (this issue https://github.com/snakemake/snakemake-wrappers/issues/150). We would need to, akin to how the CI tests changed wrappers, ensure that when a wrapper is updated it also gets a rebuild and deploy of it's container.
 - how / where to actually store the containers. Since an exec to the docker-daemon:// uri doesn't work (it requires a build with sudo) I threw together a quick quay.io organization for snakemake-wrappers, so I could push my testing container there: https://quay.io/repository/snakemake-wrappers/bwa?tab=tags. We would need to figure out how to handle the namespace, for example in the case of this container it's under `bio/bwa/mem` which is too many levels to represent in the Docker uri.

### 2. A Wrapper defines the container in meta.yaml

If you look at the [meta.yaml](https://github.com/snakemake/snakemake-wrappers/blob/adding-containers/bio/bwa/mem/meta.yaml#L4) file, the reference to the container is there!
```yaml
name: "bwa mem"
description: Map reads using bwa mem, with optional sorting using
  samtools or picard.
container: "docker://quay.io/snakemake-wrappers/bwa:mem"
authors:
  - Johannes Köster
  - Julian de Ruiter
```

so when we are setting up the rule [here](https://github.com/vsoch/snakemake/blob/add/singularity-wrapper/snakemake/workflow.py#L1204) we can easily check for the definition, and set the `ruleinfo.container_img` to use it.

```python
ruleinfo.container_img = snakemake.wrapper.get_container(
                    ruleinfo.wrapper, prefix=self.wrapper_prefix
)
```
This is triggered by the user running the workflow with `--use-singularity` of course, the container can be defined but if the user does `--use-conda` it's largely ignored. The other notable case is if if the user requests --use-singularity and no container is defined for the wrapper, a warning is issued and conda is used. This is per discussion in https://github.com/snakemake/snakemake-wrappers/issues/57#issuecomment-666524588.

### 3. The workflow is run!

At this point, we use the same wrapper logic to run a script, except since a container_img is defined, it uses it instead.
And I should also add that the way I'm testing is to cd in the bio/bwa/mem/test folder, and then update the Snakefile_samtools to be:

```
rule bwa_mem:
    input:
        reads=["reads/{sample}.1.fastq", "reads/{sample}.2.fastq"]
    output:
        "mapped/{sample}.bam"
    log:
        "logs/bwa_mem/{sample}.log"
    params:
        index="genome",
        extra=r"-R '@RG\tID:{sample}\tSM:{sample}'",
        sort="samtools",         # Can be 'none', 'samtools' or 'picard'.
        sort_order="queryname",  # Can be 'queryname' or 'coordinate'.
        sort_extra=""            # Extra args for samtools/picard.
    threads: 8
    wrapper:
        "adding-containers/bio/bwa/mem"
```
Note the wrapper is pointing to my branch where the Dockerfile, meta.yaml with the container exists! And then to run:

```bash
$ snakemake --cores 1 mapped/a.bam --use-singularity -F -s Snakefile_samtools --debug
```

## Other trivial changes / things to discuss

- we definitely need to figure out optimizing the wrapper images. Since they need snakemake, and since that image is a chonker, that would mean all wrapper images are chonkers (and maybe we can figure out how to help with that?)
- "simg" was the extension for now very old versions of singularity. I've updated it to be .sif, understanding that some previous workflows might re-pull the container.  But it should be correctly representing the sif binary so the simg extension is not correct.

So in summary, this PR will allow a snakemake-wrapper to define a container in the meta.yaml, and then use it if the user excplitly requests to `--use-singularity`. I probably missed something because I'm not totally familiar with wrappers (this is my first day looking at the code) but I wanted to at least get us started with something to discuss! I think if we can discuss a strategy for building / deploying containers for each wrapper, and then ensuring the logic is sound here (and add a test) this would be a good start. :) 

c.c. @mbhall88 who has waited so patiently to get started on this! I hope this gets us started, and sorry for the delay!


snakemake/wrapper.py
```diff
@@ -87,6 +88,40 @@ def get_conda_env(path, prefix=None):
     return path + "/environment.yaml"
 
 
+def get_container(path, prefix=None):
+    """Given a path to a rule, load the meta.yaml / meta.yml and determine if
+       a container is provided. If so, return the container URI. Otherwise
```
vsoch, line 93: I started this function mimicking the one to get the conda environment, but it's fairly messy! @johanneskoester let's talk about how to simplify this. We basically need to get a path to the meta.yaml, and read it, and return the container if it's defined.

mbhall88:
@vsoch what a start!

The biggest question I have (which is related to why the feature was requested to begin with) is what is our strategy for programs that *don't* have a conda recipe?

I guess this shows my naivety of how wrappers work but I am unsure why the container would need to have snakemake in it? My initial thought for how this would work is that the creator of the wrapper recipe would point at an image in a container repository and this would just be used as the `container:` directive when running the rule.

vsoch:
@mbhall88 I think the answer would be the same thing that would happen for some wrapper that already doesn't have conda. I'm not super experienced with the wrappers, but I haven't seen one that doesn't do it by way of installing a conda environment. @johanneskoester could you comment? If the user doesn't want conda, and doesn't want singularity, then it will fall back to executing the commands on the host? And if the recipe doesn't have conda or singularity, the we would likely want to do that too.

vsoch:
But it's not logical to me that a user would request a wrapper (with a version) and then just use some version of the software on the host, which is hugely likely not the same version, and not reproducible in terms of the workflow.

mbhall88:
> I think the answer would be the same thing that would happen for some wrapper that already doesn't have conda. I'm not super experienced with the wrappers, but I haven't seen one that doesn't do it by way of installing a conda environment. 

Yes. At the moment it is only possible to create one if a conda recipe exists. But the initial feature on the request https://github.com/snakemake/snakemake-wrappers/issues/57#issue-562606934 was asking for container support to additionally allow for wrappers in the circumstance where the program you want to run *doesn't* have a conda env, or requires a more complicated setup. I have personally run into this a few times where programs I want to run aren't in conda, and I'd like to create a wrapper for the program.

> But it's not logical to me that a user would request a wrapper (with a version) and then just use some version of the software on the host, which is hugely likely not the same version, and not reproducible in terms of sharing the workflow.

You're absolutely correct. To be clear, I am definitely not advocating running wrappers locally in any circumstance.

vsoch:
@mbhall88 I don't see why that wouldn't work, of course we would need to test! With the way it's implemented now, it would set the container image [here](https://github.com/snakemake/snakemake/pull/532/files#diff-f12d9d983123f9ffee4b9375b23b49a2R1205) and then hit [this call](https://github.com/snakemake/snakemake/blob/8194bb958f00a56baa419704068aaa2fe4b5fe6b/snakemake/wrapper.py#L152) for the wrapper, having a script, container image, but no conda. What I'd like to suggest is we figure out a strategy for the wrappers, and then create one (even if it's a test/dummy) that fits this use case, having a container but no conda environment.

mbhall88:
> What I'd like to suggest is we figure out a strategy for the wrappers, and then create one (even if it's a test/dummy) that fits this use case, having a container but no conda environment.

Sounds like a good start. We could use a tool of mine [`rasusa`][rasusa] as a test? There is no wrapper, but I've been meaning to create one. It has a [docker image][docker] and a [conda recipe][conda] so we can play around with all sorts of different methods for it. Plus the docker image is small (~6mb).  
Or would you prefer to work on a pre-existing wrapper?  

I feel like I haven't really done anything yet though so feel free to delegate tasks to me. What can I do?

[rasusa]: https://github.com/mbhall88/rasusa
[docker]: https://hub.docker.com/r/mbhall88/rasusa/tags
[conda]: https://anaconda.org/bioconda/rasusa

vsoch:
Let's work on rausa! And I'm especially excited because it's in rust (maybe you can let me help one some easy learning things, I'm new?). I think what we should do is give a shot at creating a modular wrapper, and possibly testing to see if it can live alongside code (you can imagine a repository like rausa also serving a wrapper, right?) So we have two options:

**Wrapper within source repository**
You could start with a new branch on the rausa repository, and using an example from the [snakemake-wrappers](https://github.com/snakemake/snakemake-wrappers/) create perhaps a subfolder for the wrapper (meta.yaml and wrapper.py or equivalent). Then what we can do is add the Dockerfile alongside it, and a automated build for the container and a recipe. Then we can do tests of this wrapper from a different respository (e.g., referencing it in a Snakefile).

**Wrapper repository**
On the other hand, if you want them to be entirely modular (e.g., perhaps installing your library for the wrapper after some release) then you can give a shot at creating a completely separate repository. if you go this route, I think we should go all in, have some fun, and make a snakemake-wrappers organization. So your wrapper might be at `snakemake-wrappers/rausa`, and akin to how the snakemake-wrappers repo has subfolders for topics (e.g., bio) we would have these same topics as GitHub repository topics. This would also let me test creating a global snakemake-wrappers.github.io repository that can read metadata for all of them, and then create a nice interface to explore, etc.

Rubber ducking through this now - I like the second idea better!  Let's go for - I just created the organization and invited you: https://github.com/snakemake-wrappers. @johanneskoester I invited you too, and don't worry  if the design doesn't work out we can just delete it. This will also give me the change to take a shot at converting a current wrapper there and seeing how we like it. What do you think?

vsoch:
And if you don't have bandwidth soon, I'd love to take the first shot! I get excited about pretty much anything involving rust :)

mbhall88:
Yes, I think the second option seems a better idea. It's more inline with other similar projects I guess. So just trying to get this straight in my head, each wrapper has it's own repo within the organisation? One question (further down the road) would be how would a first-time contributor create a new repo? I quite like the conda-forge approach (however I'm sure this is quite complex) when you fork the staged-recipes repo, add your recipe, and then it magically gets it's own feedstock repo created after the PR is merged.  

Anyway, as this is just playing around for now let's go with your suggestion. So I create a rasusa repo, and then add a meta.yaml and wrapper.py? And we're specifying the container in the meta.yaml either as a path or a URI?

vsoch:
> how would a first-time contributor create a new repo

I think we would give them a template, have them create on their own, and when it's ready (and it's desired on both sides) transfer to snakemake-wrappers and put them in charge as a maintainer. I love conda forge too but it is rather complex - I think for the time being we should "keep it simple stupid" and have each wrapper repository run it's own CI, deploy the container, etc. However I do think the metadata file would need to be expanded a bit, minimally to include some of the same things that the conda example has so we can easily parse and share the wrappers.

Where/who would we contact about learning more about how conda forge is working? I'm guessing there is an external server somewhere, but I'm not sure to be honest. It would be good to start learning now so if we ever decide to make it more sophisticated we are off to a good start. But for now, don't think too much about that, because we're just testing the modular wrapper repository to being with.

> Anyway, as this is just playing around for now let's go with your suggestion. So I create a rasusa repo, and then add a meta.yaml and wrapper.py? And we're specifying the container in the meta.yaml either as a path or a URI?

The container would have to be a uri (e.g., `docker://quay.io/snakemake-wrappers/<name>`) when you are ready to do the automated build ping me and I'll help set it up.

vsoch:
And you can use the branch that I was testing as a dummy example - https://github.com/snakemake/snakemake-wrappers/tree/adding-containers/bio/bwa/mem. We basically just need that content plus the test for the wrapper (a single test extracted out of [test.py](https://github.com/snakemake/snakemake-wrappers/blob/adding-containers/test.py) and then an automated build - deploy to quay.io with a robot token specifically for that repo).

johanneskoester:
> @vsoch what a start!
> 
> The biggest question I have (which is related to why the feature was requested to begin with) is what is our strategy for programs that _don't_ have a conda recipe?
> 
> I guess this shows my naivety of how wrappers work but I am unsure why the container would need to have snakemake in it? My initial thought for how this would work is that the creator of the wrapper recipe would point at an image in a container repository and this would just be used as the `container:` directive when running the rule.

Each wrapper comes with a Python, R, or Julia script (wrapper.py, wrapper.R, wrapper.jl). This script has access to the job properties via a magically inserted `snakemake` object (e.g. to access input files via `snakemake.input`). Now, in the Python case, this Snakemake object is generated via pickling from the snakemake codebase. As such, loading it requires snakemake to be available in the wrapper script. In conda world, Snakemake ensures this via adding the snakemake install path into sys.path of the wrapper script, regardless of the environment definition.

With containers, this won't work as easy anymore. There are two solutions to fix this:

1. Mounting the snakemake install dir into the container and still use sys.path.
2. Perform code generation to generate the `snakemake` object instead of pickling. The code of the Snakemake object is rather simple, and it would be easy to do that, e.g. via jinja templating.

As option 2 also gives us more flexibility regarding the eventually used python versions in the wrapper, I would prefer option 2.

johanneskoester:
> If the user doesn't want conda, and doesn't want singularity, then it will fall back to executing the commands on the host?

Yes

> And if the recipe doesn't have conda or singularity, the we would likely want to do that too.

I don't want to allow wrappers that have neither conda nor a container image. The reason is that this will make it impossible to test them automatically, and also heavily limit their reproducibility.

johanneskoester:
In general I would like to suggest the following:

1. Each wrapper must either define a conda environment, an available container image (via meta.yaml) or a Dockerfile.
2. Only one of above options is allowed at a time (otherwise we get problems with keeping them in sync).
3. Not defining a conda environment is only allowed if it is inevitable, i.e. technical or license restrictions prohibit the creation of a conda image (this should be confirmed by the author via a checkmark in the PR template).
4. In case of a conda environment definition, we generate the container image automatically (in the way you have shown at the top). Tagging and naming happens via hashing the conda environment file. This way, there is also no need for separating wrappers into individual repos, because we can simply check presence of the hashes before building. I would say that the name is a hash of the environment with removed versions, and lexicographically sorted package names. The tag is a hash of the versions, in the same order as the package names. This is inspired by the [mulled toolkit](https://galaxy-lib.readthedocs.io/en/latest/topics/mulled.html#) from the galaxy team. We might even be able to use mulled for this, although I think calculating the hash is pretty simple.

In the long run, I would like to replace 4 with a docker-buildkit based setup, which should allow us, in combination with the topologically sorted conda dependency DAG to build each conda package into its own layer. This way, we do not waste space for e.g. multiple uses samtools in separate environments.

vsoch:
Thank you @johanneskoester! How do you see allowing a meta.yaml or a Dockerfile? Wouldn’t the Dockerfile need to be built into a container that is ready to use, referenced in the meta.yaml? Or are you saying the name of the container image should be predictable without that?

vsoch:
Taking a look at mulled - wow that's neat! My one question is whether or not it needs sudo to do the build, which would make it hard to insert in the middle of a pipeline run. It's the middle of the night here but I'll definitely test this out tomorrow / later today!

vsoch:
@mbhall88 so with @johanneskoester wisdom, I think we want to take a similar approach, but no separate snakemake-wrappers repository. Maybe if you want to try doing a PR to the snakemake-wrappers repository with your wrapper organized where it belongs? I need to look into the mulled toolkit to better understand how it would work.

mbhall88:
> @mbhall88 so with @johanneskoester wisdom, I think we want to take a similar approach, but no separate snakemake-wrappers repository. Maybe if you want to try doing a PR to the snakemake-wrappers repository with your wrapper organized where it belongs? I need to look into the mulled toolkit to better understand how it would work.

Ok. Should I do this with the container directive in the meta.yaml for us to test out?

vsoch:
That would work! I need to test out the mulled toolkit to better understand how it could fit in with building a custom image at runtime (and without needing root!) It seems too good to be true to be able to build a container with Singularity (possibly on HPC) where you don't have sudo. I think the approach of having a Dockerfile that installs the same conda recipe and then deploys to the container specified in the meta.yaml is a good fallback if that doesn't work.

mbhall88:
> That would work! I need to test out the mulled toolkit to better understand how it could fit in with building a custom image at runtime (and without needing root!) It seems too good to be true to be able to build a container with Singularity (possibly on HPC) where you don't have sudo. I think the approach of having a Dockerfile that installs the same conda recipe and then deploys to the container specified in the meta.yaml is a good fallback if that doesn't work.

I've created [a PR](https://github.com/snakemake/snakemake-wrappers/pull/152) to add a `rasusa` wrapper and have added a container URI. I also added an environment.yaml just in case (and also for when it is eventually merged).

johanneskoester:
> > That would work! I need to test out the mulled toolkit to better understand how it could fit in with building a custom image at runtime (and without needing root!) It seems too good to be true to be able to build a container with Singularity (possibly on HPC) where you don't have sudo. I think the approach of having a Dockerfile that installs the same conda recipe and then deploys to the container specified in the meta.yaml is a good fallback if that doesn't work.
> 
> I've created [a PR](https://github.com/snakemake/snakemake-wrappers/pull/152) to add a `rasusa` wrapper and have added a container URI to the meta.yaml. I also added an environment.yaml just in case (and also for when it is eventually merged). Not sure if this is all in the correct format, but it's a good starting place I guess as it passes tests so the functionality is correct - now we just need to play with the container stuff I guess.

@mbhall88 this looks good to me, but the (not yet implemented) policy would be that the container image directive in the meta.yaml is only allowed if it is impossible to provide the necessary software stack via a conda package. From looking at rasusa, this does not seem to be the case. If we allow it like you do in the PR, it would be potentially hard to keep the two in sync. I'd rather say one should just define the conda environment in such a case, and the snakemake-wrappers CI pipeline would automatically build and upload a corresponding container from the conda env. Does that make sense or am I missing something?

johanneskoester:
> Thank you @johanneskoester! How do you see allowing a meta.yaml or a Dockerfile? Wouldn’t the Dockerfile need to be built into a container that is ready to use, referenced in the meta.yaml? Or are you saying the name of the container image should be predictable without that?

In case a dockerfile is provided, there is no need to specify anything in the meta.yaml. Then name under which the snakemake-wrappers CI pipeline would build and upload it should be predictable from the conda environment.

johanneskoester:
> That would work! I need to test out the mulled toolkit to better understand how it could fit in with building a custom image at runtime (and without needing root!) It seems too good to be true to be able to build a container with Singularity (possibly on HPC) where you don't have sudo. I think the approach of having a Dockerfile that installs the same conda recipe and then deploys to the container specified in the meta.yaml is a good fallback if that doesn't work.

My idea was that the container is built in the CI pipeline, not when the user uses a wrapper. Hence, root permissions are not an issue. Also, I think it should be a docker container image (or OCI) instead of singularity. Snakemake will at some point allow podman or other container runtimes instead of singularity.

vsoch:
> 
    Mounting the snakemake install dir into the container and still use sys.path.
    Perform code generation to generate the snakemake object instead of pickling. The code of the Snakemake object is rather simple, and it would be easy to do that, e.g. via jinja templating.

@johanneskoester I think by way of the wrapper calling [script](https://github.com/snakemake/snakemake/blob/master/snakemake/wrapper.py#L117) which then takes into account a singularity image and args and passes to [shell](https://github.com/snakemake/snakemake/blob/e43cb8842a60db833d8a3e26c138bf1930bf4a9c/snakemake/script.py#L346) on the execute_cmd call, we would already be handling the mount? But I do see that we are [picking](https://github.com/snakemake/snakemake/blob/e43cb8842a60db833d8a3e26c138bf1930bf4a9c/snakemake/script.py#L411) and not creating the code snippet as we would with an executor. I'll give that another test run so I can look at the snakemake content before pickling and see if I have ideas for the implementation.
 
And I think I understand the build and naming logic now! I'm going to give a shot at adding that CI to the snakemake-wrappers repository. I'm going to start simple and build containers from the conda recipe and name based on hashes of packages/versions, and we'll need to think more about how to have something like docker-buildkit with layers (I haven't used it so I might just need to read about it). Anyway, I'll get us started on something. Stay tuned!

vsoch:
@mbhall88 @johanneskoester please see https://github.com/snakemake/snakemake-wrappers/pull/153, where I started work on the wrappers repository for the automated container builds, generation of the hash, and discussion for how we can automate deploy and then interact with snakemake here. Let's have discussion over there, and then based on what we decide I can return to work on things here (and @mbhall88 you can update your wrapper PR to follow suite!)

mbhall88:
> @mbhall88 this looks good to me, but the (not yet implemented) policy would be that the container image directive in the meta.yaml is only allowed if it is impossible to provide the necessary software stack via a conda package. 

Yes, I agree. I was just using this as an example (i.e. pretending it doesn't have a conda build).

docs/conf.py
```diff
@@ -17,242 +17,237 @@
 import os
 from recommonmark.parser import CommonMarkParser
 
-source_parsers = {'.md': CommonMarkParser}
+source_parsers = {".md": CommonMarkParser}
```
vsoch, line 20: Sorry ran black on this file - I can restore to the one on master if you don't want those changes here!

snakemake/deployment/docker.py
```diff
@@ -0,0 +1,119 @@
+__author__ = "Johannes Köster"
+__copyright__ = "Copyright 2015-2020, Johannes Köster"
+__email__ = "johannes.koester@uni-due.de"
+__license__ = "MIT"
+
+import os
+import re
+import requests
+
+
```
vsoch, line 10: This is the start of a DockerContainer class that I thought would provide a minimal wrapper to parse a container uri (e.g., get the registry, repository, and version) and then do enough query to the manifest endpoint to determine if it exists or not. This class can be used by the WrapperSpec to determine if a given uri exists (or not). If it doesn't exist, we skip over using a container for the given wrapper.

snakemake/deployment/docker.py
```diff
@@ -0,0 +1,119 @@
...
+class DockerContainer:
+    """A Docker Container is a thin wrapper to a Docker Registry API to
+       primarily determine if a container exists to be used for a workflow
+    """
+
+    def __init__(self, name):
+        self.name = name
+        self.uri = None
+        if self.exists():
```
vsoch, line 19: This is fairly simple, but if a container exists, the DockerContainer will have a uri (that starts with docker:// to support how Snakemake currently expects to pull a singularity container). If the container doesn't exist, this returns None.

snakemake/deployment/docker.py
```diff
@@ -0,0 +1,119 @@
...
+            self.uri = "docker://%s" % self.name
+
+    @property
+    def version(self):
```
vsoch, line 23: The version property can take a tag (e.g., latest) or a container hash (specified with `@` as a version). For snakemake wrapper containers, the version will always be the hash of the conda environment versions. For external wrappers, or wrappers otherwise given a docker:// uri, the version can be a tag or hash.

snakemake/deployment/docker.py
```diff
@@ -0,0 +1,119 @@
...
+        elif ":" in self.name:
+            container_name, version = self.name.split(":", 1)
+        return version
+
+    @property
+    def registry(self):
+        """Return the container registry. We default to Docker Hub unless
+           another (quay.io) is defined
+        """
+        registry = "https://index.docker.io/v2"
```
vsoch, line 40: We might eventually add more registries that implement the Docker Distribution API - to start I thought Docker Hub and Quay.io were reasonable. But heads up that Docker Hub just announced they are going to start purging containers that aren't used (push or pull) after some time, so we definitely should prefer quay.

snakemake/deployment/docker.py
```diff
@@ -0,0 +1,119 @@
...
+           (quay.io/snakemake-wrappers) otherwise return False
+        """
+        # Request for the manifest
+        # GET /v2/<name>/manifests/<reference>
+        base = "%s/%s/manifests/%s" % (self.registry, self.repository, self.version)
+        headers = {"Docker-Distribution-API-Version": "registry/2.0"}
+        response = requests.get(base, headers=headers)
+
+        # Repository does not exist
+        if response.status_code == 404:
```
vsoch, line 64: I tested the 404 returned for Quay.io, and if a container doesn't exist it will return 404. I still need to test when a container does exist (if Quay also requires an authentication token). For Docker Hub, if the container doesn't exist _or_ it doesn't have a token, it returns 401 either way.

snakemake/deployment/singularity.py
```diff
@@ -76,7 +76,7 @@ def pull(self, dryrun=False):
                         "singularity",
                         "pull",
                         "--name",
-                        "{}.simg".format(self.hash),
+                        "{}.sif".format(self.hash),
```
vsoch, line 79: simg hasn't been around for years now, and most containers are sif, so I updated this.

snakemake/script.py
```diff
@@ -109,25 +109,22 @@ def log_fmt_shell(self, stdout=True, stderr=True, append=False):
...
 
-class REncoder:
-    """Encoding Pyton data structures into R."""
-
-    @classmethod
-    def encode_numeric(cls, value):
-        if value is None:
-            return "as.numeric(NA)"
-        return str(value)
+class BaseEncoder:
```
vsoch, line 112: I realized that there is a lot of redundant code for the PythonEncoder (which I added) the REncoder and the JuiaEncoder, so instead I made a BaseEncoder class that has the common functions.

snakemake/script.py
```diff
@@ -109,25 +109,22 @@ def log_fmt_shell(self, stdout=True, stderr=True, append=False):
...
+    """shared class for encoding a value. An Encoder subclass should have
+       defined: language, none_value, true_value, and false_value, all strings,
+       along with methods encode_dict, and encode_list.
+    """
 
     @classmethod
     def encode_value(cls, value):
         if value is None:
-            return "NULL"
+            return cls.none_value
```
vsoch, line 121: This also means that each encoder class provides it's own true, false, and none values.

snakemake/script.py
```diff
@@ -142,7 +139,59 @@ def encode_value(cls, value):
                     return str(value)
             except ImportError:
                 pass
-        raise ValueError("Unsupported value for conversion into R: {}".format(value))
+        raise ValueError(
+            "Unsupported value for conversion into {}: {}".format(cls.language, value)
+        )
+
+
+class PythonEncoder(BaseEncoder):
```
vsoch, line 147: Here is the PythonEncoder that I added.

snakemake/script.py
```diff
@@ -295,6 +321,9 @@ def __init__(
     def evaluate(self, edit=False):
         assert not edit or self.editable
 
+        # import code
```
vsoch, line 324: oops will take this out for the next push!

snakemake/script.py
```diff
@@ -381,42 +410,112 @@ def generate_preamble(
...
             searchpath += ", " + repr(os.path.dirname(path[7:]))
 
+        # The wrapper path should be added too
+        wrapper_path = path[7:] if path.startswith("file://") else path
+
         return textwrap.dedent(
             """
         ######## snakemake preamble start (automatically inserted, do not edit) ########
-        import sys; sys.path.extend([{searchpath}]); import pickle; snakemake = pickle.loads({snakemake}); from snakemake.logging import logger; logger.printshellcmds = {printshellcmds}; {preamble_addendum}
+        import sys, os; sys.path.extend([{searchpath}]); sys.path.insert(0, {sourcedir});from collections import namedtuple
```
vsoch, line 431: @johanneskoester here is what we talked about, removing snakemake and the picking. This isn't going to work for a lot of wrappers that use snakemake.shell, but it should work for future wrappers that don't. If the snakemake path that is added doesn't exist, I don't think it should hurt. But if it does exist, snakemake should be found akin to before.

snakemake/script.py
```diff
@@ -381,42 +410,112 @@ def generate_preamble(
...
+                        self.__dict__.update(entries)
+                    def get(self, key, default=None):
+                        if hasattr(self, key):
+                            return getattr(self, key)
+                        return default
+
+                structure = Item(items)
+                setattr(self, name, structure)
+ 
+            def source(self):
```
vsoch, line 460: I didn't understand if/how this is used - do I need it?

snakemake/script.py
```diff
@@ -381,42 +410,112 @@ def generate_preamble(
...
+                wd = os.getcwd()
+                os.chdir(self.scriptdir)
+                sys.path.insert(self.scriptdir)
+                os.chdir(wd)
+                
+            def log_fmt_shell(self, stdout=True, stderr=True, append=False):
```
vsoch, line 466: This had a format / weird error I couldn't figure out when it was represented as it was before (with the Snakemake object). When I wrote it out like this, it worked ok.

snakemake/script.py
```diff
@@ -381,42 +410,112 @@ def generate_preamble(
...
+            threads = {threads},
+            log = {log},
+            resources = {resources},
+            config = {config},
+            rule = {rule},
+            bench_iteration = {bench_iteration},
+            scriptdir = {scriptdir},
+        )
+        # If needed, add snakemake.shell
+        try:
```
vsoch, line 496: Here is where we can add snakemake functions, if snakemake can be imported.

snakemake/workflow.py
```diff
@@ -1196,11 +1196,26 @@ def decorate(ruleinfo):
...
+
+            # Using a snakemake-wrapper
             if ruleinfo.wrapper:
-                rule.conda_env = snakemake.wrapper.get_conda_env(
-                    ruleinfo.wrapper, prefix=self.wrapper_prefix
-                )
-                # TODO retrieve suitable singularity image
+
+                # Instantiate the wrapper to interact with scripts, etc.
+                rule_wrapper = WrapperSpec(ruleinfo.wrapper, prefix=self.wrapper_prefix)
```
vsoch, line 1204: The update here is that we instantiate a WrapperSpec before interacting with it.

snakemake/wrapper.py
```diff
@@ -1,18 +1,204 @@
...
+CONTAINER_PREFIX = os.environ.get(
+    "SNAKEMAKE_WRAPPER_CONTAINER_PREFIX", "quay.io/snakemake-wrappers"
+)
+WRAPPER_REPOSITORY = os.environ.get(
+    "SNAKEMAKE_WRAPPER_REPOSITORY",
+    "https://raw.githubusercontent.com/snakemake/snakemake-wrappers/",
+)
+
+
+class WrapperSpec:
```
vsoch, line 31: Here is the WrapperSpec! I had trouble writing this because I don't fully understand all the options that might be given for a path for a snakemake wrapper (e.g., git+file? file:?). I think the logic here can be hugely simplified to not have so many supporting functions - but I don't understand it enough to do it.

vsoch:
I'm honestly not sure why that test failed - there must be some issue with interacting with a script now, but I don't see it.

vsoch:
We are going to want to add a lot of tests here, I don't like the idea of making large changes without making sure everything still works.

vsoch:
Looks like the test for an empty log failed, I'll need to debug this later.

vsoch:
@johanneskoester there was a timeout, and Ive been trying to debug for hours and am having trouble reproducing on my machine. Other tests seem to fail but I can't debug them interactively, so I need to ask for your help.

vsoch:
okay I'm fixing the code smell and running into the issue that with quay.io, there isn't a reasonable way to see if an image exists via the manifest. Unlike docker, Quay requires the manifest to be the sha256. If we try to look up image ids for a repository, we can get a list of those (but then there is nothing to do with them). Here are the docs in case someone else has ideas https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/red_hat_quay_api_guide/using_the_red_hat_quay_api#accessing_the_red_hat_quay_api_from_the_command_line.

vsoch:
Asking here https://twitter.com/vsoch/status/1295429159435419648. Maybe in the meantime we just can't support quay, which means needing to push to snakemake-wrappers on Docker Hub. With their recent purge policy in place, this is not great!

snakemake/script.py
johanneskoester: Ok, so this PR needs to be combined with a PR on snakemake-wrappers that gets rid of snakemake.shell. `os.system(f"command with {flags}")` is equally readable.

johanneskoester:
> Asking here https://twitter.com/vsoch/status/1295429159435419648. Maybe in the meantime we just can't support quay, which means needing to push to snakemake-wrappers on Docker Hub. With their recent purge policy in place, this is not great!

Ok, let's wait on the answer.

vsoch:
okay, I'll get the snakemake-wrappers PR underway, and I pinged one of my old colleagues at RedHat that can hopefully help us!

vsoch:
hey @johanneskoester ! Just to update you on this PR, I just pushed changes so the quay.io "does this container exist" functionality should work.

snakemake/script.py
johanneskoester: Yes, people requested it before. Some scripts will use it.

snakemake/script.py
```diff
@@ -381,47 +410,127 @@ def generate_preamble(
...
+            from snakemake import shell
+        except:
+            print("Snakemake not found on path, skipping addition of shell.")
+            pass
```
johanneskoester, line 510:
```suggestion
        def shell(cmd):
            import inspect, os
            context = inspect.currentframe().f_back.f_locals
            os.system(cmd.format(**context))
```

snakemake/script.py
```diff
@@ -381,47 +410,127 @@ def generate_preamble(
...
 
+        # The wrapper path should be added too
+        wrapper_path = path[7:] if path.startswith(file_prefix) else path
+
         return textwrap.dedent(
             """
         ######## snakemake preamble start (automatically inserted, do not edit) ########
-        import sys; sys.path.extend([{searchpath}]); import pickle; snakemake = pickle.loads({snakemake}); from snakemake.logging import logger; logger.printshellcmds = {printshellcmds}; {preamble_addendum}
+        import sys, os; sys.path.extend([{searchpath}]); sys.path.insert(0, {sourcedir});
+        class Snakemake:
```
johanneskoester, line 432: The entire template should be put into a separate file now, also the R and Julia ones. Otherwise, this just becomes too big.

johanneskoester (changes requested):

snakemake/script.py
vsoch: Should we turn script into a folder module, akin to executors?

docs/snakefiles/modularization.rst
```diff
@@ -46,6 +46,68 @@ Examples for each wrapper can be found in the READMEs located in the wrapper sub
 The `Snakemake Wrapper Repository`_ is meant as a collaborative project and pull requests are very welcome.
 
 
+-----------------------
+Containers for Wrappers
+-----------------------
+
+As of version 5.23 and later, Snakemake wrappers support the definition of containers to be used
+alongside wrappers. The container will be used given that the user requesting the wrapper adds ``--use-singularity`` 
+to the snakemake run. There are several ways to specify a wrapper for a container, each coming down to a definition in the ``meta.yaml``,
```
johanneskoester, line 55:
```suggestion
to the snakemake run. There are several ways to specify a container for a wrapper, each coming down to a definition in the ``meta.yaml``,
```

docs/snakefiles/modularization.rst
```diff
@@ -46,6 +46,68 @@ Examples for each wrapper can be found in the READMEs located in the wrapper sub
...
+.. code-block:: yaml
+
+    container:
+      spec: Dockerfile
+      reason: |
+        It is impossible to generate this container based on a conda environment because...
+
+
+It's likely this case won't be needed and we can remove it. If a user needs to define
+ a custom Dockerfile, it's better that they just build and deploy it and provide the container name.
```
johanneskoester, line 108: Agreed, let's then keep this commented out in the docs.

johanneskoester (changes requested):

a-slide:
Hi all
I forgot about the issue but I saw that @vsoch did an amazing work to provide container support for wrappers.
This will be a great new functionality. Could we expect to see it in the next snakemake release ?
Thanks

nikostr:
Hey! What is the status of this PR? Would love to be able to integrate containers into wrappers.

johanneskoester:
So actually this is kind of obsoleted by the new containerization functionality of Snakemake. The idea is that workflow development can rely entirely on conda packages, and upon production one can automatically generate a container image. The advantage is that this works for both pure conda and pure container execution, whereas allowing containers in wrappers would lead to (a) unnecessary maintenance burden if wrappers define both a container and a conda env, and (b) wrappers becoming not usable in all scenarios if they would only provide a container.

vsoch:
ok let’s close then.

nikostr:
Perhaps this should be a separate issue, but I think it is somewhat related: a lot of useful software is available as a container, with containers being provided by the maintainer as a part of how the software is released. This is e.g. true for DeepVariant, GLNexus, but surely for a ton of other tools as well. Currently, if these are to be used in wrappers, they first need to be ported to (bio)conda, and a wrapper needs to be written that references this conda package. This adds overhead, with conda recipes of varying quality being written by maintainers of varying experience and capability to handle new releases etc. In practice this has meant that when I wanted to write a snakemake workflow using DeepVariant and GLNexus, I first had to update the (messy) DeepVariant bioconda recipe, and then update the snakemake wrapper. As no GLNexus bioconda recipe existed I ended up not using a wrapper for this, instead using the container and writing the necessary commands to be run directly in the workflow - functional, but would be neat to have as a wrapper instead.

To me it would make sense to have a possibility to write snakemake wrappers that directly reference a container rather than a conda package. Would this make any sense at all, and would it be doable?

PR #533, opened 2020-08-03
WIP: Azure Batch integration

andreas-wilm:
This is an attempt to build native Azure Batch support into Snakemake. See https://andreas-wilm.github.io/2020-07-30-azure-batch-in-snakemake/ for more info. There are quite a few kinks that need ironing out, but the code already works. I've created this PR so that others can give it a try and comment. Remaining issues (copied from the blog post above):

- Azure Batch parameters are provided to Snakemake using a yaml file with `--az-batch-config`. Instead separate arguments should be used.
- Batch autoscaling doesn’t work. I gave this only a quick try using one of the standard formulas, but the pool never increased from 0. Therefore autoscaling is currently disabled (see `create_pool()`)
- The workflow resources are not deleted from blob during shutdown()
- By default Azure Batch only places one job per node, which is inefficient
- The Azure Batch retry option is hardwired to two (see `create_job()`)
- Code cleanup: Lots of minor issues marked with `FIXME` in the code
- Untested: when using instances with extra disks, is the space made available in Docker?
- Untested: use of spot instances


johanneskoester:
> This is an attempt to build native Azure Batch support into Snakemake. See https://andreas-wilm.github.io/2020-07-30-azure-batch-in-snakemake/ for more info. There are quite a few kinks that need ironing out, but the code already works. I've created this PR so that others can give it a try and comment. Remaining issues (copied from the blog post above):
> 
> * Azure Batch parameters are provided to Snakemake using a yaml file with `--az-batch-config`. Instead separate arguments should be used.

You can do it like with tibanna: `--tibanna-config arg1=value1 arg2=value2 ...`.

> * The workflow resources are not deleted from blob during shutdown()

This is quite important I think.

> * By default Azure Batch only places one job per node, which is inefficient

We could add a todo that this can be improved in the future. Let's get it out first.

> * The Azure Batch retry option is hardwired to two (see `create_job()`)

Since Snakemake has its own restart handling ([restart_times](https://github.com/snakemake/snakemake/blob/master/snakemake/jobs.py#L1047)), I think it should be set to zero in Azure Batch.

snakemake/__init__.py
```diff
@@ -2069,6 +2094,7 @@ def adjust_path(f):
             or (args.tibanna and not args.default_resources)
             or (args.google_lifesciences and not args.default_resources)
         ):
+        # FIXME do we need default resources for az_batch as well?
```
johanneskoester, line 2097: I think that would be good in any case.

jakevc:
Would love to see this pulled in @andreas-wilm @johanneskoester.

KrisJanssen:
> Would love to see this pulled in @andreas-wilm @johanneskoester.

Same here... would be happy to help iron out some of the remaining issues. Could we get this going?

andreas-wilm:
Sorry folks, I unfortunately don't even have an Azure account anymore and switched back to AWS at work. Happy to provide contacts within Microsoft (please DM).

jakevc:
Closing this as it is now implemented in #2246

Issue #457, opened 2020-06-24
Feature requests for the google life sciences backend


johanneskoester:
The Google Life Science API backend for Snakemake, developed by @vsoch and funded by Google, has been released some time ago. While the cooperation continues, we are happy to consider feature requests regarding Google cloud support by the community.

Please post your feature requests here.

aryam7:
Hello Johannes and Vanessa,
Thank you for adding the Google Life Sciences backend to Snakemake!
Do you offer support for running rules within [preemtible VM instances](https://cloud.google.com/life-sciences/docs/reference/gcloud-examples#using_preemptible_vms), by any chance? Perhaps this could be specified as a resource, like
```
rule a:
    output:
        "test.txt"
    resources:
        preemptible=True
    shell:
        "somecommand ..."
```

vsoch:
Yes, it looks like we can! https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/projects.locations.pipelines/run#VirtualMachine I'll get in a PR to test this week.

vsoch:
Update: please see https://github.com/snakemake/snakemake/pull/549 for the preemptible instance option.

kdbinder:
@johanneskoester @vsoch 

Would be ideal to automatically log stderr into a logging bucket within GCS instead of relying on the end user to add a "log" statement in their Snakefile

aryam7:
I'll echo @kdbinder's request. The `stderr` from `gcloud beta lifesciences operations describe` is often only partial. And if the error happens prior to the execution of the rule, the log output (from using the `log` directive) will be empty. This can make it difficult to debug problems that might be occurring due to the Life Sciences Executor.
But maybe I'm missing something. @vsoch, how did you debug the executor when you were first developing it?

The Google Life Sciences API seems to automatically mount [a `/google` virtual disk with complete log output](https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/projects.locations.pipelines/run#action) whenever it executes a job. Would it be possible to copy the contents of `/google` into the storage bucket specified by the user? For example, the files could go into the `source` directory in a folder named by the life sciences operation ID, like this
```
snakemake-testing-data/source/logs/11698975339184312706
```

This option could be enabled via a command line flag, like `--google-lifesciences-logs`, which takes as an optional argument the list of rules whose log output should be copied. If no rules are specified, it could just copy the logs of all of the rules.

aryam7:
[...]

PR #549, opened 2020-08-10
Adding option for preemptible google cloud instance

vsoch:
As discussed in https://github.com/snakemake/snakemake/issues/457#issuecomment-671116746, this pull request will add support for using preemptible as a resource option in a Google Life Sciences workflow (and I suppose it would be ignored otherwise). 

I added one of the steps to be preemptible so we can test it during snakemake testing, [here](https://github.com/snakemake/snakemake/compare/master...vsoch:add/preemtible-resource-gs?expand=1#diff-f3a69272fb2bf2786691175ecdec9fc9R30).
I also checked the instance running to ensure that setting [it to true](https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/projects.locations.pipelines/run#VirtualMachine) indeed resulted in this instance type:

image

The default will be False, as I assume most users wouldn't expect this.


vsoch:
@johanneskoester I've added those additional resources to the main resources documentation! If it's not organized as you had in mind, please feel free to suggest -> commit (or of course just instruct me how to make changes). I didn't group preemptible with standard resources because it's only added for GLS, but if it's added for other cloud executors we could move it up there.

vsoch:
@johanneskoester what about pushing a dev image before we run tests on push to master and then we can use it for the google cloud executors? That way we can be sure if a previous bug was fixed. We could even delete the image tag right after the run.

aryam7:
Perhaps the docs should also caution that the preemptible option should be run with a `--restart-times` value?

Or maybe setting preemptible to true should automatically set `--restart-times` to a reasonable value for that job?

vsoch:
@aryam7 I don't see an option exposed in the Google Lifesciences API for restart times - could you point me to what you have in mind? https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/projects.locations.pipelines/run#VirtualMachine

aryam7:
ah! I was referring to Snakemake's `--restart-times`

[Preemptible instances](https://cloud.google.com/compute/docs/instances/preemptible) can be terminated by Google Cloud at any time for any reason, but they incur much lower costs. I think it would only make sense for someone to use preemptible instances if they could have Snakemake retry terminated jobs.

johanneskoester:
Mhm, can we see from life science API whether a preemptible job failed because of google killing it (because of the peemptible flag) or because of other issues? In that case, the preemption cause should lead to infinite repeats of that job.

kdbinder:
You can monitor the metadata of an instance to see if it was preempted

`curl "http://metadata.google.internal/computeMetadata/v1/instance/preempted?wait_for_change=true" -H "Metadata-Flavor: Google"`

see [here](https://cloud.google.com/compute/docs/instances/create-start-preemptible-instance#detecting_if_an_instance_was_preempted)

vsoch:
Thanks I’ll check this out!

aryam7:
If the job is preempted, it [should return an exit code of 14](https://cloud.google.com/life-sciences/docs/troubleshooting#error_codes).

One way to implement @kdbinder's approach could be to allow the preemptible resource to be a callable with the number of attempts as a parameter, like [Snakemake already does for the `mem_mb` resource](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#resources).

vsoch:
okay so we have a few options! Given that the life science worker can communicate being preemted with exit code 14, here are the options:

### 1. Preemptible as boolean with default retry.

If the user speifies `preemptible=True` this would be akin to using preemptible instances with some default number of retries, but only given that the exit code is 14. @johanneskoester you suggested infinite, but I'm weary of this because if a VM has a maximum time of 24 hours (and the user doesn't know this) it will essentially go on forever. We could have infinite retries but spit out warning messages about the maximum time, or have preemptible retry to some default value (e.g., 10?). This seems like the simplest option because the user can just request preemptible and not think about it.

### 2. Preemptible as integer to indicate retry

We could instead require the user to specify the number of retries off the bat, e.g., `preemptible=10` and this would again only be in the case that the exit code is 14. But then what would happen if the user set both preemptible and restart times? The scheduler [handles restarts](https://github.com/snakemake/snakemake/blob/cc9d4fcf8411934330801ec75c57afa7ae876e9f/snakemake/scheduler.py#L510) and this would get messy or confusing. 

### 3. Preemptible as boolean with restart times

I think the cleanest implementation would be to keep preemptible as a boolean, and:

 - case 1: if the user doesn't specify restart-times, we set it to a reasonable default (e.g., 10) for the step
 - case 2: if the user specifies --restart-times we use that with pre-emptible just as we would any other step.

The third solution above would not have any confusion between using restart times and preemptible. And if would sort of change into the first solution if the user doesn't specify a number of restarts. I think what we would want to make sure is possible is specifying the restart-times in the job resources:

```
    rule:
        input:    ...
        output:   ...
        resources:
            preemptible=True, restart-times=10
        shell:
            "..."
```
which actually makes me fall back to the other suggestion, we could simplify like this:

```
    rule:
        input:    ...
        output:   ...
        resources:
            preemptible=10
        shell:
            "..."
```
So in summary, why not allow the user to use a boolean or an integer? if an integer is provided, we set that to be the job's restart times. If not, we use a reasonable default. Thoughts?

aryam7:
Thanks for laying out the options, @vsoch!

I'm not really a fan of allowing the `preemptible` resource to be set equal to an integer. If somebody wants that behavior, they can just set it to [a callable with the number of attempts](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#resources), right? This is what I originally envisioned doing when I submitted this feature request.

I don't have strong feelings about a default number of retries as long as it can be overridden by the strategy in the callable, especially if the default is initially set to infinity.

vsoch:
I don't really have strong feelings either - even if we set it to be unlimited that would work given enough clear messages to the user about how it's working. I'll leave this decision up to @johanneskoester.

johanneskoester:
Thanks for this inspiring discussion. A different perspective: is preemptible something that should be in the workflow at all? It is not platform independent. What about this instead:

* Allowing preemption for all rules, with 10 repeats: `snakemake --preemptible default=10`
* Overwriting specific preemption for particular riöes: `snakemake --preemptible default=10 map_reads=3 call_variants=0`
* No preemption by default, but for a particular rule: `snakemake --preemptible default=0 map_reads=10`

vsoch:
@johanneskoester this part:

> A different perspective: is preemptible something that should be in the workflow at all? It is not platform independent. What about this instead:

Namely "it is not platform independent" tells me that the answer to https://github.com/snakemake/snakemake/issues/573 might in fact be that (in the case that cores is platform independent, because with drmaa we don't define any?) the workflow file should not be referencing this variable for that exact reason.

What do others think about having the preemptible args? I can refactor to fit this idea.

aryam7:
## tl;dr
> It is not platform independent.

I find this argument to be really convincing. I didn't realize that resources definitions were supposed to be defined independently from the execution environment. In that case, I think I favor the idea of having preemptible args.

## The implications
But let's just make sure we're making the implications of this decision explicit.
1. By moving the preemptible option into the arguments, you imply that the preemptibility of each rule should not be decided by the developer of the Snakefile. It should be solely the user's decision.
2. Moreover, since all other resource definitions must be platform independent, _any_ option which is specific to a particular cluster environment must be solely the user's decision. The developer should have no say because Snakefiles should be platform independent.

I am inclined to agree with the implications above. And I now have a newfound respect for the design decisions that went into Snakemake!

## A potential counter-argument
Is there any value in the developer making decisions about preemptibility before the user? I can think of at least one situation.
For the sake of the argument, let's say preemptibility has an adverse impact on long-running rules; so rules that take a long time to execute are more likely to be preempted during their lifetime than ones that execute quickly. (I have no idea whether this is true, but it seems plausible.) Thus, the developer decides to only add preemptibility to steps in the Snakefile that take a short amount of time.
**Note:** that the user may later decide to override these decisions via Snakemake's `--resources` parameter. So it's not a matter of whether the developer gets to make the choice or the user. The user always gets to make the ultimate choice.

There are undoubtedly other situations that I haven't thought of. For example, what if the developer wanted to set preemptibility for a rule based on its wildcards?
In addition, you can probably expand this counter-argument to any resource definition (as I did for implication # 2).
Also, are you planning on adding arguments to Snakemake whenever this sort of things happens in the future?

## A solution to the potential counter-argument
If you want the best of both worlds, you could allow resource definitions passed via `--resources` to be rule-specific. For example,
```
snakemake --resources preemptible=1
```
could become
```
snakemake --resources map_reads.preemptible=3
```
But then, of course, you compromise on the somewhat noble goal of ensuring that Snakefiles are utterly platform independent.

Thanks for reading till the end! I kinda got away with myself there, but this is all interesting stuff.

vsoch:
@aryam7 indeed the design is very well thought out! To your point here:

> Is there any value in the developer making decisions about preemptibility before the user? I can think of at least one situation.
For the sake of the argument, let's say preemptibility has an adverse impact on long-running rules; so rules that take a long time to execute are more likely to be preempted during their lifetime than ones that execute quickly. (I have no idea whether this is true, but it seems plausible.) Thus, the developer decides to only add preemptibility to steps in the Snakefile that take a short amount of time.
Note: that the user may later decide to override these decisions via Snakemake's --resources parameter. So it's not a matter of whether the developer gets to make the choice or the user. The user always gets to make the ultimate choice.

I would argue that there is another level of representation that a developer still could embed the needed logic. If we consider the unit of packaging to be one up from the Snakefile, and to include an executor environment (e.g., some specific provider that does/does not support preemptible instances) the developer would provide the "correct" command for the user, and/or you could imagine it being built into a container with the entrypoint having it. In other words, the snakefile should be agnostic to this setting because it's more properly defined one level up.

This is a cool idea:

```
snakemake --resources map_reads.preemptible=3
```
but my intuition says that if a user wants a preemptible instance for one rule, likely they want it for most! So this would create a very busy command:

```bash
snakemake --resources map_reads.preemptible=3 --resources rule_2.preemptible=3 --resources call_variants.preemptible=0
```
when we could have just done
```bash
snakemake --preemptible-rules map_reads=3 rule_2=3 call_variants=0
```
And I suspect if we allow passing via a resources argument on the command line, it could be confusing for users why they can't specify it via the resource in a Snakefile.

aryam7:
@vsoch sounds good! Yes, embedding logic one level above Snakemake certainly works. That's a really good point.

Yeah. When I gave that example for passing rule-specific resources via the `--resources` parameter, it was under the assumption that the developer _would be_ allowed to specify the resources in the Snakefile.
So to be clear, I was not endorsing that option. I was just trying to highlight it as a (sort-of) compromise between the two competing design philosophies. Like, a more flexible option: the developer specifies their recommended resources for each rule and the user overrides whatever they want to.
And I agree that the shorter syntax seems nicer.

Thanks for considering my ideas, at all!

vsoch:
Ah understood! And of course @aryam7, I think it’s really important that we have this discussion, and consider one another’s ideas. Hearing your thoughts is very important to me.

vsoch:
okay, PR has been updated to what we discussed!

johanneskoester:
Thanks for evaluating the implications. I think the developer might in such case rather define a job as short running. And we could allow to infer preemption for short running jobs via the command line. For this to work properly, we should define `runtime_minutes` (or another reasonable name: this is to be discussed, walltime? seconds, minutes?) as a new standard resource in the docs (under rules->resources), and add an option `--preemption-default-max-runtime MINUTES` that allows to only set the default for rules with a runtime that not exceeds the defined value.

vsoch:
@johanneskoester which comment are you responding to? I'd like to clarify what you are thinking before moving forward.

- runtime_minutes would be a resource specification independent of pre-emptibility. It's not the minutes allowed to run, but the number of minutes that the user expects the step to take? How would they know? It seems appropriate for a wrapper but maybe not a random Snakefile.
- If runtime_minutes is defined but preemption is not, it essentially does nothing
- If runtime_minutes is defined and --preemption-default-max-runtime MINUTES is defined (for the default) - we skip over jobs that are greater than the limit, the rest are made pre-emptible.
- If runtime_minutes is defined and --preemption-default-max-runtime MINUTES is not, the first again does nothing.
- If --preemption-default-max-runtime MINUTES is defined without runtime_minutes, we can't do anything either.

I think there could confusion if runtime_minutes is an estimated ETA or the allowable minutes. And for the other points we would need to put a lot of warnings so the user is aware of one setting without the other. I'm also concerned a user might think it would indicate pre-emption "but only run for this many/max number of minutes."

vsoch:
Oh, and --preemption-default-max-runtime MINUTES would have a default value of 24 hours (1440 minutes) since that's the max for any preemptimble instance I think.

johanneskoester:
> @johanneskoester which comment are you responding to? I'd like to clarify what you are thinking before moving forward.
> 
>     * runtime_minutes would be a resource specification independent of pre-emptibility. It's not the minutes allowed to run, but the number of minutes that the user expects the step to take? How would they know? It seems appropriate for a wrapper but maybe not a random Snakefile.
> 
>     * If runtime_minutes is defined but preemption is not, it essentially does nothing
> 
>     * If runtime_minutes is defined and --preemption-default-max-runtime MINUTES is defined (for the default) - we skip over jobs that are greater than the limit, the rest are made pre-emptible.
> 
>     * If runtime_minutes is defined and --preemption-default-max-runtime MINUTES is not, the first again does nothing.
> 
>     * If --preemption-default-max-runtime MINUTES is defined without runtime_minutes, we can't do anything either.
> 
> 
> I think there could confusion if runtime_minutes is an estimated ETA or the allowable minutes. And for the other points we would need to put a lot of warnings so the user is aware of one setting without the other. I'm also concerned a user might think it would indicate pre-emption "but only run for this many/max number of minutes."

Agreed, it can indeed be confused easily. runtime_minutes would still be of use without preemption, e.g. for cluster systems where the queue is selected based on runtime. And yes, it is meant as an ETA, hence probably not the best name. In any case, nothing that would need to be decided in the PR.

johanneskoester:
So, one final thing @vsoch: I think this PR is still missing a test case. Can you add one?

vsoch:
Can do! I had added a pre-emptible instance to be used for the [pack step](https://github.com/snakemake/snakemake/pull/549/files#diff-d285009a26f303fdfb0e499e7a8888a1R50) of the Google Life Sciences test - what other tests did you have in mind?

johanneskoester:
> Can do! I had added a pre-emptible instance to be used for the [pack step](https://github.com/snakemake/snakemake/pull/549/files#diff-d285009a26f303fdfb0e499e7a8888a1R50) of the Google Life Sciences test - what other tests did you have in mind?

That's enough. I did not see that, thank you!

johanneskoester (approved):

aryam7:
thank you so much for implementing this, @vsoch!

vsoch:
My pleasure @aryam7!

PR #554, opened 2020-08-12
Fixing quotes for --default-resources example

vsoch:
The quotes for any --default-resources command line arguments need to be around the whole group.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,10 +72,10 @@ setup(
         "reports": ["jinja2", "networkx", "pygments", "pygraphviz"],
         "messaging": ["slacker"],
         "google-cloud": [
-            "crc32c",
             "oauth2client",
-            "google-api-python-client==1.9.1",
-            "google-cloud-storage==1.28.1",
+            "google-crc32c",
+            "google-api-python-client",
+            "google-cloud-storage",
         ],
     },
     classifiers=[
--- a/snakemake/remote/GS.py
+++ b/snakemake/remote/GS.py
@@ -18,7 +18,7 @@ try:
     import google.cloud
     from google.cloud import storage
     from google.api_core import retry
-    from crc32c import crc32
+    from google_crc32c import Checksum
 except ImportError as e:
     raise WorkflowError(
         "The Python 3 packages 'google-cloud-sdk' and `crc32c` "
@@ -56,7 +56,7 @@ class Crc32cCalculator:
 
     def __init__(self, fileobj):
         self._fileobj = fileobj
-        self.digest = 0
+        self.checksum = Checksum()
 
     def write(self, chunk):
         self._fileobj.write(chunk)
@@ -65,14 +65,14 @@ class Crc32cCalculator:
     def _update(self, chunk):
         """Given a chunk from the read in file, update the hexdigest
         """
-        self.digest = crc32(chunk, self.digest)
+        self.checksum.update(chunk)
 
     def hexdigest(self):
         """Return the hexdigest of the hasher.
            The Base64 encoded CRC32c is in big-endian byte order.
            See https://cloud.google.com/storage/docs/hashes-etags
         """
-        return base64.b64encode(struct.pack(">I", self.digest)).decode("utf-8")
+        return base64.b64encode(self.checksum.digest()).decode("utf-8")
 
 
 class RemoteProvider(AbstractRemoteProvider):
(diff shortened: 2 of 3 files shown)


vsoch:
@johanneskoester I'm worried that we are still seeing the checksum mismatch, even with duration set to 600s. Should we adjust the Retry maximum parameter as well? Maybe 30 minutes, and it’s happening because of the memory limits of the CI?

vsoch:
Ah! So this last run gave some insight - when running on GitHub Workflows the crc32c hash is calculated and missing the last two characters, the ==. There isn’t any issue with the retry and what’s happening is that it’s just trying a million times up to the limit. Here we see that:

259F9FFB-026B-4ECF-909E-A7D7376C2F5E

So we need to figure out why the hash is missing those two characters... I can take a look again tomorrow, and I’m wondering if it’s worth trying the google library in case it wouldn’t truncate. @johanneskoester feel free to push tests or updates to this PR but let’s not merge until we have things fixed.

vsoch:
Could there be something wrong with the self.blob.crc32c? If I remember correctly the previous hash did end in ==, but the blob doesn’t have it anymore. I wonder if there were any changes to its calculation? Both the libraries we’ve tried here have been consistent with before. But that wouldn’t make sense why it’s only triggering for the workflow.

vsoch:
Here is a quick grab of a blob from May 25th (a bucket that was created but not removed?) from the command line - the hash calculation has the two `==`!

```bash
$ gsutil ls -L gs://t-2c5c1ca1-2b63-4feb-8813-52e34926519a/landsat-data.txt.bz2 | grep Hash
    Hash (crc32c):          oobFiQ==
    Hash (md5):             rO9jc1LqrvFES8il4Rqe9w==
```
Side question - Is there any reason those buckets weren't deleted? I guess I'm grateful we have them for comparison!

Here is a hash from a bucket generated today - it's the same.

```bash
$ gsutil ls -L gs://t-a4d826db-3923-4fc7-a8a0-493aa6c3b2eb/landsat-data.txt.bz2 | grep Hash
    Hash (crc32c):          oobFiQ==
    Hash (md5):             rO9jc1LqrvFES8il4Rqe9w==
```

So this means that:

 - gcloud correctly reports the hash, but the Google Storage Client blob is missing the last two `==`
 - both google_crc32c and the previous library have equivalent results - the blob ends in `==`
 - there must be some issue with how the blob object is returning it's known hash. I added a reload() which is supposed to recalculate / refresh, and if that doesn't work I'm wondering if we need to ask Google for some help.

If the hash is calculated on (Google's) server on the fly, this would mean there is some issue there. If it's calculated on our server on the fly, then it could be something to do with a limited resource (memory or otherwise?) It looks like maybe the last successful run was https://github.com/snakemake/snakemake/runs/853304393?check_suite_focus=true (July 9th) so I wonder what has changed since then. Anyhoo, off to bed now.

vsoch:
Holy crap that fixed it! So we needed the self.blob.reload() !

image

For the record it is now totally green. This is mostly for me to remember in case I wake up tomorrow and it's red again!

image
image


I'm going to do one more push (with the debugging prints removed) and fingers crossed this wasn't just some sheer luck or chance. I wonder why we never needed to reload before? Something else to look at later / in the future - [timeouts added to the blob functions directly](https://github.com/googleapis/python-storage/commit/6eeb855aa0e6a7835d1d4f6e72951e43af22ab57#diff-5d3277a5f0f072a447a1eb89f9fa1ae0) (I wonder if that's related?).

setup.py
```diff
@@ -72,10 +72,10 @@
         "reports": ["jinja2", "networkx", "pygments", "pygraphviz"],
         "messaging": ["slacker"],
         "google-cloud": [
-            "crc32c",
             "oauth2client",
-            "google-api-python-client==1.9.1",
-            "google-cloud-storage==1.28.1",
+            "google-crc32c",
+            "google-api-python-client",
+            "google-cloud-storage",
```
johanneskoester, line 78: Ok, I'm fine with that.

snakemake/remote/GS.py
```diff
@@ -190,6 +190,9 @@ def download(self):
             self.blob.download_to_file(parser)
         os.sync()
 
+        # **Important** hash can be incorrect or missing if not refreshed
+        self.blob.reload()
```
johanneskoester, line 194: Good catch!

Issue #555, opened 2020-08-12
Discovery build error with oauth2 library


vsoch:
It looks like updating the Google libraries to not pin results in updating the oauth2 lib to something > 4.0.0, resulting in this issue:
https://github.com/snakemake/snakemake/runs/976510241

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/share/miniconda/envs/snakemake/lib/python3.6/site-packages/googleapiclient/discovery_cache/file_cache.py", line 37, in <module>
    from oauth2client.locked_file import LockedFile
ModuleNotFoundError: No module named 'oauth2client.locked_file'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/share/miniconda/envs/snakemake/lib/python3.6/site-packages/googleapiclient/discovery_cache/__init__.py", line 44, in autodetect
    from . import file_cache
  File "/usr/share/miniconda/envs/snakemake/lib/python3.6/site-packages/googleapiclient/discovery_cache/file_cache.py", line 41, in <module>
    "file_cache is unavailable when using oauth2client >= 4.0.0 or google-auth"
ImportError: file_cache is unavailable when using oauth2client >= 4.0.0 or google-auth
```
It's a bit weird because it didn't show up in the testing run (which ran those same jobs) but the merge to master. I'm going to give a shot at a fix, it looks like we can disable the file cache and logging via the fix in this issue https://github.com/googleapis/google-api-python-client/issues/299#issuecomment-343255309.

vsoch:
This was fixed wit #556

PR #556, opened 2020-08-12
adding instruction to skip cache for google discovery client

vsoch:
I'm not sure why this issue popped up (#555) because it didn't happen with the full testing run last night, but these two changes (to logging and init of the discovery client) seem to be the recommended fix. Another option would be to return to the pinned Google library versions. I guess it's up to how we want to move forward - if we pin versions, we are less likely to see errors like these, but then perhaps we _do_ want to see them so we can stay up to date with the Google client updates.


vsoch:
@johanneskoester is this an issue with the latest container not being deployed with crc32c yet? If so, then the test errors are expected.

PR #565, opened 2020-08-14
add working example for updating git files

goi42:
the linked solution does not really work as advertised. This adds a working option directly to the FAQ.


johanneskoester (approved):
Awesome, thank you!

Issue #572, opened 2020-08-18
Comparison between datetime and float with Google Storage backend


hannes-brt:
**Snakemake version**
5.22.1

**Describe the bug**
I am running a workflow with `--default-remote-provider=GS` which always results in a TypeError, because Snakemake attempts to compare a `datetime.datetime` object to a float.

**Logs**
```
Traceback (most recent call last):
  File "/home/hannes/miniconda3/envs/target-id/lib/python3.6/site-packages/snakemake/__init__.py", line 704, in snakemake
    keepincomplete=keep_incomplete,
  File "/home/hannes/miniconda3/envs/target-id/lib/python3.6/site-packages/snakemake/workflow.py", line 667, in execute
    dag.init()
  File "/home/hannes/miniconda3/envs/target-id/lib/python3.6/site-packages/snakemake/dag.py", line 182, in init
    self.update_needrun()
  File "/home/hannes/miniconda3/envs/target-id/lib/python3.6/site-packages/snakemake/dag.py", line 936, in update_needrun
    update_needrun(job)
  File "/home/hannes/miniconda3/envs/target-id/lib/python3.6/site-packages/snakemake/dag.py", line 912, in update_needrun
    f for f in job.input if f.exists and f.is_newer(output_mintime_)
  File "/home/hannes/miniconda3/envs/target-id/lib/python3.6/site-packages/snakemake/dag.py", line 912, in <listcomp>
    f for f in job.input if f.exists and f.is_newer(output_mintime_)
  File "/home/hannes/miniconda3/envs/target-id/lib/python3.6/site-packages/snakemake/io.py", line 195, in wrapper
    return func(self, *args, **kwargs)
  File "/home/hannes/miniconda3/envs/target-id/lib/python3.6/site-packages/snakemake/io.py", line 419, in is_newer
    return self.mtime > time
TypeError: '>' not supported between instances of 'datetime.datetime' and 'float'
```

johanneskoester:
@vsoch mtime coming from GS must be converted into a float since epoch, or vice versa, before storing into the inventory cache. Can you have a look?

vsoch:
See https://github.com/snakemake/snakemake/pull/579. I'm not sure why this wasn't caught before!

PR #578, opened 2020-08-20
Converting GS blob datetime to timestamp

vsoch:
This should be a quick fix to convert the GS blob timestamp (when last updated) to an int (since epoch time) so we can compare. This will fix https://github.com/snakemake/snakemake/issues/572. Here is how I quickly tested:

```
blob.updated
datetime.datetime(2020, 6, 2, 10, 56, 4, 988000, tzinfo=<UTC>)

# float
blob.updated.timestamp()
1591095364.988

# int
int(blob.updated.timestamp())
1591095364
```


vsoch:
whoa this is definitely not based off the right branch!

PR #587, opened 2020-08-28
Detect and give useful error for missing "wildcards."

Sam-Tygier:
Detect the case where a user has put a wildcard in the shell commands
without prefixing with "wildcards." This allows giving a useful error
message, rather than the more generic NameError message.


Sam-Tygier:
For example:

```
rule all:
	output: "work/test_{x}.txt"
	shell: "echo test {x} > {output}"
```
now gives the message

`NameError: The name 'x' is unknown in this context. Did you mean 'wildcards.x'`

johanneskoester (approved):

johanneskoester:
Thanks!

PR #588, opened 2020-08-28
Support for panoptes monitor server

fgypas:
Feature:
- Add flag --wms-monitor to specify IP and port of panoptes server. When the flag is specified a handshake is perfomed between snakemake and panoptes. If this is sucessfull then logs are passed from snakemake to panoptes.
Info:
- To setup panoptes server please follow the instructions here: https://github.com/panoptes-organization/panoptes
- An example workflow is available here: https://github.com/panoptes-organization/snakemake_example_workflow
- Tests that could be integrated within the snakemake tests are here: https://github.com/panoptes-organization/panoptes/blob/develop/.travis.yml

Please let me know for further changes that need to be done either in snakemake or panoptes.


fgypas:
@johanneskoester I don't really get why the singularity fails.

snakemake/logging.py
```diff
@@ -14,6 +14,8 @@
 import tempfile
 from functools import partial
 import inspect
+import requests
```
johanneskoester, line 17: You need to move the requests import into the logger functions that use it. Otherwise, Python pickling will require it (even for unrelated cases) and it is not guaranteed to be there in every execution context (e.g. in a container).

johanneskoester (changes requested):

snakemake/logging.py
fgypas: Thanks a lot. Indeed this was the issue.

fgypas:
@johanneskoester Is there something else I should change or improve?

johanneskoester (changes requested):
So, actually we now need to document the REST API that Snakemake expects with --wms-monitor. Can you add documentation for that, under `docs/executing/monitoring.rst`?

fgypas:
@johanneskoester Thanks a lot for the feedback. I added some API documentation under `docs/executing/monitoring.rst` Is this enough/clear, or do you think it needs more details?

docs/executing/monitoring.rst
```diff
@@ -0,0 +1,16 @@
+.. _monitoring:
+
+========================
+Monitoring
+========================
+
```
johanneskoester, line 6: Thank you, this is a great start! Can you write it a bit more in the style of "snakemake will send the following requests to the wms monitor: ..." and then a table, which lists, and explains each request and the data that is sended along with it.

johanneskoester: Sorry for requesting an additional change, but I feel that I don't know enough about the api to modify it myself.

johanneskoester (changes requested):

docs/executing/monitoring.rst
fgypas: No problem, I understand that. We can also make changes on panoptes side if you think that this will make things cleaner. For example, I think that the long term plan would be to add full [WES](https://github.com/ga4gh/workflow-execution-service-schemas) support.
- Now we could support only GET requests and show them, but we cannot make POST from panoptes to trigger a snakemake pipeline.
- Another thing that we could refactor before merging is how to keep a unique id for pipelines that we restart. Let's say that a pipeline crashed for whatever reason. When we restart it from the command line it gets a new uuid from panoptes. Is there a unique id generated on snakemake that we could use?

I will make soon an extended table of the current documentation.

fgypas: @johanneskoester Did you have something like this in your mind? Feel free to request new changes.

johanneskoester (approved):

PR #618, opened 2020-09-13
using workflow provided snakefile before defaults for gls executor

### LINKED ISSUE #605: Google Cloud Executer does not recognize Snakefile specified by --snakefile properly
**Snakemake version**
5.23

**Describe the bug**
When spcifying a snakefile with --snakefile and there is a file called Snakefile in the same folder, the file specified is ignored and the file named Snakefile is copied to the Cloud workers
**Logs**
If applicable, any terminal output to help explain your problem.

**Minimal example**
Add a minimal example for reproducing the bug.

**Additional context**
I had a folder with Snakefile and Snakefile2
I used snakemake --snakefile Snakefile2 but kept getting errors, when inspecting the logs, within the cloud VM Snakefile was being executed instead of the Snakefile2

vsoch:
This will fix a small bug that we don't check for a custom workflow snakefile path, but instead just iterate through the default paths expected. I recall there was a reason I implemented it using the Snakefile choices, but can't remember / reproduce the reason. This will fix #605. 


snakemake/executors/google_lifesciences.py
```diff
@@ -582,10 +582,18 @@ def _set_snakefile(self):
         """
         from snakemake import SNAKEFILE_CHOICES
 
-        for snakefile in SNAKEFILE_CHOICES:
-            if os.path.exists(os.path.join(self.workdir, snakefile)):
-                self.snakefile = snakefile
-                break
+        # Case 1: snakefile specified on the command line
+        if os.path.exists(self.workflow.snakefile):
```
vsoch, line 586: One question here I'd want to know is if this could ever be None, or the attribute not defined at all? The default from `snakemake/__init__.py` seems to not allow this:
```python
    if args.snakefile is None:
        for p in SNAKEFILE_CHOICES:
            if os.path.exists(p):
                args.snakefile = p
                break
        if args.snakefile is None:
            print(
                "Error: no Snakefile found, tried {}.".format(
                    ", ".join(SNAKEFILE_CHOICES), file=sys.stderr
                )
            )
            sys.exit(1)
```
so I suspect it will be defined unless snakemake is called programatically (in python) and the snakefile explicitly set to None (does this happen?)

johanneskoester: Nope, that's not really allowed. Hence, you can remove the else.

johanneskoester (changes requested):

snakemake/executors/google_lifesciences.py
vsoch: Removed then!

snakemake/exceptions.py
```diff
@@ -256,10 +256,8 @@ class MissingOutputException(RuleException):
     def __init__(
         self, message=None, include=None, lineno=None, snakefile=None, rule=None
     ):
-        message = (
-            "Job completed successfully, but some output files are missing. {}".format(
-                message
-            )
+        message = "Job completed successfully, but some output files are missing. {}".format(
```
vsoch, line 259: hmm this suggests I am using a different version of black locally.

vsoch: okay, updated to black==20.8b1

snakemake/executors/google_lifesciences.py
```diff
@@ -580,12 +580,10 @@ def _set_snakefile(self):
         execution into a temporary directory, and the initial Snakefile
         was somewhere else on the system.
         """
-        from snakemake import SNAKEFILE_CHOICES
-
-        for snakefile in SNAKEFILE_CHOICES:
-            if os.path.exists(os.path.join(self.workdir, snakefile)):
-                self.snakefile = snakefile
-                break
+        if os.path.exists(self.workflow.snakefile):
```
johanneskoester, line 583: I think above doc comment is a bit outdated now. Also, I think if self.workflow.snakefile does not exist this should not silently exit. Just put an assertion for existence at the beginning instead of an if.

johanneskoester (changes requested):

snakemake/executors/google_lifesciences.py
vsoch: All set.

johanneskoester (approved):

PR #621, opened 2020-09-15
pulp >=2.0

johanneskoester:


mbargull:
Could you try out the following?
```patch
diff --git a/snakemake/scheduler.py b/snakemake/scheduler.py
index 61d5be6b..da9203a5 100644
--- a/snakemake/scheduler.py
+++ b/snakemake/scheduler.py
@@ -600,7 +600,7 @@ class JobScheduler:
                     [scheduled_jobs[job] * job.resources.get(name, 0) for job in jobs]
                 )
                 <= self.resources[name],
-                "Limitation of resource: {name}".format(name=name),
+                "Limitation of resource_ {name}".format(name=name),
             )
 
         # Choose jobs that lead to "fastest" (minimum steps) removal of existing temp file
```
GLPK doesn't seem to like the additional `:` in the `.lp` file
```
\* JobScheduler *\
Maximize
OBJ: 2 job_run_test_0
Subject To
Limitation_of_resource:__cores: job_run_test_0 <= 3
Limitation_of_resource:__nodes: job_run_test_0 <= 9.22337203685e+18
Binaries
job_run_test_0
End
```

mbargull:
Maybe this is relevant for this?
http://web.mit.edu/lpsolve/doc/lp-format.htmhttp://web.mit.edu/lpsolve/doc/CPLEX-format.htm
> Variables can be named anything provided that the name does not exceed 255 characters, all of which must be alphanumeric (a-z, A-Z, 0-9) or one of these symbols: ! " # $ % & ( ) / , . ; ? @ _ ` ' { } | ~. Longer names are truncated to 255 characters. A variable name cannot begin with a number or a period.

andersgs:
This breaks the conda installation of Snakemake which defines `1.6.8` as the maximum version compatible:

```        - pulp <=1.6.8  # newer pulp versions do not include a solver. Directly depending on coincbc does not work because it is not available for windows.```

I get the following error when running `snakemake 2.25.0` installed via conda:

```AttributeError: module 'pulp' has no attribute 'apis'```

Running `pip install -U pulp` fixes the issue, but makes it harder to share the environment.

mbargull:
@andersgs, we'll put out a new build on Bioconda that should fix your issue.
ref: https://github.com/bioconda/bioconda-recipes/pull/24452

PR #630, opened 2020-09-17
Fix/aggregate order priority

### LINKED ISSUE #615: Follow input list order
Given the rule:
```
rule aggregate:
    input:
        ["{dataset}/a.txt".format(dataset=dataset) for dataset in DATASETS]
    output:
        "aggregated.txt"
    shell:
        ...
```
the order in which aggregate is executed does not obey the order of the list used for ``input:``.  

The execution order is seemingly random, which is a problem for users who would like to prioritize certain ``datasets`` first (some results are needed more urgently than others).

I would like the execution order to follow the order of the input list. If users want random behavior, they can simply shuffle the input list.

Right now I'm having to partition ``DATASETS`` into high priority, medium priority, etc, and create different rules for each.

**Additional context**
None

vsoch:
This is a work in progress (things to discuss!) to ensure that jobs are run in aggregate order. The issue is described well in #615.  I first tested the following:

 - updating the ready jobs to a list
 - updating most sets in dag.py to OrderedSet

And while the second would work for cores > 1, in the case of cores == 1 it would process the jobs in the reverse order (something else going on I'm not sure about). @alexandres suggested trying to modify job priority, meaning that in the case that all jobs have priority 0, we set them to decrementing values to ensure that whatever came first is actually run first. This fixes the cores == 1 issue, but in the case of cores > 1 (e.g., == 4) we get a mixed up result still, because each core is likely being handed it's own separate list, and still writing to the same file at the same time (resulting in a weird order). To get something that appears to work for both cases, I added the orderedset along with a check (a function) that, given that all priorities are 0 (meaning the user hasn't set anything) we go through and hard code that we want the specific order:

```python
    def apply_ordered_priority(self):
        # If all jobs have priority 0, ensure executed in order
        if sum(self._priority.values()) == 0:
            idx = len(self._priority)
            for job in self._priority:
                self._priority[job] = idx
                job.rule.priority = idx
                idx -= 1
```

I'm thinking that it might make sense to have some kind of snakemake flag to run this instead of just applying blindly - let's discuss that! Because just with OrderedSet added, it will fix the issue for running with > 1 cores, which I would tend to think are most use cases.

 - This will close #615

### DIFF
--- a/docs/executor_tutorial/google_lifesciences.rst
+++ b/docs/executor_tutorial/google_lifesciences.rst
@@ -59,6 +59,15 @@ can refer to the page under iam-admin to `download your service account <https:/
     export GOOGLE_APPLICATION_CREDENTIALS="/home/[username]/credentials.json"
 
 
+The suggested, minimal permissions required for this role include the following:
+
+ - Compute Storage Admin(Can potentially be restricted further)
+ - Compute Viewer
+ - Service Account User
+ - Cloud Life Sciences Workflows Runner
+ - Service Usage Consumer
+
+
 Step 1: Upload Your Data
 ::::::::::::::::::::::::
 
--- a/docs/executor_tutorial/google_lifesciences.rst
+++ b/docs/executor_tutorial/google_lifesciences.rst
@@ -59,15 +59,6 @@ can refer to the page under iam-admin to `download your service account <https:/
     export GOOGLE_APPLICATION_CREDENTIALS="/home/[username]/credentials.json"
 
 
-The suggested, minimal permissions required for this role include the following:
-
- - Compute Storage Admin(Can potentially be restricted further)
- - Compute Viewer
- - Service Account User
- - Cloud Life Sciences Workflows Runner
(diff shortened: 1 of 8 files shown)


johanneskoester (commented):
As mentioned in the issue, I don't think that we need to enforce an order here. If I understood the issue correctly, the intention is to use it for priorititization, but Snakemake has --prioritize for that. Hence, I think we can avoid the additional performance burden that would come with introducing ordering here.

johanneskoester:
> As mentioned in the issue, I don't think that we need to enforce an order here. If I understood the issue correctly, the intention is to use it for priorititization, but Snakemake has --prioritize for that. Hence, I think we can avoid the additional performance burden that would come with introducing ordering here.

If you agree, you can close this PR (anyway thanks for giving it a try). But maybe I misunderstood, hence, I'll leave it open for now.

vsoch:
I do agree. Thanks @johanneskoester !

PR #633, opened 2020-09-18
Add cpu usage (user time + system time) in benchmark

natir:
Hello, 

Benchmark keyword record wallclock time i.e. time at end of run minus time at begin of run.

This PR add CPU usage time in the benchmark record, it can be rugly define as the sum of time each CPU spends to execute the process.

If you have better variable or column naming I'm interested!

Edit: I didn't see how I can test this new functionality, apparently benchmark test didn't check the content of the file produce by the benchmark keyword but I probably miss something


svenrahmann:
After taking a look, I find the code confusing. I certainly dont understand it enough to approve a PR on it. My confusion already applies to the state before this PR.
The already is cpu_seconds, which seems to be computed in a complicates way.
Now this also uses proc.cpu_times(), but only to compute the CPU usage (which for me would mean CPU usage in %).
The print in l. 251 seems out of place.

Isn't cpu_seconds what you were looking for in the first place?

snakemake/benchmark.py
```diff
@@ -214,6 +219,8 @@ def _update_record(self):
         check_io = True
         # CPU seconds
         cpu_seconds = 0
+        # CPU usage time
+        cpu_usage = 0
```
johanneskoester, line 223: Maybe rather call it cpu_times?

snakemake/benchmark.py
```diff
@@ -262,6 +272,7 @@ def _update_record(self):
         self.bench_record.io_in = io_in
         self.bench_record.io_out = io_out
         self.bench_record.cpu_seconds += cpu_seconds
+        self.bench_record.cpu_usage = cpu_usage
```
johanneskoester, line 275: Overwriting instead of adding is intentional here?

natir:
I'm a little bit confused by codebase too.

`cpu_seconds` is computed with [`psutil.cpu_percent()`](https://psutil.readthedocs.io/en/latest/#psutil.cpu_percent) this function returns the load average of CPU (in percent) between now and last call of `cpu_percent()`. In me understanding `cpu_seconds` is used to store and compute the CPU load average during all tasks. It's a cumulative average of CPU load. Maybe `cpu_usage` is a better name for this variable.

In this PR I want add value to show how many time (in second) CPU work on the task (`User time` and `System time` in `/usr/bin/time` command), maybe I can switch the usage of `cpu_seconds` and `cpu_usages`
Oups I forget the print sorry.

natir:
@svenrahmann did you know why this test failled ?

svenrahmann:
It seems to be a problem with the test code (scatter-gather), not with your patch. I think this has to be resolved by @johanneskoester .

johanneskoester:
@holtgrewe as the benchmarking code is yours, could you have a look at this and probably also comment on the discussion of @svenrahmann and @natir above?

holtgrewe:
I think the original code is a bit flawed. We should just record user, system, and wait times and not compute the load the way it is done.

johanneskoester:
ok, @natir, feel free to give this a major overhaul.

natir:
@johanneskoester In fact I already made some change in variable name and made some simplification (very little). Other improvements to this piece of code don't jump out at me.

natir:
@holtgrewe code is good for you ?

johanneskoester:
@holtgrewe are you ok with the changes?

holtgrewe (approved):
sorry for the delay, lgtm

johanneskoester (approved):

PR #634, opened 2020-09-19
Integrate tes v2

svedziok:
This PR adds a new executor which implements the GA4GH-TES standard. Functionality was tested using Funnel (https://ohsu-comp-bio.github.io/funnel/).


snakemake/__init__.py
```diff
@@ -1853,6 +1865,18 @@ def get_argument_parser(profile=None):
...
     )
 
+    group_tes.add_argument(
+        "--tes",
+        action="store_true",
+        help="Send workflow tasks to GA4GH TES server specified by --tes_url."
+    )
+    
+    group_tes.add_argument(
+        "--tes_url",
```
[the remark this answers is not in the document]
johanneskoester: Thinking about it, probably just have flag, --tes, which takes the URL as its first argument?

johanneskoester (changes requested):
Excellent!

snakemake/executors/__init__.py
```diff
@@ -2033,6 +2034,374 @@ def _wait_for_jobs(self):
             sleep()
 
 
+TaskExecutionServiceJob = namedtuple(
```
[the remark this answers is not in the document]
svedziok: Ok - moved tes executor code into own submodule.

snakemake/__init__.py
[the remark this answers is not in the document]
svedziok: Makes sense - removed --tes-url argument

johanneskoester (changes requested):
Here are some final remarks:
* The sonarcloud linter mentions two potential bugs and four code smells, can you fix them?
* The tests need to be explicitly activated, similar to kubernetes [here](https://github.com/snakemake/snakemake/blob/master/.github/workflows/main.yml#L112).
* Can you add a section to the docs, explaining how to use it [here](https://github.com/snakemake/snakemake/tree/master/docs/executing)?

svedziok:
> Here are some final remarks:
> 
> * The sonarcloud linter mentions two potential bugs and four code smells, can you fix them?
> * The tests need to be explicitly activated, similar to kubernetes [here](https://github.com/snakemake/snakemake/blob/master/.github/workflows/main.yml#L112).
> * Can you add a section to the docs, explaining how to use it [here](https://github.com/snakemake/snakemake/tree/master/docs/executing)?

- Fixed
- Test shall is active now
- We added a short documentation

johanneskoester (approved):

snakemake/executors/ga4gh_tes.py
```diff
@@ -0,0 +1,381 @@
+__author__ = "Johannes Köster"
+__copyright__ = "Copyright 2015-2020, Johannes Köster"
+__email__ = "koester@jimmy.harvard.edu"
```
johanneskoester, line 3: Please add your names and emails here.

johanneskoester (changes requested):

snakemake/executors/ga4gh_tes.py
svedziok: Is this required? We wouldn't mind if it stays as it is.

johanneskoester: But that is your work, at least change into `__authors__` and add a list with the three of us instead of my name only.

svedziok: ok

johanneskoester (approved):

johanneskoester:
@svedziok the tests don't work yet. The ftp test seems to access a file that for which we don't have the required permissions. And it additionally needs a folder expected-results, containing the results that shall be produced.

svedziok:
> @svedziok the tests don't work yet. The ftp test seems to access a file that for which we don't have the required permissions. And it additionally needs a folder expected-results, containing the results that shall be produced.

The ftp test was just used for development and is removed now. Sorry for the confusion. The main test at tests/test_tes shall just send a single task (which is an echo command) to a (mocked) TES server and the workflow waits for the task to be finished, which is immediately after the first status request. It does not produce any output files. Is the production of an output file required? Would it be ok to write a file manually as part of the test? Otherwise, we could also try to mock remote file storage.

johanneskoester:
> > @svedziok the tests don't work yet. The ftp test seems to access a file that for which we don't have the required permissions. And it additionally needs a folder expected-results, containing the results that shall be produced.
> 
> @johanneskoester the ftp test was just used for development and is removed now. Sorry for the confusion. The main test at tests/test_tes shall just send a single task (which is an echo command) to a (mocked) TES server and the workflow waits for the task to be finished, which is immediately after the first status request. It does not produce any output files. Is the production of an output file required? Would it be ok to write a file manually as part of the test? Otherwise, we could also try to mock remote file storage.

Thanks for clarifying. The test is fine actually. I just added an empty expected-results folder for it.

akkarapon26 (approved):

akkarapon26 (approved):

akkarapon26 (approved):

akkarapon26 (approved):

akkarapon26 (approved):

johanneskoester:
Wohoo!

matheuscburger:
I tried to use tes but I got the following error:
ModuleNotFoundError: No module named 'tes'

I had to install the package py-tes, so I think it is not in the requirements.

Issue #635, opened 2020-09-21
pulp.constants.PulpError: Variable names too long for Lp format


nick-youngblut:
**Snakemake version**

5.25.0

**Describe the bug**

Apparently, the recently update `pulp` version doesn't like the long variable names:

```
Traceback (most recent call last):
  File "/ebio/abt3_projects/software/dev/miniconda3_dev/envs/snakemake/lib/python3.6/site-packages/snakemake/__init__.py", line 735, in snakemake
    keepincomplete=keep_incomplete,
  File "/ebio/abt3_projects/software/dev/miniconda3_dev/envs/snakemake/lib/python3.6/site-packages/snakemake/workflow.py", line 972, in execute
    success = scheduler.schedule()
  File "/ebio/abt3_projects/software/dev/miniconda3_dev/envs/snakemake/lib/python3.6/site-packages/snakemake/scheduler.py", line 406, in schedule
    else self.job_selector_ilp(needrun)
  File "/ebio/abt3_projects/software/dev/miniconda3_dev/envs/snakemake/lib/python3.6/site-packages/snakemake/scheduler.py", line 625, in job_selector_ilp
    prob.solve()
  File "/ebio/abt3_projects/software/dev/miniconda3_dev/envs/snakemake/lib/python3.6/site-packages/pulp/pulp.py", line 1890, in solve
    status = solver.actualSolve(self, **kwargs)
  File "/ebio/abt3_projects/software/dev/miniconda3_dev/envs/snakemake/lib/python3.6/site-packages/pulp/apis/glpk_api.py", line 61, in actualSolve
    lp.writeLP(tmpLp, writeSOS = 0)
  File "/ebio/abt3_projects/software/dev/miniconda3_dev/envs/snakemake/lib/python3.6/site-packages/pulp/pulp.py", line 1737, in writeLP
    self.checkLengthVars(max_length)
  File "/ebio/abt3_projects/software/dev/miniconda3_dev/envs/snakemake/lib/python3.6/site-packages/pulp/pulp.py", line 1810, in checkLengthVars
    + str(long_names))
pulp.constants.PulpError: Variable names too long for Lp format
[...]

PR #648, opened 2020-09-29
Fix lpvariable

FelixMoelder:
It happens that ilp scheduling fails caused by a too long LpVariable-name(see [https://github.com/snakemake/snakemake/issues/635](https://github.com/snakemake/snakemake/issues/635)).
This occurs as the scheduler takes the name of a temp file as variable name.
Additionally, based on the file name pulp throws warning incase it contains illegal characters.

To solve both issues variables variables now consist of a prefix (preventing the illegal character warning) and a index value replacing the file names. Ensuring that variable names keep short and do not contain any special characters.

### DIFF
--- a/snakemake/scheduler.py
+++ b/snakemake/scheduler.py
@@ -566,7 +566,7 @@ class JobScheduler:
                 upBound=1,
                 cat=pulp.LpInteger,
             )
-            for idx, job in enumerate(jobs)
+            for idx, _ in enumerate(jobs)
         }
 
         temp_files = {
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):

mbargull:
This fix is probably incomplete.

https://github.com/snakemake/snakemake/issues/635#issuecomment-696184782 also reports another
> `/tmp/1476d5a6b0e34266acb10e45024241b1-pulp.lp:10: missing constraint sense`

Which means the `name` in https://github.com/snakemake/snakemake/pull/648/files#diff-3b0b40738124c8bd3c4a83b0aa580f03L638 (or another line?) needs to be handled too.

Suggestion:
- Install `glpk` in the CI tests (either generally, or probably better as additional test matrix run, to make sure Cbc is also run in the CI.),
- Add tests with long (>255 characters) input filename paths and paths with non-valid LP variable/constraint name characters,
- Carefully check all variable and constraint names that get added or mangle them as depicted in https://github.com/snakemake/snakemake/issues/635#issuecomment-696542079.

PR #650, opened 2020-09-29
Adding glpk for testing

FelixMoelder:


johanneskoester (changes requested):
We should anyway make the solver configurable. Please have a look at [this](https://coin-or.github.io/pulp/guides/how_to_configure_solvers.html), and allow to set the ilp solver via `--scheduler-ilp-solver SOLVER` with SOLVER being one of `pulp.list_solvers()`. Then, in test16, pass `scheduler_ilp_solver=GLPK_CMD`. Also, please rename test16 into test_glpk_solver.

mbargull:
> We should anyway make the solver configurable.

:+1:

> allow to set the ilp solver via `--scheduler-ilp-solver SOLVER`

Since it's only a low number of choices, tightly scoped, and the scheduler + ILP solver selection go hand in hand: I'd probably just reuse the `--scheduler` option to avoid introducing yet another one, i.e., `--scheduler=ilp=X`.

FelixMoelder:
For now I have added a separate command as @johanneskoester suggested using `GLPK_CMD` as default.
By adding `pulp.list_solvers(available_only=True)` to the argument parser it would be possible to dynamically show available solvers when calling the cli help.

snakemake/scheduler.py
```diff
@@ -419,7 +421,12 @@ def schedule(self):
                         "Ready jobs ({}):\n\t".format(len(needrun))
                         + "\n\t".join(map(str, needrun))
                     )
-                    run = self.job_selector(needrun)
+
+                    run = (
+                        self.job_selector(needrun, self.scheduler_ilp_solver)
+                        if self.scheduler_ilp_solver
+                        else self.job_selector(needrun)
+                    )
```
johanneskoester, line 429: Please keep this as before, and directly access self.scheduler_ilp_solver inside the job_selector_ilp method.

johanneskoester (changes requested):

snakemake/__init__.py
```diff
@@ -1270,6 +1275,16 @@ def get_argument_parser(profile=None):
         ),
     )
 
+    import pulp
+
+    group_exec.add_argument(
+        "--scheduler-ilp-solver",
+        default=None,
+        nargs="?",
```
mbargull, line 1283:
```suggestion
```

tests/tests.py
```diff
@@ -108,6 +108,10 @@ def test15():
     run(dpath("test15"))
 
 
+def test_glpk_solver():
+    run(dpath("test_glpk_solver"), scheduler_ilp_solver="GLPK_CMD")
```
mbargull, line 112: Please note that `GLPK_CMD` is the default anyway whenever `glpk` is installed (and no bundled Cbc/GLPK is available, which should always be the case for the conda-forge package).
Meaning, by adding `glpk` to the `test-environment.yml`, the tests will run with GLPK but not with Cbc, albeit the latter is probably the one that most (non-Windows) users will run, currently.

FelixMoelder: Good point. I just checked and the only solver available when `glpk` is not installed is `COIN_CMD`.

FelixMoelder: It appears as `COIN_CMD` is not available by default on windows. Due to the lack of a windows environment I can't check what's the default solver.

mbargull: But you should add tests for both, e.g.,
```python
def test_glpk_solver():
    run(dpath("test_coin_solver"), scheduler_ilp_solver="GLPK_CMD")

# But run the one below only for Unix, because Cbc is currently not packaged on Windows.
def test_coin_solver():
    run(dpath("test_coin_solver"), scheduler_ilp_solver="COIN_CMD")
```
Or (maybe even better?) run the tests for all solvers from `ilp_solvers = pulp.list_solvers(onlyAvailable=True)`, but then make also sure that `assert "GLPK_CMD" in ilp_solvers` and also (on Unix) `assert "COIN_CMD" in ilp_solvers`.

johanneskoester (approved):

PR #653, opened 2020-10-01
Updating google_lifesciences.py

vsoch:
We want to update the GLS executor to use the helper script provided in the repository, instead of the one in a gist. The gist is here https://gist.githubusercontent.com/vsoch/f5a6a6d1894be1e67aa4156c5b40c8e9/raw/a4e9ddbeba20996ca62745fcd4d9ecd7bfa3b311/gls.py in case anyone wants to do a comparison!

### DIFF
--- a/snakemake/executors/google_lifesciences.py
+++ b/snakemake/executors/google_lifesciences.py
@@ -675,7 +675,7 @@ class GoogleLifeSciencesExecutor(ClusterExecutor):
         commands = [
             "/bin/bash",
             "-c",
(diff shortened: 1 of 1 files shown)


johanneskoester (changes requested):
Why not even just distribute the script with Snakemake itself?

johanneskoester:
I think you are right. I was thinking that, if it is part of the snakemake package, it could be pushed as a file, but that makes little sense, because you need the script to download it then, a hen and egg problem. So, all fine here, thanks.

johanneskoester (approved):

vsoch:
:chicken: + :egg: problems!  :) Always hiding in unexpected places. Thanks!

PR #658, opened 2020-10-05
Fix target function of ilp scheduler

FelixMoelder:
This PR includes all jobs which have not been executed including jobs not being ready yet for improvement calculation.

### DIFF
--- a/snakemake/scheduler.py
+++ b/snakemake/scheduler.py
@@ -368,6 +368,11 @@ class JobScheduler:
         """ Return open jobs. """
         return filter(self.candidate, list(job for job in self.dag.ready_jobs))
 
+    @property
+    def remaining_jobs(self):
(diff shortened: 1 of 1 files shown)


snakemake/scheduler.py
```diff
@@ -647,7 +656,9 @@ def job_selector_ilp(self, jobs):
                     scheduled_jobs[job] * self.required_by_job(temp_file, job)
                     for job in jobs
                 ]
-            ) / lpSum([self.required_by_job(temp_file, job) for job in jobs])
+            ) / lpSum(
+                [self.required_by_job(temp_file, job) for job in self.remaining_jobs]
```
johanneskoester, line 660: self.remaining_jobs into a variable above, such that it does not need to be recomputed for each constraint.

johanneskoester (changes requested):

johanneskoester (approved):

PR #662, opened 2020-10-06
Optimizations to handle large workflows

mhulsman:
Large workflows (njobs>10k) can become very slow due to some code which scales O(n^2). 

    * Change candidates in update_needrun from list to set 
       (enable quick member check)

    * Use deque for faster bfs jobs traversal
       (prevents memory copy due to popping of first item)

    * Optimize open_jobs using set operations.

    * Only run open_jobs if new jobs have potentially been added to ready jobs

    * Add short-circuits to the periodicity detector to improve its speed.

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -935,11 +935,11 @@ class DAG:
         for job in candidates:
             update_needrun(job)
 
-        queue = list(filter(reason, candidates))
+        queue = deque(filter(reason, candidates))
         visited = set(queue)
         candidates_set = set(candidates)
         while queue:
-            job = queue.pop(0)
+            job = queue.popleft()
             _needrun.add(job)
 
             for job_, files in dependencies[job].items():
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -9,7 +9,7 @@ import shutil
 import textwrap
 import time
 import tarfile
-from collections import defaultdict, Counter
+from collections import defaultdict, Counter, deque
 from itertools import chain, filterfalse, groupby
 from functools import partial
 from pathlib import Path
@@ -937,6 +937,7 @@ class DAG:
 
         queue = list(filter(reason, candidates))
         visited = set(queue)
+        candidates_set = set(candidates)
         while queue:
             job = queue.pop(0)
             _needrun.add(job)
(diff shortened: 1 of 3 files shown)


snakemake/scheduler.py
```diff
@@ -355,18 +355,17 @@ def stats(self):
...
-            job not in self.running
-            and job not in self.failed
-            and (self.dryrun or (not job.dynamic_input and not self.dag.dynamic(job)))
-        )
-
     @property
     def open_jobs(self):
         """ Return open jobs. """
-        return filter(self.candidate, list(job for job in self.dag.ready_jobs))
+        jobs = self.dag.ready_jobs - (self.running | self.failed)
```
johanneskoester, line 361: Are you sure this is faster? I would've thought that the additional allocations instead make it slower. Did you benchmark that?

johanneskoester (changes requested):
Thanks a lot, this is great work!! I have one question below.

snakemake/dag.py
```diff
@@ -937,6 +937,7 @@ def update_needrun(job):
 
         queue = list(filter(reason, candidates))
         visited = set(queue)
+        candidates_set = set(candidates)
```
johanneskoester, line 940: Good catch, checking existence in a list was of course not intentional. Is there any reason not to use a set directly in line 924?

snakemake/scheduler.py
mhulsman: Yes, the set operations are written in C,  making them much faster then a custom loop. I  just did some benchmarking, which shows that the set-based approach is ~10 times as fast (set sizes do not have a large impact on this ratio). 

But you are correct that it does some additional allocations. I now rewrote this to avoid the (running | failed) allocation, which provides another 33% gain in speed.

snakemake/dag.py
mhulsman: The code before this requires the jobs to be in BFS order, so it still requires the list.

johanneskoester (approved):

johanneskoester:
Do you have an example number for how much speed gain to expect for large workflows?

mhulsman:
It can become quite extreme. I tested a 283k jobs workflow.   Doing a dry run in a empty dir:
Old: 42248 seconds   (41118 seconds in update_needrun)
New: 668 seconds   (114 seconds in update_needrun). 

- is_periodic went from 54 seconds to 24 seconds. 

The open_jobs optimization was not benchmarked here, as this requires a non-dry run.

PR #663, opened 2020-10-06
Separate metadata and incomplete job tracking

mhulsman:
Very large numbers of metadata files in the .snakemake/metadata folder can lead to slow performance on network file systems (e.g. Lustre) when checking if jobs are incomplete (--rerun-incomplete). 

1) This code separates the metadata and job-incomplete tracking, by creating an extra .snakemake/incomplete folder.
For the latter, it implements caching, as a simple file existence check suffices to determine if a job is incomplete.  The number of tracking files in the incomplete folder usually remains low, as these are deleted after a job finishes. This results in large speedups when combining --rerun-incomplete with a workflow with many files on a network file system.

Code has been added to enable backwards compatibility with the current .snakemake folder structure.

2) The PR adds also a command line option to make storing of persistent metadata optional to reduce the small file load.

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -219,7 +219,7 @@ def snakemake(
         lock (bool):                lock the working directory when executing the workflow (default True)
         unlock (bool):              just unlock the working directory (default False)
         cleanup_metadata (list):    just cleanup metadata of given list of output files (default None)
-        drop_metadata (bool):       drop metadata tracking information after job (code/input/params/version change listings will be incomplete) (default False)
+        drop_metadata (bool):       drop metadata file tracking information after job finishes (--report and --list_x_changes information will be incomplete) (default False)
         conda_cleanup_envs (bool):  just cleanup unused conda environments (default False)
         cleanup_shadow (bool):      just cleanup old shadow directories (default False)
         cleanup_scripts (bool):     delete wrapper scripts used for execution (default True)
@@ -1573,7 +1573,9 @@ def get_argument_parser(profile=None):
     group_utils.add_argument(
         "--drop-metadata",
         action="store_true",
-        help="Drop metadata after job completion.",
+        help="Drop metadata file tracking information after job finishes. "
+        "Provenance-information based reports (e.g. --report and the "
+        "--list_x_changes functions) will be empty or incomplete.",
     )
     group_utils.add_argument("--version", "-v", action="version", version=__version__)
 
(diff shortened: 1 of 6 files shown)


snakemake/__init__.py
```diff
@@ -1567,6 +1570,11 @@ def get_argument_parser(profile=None):
         action="store_true",
         help="Do not remove incomplete output files by failed jobs.",
     )
+    group_utils.add_argument(
+        "--drop-metadata",
+        action="store_true",
+        help="Drop metadata after job completion.",
```
johanneskoester, line 1576: Please also mention the missing provenance information this leads to here (for --report, --list-params-changes, etc.).

snakemake/persistence.py
```diff
@@ -157,15 +171,34 @@ def conda_cleanup_envs(self):
...
                 {"incomplete": True, "external_jobid": external_jobid},
                 f,
             )
 
-    def finished(self, job):
+    def finished(self, job, keep_metadata=True):
+        records_path = (
+            self._metadata_path
+            if self.incomplete_oldapproach
+            else self._incomplete_path
```
johanneskoester, line 194: Why do you store metadata in the incomplete_path?

snakemake/persistence.py
```diff
@@ -61,6 +63,18 @@ def __init__(
         # place to store any auxiliary information needed during a run (e.g. source tarballs)
         self.aux_path = os.path.join(self.path, "auxiliary")
 
+        self.incomplete_oldapproach = os.path.exists(
+            self._metadata_path
+        ) and not os.path.exists(self._incomplete_path)
```
johanneskoester, line 68: I think if the incomplete folder exists, we can always use the new approach. If it doesn't exist, we can do a one-time migration, by scanning through the metadata path and generating a flag file for each incomplete item in the incomplete path. Then, we can always use the new self.incomplete() version.

johanneskoester (changes requested):
Very good idea. I would suggest some slight changes below.

johanneskoester (approved):

mhulsman:
Also did a quick benchmark of this code, using the same 283k workflow as mentioned in the other pull request, but then not in an empty dir, but one at a point when it is almost complete. Total files will probably be >500k. 

Old: 30985 seconds,   of which 20311 seconds was spent in '_read_record_uncached'. 
New: 776 seconds, of which 21 seconds was spent 'check_incomplete' and 'cache_incomplete_folder'. 

Remaining difference is mostly due to IOCache framework: 
old: 8361 seconds in io.py:wrapper (of which 8271 in 'is_newer')
new:  69 seconds in io.py:wrapper.

hwalinga:
I think this PR caused #2933

The IncompleteFilesException suggest removing metadata but --cleanup-metadata does not seem to go over the new incomplete folder as well.

PR #666, opened 2020-10-06
apply latency-wait argument value to GLS job

millerdz:
The `exec_job` should apply the parameterized value, or default, of `latency_wait` that was provided to the GoogleLifeSciencesExecutor. 
The value was fixed at 0 with `--latency-wait 0`. Changing this to `--latency-wait {latency_wait}` means it will accept the user's configured `latency_wait` value or the default value of `3`.

### DIFF
--- a/snakemake/executors/google_lifesciences.py
+++ b/snakemake/executors/google_lifesciences.py
@@ -75,7 +75,7 @@ class GoogleLifeSciencesExecutor(ClusterExecutor):
             or (
                 "snakemake {target} --snakefile %s "
                 "--force -j{cores} --keep-target-files --keep-remote "
-                "--latency-wait 0 --scheduler {workflow.scheduler_type} "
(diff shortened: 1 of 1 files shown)


millerdz:
@vsoch - letting you know about this PR. 😁

vsoch:
Thank you @millerdz ! It looks like an oversight to have latency_wait provided as an argument but then hard code a value of 0! It looks like the other executors also use latency_wait=3, so that is a good default to leave as is.

millerdz:
@vsoch I'm ok with the default value of `3`. The problem this PR solves is the fact that the user provided value for `latency_wait` never makes it to GLS API - it always ends up with the value of `0`. I just opened issue #667 to elaborate on the issue and why this PR may be a solution.

vsoch:
yep, I deduced that, definitely a needed fix!

vsoch (approved):
LGTM, @johanneskoester will have to give the final approval and merge.

johanneskoester (approved):

PR #675, opened 2020-10-10
Threaded local iocache

### LINKED ISSUE #611: FileNotFoundError for files not specified in rule definition
Snakemake 5.23.0

When running a snakemake workflow on an HPC, I'm getting the following error:

```
Building DAG of jobs...
  File "<base_path>/software/miniconda/envs/snakemake/lib/python3.7/site-packages/snakemake/__init__.py", line 709, in snakemake
    keepincomplete=keep_incomplete,
  File "<base_path>/software/miniconda/envs/snakemake/lib/python3.7/site-packages/snakemake/workflow.py", line 670, in execute
    dag.init()
  File "<base_path>/software/miniconda/envs/snakemake/lib/python3.7/site-packages/snakemake/dag.py", line 177, in init
    job = self.update(self.file2jobs(file), file=file, progress=progress)
  File "<base_path>/software/miniconda/envs/snakemake/lib/python3.7/site-packages/snakemake/dag.py", line 715, in update
    progress=progress,
  File "<base_path>/software/miniconda/envs/snakemake/lib/python3.7/site-packages/snakemake/dag.py", line 792, in update_
    file.inventory()
  File "<base_path>/software/miniconda/envs/snakemake/lib/python3.7/site-packages/snakemake/io.py", line 210, in inventory
    self._local_inventory(cache)
  File "<base_path>/software/miniconda/envs/snakemake/lib/python3.7/site-packages/snakemake/io.py", line 224, in _local_inventory
    with os.scandir(path) as scan:
FileNotFoundError: [Errno 2] No such file or directory: '<analysis_patch>/tmp_batch113/codons'
```
The directory `tmp_batch113/codons` is created and deleted during the execution of another job from the same rule, *i.e.* this specific error occurs for the job dealing with batch166 which never uses this directory. I suspect this could be related to a slow file system (this error does not occur consistently) but I'm surprised Snakemake crashes because of files/directory that are never specified in the rule. For reference, here is the rule behind the jobs that generate this error:

```
rule phyloCSF_batch:
    input:
        batch_file = 'output/analyses/{assembly_code}/maf_batches/batches/{batch}.lst',
        maf_dir = rules.split_maf_contig.output,
        parameters = rules.create_parameter_set.output
    output:
        'output/analyses/{assembly_code}/phyloCSF/batches/{batch}.tsv'
    benchmark:
        'benchmarks/analyses/{assembly_code}/phyloCSF/batches/{batch}.tsv'
    log:
        'logs/analyses/{assembly_code}/phyloCSF/batches/{batch}.txt'
    conda:
        '../envs/analyses.yaml'
    threads: get_threads('phyloCSF_batch')
    resources:
        mem_mb = lambda wildcards, attempt: get_mem('phyloCSF_batch', attempt),
        runtime_s = lambda wildcards, attempt: get_runtime('phyloCSF_batch', attempt)
    params:
        batch_size = config['phyloCSF']['batch_size']
    script:
        '../scripts/phyloCSF_batch.py'
```

Any idea what could be causing this issue? Any input is appreciated!

Thanks

### LINKED ISSUE #623: Possible race condition introduced by `_local_inventory()`
**Snakemake version**
5.22.1

**Description**
I believe I'm encountering a race condition when using a rule with a temporary output directory (specified with something like `temp(directory("local_cache/{file_id}/"))`). Roughly every thousand or so rule executions I encounter a FileNotFound exception in snakemake/io.py, line 224, in _local_inventory

    with os.scandir(path) as scan:

This occurs after Snakemake logs that it has removed the temporary directory. _local_inventory() seems to have been added in commit 01cf67fd54f8c86e61a26da1d9b44518414ef20a

### LINKED ISSUE #640: PermissionError in io.py caused by unrelated file in same file tree
**Snakemake version**
```
$ snakemake --version
5.25.0
```

**Describe the bug**
Snakemake fails with a PermissionError when the folder structure contains inaccessible files, even though the files in question have no association with the current snakemake call.

We have a environment with multiple users. The users might want to execute the same pipeline on different files. Not all users have access rights to all files. The snakemake call fails when this is the case (see minimal example). This is caused by the os.scandir() call inside the _local_inventory function of the io.py script. 

**Minimal example**
We create the following example with two input files, of which only one is accessible by the current user:
```
$ cat minimal_example.sh
#!/bin/bash
mkdir -p Input/WithOUTRights
mkdir -p Input/WithRights
touch Input/WithOUTRights/File
touch Input/WithRights/File
chmod 000 Input/WithOUTRights/
```
Together with the following Snakefile:
```
$ cat Snakefile
rule all:
    input:
        'Input/{dir}/{file}'
    output:
        'Output/{dir}/{file}'
    shell:
        'cp {input} {output}'
```
When executing snakemake, an error occures:
```
$ snakemake -n Output/WithRights/File
Building DAG of jobs...
  File "/home/me/.conda/envs/snakemake/lib/python3.6/site-packages/snakemake/__init__.py", line 735, in snakemake
    keepincomplete=keep_incomplete,
  File "/home/me/.conda/envs/snakemake/lib/python3.6/site-packages/snakemake/workflow.py", line 683, in execute
    dag.init()
  File "/home/me/.conda/envs/snakemake/lib/python3.6/site-packages/snakemake/dag.py", line 177, in init
    job = self.update(self.file2jobs(file), file=file, progress=progress)
  File "/home/me/.conda/envs/snakemake/lib/python3.6/site-packages/snakemake/dag.py", line 715, in update
    progress=progress,
  File "/home/me/.conda/envs/snakemake/lib/python3.6/site-packages/snakemake/dag.py", line 792, in update_
    file.inventory()
  File "/home/me/.conda/envs/snakemake/lib/python3.6/site-packages/snakemake/io.py", line 210, in inventory
    self._local_inventory(cache)
  File "/home/me/.conda/envs/snakemake/lib/python3.6/site-packages/snakemake/io.py", line 224, in _local_inventory
    with os.scandir(path) as scan:
PermissionError: [Errno 13] Permission denied: 'Input/WithOUTRights'
```
The error refers to the directory inaccessible by the current user, even though this directory is unrelated to the current snakemake call.

**(Temporary) fix**
Following EAFP, the Error can be resolved by introducing an exception in the _local_inventory function
```
    def _local_inventory(self, cache):
        # for local files, perform BFS via os.scandir to determine existence of files
        root = cache.get_inventory_root(self)
        if root == self:
            # there is no root directory that could be used
            return
        if os.path.exists(root):
            queue = [root]
            while queue:
                path = queue.pop(0)
                # path must be a dir
                cache.exists_local[path] = True
                try:
                    with os.scandir(path) as scan:
                        for entry in scan:
                            if entry.is_dir():
                                queue.append(entry.path)
                            else:
                                # path is a file
                                cache.exists_local[entry.path] = True
                except PermissionError:
                    pass

        cache.has_inventory.add(root)
```
Resulting in a successful execution.
In case the users are unaware about their missing permission, this (only) results in a missing input file error:
```
$ snakemake -n Output/WithOUTRights/File
Building DAG of jobs...
MissingInputException in line 1 of /home/ruebsam/_Temp/snaketest/Snakefile:
Missing input files for rule all:
Input/WithOUTRights/File
```

**Additional information**
In our case the directories contain symlinks, which are accessible themselves. However, the access permissions for the linked files differ. Here a minimal example of this situation:
```
$ cat minimal_example.sh
#!/bin/bash
mkdir -p DistantDir/WithOUTRights
mkdir -p DistantDir/WithRights
touch DistantDir/WithOUTRights/File
touch DistantDir/WithRights/File
chmod 000 DistantDir/WithOUTRights/
mkdir Input
ln -s ../DistantDir/WithOUTRights/File Input/FileWithOUTRights
ln -s ../DistantDir/WithRights/File Input/FileWithRights
```
and
```
$ cat Snakefile
rule all:
    input:
        'Input/{file}'
    output:
        'Output/{file}'
    shell:
        'cp {input} {output}'
```
The error here comes from the .is_dir() call in the same function and gets fixed by the same change in the io.py script.

mhulsman:
File stat operations are slow on network filesystems such as Lustre. When stat operations are not cached,
preparing the DAG becomes very slow for large workflows, with multi-hour runs at <10% cpu usage.

This PR starts from the recently added IOCache inventory framework, and makes the following changes:

* Enable local inventory caching for absolute paths:
  (currently nonfunctional as root = path.split("/", maxsplit=1)[0] in get_inventory_root will 
   be an empty string, and therefore need_inventory will be False for these paths)
  
* Perform only an inventory of the directory of the file, and not of its subdirectories.
  This leads to more predictable behaviour, as an inventory is only obtained of directories
  with workflow files. 
  
* get_inventory_root/path is moved from IOCache to IO/Remote objects, to allow different strategies for local/remote file paths.

* Increase robustness of inventory by handling FileNotFoundErrors (fixes #611, fixes #623) and PermissionErrors (fixes #640)

* Add caching constants (e.g. IOCACHE_BROKENSYMLINK) and handling code to enable error reporting by cached responses.

* Add mtime_target caching dict which contains the mtime of the target of a symlink. This enables caching of the
  specific behaviour present in the is_newer function. 
  
* Adds background threading infrastructure to IOCache

* Enable two-staged caching, in which e.g. the file existence cache is already filled, but mtime/size is set to IOCACHE_DEFERRED.
  This two-staged approach is used for local inventory, where file existence inventory (scandir, cheap) is performed by the main  
  thread, while file stat inventory (mtime/size, expensive) is performed by background threads in IOCache.
  
* The recently added max_wait_time check was not kept for the local inventory. In my own use case, it would disable the
  inventory when it is needed most, i.e. in a large workflow with huge number of files. Also, with the non-recursive 
  inventory and background-threading, the inventory operation is still reasonably fast. 
  (Maybe an --disable-inventory flag can be considered for very small workflows that are run in huge directories.)

Together, this results in my use case (very large workflows on Lustre system) in a ~10 fold speedup.

### DIFF
--- a/snakemake/io.py
+++ b/snakemake/io.py
@@ -33,10 +33,11 @@ from inspect import isfunction, ismethod
 
 from snakemake.common import DYNAMIC_FILL, ON_WINDOWS
 
-#IO cache constants. Object to get unique id()
-IOCACHE_DEFERRED = object() #value is unknown during inventory operation
+# IO cache constants. Object to get unique id()
+IOCACHE_DEFERRED = object()  # value is unknown during inventory operation
 IOCACHE_BROKENSYMLINK = object()
 IOCACHE_NOTEXIST = object()
+
 
 def lutime(f, times):
     # In some cases, we have a platform where os.supports_follow_symlink includes stat()
@@ -179,12 +180,20 @@ def IOFile(file, rule=None):
 def cached_abspath(path):
     return os.path.abspath(path)
 
+
 class _IOFile(str):
     """
     A file that is either input or output of a rule.
     """
 
-    __slots__ = ["_is_function", "_file", "rule", "_regex", "inventory_root", "inventory_path"]
+    __slots__ = [
+        "_is_function",
+        "_file",
+        "rule",
+        "_regex",
+        "inventory_root",
+        "inventory_path",
+    ]
 
     def __new__(cls, file):
         # Remove trailing slashes.
@@ -210,7 +219,7 @@ class _IOFile(str):
 
             if file.endswith("/"):
                 obj.inventory_path = os.path.join(directory, name + "/")
-            else: 
+            else:
                 obj.inventory_path = os.path.join(directory, name)
             obj.inventory_root = directory
 
@@ -223,18 +232,24 @@ class _IOFile(str):
                 iocache = self.rule.workflow.iocache
                 if iocache.active:
                     cache = getattr(iocache, func.__name__)
-                    if (self.inventory_path in cache):  # first check if file is present in cache
+                    # first check if file is present in cache
+                    if self.inventory_path in cache:
                         res = cache[self.inventory_path]
-                    elif iocache.in_inventory(self.inventory_root):  # check if the folder was cached
+                    # check if the folder was cached
+                    elif iocache.in_inventory(self.inventory_root):
                         # as the folder was cached, we do know that the file does not exist
                         if raise_error:
-                             # make sure that the cache behaves the same as non-cached results
-                            raise FileNotFoundError("No such file or directory: {}".format(self.file))
+                            # make sure that the cache behaves the same as non-cached results
+                            raise FileNotFoundError(
+                                "No such file or directory: {}".format(self.file)
+                            )
                         else:
                             return False
                     elif self._is_function:
-                        raise ValueError("This IOFile is specified as a function and "
-                                                     "may not be used directly." )
+                        raise ValueError(
+                            "This IOFile is specified as a function and "
+                            "may not be used directly."
+                        )
                     else:
                         res = IOCACHE_DEFERRED
 
@@ -260,7 +275,6 @@ class _IOFile(str):
             return wrapper
 
         return inner_iocache
-
 
     def _refer_to_remote(func):
         """
@@ -314,9 +328,14 @@ class _IOFile(str):
             # this function is only called for local objects
             cache.exists_remote[path] = False
 
-            if ((not cache.mtime.get(path, IOCACHE_DEFERRED) is IOCACHE_DEFERRED)
+            if (
+                (not cache.mtime.get(path, IOCACHE_DEFERRED) is IOCACHE_DEFERRED)
                 and (not cache.size.get(path, IOCACHE_DEFERRED) is IOCACHE_DEFERRED)
-                and (not cache.exists_local.get(path, IOCACHE_DEFERRED) is IOCACHE_DEFERRED)):
+                and (
+                    not cache.exists_local.get(path, IOCACHE_DEFERRED)
+                    is IOCACHE_DEFERRED
+                )
+            ):
                 continue
 
             try:
@@ -348,7 +367,6 @@ class _IOFile(str):
                     else:
                         raise  # this will stop the thread, thereby slowing/disabling caching
 
-
     def _local_inventory(self, cache):
         # for local files, perform BFS via os.scandir to determine existence of files
         # obtaining mtime and size of the files is deferred for parallel execution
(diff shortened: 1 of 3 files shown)


johanneskoester:
Awesome work, thanks a lot! This is really valuable for me, as I do not use any slow network file systems and hence could only guess about the fine tuning that you have done here.

johanneskoester (approved):

johanneskoester:
@mhulsman I would like to give you credit in the next twitter announcement. Do you have a twitter account, or can at least give me full name and association?

mhulsman:
Thanks!  Twitter: HulsmanMarc

PR #676, opened 2020-10-11
[docs] Update mamba github link

kinow:
Following the docs after a co-worker mentioned Snakemake in an internal chat :+1: and found that the Mamba repository had changed, and GitHub is taking care to redirect users, but probably safer to just use the new URL.

Thanks for the great docs BTW!

Bruno


johanneskoester (approved):

johanneskoester:
Thanks a lot, and welcome!

Issue #598, opened 2020-09-03
Non-linear runtime behavior for complex workflows?


tobiasmarschall:
I'm using Snakemake 5.22.1. For my workflow, Snakemake becomes prohibitively slow (~4h until the first jobs are started) and I believe this might be unnecessary. My workflow looks like this. There is a list of around 3000 input files (A), each of which gets turned into one intermediate file for a total of 3000 intermediate files (B). Then there is a rule creating around 3000 further intermediate files (C), but this time, each of these rule executions depends on ALL of the files B. Then there is a rule that creates one desired final output file D using all the files C. After a partial run of Snakemake, all files B have already been created and I am invoking Snakemake again to create file D. From the runtime behavior I am guessing that Snakemake checks for presence and time stamp of all files B again and again for every file in C, creating a quadratic runtime behavior. I don't have time right now to investigate this further, but wanted to record the issue here.

PR #690, opened 2020-10-15
Seed up DAG resolution

jmeppley:
This PR is inspired by issue #598 and personal experience with slow DAG builds. It does two things to speed the DAG up:

### prevent update update_ recursion

First, in `dag.update()` and `dag.update_()` an new dictionary is introduced to keep track of input files that already have had producers found. So if a second (or more) rule needs the same input, dag.update() does not get called again. 

### skip useless calls to apply_wildcards

Second, when an new `_IOFile` is created, the string is scanned once for wildcards to flag the file as 'concrete' or not. Now, `_IOFile.apply_wildcards` just returns `self` for concrete files instead of calling the root `apply_wildcards()` and cloning the flags.

## test case
Following #598, I created this Snakefile:
```
N = config.get('N', 100)

# compute this list once, so that's not part of the slowness
step_1_files = expand('step_1/file_{n}', n=range(N))
rule all:
    input: expand('step_2/file_{n}', n=range(N))

rule step_1:
    output: "step_1/file_{n}"
    shell: "echo {wildcards.n} > {output}"
        
rule step_2:
    input:
        primary='step_1/file_{n}',
        alln=step_1_files
    output: 'step_2/file_{n}'
    shell: 'touch {output}'
```

Using the current snakemake, running for progressively larger N's gives non-linear increases in runtimes. The following are from dry runs with no outputs yet created:
```
N=25
CPU times: user 7.32 ms, sys: 4.73 ms, total: 12.1 ms
Wall time: 761 ms

N=50
CPU times: user 7.62 ms, sys: 8 ms, total: 15.6 ms
Wall time: 984 ms

N=100
CPU times: user 24.4 ms, sys: 5.17 ms, total: 29.5 ms
Wall time: 2.03 s

N=200
CPU times: user 92.7 ms, sys: 15.9 ms, total: 109 ms
Wall time: 7.54 s

N=400
CPU times: user 459 ms, sys: 89 ms, total: 548 ms
Wall time: 43 s

N=800
CPU times: user 3.02 s, sys: 581 ms, total: 3.6 s
Wall time: 4min 31s
```

With this PR:
```
N=50
CPU times: user 54.4 ms, sys: 2.02 ms, total: 56.4 ms
Wall time: 55.2 ms

N=100
CPU times: user 134 ms, sys: 1.99 ms, total: 136 ms
Wall time: 136 ms

N=200
CPU times: user 407 ms, sys: 7 ms, total: 414 ms
Wall time: 415 ms

N=400
CPU times: user 1.43 s, sys: 26.8 ms, total: 1.45 s
Wall time: 1.46 s

N=800
CPU times: user 4.98 s, sys: 28.9 ms, total: 5.01 s
Wall time: 5.02 s
```


jmeppley:
This is not quite ready, I'm taking it off the table for now.

jmeppley:
Some final thoughts for now:

The speedup I saw initially was false hope. I was able to almost double the speed in the end, but I'm still seeing non-linear behavior. One of the problems is that a new _IOFile object is being created for every job/file combination. When files appear in multiple rules, this escalates quickly. 

It might be (partially) solved by re-using IOFile objects when the point to a concrete resource, maybe separating IOFile into two classes: one for the more abstract and mutable rule inputs and outputs, and another for job inputs and outputs that point to specific resources.

PR #681, opened 2020-10-12
a suggested fix for directory support in Snakefile for GLS

vsoch:
This fix cannot be tested because I currently dont have access to a Google Cloud account, but Ill open it for review and discussion. This PR also depends on #665 - we won't be able to get to the remote upload step given that the path is not flagged as a directory.

Signed-off-by: vsoch <vsochat@stanford.edu>

PR #691, opened 2020-10-16
Test/gls directory output

vsoch:
This fix cannot be tested because I currently dont have access to a Google Cloud account, but Ill open it for review and discussion. This PR also depends on #665 - we won't be able to get to the remote upload step given that the path is not flagged as a directory.

This PR will replcae #681 so we can have testing of the directory upload - this isn't possible to easily debug because we currently don't have extra funds to spend on Google Cloud so we are hoping the test will be enough :)


vsoch:
@johanneskoester you are probably aware, but it looks like all testing that include Google Cloud is going to fail without the buckets, project, etc. Let me know if I can help further / what else we can do.

johanneskoester:
@vsoch, I suggest to create the buckets on the fly, similar to the kubernetes executor tests (in kubernetes tests, there is a fixture that can be reused/modified for that). Can you do that?

vsoch:
Error is:
```
[Mon Oct 19 16:34:00 2020]
Error in rule directory:
    jobid: 0
    output: ***-7aicwgdx/test_google_lifesciences/testdir
    log: ***-7aicwgdx/test_google_lifesciences/logs/directory.log (check log file(s) for error message)
    shell:
        mkdir -p ***-7aicwgdx/test_google_lifesciences/testdir; touch ***-7aicwgdx/test_google_lifesciences/testdir/test.txt
        (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)
```
I'm not sure how to work on this / debug just seeing the error in creating the directory.

johanneskoester:
@vsoch, Johanna has organized the credits. Please feel free to test in our shared gcloud project as before. Just make sure to delete instances and buckets as soon as possible once you are done.

vsoch:
okay here is how I tested! I created a Snakefile in the root of my snakemake work here (snakework?)
```
rule all:
    input:
        "testdir"

rule directory:
    output:
        directory("testdir")
    log:
        "logs/directory.log"
    shell:
        "mkdir -p {output}; touch {output}/test.txt"
```
 ensured that we had the snakemake-testing bucket created, exported my `GOOGLE_APPLICATION_CREDENTIALS` and then issued this command:

```
snakemake --default-remote-prefix snakemake-testing/dirtest --google-lifesciences --default-resources "machine_type=n1" --container-image snakemake/snakemake:latest
```
The basic idea is that this will run a worker for the test Snakefile, fail, but then I can inspect the worker to get the archive to download, create a instance on Compute Engine to mimic the worker, and then interactively debug. I've written up a gist that shows the steps [here](https://gist.github.com/vsoch/b4b9b6f27517209f3b9036726b166b79).

Now, let's interactively debug what happens on the upload. First, we confirm that the testdir and the file touched inside exist:

```python
ls snakemake-testing/dirtest/testdir/
test.txt
```
Okay, the first file for upload is the log file, and this works fine.

```python
'snakemake-testing/dirtest/logs/directory.log'
```
Now when I hit the directory, I'm able to fix the extra prefix of the bucket and have it upload as expected. But when we go back to the exists function, this is where the error arises - the blob does not exist.

```
self.blob
<Blob: snakemake-testing, dirtest/testdir>

self.blob.name
'dirtest/testdir'

self.blob.exists()
False
```

It doesn't make a difference if it ends with the slash or not. 

Okay just finished an hour or so of interactive debugging - the issue was deeper than we've been discussing. It's not just a matter of setting a new upload_blob in the remote upload function(), because self.blob (the blob attribute) needs to have the trailing slash *and* have the upload done using it for the later self.exists() to return True. Since self.blob is an attribute we can't update it, but instead need to have custom logic for the self.key. self.key also couldn't be a lazy property, because we would need it to update to include the trailing slash once the file exists and is determined to be a directory.

And I realize now we can't test it here because the container that the worker needs to use isn't pushed or developed yet! So what I did is built this branch into a container, and tested locally. But now the test that was working on the instance is failing locally. I've spent about 4 hours on this today and I'm really tired, so I need to stop the instances and try again another day.

vsoch:
I tested again, and found a (fairly silent) error when running this manually:

```
~/snakemake/dag.py in <module>
----> 1 f.mtime

~/snakemake/io.py in wrapper(self, *args, **kwargs)
    263                 iocache = self.rule.workflow.iocache
    264                 if not iocache.active or self.inventory_path is None:
--> 265                     return func(self, *args, **kwargs)
    266 
    267                 cache = getattr(iocache, func.__name__)

~/snakemake/io.py in wrapper(self, *args, **kwargs)
    321             if self.is_remote:
    322                 if hasattr(self.remote_object, func.__name__):
--> 323                     return getattr(self.remote_object, func.__name__)(*args, **kwargs)
    324             return func(self, *args, **kwargs)
    325 

~/snakemake/remote/GS.py in mtime(self)
    178             self.update_blob()
    179             t = self.blob.updated
--> 180             return t.timestamp()
    181         else:
    182             raise WorkflowError(

AttributeError: 'NoneType' object has no attribute 'timestamp'
```
Okay to step back, like 634 of dag.py, we start with a file for the remote folder, and it is shown to exist:

```
n [1]: f
Out[1]: 'snakemake-testing/dirtest/testdir'

In [2]: f.is_remote
Out[2]: <snakemake.remote.GS.RemoteObject at 0x7fe50055a340>

In [3]: f.remote_object.blob
Out[3]: <Blob: snakemake-testing, dirtest/testdir>

In [4]: f.remote_object.blob.exists()
Out[4]: True

In [5]: f.remote_object.key
Out[5]: 'dirtest/testdir/'

In [6]: f.remote_object.blob
Out[6]: <Blob: snakemake-testing, dirtest/testdir>

In [7]: f.remote_object.exists()
Out[7]: True

In [8]: f.remote_object._blob
Out[8]: <Blob: snakemake-testing, dirtest/testdir>

In [9]: f.remote_object._blob.exists()
Out[9]: True
```
So why would calling mtime mess that up? Let's walk through the steps manually. Here is the issue - calling update_blob on the directory makes it invalid (it no longer exists).

```python
f.remote_object.exists()
True
f.remote_object.update_blob()
f.remote_object.exists()
False
```
The reason I think comes down to the derivation of the key - it must be that the local files no longer exist after we "finish" so the key is not updated to be correct, and then the blob is updated to something that doesn't exist, the timestamp obviously can't be returned. I tried a derivation of the key function to not rely on the file existing:

```
    @property
    def key(self):
        key = self.parse().group("key")
        f = self.local_file()
        if snakemake.io.is_flagged(f, "directory"):
            key = key if f.endswith("/") else key + "/"
        return key
```
And then building a fresh container locally, pushing, and running again locally, now I have a new error that the input directory doesn't exist. Is this related to some change?

```bash
$ snakemake --default-remote-prefix snakemake-testing/dirtest --google-lifesciences --default-resources "machine_type=n1" --container-image vanessa/snakemake:dirtest
Building DAG of jobs...
Using shell: /bin/bash
Job counts:
	count	jobs
	1	all
	1
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/__init__.py", line 667, in snakemake
    success = workflow.execute(
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/workflow.py", line 986, in execute
    success = scheduler.schedule()
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/scheduler.py", line 446, in schedule
    run = self.job_selector(needrun)
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/scheduler.py", line 723, in job_selector_greedy
    c = list(map(self.job_reward, jobs))  # job rewards
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/scheduler.py", line 806, in job_reward
    input_size = job.inputsize
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/jobs.py", line 385, in inputsize
    self._inputsize = sum(f.size for f in self.input)
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/jobs.py", line 385, in <genexpr>
    self._inputsize = sum(f.size for f in self.input)
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/io.py", line 265, in wrapper
    return func(self, *args, **kwargs)
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/io.py", line 323, in wrapper
    return getattr(self.remote_object, func.__name__)(*args, **kwargs)
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/remote/GS.py", line 191, in size
    return self._iofile.size_local
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/io.py", line 680, in size_local
    self.check_broken_symlink()
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/io.py", line 685, in check_broken_symlink
    if not self.exists_local and os.lstat(self.file):
FileNotFoundError: [Errno 2] No such file or directory: 'snakemake-testing/dirtest/testdir'
```
I've tried changing it to output, adding directory() as well, it either reports there is nothing to be done (when there is) or issues this error. @johanneskoester any ideas?

johanneskoester:
Mhm, no immediate idea as well yet. I'll think about it. In the meantime, could you separate your test case fixes into another PR. Then, we could at least first fix master again.

vsoch:
Sure thing! See https://github.com/snakemake/snakemake/pull/698.

vsoch:
More testing! I updated the Snakefile to be slightly different, and this time I was able to get the job to run:
```
rule results:
    input:
        directory('data/test/')

rule minimal_test:
    resources:
        machine_type      = 'n1-standard-4', 
        disk_mb           = 10000           
    output:
        directory('data/test/')
    shell:
        '''
        mkdir -p {output}
        for i in $(seq 1 5); do echo $i > {output}/$i.txt; done
        '''
```
It gives me an error that a directory can't be used for input
```
The flag 'directory' used in rule results is only valid for outputs, not inputs.
Building DAG of jobs...
```
but the job still runs. Lo and behold, it's successful (there is no error) but it [fails on job.postprocess](https://github.com/snakemake/snakemake/blob/master/snakemake/jobs.py#L1001) because the directory doesn't exist locally. So I added a check to only check between file types given that the output isn't a remote object (because we don't have a good way to tell):

```
        # Ensure that outputs are of the correct type (those flagged with directory()
        # are directories and not files and vice versa). We can't check for remote objects
        for f in expanded_output:
            if (f.is_directory and not f.remote_object and not os.path.isdir(f)) or (
                not f.remote_object and os.path.isdir(f) and not f.is_directory
            ):
                raise ImproperOutputException(job.rule, [f])
```
(The not f.remote_object is what I added)

And then with that I was able to hit the download() function for getting the remote outputs, here we also needed to account for finding directories. I updated GS.py to handle this, adding a shared function to handle download of a blob (and only doing the checksum given it's not a folder). The changes are reflected in the PR here. Then I was able to do the run once more cleanly:

```
 snakemake --default-remote-prefix snakemake-testing/dirtest --google-lifesciences --default-resources "machine_type=n1" --container-image vanessa/snakemake:dirtest
The flag 'directory' used in rule results is only valid for outputs, not inputs.
Building DAG of jobs...
Using shell: /bin/bash
Job counts:
	count	jobs
	1	minimal_test
	1	results
	2

[Wed Oct 21 13:51:35 2020]
rule minimal_test:
    output: snakemake-testing/dirtest/data/test
    jobid: 1
    resources: machine_type=n1-standard-4, disk_mb=10000

Get status with:
gcloud config set project snakemake-testing
gcloud beta lifesciences operations describe projects/snakemake-testing/locations/us-central1/operations/11158340832637682533
gcloud beta lifesciences operations list
Logs will be saved to: snakemake-testing/dirtest/google-lifesciences-logs

[Wed Oct 21 13:53:15 2020]
Finished job 1.
1 of 2 steps (50%) done

[Wed Oct 21 13:53:15 2020]
localrule results:
    input: snakemake-testing/dirtest/data/test
    jobid: 0
    resources: machine_type=n1

Downloading from remote: snakemake-testing/dirtest/data/test
Finished download.
[Wed Oct 21 13:53:16 2020]
Finished job 0.
2 of 2 steps (100%) done
Complete log: /home/vanessa/Desktop/Code/snakemake/snakemake/.snakemake/log/2020-10-21T135128.015547.snakemake.log
```
And verify that we have all the output from the remote directory (including the hidden file)

```
$ ls snakemake-testing/dirtest/data/test/.*
snakemake-testing/dirtest/data/test/.snakemake_timestamp

snakemake-testing/dirtest/data/test/.:
1.txt  2.txt  3.txt  4.txt  5.txt
```
So I went back to the original recipe and updated it to look like this one:

```
rule all:
    input:
        directory("testdir/")

rule directory:
    output:
        directory("testdir/")
    log:
        "logs/directory.log"
    shell:
        "mkdir -p {output}; touch {output}/test.txt"
```
And that seemed to work this time! 
```
$ snakemake --default-remote-prefix snakemake-testing/testdir --google-lifesciences --default-resources "machine_type=n1" --container-image vanessa/snakemake:dirtest
The flag 'directory' used in rule all is only valid for outputs, not inputs.
Building DAG of jobs...
Using shell: /bin/bash
Job counts:
	count	jobs
	1	all
	1	directory
	2

[Wed Oct 21 13:56:09 2020]
rule directory:
    output: snakemake-testing/testdir/testdir
    log: snakemake-testing/testdir/logs/directory.log
    jobid: 1
    resources: machine_type=n1

Get status with:
gcloud config set project snakemake-testing
gcloud beta lifesciences operations describe projects/snakemake-testing/locations/us-central1/operations/10968913311222818404
gcloud beta lifesciences operations list
Logs will be saved to: snakemake-testing/testdir/google-lifesciences-logs

[Wed Oct 21 13:57:50 2020]
Finished job 1.
1 of 2 steps (50%) done

[Wed Oct 21 13:57:50 2020]
localrule all:
    input: snakemake-testing/testdir/testdir
    jobid: 0
    resources: machine_type=n1

Downloading from remote: snakemake-testing/testdir/testdir
Finished download.
[Wed Oct 21 13:57:50 2020]
Finished job 0.
2 of 2 steps (100%) done
Complete log: /home/vanessa/Desktop/Code/snakemake/snakemake/.snakemake/log/2020-10-21T135603.058453.snakemake.log
```
Now I have output for both
```
$ ls snakemake-testing/
dirtest/ testdir/ 

$ ls snakemake-testing/testdir/
testdir

$ ls snakemake-testing/testdir/testdir/
test.txt
```
The reason for the doubled "testdir" is because one is the prefix I asked for, the second is specified in the Snakefile. So I think we are almost there? Perhaps we need some more testers?

johanneskoester:
Ok, that looks good indeed. You don't need to specify `directory()` when refering to the output from the rule all. The flag is only needed when specifying the output.

johanneskoester (approved):

johanneskoester:
Maybe the missing directory flag causes the issues in the rule all then, because it will try to download the path as a file, right?

johanneskoester:
When downloading instead of determining whether it is a directory via the flag, we should rather see that from the google storage. I think what makes sense is to check whether <bucket/key/.snakemake_timestamp> exists. If that is the case, consider it to be a directory.

vsoch:
Okay so I tried updating snakemake to not have the check for remote at the end, first with removing the directory() around the input - it ran but only the first step.

```
rule all:
    input:
        "testdir"

rule directory:
    output:
        directory("testdir/")
    log:
        "logs/directory.log"
    shell:
        "mkdir -p {output}; touch {output}/test.txt"
```
```
$ snakemake --default-remote-prefix snakemake-testing/testdir --google-lifesciences --default-resources "machine_type=n1" --container-image vanessa/snakemake:dirtest
Building DAG of jobs...
Using shell: /bin/bash
Job counts:
	count	jobs
	1	all
	1

[Thu Oct 22 09:13:21 2020]
localrule all:
    input: snakemake-testing/testdir/testdir
    jobid: 0
    resources: machine_type=n1

[Thu Oct 22 09:13:21 2020]
Finished job 0.
1 of 1 steps (100%) done
Complete log: /home/vanessa/Desktop/Code/snakemake/snakemake/.snakemake/log/2020-10-22T091313.980538.snakemake.log
```
Note only running the all step in the above, and nothing else seems to happen.

```

I then tried changing it to specify "output" for the same, and it told me "nothing to be done" 

```
$ snakemake --default-remote-prefix snakemake-testing/testdir --google-lifesciences --default-resources "machine_type=n1" --container-image vanessa/snakemake:dirtest
Building DAG of jobs...
Nothing to be done.
```
which doesn't make sense because we don't have the directory locally or remotely.

```
rule all:
    output:
        "testdir"

rule directory:
    output:
        directory("testdir/")
    log:
        "logs/directory.log"
    shell:
        "mkdir -p {output}; touch {output}/test.txt"
```

I think I had added the directory() specification to "fix" this issue. What am I missing?

**Comment by johanneskoester** [comment]
I think the issue is that you have a trailing slash in the output of rule directory, but not for rule all. In the lastest master, this has been fixed to be always normalized away. But the version you used for testing presumably did not have this fix. Changing from input to output for rule all does not make sense (though I agree that it is unintuitive that it does not complain about the missing output in that case, I'll put that on my todo).

**Comment by vsoch** [comment]
@johanneskoester that might be it! I just tested with the adopted snakefile:
```
rule all:
    input:
        "testdir"

rule directory:
    output:
        directory("testdir")
    log:
        "logs/directory.log"
    shell:
        "mkdir -p {output}; touch {output}/test.txt"
```
And using the custom built container
```
snakemake --default-remote-prefix snakemake-testing/dirtest --google-lifesciences --default-resources "machine_type=n1" --container-image vanessa/snakemake:dirtest
```
That didn't seem to work (this is in red). Both with and without the slash currently fail.
```
ImproperOutputException in line 5 of /home/vanessa/Desktop/Code/snakemake/snakemake/Snakefile:
Outputs of incorrect type (directories when expecting files or vice versa). Output directories must be flagged with directory(). for rule directory:
snakemake-testing/dirtest/testdir
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/executors/__init__.py", line 814, in handle_job_success
  File "/home/vanessa/Desktop/Code/snakemake/snakemake/snakemake/executors/__init__.py", line 247, in handle_job_success
```
So we definitely need more testers.

**Comment by johanneskoester** [comment]
So, it seems like your change [here](https://github.com/snakemake/snakemake/pull/691/files#diff-7898111b777083eaffbb7e986fdf700073f388165a2f752a1477b374eff7ee52R488) does not properly work. That is the only case where ImproperOutputException is thrown.

**Comment by vsoch** [comment]
I'm not really sure what to think, I've poured many hours into this and tried debugging this so many times and it seems to be inconsistent. Maybe someone else can take a look.

**Comment by johanneskoester** [comment]
Ok, I will have a look now.

**Comment by johanneskoester** [comment]
Ok, just tried the following and it worked fine:
```
snakemake --default-remote-provider GS --default-remote-prefix snakemake-testing-jk -j1
```

johanneskoester:
Waiting now for the tests on the latest version of this branch. If they don't pass, it must be something with the life science backend. Maybe just the container image being not in sync again.

vsoch:
There is no way the tests could pass because they are using an older container, What I typically have to do is specify --container-image with a container I just built and pushed for the branch.

johanneskoester:
makes sense...

johanneskoester:
The dirtest-labelled container image that you used seems quite old to me. Maybe just update to the latest version of this branch?

vsoch:
Would it be feasible for PRs to build and push a development image, built from the codebase being tested directly? Of course named by something like the commit. Then of course delete it / clean up after.

johanneskoester:
Yes, that would be great for such cases. For others it would be an enormous overhead though...

johanneskoester:
Maybe upon request.

johanneskoester:
We could check for a label in the PR, and only trigger the build and push step if the label is present.

johanneskoester:
Anyway I think I found the problem. Your dirtest container is 7 days old. But 6 days ago you modified the if clause which decides about the ImproperOutputException. So I think the container image just needs to be updated.

vsoch:
Okay I’ll rebuild and try again when I officially wake up, still before 6am here :)

johanneskoester:
Wait first, I am currently trying to enable the docker build and push upon label strategy.

johanneskoester:
Yay, all passed!

johanneskoester:
Thanks a lot @vsoch!

vsoch:
Thank you to *you*! I really appreciate the help, and I'm so glad you like the idea of adding the development container build for the test - it's going to make it hugely easier for developers that want it.

Issue #417, opened 2020-05-29
`--list-code-changes` does not work with `include`


haje01:
I have two Snakefiles, one for common feature engineering and the other for predictive models. Each is written in a separate directory.

Predictive model's Snakefile includes feature engineering one. The problem occurs when I call `snakemake --list-code-changes` in predictive model directory.

```
$ snakemake --list-code-changes

Building DAG of jobs...
Traceback (most recent call last):
  File "/Users/haje01/.pyenv/versions/3.7.2/envs/ml2/lib/python3.7/site-packages/snakemake/__init__.py", line 654, in snakemake
    keepincomplete=keep_incomplete,
  File "/Users/haje01/.pyenv/versions/3.7.2/envs/ml2/lib/python3.7/site-packages/snakemake/workflow.py", line 764, in execute
    items.extend(list(j.outputs_older_than_script_or_notebook()))
  File "/Users/haje01/.pyenv/versions/3.7.2/envs/ml2/lib/python3.7/site-packages/snakemake/jobs.py", line 208, in outputs_older_than_script_or_notebook
    assert os.path.exists(path)  # to make sure lstat works
AssertionError
```

I guess it's because Snakemake can't find the source code under the included directory. Is this an intended behavior?

Thank you in advance.

My environment:
Snakemake 5.15.0
macOS Catalina 10.15.4

jbloom:
This issue should be fixed in #695

johanneskoester:
Thanks for reporting, fixed in master now.

PR #695, opened 2020-10-19
fix `--list-code-changes` for included rules

jbloom:
Previously the `--list-code-changes` option did not work for included rules  with the `script` or `notebook` directive if the rule was included in a subdirectory

The reason was that the code looked for the included file relative to the top directory rather than the rules subdirectory.

This commit fixes this, thereby addressing #417


johanneskoester (approved):

johanneskoester:
Thank you!

Issue #682, opened 2020-10-12
bug in report for version 5.26.1


jbloom:
For version 5.26.1, I get the following error sometimes when trying to build a report:
```
snakemake/report/__init__.py", line 295, in code
    return [
TypeError: 'NoneType' object is not iterable
```

The problem seems to be when the language cannot be identified, which leaves the variable `source` assigned to `None` in `RuleRecord.code`. In the past (version 5.25), this simply lead to a simple return of a string [as here](https://github.com/snakemake/snakemake/blob/v5.25.0/snakemake/report/__init__.py#L282). But now in version 5.26.1, it tries to build a return object by iterating over `source` (see [here](https://github.com/snakemake/snakemake/blob/v5.26.1/snakemake/report/__init__.py#L295-L298)), which causes an error as `source` is `None` and so is not iterable.

johanneskoester:
Should be fixed in current master.

johanneskoester:
A new release will come soon.

PR #696, opened 2020-10-19
code no longer crashes on report for run rules

jbloom:
This commit fixes a bug in v5.26.* that was causing `snakemake --report` to crash when there were `run` rules.

The code has never handled this well: it seems like the desired behavior would be to make the "source" in the report for `run` rules to actually be the Python code being run. In < v5.26, instead in just printed "source" instead of the Python code. But then in v5.26.*, it was changed so that the code was trying to iterate over a `None` object, so the `snakemake --report` directive actually crashed when there was a `run` rule. This commit makes it go back to doing the old behavior of just printing "source", so at least it doesn't crash even if the report is uninformative as to the code being run.

Ideally, someone who knows how to actually access the code for `run` cells could fix this better.

Addresses issue #682.

### DIFF
--- a/snakemake/report/__init__.py
+++ b/snakemake/report/__init__.py
@@ -288,7 +288,7 @@ class RuleRecord:
             # change simply makes the report just put "source" in place of the
             # the actual code source, rather than throwing an error which
             # crashes the entire generation of the `snakemake` report.
-            sources = ['source']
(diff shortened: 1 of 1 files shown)


johanneskoester:
Thanks a lot! I have solved this in a slightly different way in PR #704.

Issue #292, opened 2020-03-25
Checkpoints slow in large workflows


jonaszierer:
**Is your feature request related to a problem? Please describe.**
Updating of depending rules takes really long (and doesn't seem to scale linearly) with many targets. As an example: Take the checkpoint example from the [docs](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#data-dependent-conditional-execution) and change the all rule to 

```
rule all:
    input: expand("aggregated/{sample}.txt", sample = range(10000))
```

When having in the order of 1000s samples here the updating of the dependencies takes some minutes, however when moving to the 10000s every single step takes seconds, so just updating the dependencies would take hours. From the logs:

```
Updating job 49 (aggregate).
Replace aggregate with dynamic branch aggregate
updating depending job all
```
It seems that the updating depending job all is the slow step...

**Describe the solution you'd like**
Update the DAG faster even for >10k 'samples'

**Describe alternatives you've considered**
One could run a single job to process all 'samples' in parallel, though then one looses all the snakemake magic of checking if output was created correctly etc....

**Additional context**
'sample' from the example doesn't have to be samples. If one want's to run one analysis per gene/transcript/protein etc it is easy to create workflows with >10k targets

epruesse:
I'm experiencing the same issue on with 5.20 to 5.24. The checkpoint feature is extremely slow. I've got only a few hundred samples, but before a half-finished workflow even starts, I'm seeing a large number of "Updated job NNN (joining_job_name)" messages come in, each taking approx. 2 seconds.

@johanneskoester I love this new feature! What is it doing there that's taking so long? Any way to accelerate it? It feels like this should incur less effort.

jmeppley:
I've seen similar behavior, too, and would vote for an improvement here.

epruesse:
The problem is that the loop below checks a lot of other jobs for each checkpoint job as far as I can tell. I did some benchmarking with yappi and it looks like that loop causes O(n*n) behavior. Best I can say for now. Also not something easy to fix unless you really understand the rest of the code.

https://github.com/snakemake/snakemake/blob/0fb0a2adaf561cb6cd0c827983c93246947baa85/snakemake/dag.py#L1220-L1239

jmeppley:
PR #697 might help this somewhat. I was able to remove a little of the O(n*n) behavior from `dag.update_needrun()` which is called by `dag.postprocess()`.

[...]

PR #697, opened 2020-10-19
Seed up DAG resolution for complex workflows

jmeppley:
This PR is inspired by issue #598 and personal experience with slow DAG builds. It does a few things to speed the DAG up. Combined, they halve the run time for my needlessly complex test workflow:

### prevent update update_ recursion
First, in `dag.update()` and `dag.update_()` an new dictionary is introduced to keep track of input files that already have had producers found. So if a second (or more) rule needs the same input, `dag.update()` does not get called again.

### re-use _FileIO object when already concrete
Second, when an new `_IOFile` is created, the string is scanned once for wildcards to flag the file as 'concrete' or not. Now, `_IOFile.apply_wildcards` just returns `self` for concrete files instead of calling the root apply_wildcards() and cloning the flags.

### reduce duplicate calls to missing_output in update_needrun
By keeping track of which files have already been checked, we can reduce the time spent in the two calls to `dag.update_needrun()`

### speed up missing_output
I added a second version of `job.missing_output()` specific for the case where it's only checking a subset of files

## test case
I created this Snakefile that has the behavior described in #598:
```
N = config.get('N', 100)

# compute this list once, so that's not part of the slowness
step_1_files = expand('step_1/file_{n}', n=range(N))
rule all:
    input: expand('step_2/file_{n}', n=range(N))

rule step_1:
    output: "step_1/file_{n}"
    shell: "echo {wildcards.n} > {output}"
        
rule step_2:
    input:
        primary='step_1/file_{n}',
        alln=step_1_files
    output: 'step_2/file_{n}'
    shell: 'touch {output}'
```

Using the current snakemake:
```
N = 50
Wall time: 388 ms
N = 100
Wall time: 968 ms
N = 200
Wall time: 4.49 s
N = 400
Wall time: 24.2 s
```

With this PR:
```
N = 50
Wall time: 313 ms
N = 100
Wall time: 572 ms
N = 200
Wall time: 2.07 s
N = 400
Wall time: 8.3 s
```

Runtime is still proportional to N*N, but it's significantly faster for all sizes of N.


jmeppley:
I think this would also help with #292

snakemake/jobs.py
```diff
@@ -551,6 +542,41 @@ def output_mintime(self):
 
         return mintime
 
+    def missing_requested_output(self, requested_output):
```
johanneskoester, line 545: I do not yet understand why it is beneficial to have this quasi duplicated method. Why not just take the original one. Or alternatively, reimplement the original one while specifying job.output as requested_output.

johanneskoester (changes requested):
Awesome, thanks a lot! Two questions below.

snakemake/jobs.py
jmeppley: The original method was is called in two different scenarios.

In the first, all the outputs are checked. In this case, the original method is a little more efficient.

In the second, only one file or a few files is/are checked at a time. This is what I wrote the new method for.  

If the first case is not a bottleneck, it would be OK to use the new method for both, but I haven't checked that.

snakemake/dag.py
```diff
@@ -938,18 +964,28 @@ def update_needrun(job):
...
             for job_, files in dependencies[job].items():
-                missing_output = job_.missing_output(requested=files)
+                # assume a give file can only come from one job
+                #  only check files we haven't seen before from this job
+                unknown_files = files.difference(known_files)
+                missing_output = job_.missing_requested_output(unknown_files)
+
+                # save newly found missing files to reason and known dict
                 reason(job_).missing_output.update(missing_output)
+                known_files.update({f: (f in missing_output) for f in unknown_files})
```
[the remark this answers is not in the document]
jmeppley: I initially expected to need to know whether the previously checked files existed or not. It looks like that never gets used and a set would be fine.

snakemake/jobs.py
johanneskoester: I have set up a comprehensive benchmark for PR #710. Would it be ok for you if I pick individual changes from this PR and include them manually? Of course, I will give you the credit in the commit messages.

johanneskoester: And in the release.

jmeppley:
Absolutely! Please incorporate these (some or all) in whatever way works
best for you.

johanneskoester:
Closing in favor of PR #717. Thanks a lot, this was of great help!

PR #699, opened 2020-10-22
Rewrite of inventory functionality with dramatically reduced number of stat calls.

johanneskoester:
* Also make remote and local inventory run asyncronously via coroutines.
* Asynchronously retrieve mtimes (via coroutines) before needrun computation.

This rewrite has been greatly inspired by PR #675 by @mhulsman. It this previously merged PR with a simplified codebase, uses coroutines instead of threads and futher reduces the number of stat calls.

This PR also changes Snakemake's required Python version to >=3.7 because of the coroutine usage.


mhulsman:
Very nice idea to use the job files to further reduce stat usage.  I quickly scanned through the code, and I noticed that is_newer is not cached now. This is actually where  the majority of the 'stat' time is spent (see last comment on PR #663 for some benchmarking).  In PR #675, this required a separate mtime_target cache, due to the handling of mtime of symbolic links and their targets.

johanneskoester:
I'll have a look.

johanneskoester:
Thanks for pointing me to that. Now, is_newer properly uses mtime, and mtime is cached for local, local_target and remote in one object. The rewrite not only leads to properly cached is_newer, it also reduces the number of stat calls for that to 1 instead of 4.

Issue #680, opened 2020-10-12
New CLI argument to disable config argument type inferencing.


haje01:
I have a Snakemake based ETL process which takes a config parameter for date to process. It runs like this:

    snakemake --config date=2020-10-09 -j

In this case, the type of date parameter is `string`, and the value is '2020-10-09'. So, I implemented the code for that.

But with following command:

    snakemake --config date=2020-10-10 -j

My code raises an error. Because, now the type of date parameter is `int`, and the value is 2000. I guess some kind of smart implicit type inferencing and evaluation is working here. 

I think explicit is better than implicit. But, you may have reasons. So, I suggest simple extra command line argument to disable type inferencing:

    snakemake --config date=2020-10-10 --disable-config-type-inferencing -j

I would like to hear you opinions. 

Thank you.

nhartwic:
I agree that passing "config" values through the CLI sucks. I don't think an option to disable type inference really solves anything though. The real problem is that config values are passed using an unspecified and ad hoc format. The sollution to this is to switch to a widely used format that snakemake already uses and that can work at the command line: json. Your command would be...

```
snakemake --config ' {"date": "2020-10-10"} ' -j
```

No more ambiguity. Works super cleanly with configfiles. Only drawback is that json is a bit ugly and needs to be quoted properly

nhartwic:
I can probably put together a pull request to implement this, but it seems like a large change to the CLI and I'm hesitant to do so without input from other, more experienced snakemake contributors.

Does anyone have any thoughts about switching to json for passing CLI config options? Was anyone around when the current system was being implemented? Since it seems like using json is the obvious thing to do, I have to ask if there is a reason this wasn't done before?

dariober:
@nhartwic 

```
snakemake --config ' {"date": "2020-10-10"} ' -j
```

This doesn't work for me on snakemake 5.14.0:

```
snakemake --config ' {"date": "2020-10-10"} ' -j
Invalid config definition: Config entries have to be defined as name=value pairs.
```

(I can't/don't want to upgrade snakemake because of this issue https://github.com/snakemake/snakemake/issues/575)

nhartwic:
You can't use a json string to pass config values via the 'config' option in any existing version of snakemake. This issue is about an enhancement proposal, a change to current functionality.

In the meantime, you will probably need to write your json string to a file and then pass that file to snakemake using the 'configfile' option. 

@dariober

bingxiao:
Hi Authors,

[...]

PR #709, opened 2020-10-29
replace 'eval' parser with 'yaml' for command line config

bingxiao:
For user input parsing, `eval` is a powerful function but not recommended in
principle. In daily scenario, it does unnecessary data parsing or math
calculation even when not explicitly asked by users, like '1+1' -> 2, and '1_1'
-> 11. As reported in issue: https://github.com/snakemake/snakemake/issues/680

If simply remove 'eval' parser, the data conversion and security issue will be
instantly addressed. However, to continue enabling complex data structure inputs
such as list and dictionary, here I propose `yaml BaseLoader ` as a safe
replacement since parsing user input has been a well established field in yaml.
Comparing to `json`, yaml also have some advantage of less typing.


johanneskoester (approved):
This is indeed a very good idea. Thank you!

Maarten-vd-Sande:
This actually changes the behaviour of CLI config parsing.

Let's pretend we have a snakefile and call snakemake like this: `snakemake -j1  --config mybool=True`

If we were to print our config it would look like this:
`{'mybool': 'True'}`

And with snakemake version 5.27.4 it would print:
`{'mybool': True}`

This is probably not an intended side-effect of this change?

bingxiao:
@Maarten-vd-Sande 
Thank for considering the same question. I've been searching for a better `yaml`-based solution recently for a while. For data type in list and dict, the `BaseLoader`, though safer than `eval`, loads any scalar value (`int, float, bool`) as `str`. While the other loaders behave more consistently, it may also have more aggressive type-conversions than users may want. So one have to evaluate situations and make a sensible and predictable agreement.

I'm now considering using a `SafeLoader` with DateTime auto-conversion switched-off. Likely: https://stackoverflow.com/a/37958106

About YAML BaseLoader and SafeLoader: https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

PR #717, opened 2020-11-02
Further speed improvements inspired by PR #697

johanneskoester:
Some of the optimizations in PR #697 have become superfluous due to previous optimizations from PR #710. The rest has been included here, in slightly different ways but the ideas are the same.

# TODO
In the squash merge, mention @jmeppley!

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -733,7 +733,7 @@ class DAG:
                 self.update_(
                     job,
                     visited=set(visited),
-                    known_producers=None,
+                    known_producers=known_producers,
                     skip_until_dynamic=skip_until_dynamic,
                     progress=progress,
                     create_inventory=create_inventory,
@@ -845,6 +845,7 @@ class DAG:
                         res.jobs,
                         file=res.file,
                         visited=visited,
+                        known_producers=known_producers,
                         skip_until_dynamic=skip_until_dynamic
                         or res.file in job.dynamic_input,
                         progress=progress,
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -16,7 +16,7 @@ from pathlib import Path
 import uuid
 import math
 
-from snakemake.io import PeriodicityDetector, wait_for_files, is_flagged
+from snakemake.io import PeriodicityDetector, wait_for_files, is_flagged, IOFile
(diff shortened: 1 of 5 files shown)


jmeppley:
Yay! I'm glad that was still helpful, even after #710. 

There were two other ideas I played with, that showed promise. They were not simple, and I haven't checked how they effect things after this PR and #710. 

You may have already considered these, but just in case it's helpful:

 ***Re-using IOFile objects***: Files used in multiple jobs get a new _IOFile object each time. The iocache solves some of this by saving the actual file states, but I think there is still some room here. I am picturing splitting _IOFile into _AbstractIOFile for rules and _ConcreteIOFile for jobs (and maybe explicit locations in rules). The major sticking point was how to handle flags set by different rules/jobs on the same file.

 ***Postpone logger str formatting***: Unlike the python logger, the snakemake logger does not let you pass a template and values instead of a full message. So the job of building up messages gets done even if they are never going to be printed. I don't think there's a ton of time lost to unused debug messages in normal mode, but there are enough info messages that delaying formatting might speed up runs with --quiet.

johanneskoester:
> Yay! I'm glad that was still helpful, even after #710.
> 
> There were two other ideas I played with, that showed promise. They were not simple, and I haven't checked how they effect things after this PR and #710.
> 
> You may have already considered these, but just in case it's helpful:
> 
> _**Re-using IOFile objects**_: Files used in multiple jobs get a new _IOFile object each time. The iocache solves some of this by saving the actual file states, but I think there is still some room here. I am picturing splitting _IOFile into _AbstractIOFile for rules and _ConcreteIOFile for jobs (and maybe explicit locations in rules). The major sticking point was how to handle flags set by different rules/jobs on the same file.
> 
> _**Postpone logger str formatting**_: Unlike the python logger, the snakemake logger does not let you pass a template and values instead of a full message. So the job of building up messages gets done even if they are never going to be printed. I don't think there's a ton of time lost to unused debug messages in normal mode, but there are enough info messages that delaying formatting might speed up runs with --quiet.

Indeed, logging will be overhauled soon, making it both more concise as well as faster (the logger code is a historically grown mess).

Regarding IOFile sharing, this will indeed be very hard to get right. At least they are already shared to a large extend because of Pythons string interning. So only the layer with the flags and some booleans is actually separate per object.

PR #721, opened 2020-11-03
update conda env command

KasperThystrup:
`conda env` does not work, corrected to `conda-env`.

### DIFF
--- a/docs/tutorial/setup.rst
+++ b/docs/tutorial/setup.rst
@@ -134,7 +134,7 @@ The ``environment.yaml`` file can be used to install all required software into 
 
 .. code:: console
 
-    $ conda env create --name snakemake-tutorial --file environment.yaml
+    $ conda-env create --name snakemake-tutorial --file environment.yaml
 
 Step 4: Activating the environment
 ::::::::::::::::::::::::::::::::::


mbargull:
@KasperThystrup, both `conda env` and `conda-env` should work.
`conda-env` is a separate program that is installed alongside `conda`.
`conda` also delegates any non-builtin subcommand `some-command` to an executable `conda-some-command` if it exists., e.g., `conda env` will run `conda-env`.
If `conda env` doesn't work, it indicates that your setup is in an unexpected state

----
As for the change from this PR itself:
Usually it's fine to run `conda-env` directly. There can be a subtle difference in some cases: If `conda` is setup up as intended and called from a shell, then the `conda` shell function is used. If you install `conda` into another activated environment than `base` but have the `conda` shell function from `base` active, then `conda env` (invoked from the shell) will run the `conda-env` from the `base` environment but `conda-env` (also, if invoked from the shell) will run the `conda-env` from the activated environment.
Usually you want to run the `conda`/`conda-env` from the `base` environment and as such probably want `conda env` instead of `conda-env`.
(I know, it's a bit confusing and not obvious at all, unfortunately.)

KasperThystrup:
@mbargull Thank you so much for taking your time to provide such a detailed description. I was not aware that my (mini)conda setup could be flawed, I definitely will have to look into that!

Issue #725, opened 2020-11-04
Asyncio issue with python < 3.7


cclienti:
With the snakemake version 5.27.4 and python 3.6 , I got the following error:

```
Building DAG of jobs...
Traceback (most recent call last):
  File ".../lib/python3.6/site-packages/snakemake/__init__.py", line 751, in snakemake
    keepmetadata=keep_metadata,
  File ".../lib/python3.6/site-packages/snakemake/workflow.py", line 696, in execute
    dag.init()
  File ".../lib/python3.6/site-packages/snakemake/dag.py", line 191, in init
    self.update_needrun(create_inventory=True)
  File ".../lib/python3.6/site-packages/snakemake/dag.py", line 900, in update_needrun
    self.workflow.iocache.mtime_inventory(self.jobs)
  File ".../lib/python3.6/site-packages/snakemake/io.py", line 136, in mtime_inventory
    async_run(self._mtime_inventory(jobs))
  File ".../lib/python3.6/site-packages/snakemake/common.py", line 34, in async_run
    return loop.run_until_complete(coroutine)
  File ".../lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
    return future.result()
  File ".../lib/python3.6/site-packages/snakemake/io.py", line 156, in _mtime_inventory
    tasks = [asyncio.create_task(worker(queue)) for _ in range(n_workers)]
  File ".../lib/python3.6/site-packages/snakemake/io.py", line 156, in <listcomp>
    tasks = [asyncio.create_task(worker(queue)) for _ in range(n_workers)]
AttributeError: module 'asyncio' has no attribute 'create_task'
```

I've checked and the asyncio.create_task() method has been added in python 3.7.

Maybe we can just use the following syntax:
```python
asyncio.get_event_loop().create_task()
```

johanneskoester:
Thanks for reporting. Fixed with PR #726.

PR #726, opened 2020-11-05
Fix asyncio create_task syntax for compatibility with python < 3.7

cclienti:
The `asyncio.create_task` method was introduced with python 3.7. To be
compatible with older version, we should use
`asyncio.get_event_loop().create_task`

https://github.com/snakemake/snakemake/issues/725


johanneskoester:
Thanks a lot!

PR #731, opened 2020-11-06
add and use compat shims for python3.6

graingert:


johanneskoester (approved):

johanneskoester:
Thanks a lot, good catch!

PR #753, opened 2020-11-18
parse_config recognizes bools

Maarten-vd-Sande:
With the changes made in https://github.com/snakemake/snakemake/pull/709 booleans (True/False) do not get recognized anymore as booleans, but as their string representation. My guess is that was an unintended side-effect. This should fix that.

example command: `snakemake -j1 --config mybool="True" --dryrun --quiet` and if we would print the config in the Snakefile

5.27.4
{'mybool': True}

5.28.0
{'mybool': 'True'}

this PR
{'mybool': True}

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -873,12 +873,20 @@ def parse_key_value_arg(arg, errmsg):
     return key, val
 
 
+def _bool_parser(value):
+    if value == "True":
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):

johanneskoester:
Thank you!

Issue #741, opened 2020-11-12
error during report generation for run rules


hunter-cameron:
**Snakemake version**
5.27.4

**Describe the bug**
When trying to generate a report for workflows that have a "run" rule in them, I got a `TypeError: '<' not supported between instances of 'NoneType' and 'int'` error. This occurs because there is no start time recorded for run rules. 

I did some digging and it seems that run rules might be getting scheduled twice which clobbers the file snakemake uses to get the start time for the stats because one run finishes and cleans up the file and then the next run cannot read the start time (but this run writes the metadata). Since it cannot find the file, it becomes None in the metadata which, when compared to an integer, throws the error. I verified the double scheduling wasn't happening with shell directives, only with run.

However, I haven't been able to figure out where to fix this so I can't submit a merge request. 

**Logs**
```
user@:/analysis$ snakemake --cores 1 -pr --report test.html all
Building DAG of jobs...
Creating report...
Adding shell_rule.txt (0 MB).
Adding run_rule.txt (0 MB).
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/snakemake/__init__.py", line 668, in snakemake
    success = workflow.execute(
  File "/usr/local/lib/python3.8/dist-packages/snakemake/workflow.py", line 816, in execute
    auto_report(dag, report, stylesheet=report_stylesheet)
  File "/usr/local/lib/python3.8/dist-packages/snakemake/report/__init__.py", line 758, in auto_report
    rec.starttime = min(rec.starttime, meta["starttime"])
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```
**Minimal example**

Snakefile used:
```
rule shell_rule:
    """ A rule that runs a shell command """
    output: report("shell_rule.txt")
    shell:
        "touch {output}"

rule run_rule:
    """ A rule that runs a 'run' command """
    output: report("run_rule.txt")
    run:
        with open(output[0], 'w'):
            pass


rule all:
    input:
        rules.shell_rule.output,
        rules.run_rule.output
```

**Additional context**

lparsons:
I'm running into the same issue with `5.27.4`

lparsons:
Did not experience the error with `5.26.1`

lparsons:
I believe the problem here is this change:
https://github.com/snakemake/snakemake/pull/663/files#diff-ea35c3ed2e47884b4280b7d8d7be5bcf8672ddf661fcbe0b1e2268f17de1e239L178-R225

[...]

PR #767, opened 2020-11-24
Fix metadata starttime

lparsons:
The starttime metadata for run rules would be set to None since the `_incomplete_path` did not seem to exist for them. Instead, fall back to using the `_metadata_path`.

I do not fully understand the changes that caused this bug, and thus, while this seems to fix things to me, I don't know that this is the right way to fix things.  @mhulsman or @johanneskoester You two worked on the PR related to this, perhaps you could help me understand the proper way to fix things?

Resolves: #741

### DIFF
--- a/snakemake/persistence.py
+++ b/snakemake/persistence.py
@@ -231,9 +231,10 @@ class Persistence:
         fallback_time = time.time()
         for f in job.expanded_output:
             rec_path = self._record_path(self._incomplete_path, f)
-            if not os.path.exists(rec_path):
-                rec_path = self._record_path(self._metadata_path, f)
             starttime = os.path.getmtime(rec_path) if os.path.exists(rec_path) else None
+            # Sometimes finished is called twice, if so, lookup the previous starttime
+            if not os.path.exists(rec_path):
+                starttime = self._read_record(self._metadata_path, f).get("starttime", None)
             endtime = f.mtime.local_or_remote() if f.exists else fallback_time
             self._record(
                 self._metadata_path,
--- a/snakemake/persistence.py
+++ b/snakemake/persistence.py
@@ -231,6 +231,8 @@ class Persistence:
         fallback_time = time.time()
         for f in job.expanded_output:
             rec_path = self._record_path(self._incomplete_path, f)
+            if not os.path.exists(rec_path):
+                rec_path = self._record_path(self._metadata_path, f)
             starttime = os.path.getmtime(rec_path) if os.path.exists(rec_path) else None
             endtime = f.mtime.local_or_remote() if f.exists else fallback_time
             self._record(
--- a/snakemake/persistence.py
+++ b/snakemake/persistence.py
@@ -234,7 +234,9 @@ class Persistence:
             starttime = os.path.getmtime(rec_path) if os.path.exists(rec_path) else None
             # Sometimes finished is called twice, if so, lookup the previous starttime
             if not os.path.exists(rec_path):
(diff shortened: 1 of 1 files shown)


snakemake/persistence.py
```diff
@@ -231,6 +231,8 @@ def finished(self, job, keep_metadata=True):
         fallback_time = time.time()
         for f in job.expanded_output:
             rec_path = self._record_path(self._incomplete_path, f)
+            if not os.path.exists(rec_path):
+                rec_path = self._record_path(self._metadata_path, f)
```
mhulsman, line 235: I think this addresses the problem of calling finished twice. 
However, small remaining issue is that after the second 'finished', I believe this will give an incorrect starttime equal to the endtime of the job, as the flow becomes:

finished 1:
write metadata-tracker file (with correct starttime from creation time of incomplete-tracker file, etc).
remove incomplete-tracker file

finished 2:
do not find incomplete-tracker file, use metadata-tracker file as source.
write new metadata-trakcer file, but now with incorrect starttime from creation time of metadata-tracker file. 

To solve this, in finished 2 you could parse (_read_record) the metadata-tracker file, and obtain the starttime from it.

mhulsman:
Hi, thanks for looking at this. I assumed that a job can only finish once, but that is apparently not the case. I'm not sure if that is intended, but at least I think it is good to make this code robust against it. 

Additionally, I'm not sure why the test-suite didnt pick this up, so I agree it might be a good idea to add a test for this.

snakemake/persistence.py
lparsons: > To solve this, in finished 2 you could parse (_read_record) the metadata-tracker file, and obtain the starttime from it.

I'm happy to make that change, but I'm a bit unclear at the moment as to how to do it. Could you post some suggested code? I'll also start poking around, but I'm not familiar with the internals here, so I'm a bit lost.

Thanks for talking a look at this. I would agree that if `finished` is being called twice, that is likely an underlying problem (and should probably be solved). However, I'm not sure how that is happening, and thus I'm not able to debug at the moment.

lparsons:
> Additionally, I'm not sure why the test-suite didnt pick this up, so I agree it might be a good idea to add a test for this.

The test suite only runs snakemake once per test. For the report tests, the test directory contains the workflow already completed, along with metadata. Thus, the test suite is not actually testing the generation of the metadata. I'm not sure how to validate the metadata once it's generated, other than running a report, which parses it, so I don't really know the best way to go about writing a test for this. Any suggestions would be welcome.

lparsons:
@mhulsman I took a stab at reading the starttime from the metadata file in the case when the incomplete file doesn't exist. Hopefully this will get the correct starttime.

mhulsman:
Yes, that looks good.  

W.r.t. to the test: sorry cant help right now, maybe next week.  You can have multiple snakemake calls  (multiple calls to 'run' function in tests.py). But maybe an option needs to be added to make the report call run in the same tmpdir as the workflow call.

johanneskoester:
> Yes, that looks good.
> 
> W.r.t. to the test: sorry cant help right now, maybe next week. You can have multiple snakemake calls (multiple calls to 'run' function in tests.py). But maybe an option needs to be added to make the report call run in the same tmpdir as the workflow call.

Exactly! I'll merge this fix to have it working in the master branch for the upcoming release, but I would appreciate a follow-up PR with a testcase a lot.

johanneskoester (approved):

PR #772, opened 2020-11-26
Fix #488

### LINKED ISSUE #488: Snakemake doesn't ensure read permission of jobscript
We found out, that Snakemake doesn't ensure that the u+r bit is set, when it writes a jobscript to call qsub with. It only ensures that u+x is set, but this is not enough. Instead it should be u+rx, because in case of weird permission inheritance, it can occur, that the permissions as reported by os.stat().st_mode are 32768 aka no permissions. You should be able to reproduce this by running a `umask 0666` prior to executing a qsub-Snakemake file.

Simply changing
https://github.com/snakemake/snakemake/blob/9d2d64d7be385bf88f262d6dd201d06ed4ca7570/snakemake/executors/__init__.py#L739
to
```python
        os.chmod(jobscript, os.stat(jobscript).st_mode | stat.S_IXUSR | stat.S_IRUSR)
```

should be enough to fix this.

Phhere:
Enforce read and execut bit on job script.
Weird acl inheritance can cause that os.stat() reports no permissions but read and execute are required.

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -779,7 +779,7 @@ class ClusterExecutor(RealExecutor):
         logger.debug("Jobscript:\n{}".format(content))
         with open(jobscript, "w") as f:
             print(content, file=f)
-        os.chmod(jobscript, os.stat(jobscript).st_mode | stat.S_IXUSR)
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):

johanneskoester:
Thank you!

Issue #783, opened 2020-12-04
Subworkflows only use one node with cluster backend


votti:
**Snakemake version**
5.30.1
**Describe the bug**
Using a cluster backend (Slurm via https://github.com/Snakemake-Profiles/slurm) I noticed that jobs in Subworkflows are not run in parallel.

With local execution, rules parallelize as expected.

**Logs**
-

**Minimal example**
Folder structure:
- Snakefile:
```
subworkflow subworkflow1: 
    workdir: 'subworkflow1'

rule all:
    input: subworkflow1('all.done')
```
- subworkflow1/Snakefile:
```
rule:
    input: [f'{i}.txt' for i in range(10)]
    output: touch('all.done') 

rule:
    output: '{nr}.txt' 
    shell:
        '''
        sleep 100
        touch {output[0]}
        '''
```

Local execution:
```
snakemake --cores 10
```
-> Runs jobs in parallel

Cluster execution:
```
snakemake --jobs 10 --profile slurm
```
-> Runs jobs sequentially

**Additional context**
To me this seems very related to https://github.com/snakemake/snakemake/issues/208

**Proposed fix**
Looking at the code I think the reason could be that the 'nodes' argument is not passed to `subsnakemake` in https://github.com/snakemake/snakemake/blob/master/snakemake/workflow.py

Is there any reason this should not be passed akin to the cores argument?

I quickly tried this on a branch and this seems to fix this issue: https://github.com/BodenmillerGroup/snakemake/tree/feature-swf-nodes

PR #723, opened 2020-11-03
enable (local) multithreaded subworkflows

goi42:
Passes cores to subworkflows at evaluation time. Resolves #208.

PR #784, opened 2020-12-04
Fix Issue #783: Pass `nodes` to subworkflows to allow parallel execution on cluster backend.

votti:
In my hands this fixes the behavior observed in issue #783, allowing subworkflows jobs to also be executed in parallel .

This was achieved by passing the 'nodes' argument to subworkflows.

This fix is essentially a copy of: https://github.com/snakemake/snakemake/pull/723

Is there anything else that would similarly profit from being passed to `subworkflows`?


johanneskoester:
Should be fine like this, thanks!

johanneskoester (approved):

PR #776, opened 2020-11-27
Implement Paramspace helper and testcase.

johanneskoester:
This is inspired by the work on JUDI from @soumitrakp. Thanks a lot Soumitra!
Key idea:
Snakemake already brings everything to handle parameter spaces in any way (sub-spacing for certain rules and even depending on wildcard values, the ability to read or generate spaces on the fly or from files via pandas, etc. However, there was so far no easy way to translate a space into wildcard patterns, and translate it back into concrete parameters for scripts and commands. This is now solved via the new `Paramspace` helper, which can be used as follows:

```python
from snakemake.utils import Paramspace
import pandas as pd

# declare a dataframe to be a paramspace
paramspace = Paramspace(pd.read_csv("params.tsv", sep="\t"))


rule all:
    input:
        # Aggregate over entire parameter space (or a subset thereof if needed)
        # of course, something like this can happen anywhere in the workflow (not 
        # only at the end).
        expand("results/plots/{params}.pdf", params=paramspace.instance_patterns)


rule simulate:
    output:
        # format a wildcard pattern like "alpha~{alpha}/beta~{beta}/gamma~{gamma}" 
        # into a file path, with alpha, beta, gamma being the columns of the data frame
        f"results/simulations/{paramspace.wildcard_pattern}.tsv"
    params:
        # automatically translate the wildcard values into an instance of the param space
        # in the form of a dict (here: {"alpha": ..., "beta": ..., "gamma": ...})
        simulation=paramspace.instance
    script:
        "scripts/simulate.py"


rule plot:
    input:
        f"results/simulations/{paramspace.wildcard_pattern}.tsv"
    output:
        f"results/plots/{paramspace.wildcard_pattern}.pdf"
    shell:
        "touch {output}"
```

Given that `params.tsv` contains:
```
alpha	beta	gamma
1.0	0.1	0.99
2.0	0.0	3.9
```

This workflow will run as follows:
```
[Fri Nov 27 20:57:27 2020]
rule simulate:
    output: results/simulations/alpha~2.0/beta~0.0/gamma~3.9.tsv                                                                                                                           
    jobid: 4                                                                                                                                                                               
    wildcards: alpha=2.0, beta=0.0, gamma=3.9                                                                                                                                              

[Fri Nov 27 20:57:27 2020]
rule simulate:
    output: results/simulations/alpha~1.0/beta~0.1/gamma~0.99.tsv                                                                                                                          
    jobid: 2                                                                                                                                                                               
[...]

PR #787, opened 2020-12-06
Paramspace patterns

HenningTimm:
This PR adds a parameter allowing custom patterns to the Paramspace class implemented in #776 
A string formatting pattern like `"{}/{}_{}"` can be passed to achieve different connections between parameter values. This allows for shallower directory structures when many parameters are used.
Currently, in this PR there is no verification that the supplied pattern contains the right number of formatting fields. Too many format fields result in `IndexError: tuple index out of range`, but too few fields could go unnoticed. 

The only solid way to achieve this that I could come up with would be adding the following steps to the `__init__`:
1. Try to format the pattern string using an amount of dummy values equal to the number of supplied params. If this fails, raise an error since there are too few params.
2. Try to format the pattern string using exactly one dummy parameter less than before. If this fails, the expected amount of format fields was supplied and everything is ok. If this formatting works, there were too many parameters: Raise an error.

An easier solution would be to count the number of opening and closing curly braces. However, this fails to properly report the number of fields for patterns that contain escaped curly braces. While I would consider this to be an unlikely case, using this approach could lead to arcane and hard to debug errors.


johanneskoester:
As mentioned in the other PR, can we make the placeholders in the pattern named instead of plain `{}` placeholders?

johanneskoester:
On the other hand, if there are many columns it can be cumbersome to specify the whole pattern, and even impossible to do so. Maybe it is better to just offer
```
Paramspace(df, filename_params=["alpha", "beta"])
```
and if filename_params is None, Paramspace will just use the last column for the filename?

HenningTimm:
Worked on a first draft of this that seems to work. Tests etc. are still missing.

HenningTimm:
The last commits add a parameter to select the symbol used to join the parameter name and value. The default remains `'~'`. There are currently some errors, but since they also occurred on the main branch, I am confident that this code is not to blame ;)

Testing locally I observed the inverse problem: The test cases I wrote finished without issue even though the generated files did not match the structure in `expected-results`. Maybe I am doing something wrong there.

johanneskoester:
Can you merge in latest master from upstream?

HenningTimm:
Thank you, that solved the test errors. I still need to add a test for the join string parameter before this is ready to merge.

johanneskoester (approved):

johanneskoester:
@HenningTimm any chance to get the final testcase today? And a merge with upstream master? I have already included paramspace into the paper, and hope it to appear online soon.

HenningTimm:
@johanneskoester  OK, no problem. I will get this done today.

HenningTimm:
@johanneskoester Everything runs as intended now and the test seem to work. However, the comparison with the content of the "expected-results" folder does not seem to work. I will open a new PR for that.

Issue #743, opened 2020-11-12
subworkflow order enforcing


DrYak:
**Is your feature request related to a problem? Please describe.**

We have use cases where several sub-workflows need to be chained together, in our case: a first sub-workflow for downloading and preformatting sequencing data, another to process the data and find variants, and the finally the main workflows analyses the result of the later. The downloading is specific to this experiment (other use of the processing pipeline could uses other sources) and thus it cannot be made into a subworkflow of _that_ subworkflow.

Current [documentation of Snakemake state explicitely that all sub-workflow are executed prior to the main workflow](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html#sub-workflows):
>  A sub-workflow is executed independently before the current workflow is executed. 

Thus it is not possible to enforce some order, e.g.: using rules in the main workflow.

It turns out that snakemake stores its subworkflow in a dictionnary, and [starting with Python 3.7 according to documentation](https://docs.python.org/3.7/library/stdtypes.html#typesmapping):
> Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.

Thus it happens that the order in which subworkflow are mentionned in the main workflow will influence their execution order, so we could have a temporary workaround solution working for us, BUT...

**Describe the solution you'd like**

...but, the above ordering happens to be a lucky accident of the peculiar implementation of snakemake (dictionaries) and the peculiar platform it's currently running on (python >= 3.6).

It would be better with snakemake could have some `subworkfloworder:` directive, that either:
 - gave an explicit order, and snakemake runs first these subworkflow in that order, and then any remaining subworkflow
 - gave a list of `>` compared elements, that could then in turn be used to guide a `sort` of the keys (I would have said _a custom comparing function in a sort_ using [cmp_to_key python 3 adapter](https://docs.python.org/3/library/functools.html#functools.cmp_to_key) )

**Describe alternatives you've considered**

As mentionned, it happens to accidentally work in the current implementation, but it's relying on a happy coincidence that might get broken in future implementations of snakemake.
Having an official subworkflow reordering would make it easier.

**Additional context**

We're currently a bit busy with research into the current pandemic, so there won't be pull-request with suggestions for implementing that feature soon. Please be patient.

nhartwic:
[...]

PR #791, opened 2020-12-08
Extend subworkflow functionality

kpj:
As discussed in #743, this PR will extend the functionality of subworkflows to make them behave more rule-like.

Among other things, this will include:
* allowing execution of subworkflows not only at the beginning of the main workflow
* allowing to define subworkflow input files
* including subworkflows in the DAG visualization


kpj:
After taking a look at the code and thinking some more, I believe there are three general ways of implementing this feature:

* integrate subworkflow DAG during DAG generation of the main workflow

* use a checkpoint-like approach to re-generate the main-DAG by adding the sub-DAG to it, when the respective rule is executed

* spawning a whole new Snakemake process, when the respective rule is executed

It seems to me like the first item would be the cleanest but most difficult approach, while the last item might be suboptimal but rather easy to figure out (I hope?).

DrYak:
I would add that the last point isn't dissimilar from how snake jobs are handled on clusters:
The execution node currents gets to run its own snakemake session locally, for one single rule instance.
While the main snakemake scheduler patiently waits for the files to pop-up.

snakemake/parser.py
```diff
@@ -611,6 +611,22 @@ def args(self):
         )
 
 
+class Subworkflow_RuleLike(Script):
```
nh13, line 614:
```suggestion
class SubworkflowRuleLike(Script):
```
Not sure what naming convention this was, but it should be upper camel case for class names

snakemake/parser.py
```diff
@@ -611,6 +611,22 @@ def args(self):
...
+    start_func = "@workflow.subworkflow"
+    end_func = "subworkflow"
+
+    def args(self):
+        # basedir
+        yield ", {!r}".format(os.path.abspath(os.path.dirname(self.snakefile.path)))
```
nh13, line 620: f-strings are faster and more concise

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+    bench_record,
+    jobid,
+    bench_iteration,
+    cleanup_scripts,
+    shadow_dir,
+    workflow,
+):
+    # extract (and normalize) parameters
+    basedir = os.path.normpath(basedir)
+    sub_workdir = os.path.normpath(params.get("workdir", f"subworkflow_{rulename}"))
```
nh13, line 32: For many sub-workflows, this could get messy in the working directory.  How about:
```suggestion
    sub_workdir = os.path.normpath(params.get("workdir", f"subworkflows/{rulename}"))
```

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+    sub_configfile = params.get("configfile", "")
```
nh13, line 33:
```suggestion
    sub_configfile = params.get("configfile")
```
don't use an empty string when you really want `None`

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # execute sub-workflow
+    logger.info("Executing sub-workflow")
+
+    workflow.subsnakemake(
+        path,
+        workdir=sub_workdir,
+        targets=output_normalized,
+        cores=threads,
+        configfiles=[sub_configfile] if sub_configfile else None,
```
nh13, line 63:
```suggestion
        configfiles=[sub_configfile] if sub_configfile is not None else None,
```

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+    # extract (and normalize) parameters
+    basedir = os.path.normpath(basedir)
+    sub_workdir = os.path.normpath(params.get("workdir", f"subworkflow_{rulename}"))
+    sub_configfile = params.get("configfile", "")
+
+    # the sub-workflow needs to run in a different directory than the
+    # main-workflow. This is necessary due to Snakemake's locking mechanism
+    if basedir == sub_workdir:
+        raise WorkflowError(
+            "Sub-workflow workdir must be different from main-workflow workdir"
```
nh13, line 39: consider including the two directories in the message

nh13: Also, you control the `sub_workdir` so should this ever happen (isn't it a bug if it does?)

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+    # all output files must use the sub-workflow's workdir as their prefix
+    # otherwise file dependencies cannot be easily handled
```
nh13, line 43: I don't get why this is the case?  Why is this a requirement?  

I can see many cases where I'll have many sub-workflows all outputting to the same output directory.

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+            raise WorkflowError(
+                "All output files must begin with "
+                f"(normalized) sub-workflow workdir ({sub_workdir})"
+            )
+
+        fname_rel = os.path.relpath(fname, start=sub_workdir)
+        output_normalized.append(fname_rel)
+
+    # execute sub-workflow
+    logger.info("Executing sub-workflow")
```
nh13, line 56:
```suggestion
    logger.debug("Executing sub-workflow")
```

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+
+    workflow.subsnakemake(
+        path,
+        workdir=sub_workdir,
+        targets=output_normalized,
+        cores=threads,
+        configfiles=[sub_configfile] if sub_configfile else None,
+    )
+
+    logger.info("Finished execution of sub-workflow")
```
nh13, line 66:
```suggestion
    logger.debug("Finished execution of sub-workflow")
```

snakemake/subworkflow.py
nh13: Probably want the subworkflow name in here too (and down below)

nh13 (changes requested):
I think we should be very careful about how subworkflows are implemented and not rush things.  On my snakemake wishlist has been first-class support for subworkflows to enable modular composition.  In my view, a subworkflow should behave exactly like a rule.  In fact, I don't want them to be very different from an API/DSL perspective.  They have inputs/outputs and other directives.  It's just that a subworkflow has a set of rules to that it may execute to make the subworkflow's outputs from the subworkflow's inputs, versus a single rule having a shell/run/etc directive.

I am a bit concerned that the requirement in this PR that the subworkflow outputs must be written to the subworkflow's working directory.  I can see many cases where the subworkflow, and even rules inside the subworkflow, write to wherever I please (i.e. not in the subworkflow's working directory).  I think there is an issue if a rule outside the subworkflow and a rule inside the subworklow can (and do) produce the same output file.  I am not sure what to do about that.

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+    input,
+    output,
+    params,
+    wildcards,
+    threads,
+    resources,
+    log,
+    config,
+    rulename,
+    conda_env,
```
nh13, line 19: do rules within a sub-worfklow overide the conda environment?

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+    container_img,
```
nh13, line 20: do rules within a sub-worklow override this, or run docker in docker

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+
+def subworkflow(
+    path,
+    basedir,
+    input,
+    output,
+    params,
+    wildcards,
+    threads,
+    resources,
```
nh13, line 15: are resources enforced in sub-workflows?

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def subworkflow(
+    path,
+    basedir,
+    input,
+    output,
+    params,
+    wildcards,
+    threads,
```
nh13, line 14: are threads enforced in sub-workflows?

snakemake/subworkflow.py
```diff
@@ -0,0 +1,66 @@
...
+from snakemake.logging import logger
+from snakemake.exceptions import WorkflowError
+
+
+def subworkflow(
+    path,
+    basedir,
+    input,
+    output,
+    params,
```
nh13, line 12: how are params passed to rules in a subworkflow? Are they exposed in the config dict?

snakemake/parser.py
kpj: There was actually no naming convention here :D
I was only looking for a placeholder which would not clash with the already existing `Subworkflow` class (assuming that this class would porbably be deprecated at some point).

snakemake/parser.py
kpj: Absolutely!

I used `format` because other classes also use this style (`CWL`, `Script`, ...).
If wanted, I could change all of them to use f-strings.

snakemake/subworkflow.py
kpj: Agreed, great idea!

snakemake/subworkflow.py
kpj: The user can specify the workdir by setting `params.workdir` in the respective rule.
So checking this still seems sensible (at least in the current approach).

snakemake/subworkflow.py
kpj: I'll discuss this further in a reply in the main thread.

kpj:
@nh13: Thanks a lot for your suggestions! I adapted most and commented on a few.

I'll address the other points you raised below.

> I think we should be very careful about how subworkflows are implemented and not rush things.

I do very much agree. This PR's main purpose was to get the ball rolling, and hopefully come up with a neat idea and implementation.

> I am a bit concerned that the requirement in this PR that the subworkflow outputs must be written to the subworkflow's working directory. I can see many cases where the subworkflow, and even rules inside the subworkflow, write to wherever I please (i.e. not in the subworkflow's working directory).

My thought was that (i) a subworkflow can never be executed in the same workdir as the main workflow (due to Snakemake's directory locking mechanism) and (ii) the rule output files have to match the subworkflow output files.
In fact, the output files do not have to reside in the sub-workflows workdir (think of `subworkflow/sub_workdir/../../../letsgowherever/foo.txt`). This if of course not very intuitive and can cause issues when using absolute paths (but who would use those, right?).
However, by enforcing this requirement workdirs will definitely (most likely) differ and matching subworkflow files to rule output files should be somewhat clear.

However, I'd be very open to new and better approaches for how to solve this.

> I think there is an issue if a rule outside the subworkflow and a rule inside the subworklow can (and do) produce the same output file. I am not sure what to do about that.

That is another very good point. This could be trivially handled if we would simply integrate the subworkflow's DAG into the main DAG. But that would make this whole implementation more difficult. I am currently also not sure what the best way to go would be.

---

On a different note, I'd also be happy to modify the DAG visualization of rule-like subworkflows. Firstly, one could simply change their node style to distinguish them from normal rules. Secondly, one could refactor the plotting functions to be able to embed the subworkflow's DAG in the main DAG visualization.

snakemake/subworkflow.py
kpj: You raise a few good points about which directives to propagate.
I have not thought about this in details.

lasflo:
It is probably included in the idea of making subworkflows behave like rules, but the ability to add wildcards both in workdir and config file could be neat. This would allow the possibility of having another rule make the configfile for a subworkflow.

snakemake/subworkflow.py
kpj: The `params` directive could generally be used to modify the sub-workflows setup (as `params. configfile` and `params.workdir` currently do).
Furthermore, `params.config` could be merged into the sub-workflows configuration, `params. restart_times` could set sub-workflow specific restart times, etc.

snakemake/subworkflow.py
kpj: They should be (I hope), because I set `cores=threads` further below.

snakemake/subworkflow.py
kpj: Is there something I'd have to do manually? When executing on a cluster, the cluster profiles should take care of this, shouldn't they?

nh13:
I was thinking about this more, and my ideal would be a subworkflow defines the scope of the rules that it contains, such that those rules that it contains can only be applied to generate the subworkflow outputs from the required subworkflow inputs.  The main or top-level workflow (Snakefile) can contain rules that apply to its scope, while subworkflows has rules that only apply in its own scope.  Then nested sub-workflows are possible.  A subworkflow can then be viewed like any other `rule`: it generates outputs from the required inputs, and so on.  When a subworkflow is encountered, Snakemake can just run the same solver it uses for the full workflow, but only using the rules found in the subworkflow (and any other subworkflow).  This makes solving very easy, as it in effect acts as constraints on which rules can be applied in which scope.  It also has the effect that the dag is (almost) the same as if all rules (including rules in subworkflows) were available and fully embedded in the DAG.  

@johanneskoester I think fleshing the idea of subworkflows to in effect limit the scope of rules and make subworkflows "look like" rules is a big win in modular composition of workflows.  Thoughts?

kpj:
@nh13 Do I understand your comment correctly, that you propose to integrate the sub-workflow's DAG into the main-workflow's DAG? At the moment I do not know what the best way of implementing this would be.

Have any further thoughts transpired?

johanneskoester:
These are all very interesting and reasonable ideas. I need to find some time to think about it, before deciding into one direction prematurely. Ideally, I'd only change the subworkflow implementation once, and then in a way that is perfect and covers all use cases.

kpj:
I agree, sounds very reasonable!
Looking forward to your feedback :-)

johanneskoester:
So, thanks again for drafting this, and for the discussion. This was really incredibly helpful! All these thoughts have now lead to the new module system of Snakemake, just merged from PR #888. This makes any further changes to subworkflows unnecessary.

kpj:
That is great, thanks a lot!
I am very much looking forward to using the new module system :-)

mschubi:
Hi, I have a main Snakefile and several subworkflows running in independent subdirectories (with paths relative to their own directories). I would like to have these subworkflows integrated in the main dag, so if I understand correctly this is not possible and will not be implemented ?  After reading your comment, I tried the module system. Maybe I don't use it correctly, but it doesn't seem to work for my use case. If I import a rule from a subdirectory it complains that there are missing inputs. It doesn't find the scripts because they are in the subdirectory and not in the main directory. Is that the expected behaviour ?

PR #795, opened 2020-12-09
Allow Paths wherever wildcards are applied

sappjw:
This includes expanding params, inputs, and log files. This fixes a bug when providing a Path or list of Path as the return value of an input function, for example.


johanneskoester (approved):

johanneskoester:
thank you!

Issue #799, opened 2020-12-13
Update to panoptes to be logger proper


vsoch:
Hi @johanneskoester ! I'm working with @fgypas and we are putting together a panoptes [Monitor Schema](https://github.com/panoptes-organization/monitor-schema/) so that instead of being a one off implementation of panoptes here in Snakemake, any server tool that implements the [spec](https://github.com/panoptes-organization/monitor-schema/blob/main/spec.md) can serve to monitor snakemake jobs. Likewise, any tool that implements the client (like snakemake) can interact with any server.

Here is a proposed timeline for how we'd like to go about doing this, and we are looking for your feedback and (hopefully) blessing!

1. First we'd ask for your feedback on the [Monitor schema](https://github.com/panoptes-organization/monitor-schema). You can see our discussion of changes in [this issue](https://github.com/panoptes-organization/monitor-schema/issues/1). The design inspiration is from the opencontainers distribution spec, namely having a rooted namespace (/m1/) and then a set of well defined endpoints and interactions that a server should implement, and a client should expect.
2. Then pin the version of panoptes here in snakemake to the current so we can work on a new one without breaking anything.
3. Next do slight tweaks to panoptes so that it confirms to the monitor spec, and do a new release.
4. Finally, we'll PR here to update snakemake for the new release. @fgypas has suggested that the wms_monitor should be a Logger proper (e.g., instead of an [if statement here](https://github.com/snakemake/snakemake/blob/3705d94d87e14636fe5d58ab03542517026f2581/snakemake/logging.py#L511) it would be akin to the Slack logger) and I agree. 

There are some more advanced features I'd like to see, for example adding simple authentication tokens so when a workflow is submit, the endpoint will only work for that job (and someone on the shared resource that discovers it can't ping it maliciously). The Snakemake Interface I'm working on will plug into this standardized set of endpoints easily, and can also issue tokens in the header to better secure things.

Let us know your feedback! We are coming up on break so I'm very happy and excited to contribute to all of the above!

johanneskoester:
Hi @vsoch and @fgypas this is a great plan and I fully support it!

vsoch:
@johanneskoester and @fgypas I'm going to start working on snakemake updates soon and want to confirm a few things first!

 - @fgypas I'm not going to be making any changes to the current API since that is under discussion, I'll just be refactoring what already exists into a logger.
[...]

PR #800, opened 2020-12-14
Adding future note about panoptes monitoring

vsoch:
This is a tiny note to add to the monitor docs to prepare for the refactor of Panoptes as a logger proper. Of course when some new version is developed, I'll come back for another PR to make the changes and further update these docs! This is related to #799 


vsoch:
@johanneskoester there aren't any changes to code here so I'm wondering why this is coming up. It looks like the core error is:
```
ERROR    snakemake.logging:logging.py:402 AttributeError in line 6 of /tmp/snakemake-wdokifod/Snakefile:
'_patch' object has no attribute 'is_local'
```
And this might be because of referencing a property on a mocked class? https://stackoverflow.com/questions/31709792/patching-a-class-yields-attributeerror-mock-object-has-no-attribute-when-acce. Was this working before?

vsoch:
I'm running it again in case it was transient, but likely not.

johanneskoester:
> @johanneskoester there aren't any changes to code here so I'm wondering why this is coming up. It looks like the core error is:
> 
> ```
> ERROR    snakemake.logging:logging.py:402 AttributeError in line 6 of /tmp/snakemake-wdokifod/Snakefile:
> '_patch' object has no attribute 'is_local'
> ```
> 
> And this might be because of referencing a property on a mocked class? https://stackoverflow.com/questions/31709792/patching-a-class-yields-attributeerror-mock-object-has-no-attribute-when-acce. Was this working before?

it seems to be an incompatibility in moto. I hope it will go away with a future release. For now, I skip the test.

johanneskoester (approved):

Issue #727, opened 2020-11-05
Jobs updated by checkpoint are not getting run


jmeppley:
**Snakemake version**
works:
 * 5.20.0
 * 5.25.0
 * 5.26.0
 * 5.26.1

fails:
 * HEAD (5.27.4+1.g4a63a269) 
 * 5.27.4
 * 5.27.0

**Describe the bug**
The test run for our [DTR-phage-pipeline](https://github.com/nanoporetech/DTR-phage-pipeline) has stopped working recently. It has a few checkpoints and 160+ total jobs, so it's not the simplest test case, but it did work in earlier version.

Instead of completing in one go, it stops after each checkpoint. DEbugging output indicates that the downstream jobs are getting updated, but somehow it stops anyway. Restarting snakemake continues where it left off.

**Logs**
Here is the end of the output from the first run:
```
checkpoint generate_bins_and_stats:
    input: test/test-output/25m/1D/v2/kmer_binning/bin_membership.tsv
    output: test/test-output/25m/1D/v2/kmer_binning/bin_stats.csv, test/test-output/25m/1D/v2/kmer_binning/bins
    jobid: 11
Downstream jobs will be updated after completion.

[Thu Nov  5 14:15:24 2020]
rule plot_umap_freq_map_bins:
    input: test/test-output/25m/1D/v2/kmer_binning/bin_membership.tsv
    output: test/test-output/25m/1D/v2/kmer_binning/kmer_comp.umap.bins.png
    jobid: 9
                                                                                                                                                                                                                 
python /mnt/data0/jmeppley/projects/DTR/TBD/DTR-phage-pipeline/scripts/plot_umap_with_hdbscan_labels.py -o test/test-output/25m/1D/v2/kmer_binning/kmer_comp.umap.bins.png -s 7 -a 0.2 test/test-output/25m/1D/v2/kmer_binning/bin_membership.tsv
No kaiju DB found in /media/sfodata/databases/kaiju/nr_euk.
Skipping Kaiju
No kaiju DB found in /media/sfodata/databases/kaiju/nr_euk.
Skipping Kaiju
Job counts:
        count   jobs
        1       generate_bins_and_stats
        1
Activating conda environment: /mnt/data0/jmeppley/projects/DTR/TBD/DTR-phage-pipeline/.snakemake/conda/53e51836
Updating job 12 (combine_bin_cluster_read_info).
Replace combine_bin_cluster_read_info with dynamic branch combine_bin_cluster_read_info
[...]

PR #801, opened 2020-12-15
Possible fix for issue #727

jmeppley:
This adds `finished` property to `jobs.Reason` so that jobs don't get flagged when `DAG.udpate_needrun()` is triggered by a finished checkpoint.


jmeppley:
@johanneskoester 

I think that a job's `Reason` staying `True` after the job completes is the root of at least some of the checkpoint problems, but I'm sure there are other ways to address it than my `finished` property. The key is that the job's reason needs to resolve to False after it's run successfully. Some other approaches I considered:

 * replacing the `Reason` with an new (Fasle/empty) one when a job finishes
 * adding a method to reset the Reson's state (called when the job finishes)
 * patching `update_needrun` to account for the possibility of stale Reasons
 * re-evaluating all Reasons after completed checkpoints

The `finished` property seemed logical and simple to me, but I'm still figuring out how everything is put together.

johanneskoester (approved):

johanneskoester:
Excellent work, thanks a lot. This makes sense to me.

PR #812, opened 2020-12-18
remove R_LIBS, PYTHONPATH, etc. when using conda env

johanneskoester:
(can be deactivated via the command line).


karldw:
It's maybe worth noting that R also uses `R_LIBS_USER` and `R_LIBS_SITE` variables, in addition to `R_LIBS`.
Docs: https://stat.ethz.ch/R-manual/R-devel/library/base/html/libPaths.html

Issue #817, opened 2020-12-22
Unassigned jobid after checkpoint update


jmeppley:
**Snakemake version**
the lastest from github: 5.30.2+6.g674a91ca

**Describe the bug**
After a checkpoint completes in some complicated workflows, an exception is thrown because a jobid is requested for a job that does not have a jobid yet.

**Logs**
This is the exception:
```
Traceback (most recent call last):
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.272/DTR-phage-pipeline/snakemake/snakemake/__init__.py", line 687, in snakemake
    success = workflow.execute(
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.272/DTR-phage-pipeline/snakemake/snakemake/workflow.py", line 699, in execute
    dag.update_checkpoint_dependencies()
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.272/DTR-phage-pipeline/snakemake/snakemake/dag.py", line 1300, in update_checkpoint_dependencies
    logger.info("Updating job {} ({}).".format(self.jobid(j), j))
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.272/DTR-phage-pipeline/snakemake/snakemake/dag.py", line 704, in jobid
    return self._jobid[job]
KeyError: align_dtr_seqs_to_polished_refs
```

**Minimal example**
A reproducible way to generate this error is to download the [DTR-phage-pipeline](https://github.com/nanoporetech/DTR-phage-pipeline) and run the test. Unfortunately, I have not been able to reproduce the error with a simpler workflow.

pb-cdunn:
```
Snakemake version
the lastest from github: 5.30.2+6.g674a91ca
```
The fix was #727 was not in `5.30.2`, but it's in `5.31.1`. Have you tried the latest version from PyPI? Or the tip of `master` here?

I think a test was added for that other fix (mentioned in your #818). So if you revert that fix and apply your own, you might have a better solution.

jmeppley:
@pb-cdunn Yes, this was found in the development branch for the #727 bugfix (#801). I reported the version wrong, I think.

I actually did try reverting 71b5aeb and applying my fix, but it seems to slow things down. Not surprisingly (in retrospect) assigning all the jobids at once is significantly faster, but somehow a couple jobs slip through in rare cases, so adding a fall back fixed this bug for me.

[...]

PR #818, opened 2020-12-22
Fix issue #817: assign job id if it doesn't exist in DAG.jobid()

jmeppley:
This simple fix solves #817, but it may break the jobid consistency created by 71b5aeb35b0919f906189d5a945a7d7e189ed90e. I suspect undoing that commit would also solve this bug.

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -232,9 +232,16 @@ class DAG:
         )
 
     def update_jobids(self):
+        logger.debug("running DAG.update_jobids()")
         for job in self.jobs:
             if job not in self._jobid:
-                self._jobid[job] = len(self._jobid)
+                self.assign_jobid(job)
+
+    def assign_jobid(self, job):
+        jobid = len(self._jobid)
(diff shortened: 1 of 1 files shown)


johanneskoester:
Thanks a lot for pointing me to this issue! The jobid information during the checkpoint update is not so important, hence I think it is simpler and faster to just omit it. Closing in favor of #839.

epruesse:
This may yet be needed. While I was trying to come up with an MWE for #855 I managed to trigger the key error. If you have two rules using the same checkpoint and directly depending on one another, there is an update/replace_job before the job is created.

jbloom:
I am getting this error too in a complicated workflow where the same checkpoint is being referred to multiple times. I don't really understand the error or all the discussion above, but I'm getting it in 5.32.1 which makes me suspect @epruesse is correct and it's not fixed by @johanneskoester's #839:
```
snakemake/dag.py", line 700, in jobid
    return self._jobid[job]
KeyError
```

epruesse:
@jbloom - Can you come up with an MWE and open a new issue?

PR #819, opened 2020-12-27
formulating wms monitor into logger proper

vsoch:
This is the first round of work to address #799. Specifically, we have the WMS monitor as a proper logger, and don't need to do any customization in the setup_logger function. We are currently using the defined API schema (defined by panoptes) although the project is thinking about changing this to some spec (see https://github.com/panoptes-organization/monitor-schema/pull/7#issuecomment-750897929). The spec in question doesn't have an update workflow endpoint, so I'm not sure how that would work yet, but I think @fgypas perhaps has something in mind.

In addition, we have added authentication, which is optional and enforced by the WMS monitor server. If authentication is required, the user must export `WMS_MONITOR_TOKEN` in their environment, which is provided by the server. We could do more complex interactions like allow a user to specify username:password for basic auth and then have the server issue a temporary token for each request, but looking at the spec that @fgypas has in mind, the [authorization part](https://ga4gh.github.io/workflow-execution-service-schemas/docs/#section/Authorization-and-Authentication) only mentions a url to go to learn about how to authenticate - I don't see any protocol that is standard for these kinds of APIs to return 401, and then the client knows to authenticate, etc. Unless that is supposed to be provided in the ` auth_instructions_url` field? (note that it's typically provided in an Www-Authenticate header.

## Discussion

I think for the create_workflow endpoint that snakemake should be sending more metadata about the run to the server. Right now it's basically sending nothing. Minimally I'd like to suggest we add:

 - snakefile path
 -  working directory 
 - any snakemake identifiers?
 - some kind of name? 

For example, for a user that creates the workflow in the interface (e.g., for snakeface) these fields are derived on that step, so when we issue the command the additional id provided hooks up to all of that. But if we run the workflow on the command line and haven't created it in the interface, a new workflow is created on behalf of the user (and it's hugely missing metadata).

Also, I think this will need to be done in stages. The first stage is keeping things as they are now, but adding the option for the token. The second stage will be the refactor of the API, with whatever spec @fgypas decides. I'm going to need to update snakeface too so I'm hoping to help with whatever I can to move this forward.

cc @johanneskoester @fgypas 

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -2246,6 +2246,19 @@ def get_argument_parser(profile=None):
     return parser
 
 
+def generate_parser_metadata(parser, args):
+    """Given a populated parser, generate the original command along with
+    metadata that can be handed to a logger to use as needed.
+    """
+    command = "snakemake %s" % " ".join(
+        parser._source_to_settings["command_line"][""][1]
+    )
+    workdir = os.getcwd()
+    metadata = args.__dict__
+    metadata.update({"command": command})
+    return metadata
+
+
 def main(argv=None):
     """Main entry point."""
 
@@ -2562,7 +2575,11 @@ def main(argv=None):
             log_handler.append(slack_logger.log_handler)
 
         elif args.wms_monitor or args.log_service == "wms":
-            wms_logger = logging.WMSLogger(args.wms_monitor, args.wms_monitor_arg)
+            # Generate additional metadata for server
+            metadata = generate_parser_metadata(parser, args)
+            wms_logger = logging.WMSLogger(
+                args.wms_monitor, args.wms_monitor_arg, metadata=metadata
+            )
             log_handler.append(wms_logger.log_handler)
 
         if args.edit_notebook:
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -21,7 +21,7 @@ from importlib.machinery import SourceFileLoader
 from snakemake.workflow import Workflow
 from snakemake.dag import Batch
 from snakemake.exceptions import print_exception, WorkflowError
-from snakemake.logging import setup_logger, logger, SlackLogger
+from snakemake.logging import setup_logger, logger, SlackLogger, WMSLogger
 from snakemake.io import load_configfile
 from snakemake.shell import shell
 from snakemake.utils import update_config, available_cpu_count
@@ -123,7 +123,6 @@ def snakemake(
     updated_files=None,
     log_handler=[],
     keep_logger=False,
-    wms_monitor=None,
     max_jobs_per_second=None,
     max_status_checks_per_second=100,
     restart_times=0,
@@ -269,7 +268,6 @@ def snakemake(
                                     whether to clean up conda tarballs after env creation (default None), valid values: "tarballs", "cache"
         singularity_prefix (str):   the directory to which singularity images will be pulled (default None)
         shadow_prefix (str):        prefix for shadow directories. The job-specific shadow directories will be created in $SHADOW_PREFIX/shadow/ (default None)
-        wms-monitor (str):          workflow management system monitor. Send post requests to the specified (server/IP). (default None)
         conda_create_envs_only (bool):    if specified, only builds the conda environments specified for each job, then exits.
         list_conda_envs (bool):     list conda environments and their location on disk.
         mode (snakemake.common.Mode): execution mode
@@ -471,7 +469,6 @@ def snakemake(
             use_threads=use_threads,
             mode=mode,
             show_failed_logs=show_failed_logs,
-            wms_monitor=wms_monitor,
         )
 
     if greediness is None:
@@ -1336,10 +1333,22 @@ def get_argument_parser(profile=None):
         action="store",
         nargs="?",
         help=(
-            "IP and port of workflow management system to monitor the execution of snakemake (e.g. http://127.0.0.1:5000"
+            "IP and port of workflow management system to monitor the execution of snakemake (e.g. http://127.0.0.1:5000)"
+            " Note that if your service requires an authorization token, you must export WMS_MONITOR_TOKEN in the environment."
         ),
     )
-
+    group_exec.add_argument(
+        "--wms-monitor-arg",
+        nargs="*",
+        metavar="NAME=VALUE",
+        help=(
+            "If the workflow management service accepts extra arguments, provide."
+            " them in key value pairs with --wms-monitor-arg. For example, to run"
+            " an existing workflow using a wms monitor, you can provide the pair "
+            " id=12345 and the arguments will be provided to the endpoint to "
+            " first interact with the workflow"
+        ),
+    )
     group_exec.add_argument(
         "--scheduler-ilp-solver",
         default=recommended_lp_solver,
@@ -1911,10 +1920,10 @@ def get_argument_parser(profile=None):
     group_behavior.add_argument(
         "--log-service",
         default=None,
-        choices=["none", "slack"],
+        choices=["none", "slack", "wms"],
         help="Set a specific messaging service for logging output."
         "Snakemake will notify the service on errors and completed execution."
-        "Currently only slack is supported.",
+        "Currently slack and workflow management system (wms) are supported.",
     )
 
     group_cluster = parser.add_argument_group("CLUSTER")
@@ -2552,6 +2561,10 @@ def main(argv=None):
             slack_logger = logging.SlackLogger()
             log_handler.append(slack_logger.log_handler)
 
+        elif args.wms_monitor or args.log_service == "wms":
+            wms_logger = logging.WMSLogger(args.wms_monitor, args.wms_monitor_arg)
+            log_handler.append(wms_logger.log_handler)
+
         if args.edit_notebook:
             from snakemake import notebook
 
@@ -2679,7 +2692,6 @@ def main(argv=None):
             cluster_status=args.cluster_status,
             export_cwl=args.export_cwl,
             show_failed_logs=args.show_failed_logs,
-            wms_monitor=args.wms_monitor,
             keep_incomplete=args.keep_incomplete,
             keep_metadata=not args.drop_metadata,
             edit_notebook=args.edit_notebook,
--- a/snakemake/logging.py
+++ b/snakemake/logging.py
@@ -10,7 +10,6 @@ import datetime
 import sys
 import os
 import json
-import requests
 import threading
 import tempfile
 from functools import partial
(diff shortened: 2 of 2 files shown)


vsoch:
Another quick question - why does the msg field here have repr? That means that if it's a dictionary, for example, it gets sent as a string representation of the dict (and not a dict as it should):

```
server_info = {
        "msg": repr(msg),
        "timestamp": time.asctime(),
        "id": self.server["id"],
    }
```
Unless there is something I'm missing, I would do:

```
server_info = {
    "msg": msg,
    "timestamp": time.asctime(),
    "id": self.server["id"],
}
```

fgypas:
@vsoch Thanks a lot. I'll test it with panoptes and come back to you.

fgypas:
@vsoch This works fine with panoptes. 

I am not sure it will work out of the box if you remove the `repr`. I remember we had some issue with serialization/deserialization, but I think a simultaneous panoptes pull request would fix this.

vsoch:
It looks like you have to use eval on the message to get it back to it's original state: https://github.com/panoptes-organization/panoptes/blob/b5add3e98e7dbfb8bc64e51b2fb3d24a0fc6deca/panoptes/server_utilities/db_queries.py#L14, which I think we could possibly avoid. Could we perhaps test removing it, or having a try / except for the error that you were trying to solve? Or what about just dumping it to a string that could be loaded instead of evaluated? Let me know what you think, and how I can help!

vsoch:
And @fgypas I have a quick question! If you start a workflow from the web interface, how do you handle environment.yamls that might be alongside it? (e.g., if they aren't installed and activated it won't work).

fgypas:
> And @fgypas I have a quick question! If you start a workflow from the web interface, how do you handle environment.yamls that might be alongside it? (e.g., if they aren't installed and activated it won't work).

Do you mean yamls that are required to install the dependencies (e.g. snakemake) or yamls that are required within a snakemake pipeline and specified with the `conda` argument?

vsoch:
This step 👉 https://snakemake.readthedocs.io/en/stable/tutorial/setup.html#step-3-creating-an-environment-with-the-required-software

vsoch:
Here is the issue with using repr - we get strings that look like this:

```
{'level': 'dag_debug', 'status': 'candidate', 'job': all}
{'level': 'dag_debug', 'status': 'candidate', 'job': plot_quals}
{'level': 'dag_debug', 'status': 'candidate', 'job': bcftools_call}
{'level': 'dag_debug', 'status': 'candidate', 'job': samtools_sort}
{'level': 'dag_debug', 'status': 'candidate', 'job': bwa_map}
{'level': 'dag_debug', 'status': 'selected', 'job': bwa_map}
```
which we can't just easily evaluate, especially when there are python representations of objects that are undefined or custom functions. In this case:

```python
{'level': 'dag_debug', 'status': 'candidate', 'job': all}
{'level': 'dag_debug', 'status': 'candidate', 'job': <built-in function all>}
{'level': 'dag_debug', 'status': 'candidate', 'job': plot_quals}
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-18-f903b09c2e84> in <module>
      2     if isinstance(ws.msg, str):
      3         print(ws.msg)
----> 4         print(eval(ws.msg))
      5 
      6 

<string> in <module>

NameError: name 'plot_quals' is not defined
```
So I think we would be best to do parsing on the level of snakemake that handles these objects however is appropriate to send to the update endpoint as a dictionary proper, so there is no `repr()` and no `eval()` needed. Without that we will wind  up with cases of strings that cannot be evaluated and it should be a consistent strategy. Let me know your thoughts and what you'd like to try.

vsoch:
I've just pushed changes so we more properly parse the message into something that is json serializable instead of using repr() and then eval(). But actually, using eval would still work, as an example, here is an eval on a json dump of a dictionary (what we are doing here)

```python
eval(json.dumps({'one': 'one'}))
```
The difference is that now on the server I can do:

```python
# The message should be json dump of attributes
message = json.loads(request.POST.get("msg", {}))
```
Which I feel much better about than doing eval. I'm going to see if I can add some metadata to the request next.

fgypas:
> This step 👉 https://snakemake.readthedocs.io/en/stable/tutorial/setup.html#step-3-creating-an-environment-with-the-required-software

So, for this, I think there are different ways to handle depending on the repo/pipeline that you want to execute and the infrastructure you have. Maybe @johanneskoester or someone else here has a better idea than me, but my opinion is that if you have a specific structure of the repo that you want to execute then things become simpler. My personal preference is to have only snakemake as a dependency and then put all other dependencies within the Snakefile as conda or singularity. The way I can think of it visually is to manage the virtual environments with something similar to the anaconda gui manager. 



As far as I know there is no golden standard of the structure of a snakemake pipeline, but as a rule of thumb you always have a Snakefile and a config.

fgypas:
> I've just pushed changes so we more properly parse the message into something that is json serializable instead of using repr() and then eval(). But actually, using eval would still work, as an example, here is an eval on a json dump of a dictionary (what we are doing here)
> 
> ```python
> eval(json.dumps({'one': 'one'}))
> ```
> 
> The difference is that now on the server I can do:
> 
> ```python
> # The message should be json dump of attributes
> message = json.loads(request.POST.get("msg", {}))
> ```
> 
> Which I feel much better about than doing eval. I'm going to see if I can add some metadata to the request next.

Thanks a lot for testing. I agree that we should avoid using `repr`/`eval`. If you have time, can you please try to open a PR on panoptes? Anyway, we don't use (at least for now) the `dag_debug` that seems to give rise to the problem.

vsoch:
> So, for this, I think there are different ways to handle depending on the repo/pipeline that you want to execute and the infrastructure you have. Maybe @johanneskoester or someone else here has a better idea than me, but my opinion is that if you have a specific structure of the repo that you want to execute then things become simpler. My personal preference is to have only snakemake as a dependency and then put all other dependencies within the Snakefile as conda or singularity. The way I can think of it visually is to manage the virtual environments with something similar to the anaconda gui manager.

I'm following the tutorial (which has this step) trying to reproduce using the interface, and it hits the error that we haven't installed some dependencies in the environment. I think if it were specified under conda then snakemake would install and source on our behalf so it wouldn't be an issue. I might just modify the example to do that.

And another update! I just pushed changes so that the original create_workflow endpoint sends along minimal metadata (snakefile, working directory, and command). This way, the user has more information than a workflow id and the output in the interface.

> Thanks a lot for testing. I agree that we should avoid using repr/eval. If you have time, can you please try to open a PR on panoptes? Anyway, we don't use (at least for now) the dag_debug that seems to give rise to the problem.

Yes definitely! With this change, we should be able to do the same json.loads() instead of eval.

fgypas:
> I'm following the tutorial (which has this step) trying to reproduce using the interface, and it hits the error that we haven't installed some dependencies in the environment. I think if it were specified under conda then snakemake would install and source on our behalf so it wouldn't be an issue. I might just modify the example to do that.

Ah ok, this requires the installation of `Flask`. Since most users don't use the standard web interface of snakemake, I think it makes sense that it is not installed by default.

fgypas:
> And another update! I just pushed changes so that the original create_workflow endpoint sends along minimal metadata (snakefile, working directory, and command). This way, the user has more information than a workflow id and the output in the interface.

I think this is very helpful. Do you send the path of the Snakefile, or the content of it? I am mainly asking, as in many cases the Snakefile is split into multiple files (`.smk`) that are included in the main Snakefile.

vsoch:
>  Do you send the path of the Snakefile, or the content of it? I am mainly asking, as in many cases the Snakefile is split into multiple files (.smk) that are included in the main Snakefile.

We just send the path of the snakefile, as Snakemake sees it.

vsoch:
Another suggestion - if we forsee needing to launch workflows in some way that doesn't afford access to an environment with a token (or we just don't want to use the environment, maybe for some reason) we can allow the token to be defined on the command line too:

```bash
snakmake --cores 1 --wms-monitor http://127.0.0.1:5000 --wms-monitor-arg token=12345
```
That would be read from the args like this:

```python
 # A token is suggested but not required, depends on server
self.token = os.getenv("WMS_MONITOR_TOKEN") or args.get('token')
```

vsoch:
Update for @fgypas and @johanneskoester - this API with the token is currently working for Snakeface (https://github.com/snakemake/snakeface). @johanneskoester I can make this public if it's okay with you, I have a few quick questions:

1. Can we make it public? Once I finish the core features (likely for having notebook + different executors) I'll add deployment types per user request (e.g., starting with docker-compose with a more substantial task runner and then adding the different auth plugins for saml, ldap, pam, etc.)
2.  - can I use the Snakemake cloud account to test setting up an executor for Snakeface? I'll use the dummy example workflow so the costs should be minimal. 
3. Given 1, can we also deploy the docs to readthedocs? They should be ready to go. Let me know if there is any issue because we need to install Django and run `django.setup()` to be able to import the modules. It should work given it's a dependency of snakeface but you never know!

Our break is over January 4th (Monday) so I'm pretty happy to get the skeleton version (a snakemake notebook) mostly finished. It would be great to get feedback about changes / direction before I implement more complex deployment types. But of course no rush, once I go back I won't have as much time for fun projects :P 

Happy New Year! I should umm, attempt to do something that isn't programming :laughing:

johanneskoester (approved):

johanneskoester:
Great work, @vsoch!


>     1. Can we make it public? Once I finish the core features (likely for having notebook + different executors) I'll add deployment types per user request (e.g., starting with docker-compose with a more substantial task runner and then adding the different auth plugins for saml, ldap, pam, etc.)

Yes :-)!

> 
>     2. Can I use the Snakemake cloud account to test setting up an executor for Snakeface? I'll use the dummy example workflow so the costs should be minimal.

Of course :-)!

> 
>     3. Given 1, can we also deploy the docs to readthedocs? They should be ready to go. Let me know if there is any issue because we need to install Django and run `django.setup()` to be able to import the modules. It should work given it's a dependency of snakeface but you never know!

Yes, let me know (best via mail) if you need any help!

johanneskoester:
Great work, @vsoch!


>     1. Can we make it public? Once I finish the core features (likely for having notebook + different executors) I'll add deployment types per user request (e.g., starting with docker-compose with a more substantial task runner and then adding the different auth plugins for saml, ldap, pam, etc.)

Yes :-)!

> 
>     2. Can I use the Snakemake cloud account to test setting up an executor for Snakeface? I'll use the dummy example workflow so the costs should be minimal.

Of course :-)!

> 
>     3. Given 1, can we also deploy the docs to readthedocs? They should be ready to go. Let me know if there is any issue because we need to install Django and run `django.setup()` to be able to import the modules. It should work given it's a dependency of snakeface but you never know!

Yes, let me know (best via mail) if you need any help!

PR #820, opened 2020-12-28
fix #821: enable --no-subworkflows command when using --cluster

### LINKED ISSUE #821: `--no-subworkflows` command does nothing when submitting using `--cluster`
**Snakemake version**
5.28.0

**Describe the bug**
If I call `snakemake --nosw -n`, I see that the DAG is built without sub-workflows, as expected.
If I call using `--cluster` and then look at the generated logs for each job, I see that the subworkflow _does_ get evaluated as part of the DAG. 
This appears to be caused by the `--cluster` command effectively resulting in a limited re-evaluation of the workflow for each job.

goi42:
Changes the `executesubworkflows` argument in `Workflow.execute` to the `execute_subworkflows` attribute of `Workflow`, enabling `get_additional_args` to add the `--no-subworkflows` command to the cluster submission script when appropriate.

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -587,6 +587,7 @@ def snakemake(
             envvars=envvars,
             max_inventory_wait_time=max_inventory_wait_time,
             conda_not_block_search_path_envvars=conda_not_block_search_path_envvars,
+            execute_subworkflows=execute_subworkflows,
         )
         success = True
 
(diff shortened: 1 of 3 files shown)


goi42:
Resolves #821.

johanneskoester (approved):

johanneskoester:
Thanks!

PR #822, opened 2020-12-28
ensure that singularity with GLS exits with a clear error message

vsoch:
GLS does not support Singularity, but we currently issue a warning and continue. We should be raising a workflow exception and giving a clear message to the user that Singularity is not supported. See https://github.com/snakemake/snakemake/pulls#issuecomment-751803456.


johanneskoester (approved):

johanneskoester:
Thanks!

PR #826, opened 2021-01-06
Add recursive file discovery for tests

HenningTimm:
This PR adds recursive discovery for files in the expected-results folders allowing to validate files placed within directory structures.

Some tests did not work properly, since the expected_results folder was not evaluated recursively. This PR adds recursive comparison of files in said folder and fixes some (potential) errors in affected tests.
As far as I can tell, this bug only masked one actual error in the code (affecting the `instance` method of the recently added `Paramspace` helper object), which is also fixed by this PR. The remaining (newly) failing tests were due to minor errors in tests.

## Reasoning
I noticed during testing paramspaces that some files were not tested properly. Properly means, in this case, that tests passed even if required files were not present and those that were present could have the wrong content.

This was due to the fact that only files in the `expected_results` subfolder were considered. This folder, however, was not recursively explored, leading to files in subdirectories being ignored.

I fixed this by employing a glob pattern instead of `os.listdir(results_dir)`.
As a result of this, a total of six tests now failed (testing locally) due to missing files or files with the wrong content. These were:

- test_directory
- test_pathlib
- test_checkpoints_dir
- test_output_file_cache
- test_metadata_migration
- test_paramspace

Since I am not familiar with all features that are touched in these tests, I mentioned the people who originally wrote them for help. Sorry for the inconvenience :(

### test_directory
This tests does not create the file `some/dir-child` which is contained in the expected results folder. This is due to the case, that the rule "not_child_to_other" is not speciefied as a target in `tests/tests.py::test_directory`.

If I read correctly, this rule should be run and complete without error. Therefore I added it to the cases which should complete. I hope I interpreted this correctly and would appreciate feedback from @rasmusagren who last touched this test case.

### test_pathlib ✔️
This tests does not create two files that are expected: `outdir/log_dynamic` and `outdir/log_other_rule_output`. There is no rule that could create `log_other_rule_output`, however I guess this file is only missing the wildcard name and should actually be `log_other_rule_output_existing`. I renamed the expected file accordingly.
   
For `outdir/log_dynamic` I identified `log_temp` as the potentially intended target. I renamed `log_dynamic` to `log_temp`, changed its content accordingly, and added the target rule all back in.
Maybe @boulund  can confirm that this was the right cause of action?

### test_checkpoints_dir

`aggregated/a.txt` and `aggregated/b.txt` are both expected to contain the lines

```
2
1
3
```
The generated files contain them in the order
```
1
3
2
```
This happens consistently and is caused by the order in which `snakemake/io.py::glob_wildcards` returns the wildcards. Since the order of input files should not be relevant (please correct me if this is not true) I changed the expected content of the files to fit the test.

Since this test was originally written by @johanneskoester I would relay the final decision to him.

**Update:** This generates different results for different build environments and different runs. The [first run](https://github.com/snakemake/snakemake/runs/1656316151#step:11:337) created the initially expected result on the linux build (`2 1 3`), but `1 2 3` [on windows](https://github.com/snakemake/snakemake/runs/1656316126#step:6:290), while the [next one](https://github.com/snakemake/snakemake/runs/1656410161#step:11:337) created another order (`2 3 1`) and again `1 2 3` [on windows](https://github.com/snakemake/snakemake/runs/1656410236#step:6:290).

This requires a different test setup.

### test_output_file_cache
*Caveat:* I am not familiar with this feature (I am surely going to use this from now on) of snakemake and hope that I understood the test correctly. I added a docstring to the file to clarify what I think this test does.

For this test, different files than expected were generated using the caching directive. This test shares output files with the similar (skipped) test for `test_output_file_cache_remote` however, the expected results for this test were not upgraded when the
feature `multiext` was added.
The `multiext` helper results in an additional file being cached and changed the signature of the rule and thereby its hash.

I checked which files should be generated and replaced the deprecated ones in the `expected-results/cache` folder.

Additionally I had trouble getting this test to work locally. I had to make the `cache` folder set in the `os.environ["SNAKEMAKE_OUTPUT_CACHE"]` variable a relative path for it to work. This might not work using Github Actions. If that is the case, I will revert this change.

@johanneskoester could you verify that this test now works as intended?

### test_metadata_migration
This test expects the file `path.txt` with the very long file name to contain the string `"updated"`, while the generated file is empty.
Here I am at a loss, what exactly is supposed to happen. The test documentation (test.py line 1119f) reads like the test case is expected to fail, however the corresponding flag is not set.

Depending on the feedback from @johanneskoester I will add this flag, removed the content of the expected result file, or look further into why this happens.

### test_paramspace
This one's mine. I deliberately added an error in a path to check if my tests would fail through that. Turns out, I wouldn't have needed to, since I unintentionally added another typo in the `full_reorder` path. Said error(s) also served as TDD test cases for this patch. I fixed all typos in the filenames to resolve this.

Additionally, this revealed the only actual error in the code. The output of the `instance` method was always an empty dictionary due to a missing `value` in the code. This error has gone unnoticed, since files were not evaluated recursively. I fixed this bug and changed the content of all expected .tsv files to reflect the new output pattern including types.

## Conclusion
Testing locally, the tests mentioned above were the only ones failing. I hope not too many additional one are affected, once this runs on GitHub Actions.


boulund:
The proposed changes to the pathlib-related test look ok to me.

HenningTimm:
Thank you for taking the time to review this @boulund

HenningTimm:
Concerning the `test_checkpoints_dir` test I ended up sorting the generated input files, so that they appear in the same order on all systems. I hope this does not undermine the intention of this test.

johanneskoester (approved):
LGTM thanks a lot, great work! Shall I help with the final failing test, or do you already have a clue?

HenningTimm:
Unfortunately I do not quite have a handle on that one yet. I would appreciate some help as I am not yet familiar with the metadata folder/ files. I am not sure if I understood the error with long paths correctly, so I am not certain, what this tests actually tests for. From what I have read I gather the test does the following:

1. Generate an empty file `path.txt` at `outpath`
2. (Incomplete) Metadata are stored in the accompanying snakemake folder
3. An expected file containing the string `"updated"` is created an `exppath`
4. The workflow is run. It should find a set of metadata (on old v1 format, which is automatically migrated) for the rule flagging it as incomplete, thus flagging the rule as incomplete and triggering a rerun. This rerun should write `"updated"` into `path.txt`. However, this doesn't happen (see [build log](https://github.com/snakemake/snakemake/runs/1670866804#step:11:430))

When I compare both versions of the metadata file (in `expected-results` in the repo and in `/tmp`) locally, they are identical before and after the migration. I have not found the time to look into that further yet.

johanneskoester:
> Unfortunately I do not quite have a handle on that one yet. I would appreciate some help as I am not yet familiar with the metadata folder/ files. I am not sure if I understood the error with long paths correctly, so I am not certain, what this tests actually tests for. From what I have read I gather the test does the following:
> 
>     1. Generate an empty file `path.txt` at `outpath`
> 
>     2. (Incomplete) Metadata are stored in the accompanying snakemake folder
> 
>     3. An expected file containing the string `"updated"` is created an `exppath`
> 
>     4. The workflow is run. It should find a set of metadata (on old v1 format, which is automatically migrated) for the rule flagging it as incomplete, thus flagging the rule as incomplete and triggering a rerun. This rerun should write `"updated"` into `path.txt`. However, this doesn't happen (see [build log](https://github.com/snakemake/snakemake/runs/1670866804#step:11:430))
> 
> 
> When I compare both versions of the metadata file (in `expected-results` in the repo and in `/tmp`) locally, they are identical before and after the migration. I have not found the time to look into that further yet.

ok, looking at it now, thanks!

johanneskoester:
Should be fixed now.

PR #835, opened 2021-01-12
Implement AutoRemoteProvider to automatically select Providers for remote files

### LINKED ISSUE #260: Add method to automatically choose an appropriate `RemoteProvider`
When specifying remote input files, one needs to pay attention to the URL scheme (e.g. `http` vs `ftp`) and select an appropriate `RemoteProvider`.

It would be nice to have a wrapper which automatically chooses the correct provider.
An (incomplete) implementation could look something like this:
```python
from urllib.parse import urlparse

from snakemake.remote.HTTP import RemoteProvider as HTTPRemoteProvider
from snakemake.remote.FTP import RemoteProvider as FTPRemoteProvider

HTTP = HTTPRemoteProvider()
FTP = FTPRemoteProvider()

def auto_remote(url):
    o = urlparse(url)
    if o.scheme in ('http', 'https'):
        return HTTP.remote(url)
    elif o.scheme == 'ftp':
        return FTP.remote(url)
    else:
        raise RuntimeError(f'Invalid url: "{url}"')
```

At the moment I manually implement this method in my `Snakemake` files.

kpj:
This PR adds the class `AutoRemoteProvider` which allows you to transform
```python
from snakemake.remote.FTP import RemoteProvider as FTPRemoteProvider
from snakemake.remote.HTTP import RemoteProvider as HTTPRemoteProvider
FTP = FTPRemoteProvider()
HTTP = HTTPRemoteProvider()

rule manual_provider:
    input:
        fname_ftp=FTP.remote('ftp://ftp.ncbi.nlm.nih.gov/pub/taxonomy/taxcat.tar.gz'),
        fname_http=HTTP.remote('https://github.com/hetio/hetionet/raw/master/hetnet/tsv/hetionet-v1.0-nodes.tsv')
    output: touch('foo')
```
into
```python
from snakemake.remote import AUTO

rule auto_provider:
    input:
        fname_ftp=AUTO.remote('ftp://ftp.ncbi.nlm.nih.gov/pub/taxonomy/taxcat.tar.gz'),
        fname_http=AUTO.remote('https://github.com/hetio/hetionet/raw/master/hetnet/tsv/hetionet-v1.0-nodes.tsv')
    output: touch('bar')
```

Instead of having to manually choose which provider to use, `AutoRemoteProvider` automatically detects it from the URL's scheme. Fixes #260.

Some open questions remain, e.g., what the best way of automatically retrieving all available providers is.


snakemake/remote/__init__.py
```diff
@@ -294,3 +295,46 @@ def local_path(self):
...
+
+
+class AutoRemoteProvider:
+    def __init__(self, *args, **kwargs):
+        self.constructor_args = args
+        self.constructor_kwargs = kwargs
+
+    @property
+    def protocol_mapping(self):
+        # TODO: how to include all providers automatically (AbstractRemoteProvider.__subclasses__)?
```
johanneskoester, line 307: Sounds reasonable, can you try that out? We might need to add a static field to each remote provider that tells whether it is eligible for being automated like this (some might require special args).

johanneskoester (changes requested):
Good idea!

snakemake/remote/__init__.py
kpj: I implemented one possible, though maybe not robust, way of gathering all implemented RemoteProviders.
We could add a new argument to `AutoRemoteProvider.remote` which allows passing special args to each instance.

johanneskoester (changes requested):
LGTM. Can you also add a section to the remote provider docs?

kpj:
Added a brief section with example code to the docs!

johanneskoester (approved):

PR #841, opened 2021-01-18
Remove f-strings

mbhall88:
### Fixed

- f-strings in runtime code were replaced with `format` syntax as we still support python 3.5

There are also f-strings in the following files but I didn't touch them as I think they are in files that will only be executed in dev environments.

```
tests/common.py: line 198
docs/snakefiles/rules.rst: lines 1374, 1385, 1387
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+Unreleased
+==========
+Fixed
+-----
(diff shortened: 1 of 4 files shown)


johanneskoester (approved):

johanneskoester:
Thank you!

PR #842, opened 2021-01-18
Add rule name to warning about wildcard constraints in inputs

jheuel:
This change makes it easier to find and fix the offending rules:

> Wildcard constraints in inputs are ignored. 

is changed to

> Wildcard constraints in inputs are ignored. (rule: my_rule_name)


johanneskoester (approved):

johanneskoester:
Thank you!

PR #852, opened 2021-01-25
Add rule dependency for local container image files

### LINKED ISSUE #848: Conda environments and container image files as rule dependencies
The singularity image files which are used in my workflow are created as part of the workflow. The same applies to the conda environment files. Image and environment files are outputs of a rule.

In both cases rule dependencies are not created.

It would be really helpful if both conda environment files and (local) singularity image files would be treated similar to rule inputs, i.e. by establishing rule dependencies.

Files can be added as inputs to other rules but this is a hack.

I also tried to add the dependency e.g. in the [`@container_img.setter`](https://github.com/snakemake/snakemake/blob/master/snakemake/rules.py#L338-L340) with `self.dependencies[container_img] = container_img.rule` but this does not establish the dependency.

Setting it as an input item (`self._set_inoutput_item(container_img, name="__container_img__")`) establishes the rule dependency but also adds another input to the rule which is not desired.

**Additional context**
I would try to implement this feature but need more guidance on how to do it.

bilke:
Closes #848 by implementing this for local Singularity image files.

For Conda environments for me it already seems that dependencies are setup properly.

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -13,6 +13,7 @@ from collections import defaultdict, Counter, deque, namedtuple
 from itertools import chain, filterfalse, groupby
 from functools import partial
 from pathlib import Path
+from urllib.parse import urlparse
 import uuid
 import math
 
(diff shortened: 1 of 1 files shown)


bilke:
Same CI error as on [master](https://github.com/snakemake/snakemake/runs/1758507304?check_suite_focus=true).

johanneskoester:
I don't think a container image should be an input file to a rule. We also don't do that for conda environments. Input files should be data. Code or software environment changes should not automatically trigger reruns, but rather allow for optional reruns, as for example you can do it with `snakemake -R $(snakemake --list-code-changes)` etc. That is, of course it is a good idea to add an additonal flag like `snakemake --list-software-stack-changes` that would list all output files for which either the container image or the conda environment has changed.

PR #856, opened 2021-01-28
Use temp file for long scripts (fixes #854)

### LINKED ISSUE #854: Maximum shell script length not checked or handled
**Snakemake version**

5.32.0

**Describe the bug**

- Long `shell:` scripts (after wildcard expansion) cause Snakemake to fail
- This can easily happen due to long file names and many files
- Snakemake falsely claims `one of the commands exited with non-zero exit code;`

**Minimal example**

```
arg_max = int(subprocess.check_output("getconf ARG_MAX", shell=True))
string = "x" * arg_max
rule all:
  shell: "echo {string} >/dev/null"
```

output:

```
[command printed, many lines]
 (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)

Exiting because a job execution failed. Look above for error message.
```

**Desired Behavior**

- Present correct error message if `Popen()` in `shell.py` raises unexpected exceptions.
- Pass the script into the shell command via STDIN or as file, at least if `command` is longer than `getconf ARG_MAX`
- (Also, ellipsize the command executed before printing it as error.)

**Background**

There is a limit to the size of arguments that may be passed to a command. It can be retrieved with `getconf ARG_MAX`. Typical values are currently 2MB (Linux) or 256kB (OSX). On Linux, it can be increased with `ulimit -s` as it is dependent on the size of the stack.

Bash scripts written "safely" avoid any `command $LONG_ARGUMENT_LIST`, and instead use pipes when the argument list is unbounded, e.g. with constructs such as `echo $LONG_ARGUMENT_LIST | xargs command`. The latter works safely because `echo` is a bash "builtin", so no external command is run and the ARG_MAX limit is not relevant.

In Snakemake, these techniques will not work as the whole "script" is passed as argument. The workaround using `run:` instead of `shell:` are cumbersome.

Is there any reason why Snakemake couldn't *always* use a temporary file to run `shell:` commands? There are security implications as well. Secrets in any command will be visible on `ps` and that may not be expected. The performance impact should be in the milliseconds per job executed, so not really relevant.

epruesse:
This has `shell(cmd)` create a temporary file if the length of `cmd` with all modifications is too long to be passed as an argument on the command line.

Note that this maximum is NOT the `getconf ARG_MAX` limit that can be changed with `ulimit -s`, but rather a hardcoded define in Linux that allows 32 pages, i.e. typically 128kb.

On Windows and on OSX, the limits are different. Empirically, 251kb is the maximum on OSX. Since it's difficult to determine the exact maximum length, this patch sets the maximum as 128k-1 characters.

Conda and modulefiles environments are passed into the script. Singularity is wrapped after the shell command is replaced with a temporary script file.

Fixes #854

**Notes**

- `snakemake.utils.argvquote` is possibly unnecessary. `sp.Popen` will do this itself, **IF** `args` is a list. 
- The patch will execute `bash -c 'bash .snakemake/tmpscriptXXXX/script.sh'` if singularity is not enabled, to allow for `bash -c 'singularity .... "bash .snakemake/tmpscriptXXX/script.sh"` without rewriting half the module.
- Windows is a pain.


snakemake/deployment/conda.py
```diff
@@ -463,8 +463,6 @@ def bin_path(self):
         return os.path.join(self.prefix_path(), "bin")
 
     def shellcmd(self, env_path, cmd):
-        from snakemake.shell import shell
-
```
epruesse, line 467: dead code

snakemake/shell.py
```diff
@@ -17,7 +19,6 @@
 from snakemake.deployment import singularity
 from snakemake.deployment.conda import Conda
 from snakemake.exceptions import WorkflowError
-import snakemake
```
epruesse, line 20: dead code

snakemake/shell.py
```diff
@@ -117,7 +126,6 @@ def __new__(
         if not context.get("is_shell"):
             logger.shellcmd(cmd)
 
-        env_prefix = ""
```
epruesse, line 120: dead code

snakemake/shell.py
```diff
@@ -134,6 +142,15 @@ def __new__(
         if conda_env:
             cmd = Conda(container_img).shellcmd(conda_env, cmd)
 
+        tmpdir = None
+        if len(cmd.replace("'", r"'\''")) + 2 > MAX_ARG_LEN:
```
epruesse, line 146: Awkward,  but no good alternative without changing `singularity.shellcmd` synopsis. Should get a useful actual length for the script as an argument.

snakemake/shell.py
```diff
@@ -29,6 +30,14 @@
     STDOUT = None
 
 
+# There is a max length for a command executed as well as a maximum
+# length for each argument passed to a command. The latter impacts us
+# especially when doing `sh -c 'long script from user'`. On Linux, it's
+# hardcoded in the kernel as 32 pages, or 128kB. On OSX it appears to be
+# close to `getconf ARG_MAX`, about 253kb.
+MAX_ARG_LEN = 16 * 4096 - 1
```
epruesse, line 38: Using 16 * 4096 instead of 32*4096 to be conservative. A script that's 64kb long we may as well pass as a file. Personally, somewhere at 1kb I'd say please don't spam this to the log...

epruesse:
@johanneskoester `test_cwl` appears broken currently it failed in other PRs as well.

epruesse:
@johanneskoester ready for review

johanneskoester (changes requested):
Thanks a lot, good idea! 
You can remove the cwl test skips, I've fixed them in the master.

johanneskoester (approved):

PR #893, opened 2021-02-26
Add environment variables for cluster status checks (fixes #891)

### LINKED ISSUE #891: Pass --cluster-status by environment
Especially new users to not know that --cluster-status may be required on an hpc system to correctly track job states. It would be nice t have an option to set this as site administrator.

Maybe it would be possible to pass SNAKEMAKE_CLUSTER_STATUS as environment to set this parameter

Any other solution to set the parameter as site administator is good.

**Additional context**
Most Schedulers need a --cluster-status script to correctly handle jobs states and it would be nice to have a system option to pass this script.

LukeLR:
The `--conda-prefix` argument already has a [nice way](https://github.com/snakemake/snakemake/blob/bd5a9239b2684b57ab0143bb709e77afa1493cdc/snakemake/__init__.py#L2177) of fetching the corresponding environment variable for a command line argument. I've implemented the same functionality for the `--cluster-status` and `--max-status-checks-per-second` arguments, using the environment variables `SNAKEMAKE_CLUSTER_STATUS` and `SNAKEMAKE_MAX_STATUS_CHECKS_PER_SECOND`. This allows HPC admins to provide a default status script that is compatible with their cluster setup, so users don't have to take care of that. The default value of 10 for `--max-status-checks-per-second` is kept when the environment variable is unset.

I've added documentation for this within the help string and the corresponding `additional_features.rst` doc file.

This fixes #891.

### DIFF
--- a/docs/executing/cli.rst
+++ b/docs/executing/cli.rst
@@ -120,6 +120,7 @@ Under https://github.com/snakemake-profiles/doc, you can find publicly available
 Feel free to contribute your own.
 
 The profile folder can additionally contain auxilliary files, e.g., jobscripts, or any kind of wrappers.
+It can be supplied via the environment variable ``SNAKEMAKE_PROFILE`` as well.
 See https://github.com/snakemake-profiles/doc for examples.
 
 
--- a/docs/tutorial/additional_features.rst
+++ b/docs/tutorial/additional_features.rst
@@ -264,7 +264,7 @@ The following (simplified) script detects the job status on a given SLURM cluste
     else:
       print("failed")
 
-To use this script call snakemake similar to below, where ``status.py`` is the script above.
+To use this script call snakemake similar to below, where ``status.py`` is the script above, or set the environment variable ``SNAKEMAKE_CLUSTER_STATUS``.
 
 .. code:: console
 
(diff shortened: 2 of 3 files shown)


johanneskoester:
Mhm, HPC admins should actually provide an entire profile, which also can set these values. Wouldn't that work as well?

johanneskoester:
I have further thought about this, and I think there is a generic solution to this:
We already use https://pypi.org/project/ConfigArgParse, for the profiles. It further allows to specify environment variable names for arguments via the `env_var` keyword. Can you modify this PR to use that keyword instead, also for the other places where environment variables are used in the arg parser (e.g. the --conda-prefix flag)?

johanneskoester (changes requested):
I have further thought about this, and I think there is a generic solution to this:
We already use https://pypi.org/project/ConfigArgParse, for the profiles. It further allows to specify environment variable names for arguments via the `env_var` keyword. Can you modify this PR to use that keyword instead, also for the other places where environment variables are used in the arg parser (e.g. the --conda-prefix flag)?

LukeLR:
Thanks for getting back to this! I’ll look into this as soon as I can, shouldn’t be too hard, I guess.

LukeLR:
Hi, I got back to your comment regarding the profile. I created a profile for our HPC cluster now. Similar to my previous approach with the cluster status script and the max status checks per second, I want to set the profile using an environment variable for all users, so the individual users don't have to do `--profile /full/path/to/our/profile/on/the/central/storage/` but can simply invoke `snakemake` instead. I therefore added an environment variable `SNAKEMAKE_PROFILE`. What do you think?

LukeLR:
Closed, as this appears to be implemented in v6.5.0

PR #898, opened 2021-03-02
Refactor f-strings

mbhall88:
Refactored some f-strings to restore python 3.5 compatibility.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,6 @@
+unreleased
+==========
+- Restored Python 3.5 compatibility by removing f-strings (@mbhall88)
 
 [6.0.1] - 2021-02-26
 ====================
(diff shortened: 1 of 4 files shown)


mbhall88:
I guess that given python 3.5 is now deprecated we might want to think about when we stop support (if we do)?

johanneskoester (approved):

johanneskoester:
Thank you!

johanneskoester:
We should probably also run the test suite with python 3.5. At least weekly or so.

mbhall88:
> We should probably also run the test suite with python 3.5. At least weekly or so.

Looks like they still have it available in Actions, so should be pretty easy to add https://github.com/actions/python-versions/blob/17b1cd02cabff1d89e008c8356ecf41eb6f388de/versions-manifest.json#L1633

PR #924, opened 2021-03-19
Pass wait-file list in file instead of commandline parameter list

### LINKED ISSUE #918: "Argument list too long" crash with long rule input file list
**Snakemake version**
`v6.0.5`

**Describe the bug**
I have a rule like this:
```python
rule aggregate_results:
    input:
        fname_list = expand('results/data_{accession}.csv.gz', accession=accession_list),
    output:
        fname = 'results/output.csv.gz',
    run:
        print('do work here')
```

which is executed on the cluster with `snakemake --profile lsf -pr --use-conda`.
The variable `accession_list` stores `>100000` samples.

The job then crashes with:
```bash
../.snakemake/tmp.27se5e40/snakejob.aggregate_results.631976.sh: line 4: ../python/3.8.5/x86_64/bin/python3.8: Argument list too long
```

This might be somewhat related to #854 but was apparently not fixed by #856.

When looking at the jobscript, it seems like the issue is that all input files are passed as an argument for the `--wait-for-files` parameter. Maybe one could follow a similar approach as in #854 to fix this.

kpj:
When creating a jobscript, wait-files are passed using the `--wait-for-files` parameter.
This can lead to the error "Argument list too long" when there are many files.

This PR introduces a new commandline parameter `--wait-for-files-file` (name is up for debate) which allows to pass a filename which contains one file-to-be-waited-for per line and fixes #918. This fix was required to run our [large-scale analysis of SARS-CoV-2 samples](https://github.com/cbg-ethz/SARS-CoV-2_Analysis).


epruesse:
- For this case, another option would be doing multiple --wait-for-files options. The maximum length for a single argument is significantly shorter than the maximum length for the entire argument list. IIRC on Linux the total can be 2MB, and increased further by increasing stack size. That's a lot of very long file names. Would avoid using a secondary temporary file.
- Where does the temp file you create get deleted?
- Can you add a check that only triggers the temp-file if it is required?
- The file can't just be `open()`ed, but should be using the NFS timeout. Some of us have crappy Isilon filers that have a worst case latency of 600+ seconds for files to appear across the cluster.

(Just my $0.02 - wait for johannes to say how he wants this fixed)

kpj:
@epruesse Thanks for your feedback!

* While I do not know the various maximum lengths, I feel like using multiple `--wait-for-files` options is just delaying the problem instead of solving it.
* At the moment I do not explicitly delete them, because I assumed the whole temporary directory it and the jobscript are located in will be deleted eventually.
* Certainly! What string length should I use as a threshold?
* What exactly do you mean by "should be using the NFS timeout"?

epruesse:
> * While I do not know the various maximum lengths, I feel like using multiple `--wait-for-files` options is just delaying the problem instead of solving it.

There are always limits. A file name can be up to 255 bytes, so with a directory structure, let's assume 512 bytes per file on average (probably a lot less most of the time). The maximum command line length is operating system dependent. On modern Linux, there are ARG_MAX and MAX_ARG_STRLEN. ARG_MAX is 1/4 of the stack size, so 2MB (4k files), but only by default. It can be increased using `ulimit` as much as you like (if you are root and have the RAM). MAX_ARG_STRLEN is hardcoded to 32 pages, so 128kb. That's the length for any one argument, and the one that my PR worked around (yes, also for covid datasets).

This means that at least under Linux, given willing sysadmins, you can use the multiple-options approach to do arbitrary number of arguments (limited by RAM). On OSX, the figure is 256kb total, but no limit on individual arguments (as far as I can tell). 

An easy way to see the limits is `xargs --show-limits` if you have GNU xargs installed, by the way. 

> * At the moment I do not explicitly delete them, because I assumed the whole temporary directory it and the jobscript are located in will be deleted eventually.

Ok, that's probably enough. The code deleting is in the same file, right? Just in case someone changes it later, so they have a chance to see it. 

> * Certainly! What string length should I use as a threshold?

Check my PR. I'd use the same variable, as it's the same problem.

I think it's OK to underestimate what is possible. For a job with 100 files, writing and reading another file won't make a big difference. 10000 jobs with 1 file, however, would likely suffer from adding another file committed to disk. 

Wait for Johannes on this though. He may prefer less complexity and just always use a file to transfer `wait-for-files`. Or maybe use the persistence system that already marks incomplete files somehow. 

> * What exactly do you mean by "should be using the NFS timeout"?

Short explanation:
Wait for the file to become available for `--latency-wait` seconds as is done for the `wait-for-files` themselves. 

Long explanation:

Like almost everything these days, distributed file systems use caches to improve performance. Like every cache, you have to consider coherency. Perfect cache coherence, ensuring that every cache on every client is at all times coherent (=contains the same data) is expensive (=slow), particularly if there are many caches. On a classical compute cluster with NFS shared file system, each compute node has a cache for NFS. This cache stores e.g. attributes, time stamps and so on. The NFS client will only check for new files in a directory if the directory time stamp has changed (AFAIK). So the timeout (that's the NFS timeout I meant) that governs how long directory attributes are cached governs also how long one must worst-case wait until a file appears on the client. The Linux NFS client has the values `acdirmin` and `acdirmax` for this, which default to 30 and 60 seconds, respectively.

So in effect, when you are on NFS (or some other distributed file system with imperfect cache coherency), you can have a scenario where a job completes, Snakemake learns of this from the cluster management engine (e.g. SLURM), but the files aren't there yet. It will then loop and wait for those files for the time you specified with `--latency-wait` before it decides that your job must have actually failed and not produced the files it was supposed to produce.

Similarly, a job that is submitted could be suffering from the same effects. So Snakemake also waits for files in the jobs submitted to the cluster. I think --wait-for-files does just that.

When you create a temporary file in the new folder, in theory it should not be in the cache on the cluster node when the job starts, so it probably works, at least on NFS. It would be safer, though, to use the same mechanism that Snakemake uses for the other files, to make sure that this file is there, before failing when opening it. 

One example of particularly frustrating storage systems is the "Dell Isilon" we are forced to use by our IT department. Besides being absurdly expensive and other issues, it exhibits worst case cache coherency timeouts of 600+ seconds, and IT and Dell have been unable to fix this. Basically, if you write something on one node, you can't just expect it to be there immediately on another node, it might take 10 minutes, sometimes more (don't know if there is a bound).

So write defensively, wait for your file just in case it takes a while to become visible. It should not. But it might rarely. And if you are running thousands of samples, "might rarely" becomes 90% of your work...

epruesse:
@johanneskoester Can you comment on how @kpj needs to implement this to be acceptable for merging?

kpj:
@epruesse Thank you so much for your detailed and insightful response! It was very helpful.


> Ok, that's probably enough. The code deleting is in the same file, right? Just in case someone changes it later, so they have a chance to see it.

Yes, the temp-directory is deleted in `ClusterExecutor.shutdown`.


> Check my PR. I'd use the same variable, as it's the same problem.

Thanks! That would be `MAX_ARG_LEN = 16 * 4096 - 1`.


> Wait for the file to become available for `--latency-wait` seconds as is done for the `wait-for-files` themselves.

But this would only be relevant for reading (and not writing the file), right?
And this I am already doing by calling `snakemake.io.wait_for_files` [here](https://github.com/kpj/snakemake/blob/3bf3ece334e802cbb3a8a5421999536e8d63d6dc/snakemake/__init__.py#L2602).

epruesse:
> But this would only be relevant for reading (and not writing the file), right?

Correct

> And this I am already doing by calling `snakemake.io.wait_for_files` [here](https://github.com/kpj/snakemake/blob/3bf3ece334e802cbb3a8a5421999536e8d63d6dc/snakemake/__init__.py#L2602).

Ah, I missed that part. Yes, looks good to me.

johanneskoester (changes requested):
Hi Elmar and Kim. Thanks a lot for looking into this, and sorry for the delay. I think the file approach is ok if we wait on the file in the same way as we later wait on its contents. Regarding the overhead, what about simply looking whether the number of files to wait for exceeds some absolutely safe threshold, say 20? This way, for 99% of the jobs we won't create the temporary file.

kpj:
That sounds like a reasonable solution to me!

Do you have a suggestion what the best way of implementing this would be?
I tried something along the lines of
```python
# Only create extra file if we have more than 20 input files.
# This should not require the file creation in most cases.
if len(wait_for_files) > 20:
    wait_for_files_file = self.get_jobscript(job) + ".waitforfilesfile.txt"
    with open(wait_for_files_file, "w") as fd:
        fd.write("\n".join(wait_for_files))

    waitfiles_parameter = format("--wait-for-files-file {wait_for_files_file}",
        wait_for_files_file=wait_for_files_file
    )
else:
    waitfiles_parameter = format("--wait-for-files {wait_for_files}",
        wait_for_files=wait_for_files
    )

format_p = partial(
    self.format_job_pattern,
    job=job,
    properties=job.properties(cluster=self.cluster_params(job)),
    latency_wait=self.latency_wait,
    waitfiles_parameter=waitfiles_parameter,
    path=path,
    **kwargs,
)
```

but passing `waitfiles_parameter=waitfiles_parameter` to `self.format_job_pattern` seems to quote the whole thing in the final jobscript:
```python
[..]
snakemake bar --snakefile Snakefile \
--force -j --keep-target-files --keep-remote --max-inventory-time 0 \
'--wait-for-files .snakemake/tmp.cizqyh76 foo' --latency-wait 5 \
[..]
```
This is probably due to `_quote_all=True` being passed to `self.format_job`.

kpj:
@johanneskoester Do you have any suggestion regarding the implementation? :-)

johanneskoester:
Sorry for the delay. You need to include the item with `{waitfiles_parameter:u}`, forcing that it is not quoted.

kpj:
Thanks a lot, that did the trick!

johanneskoester (approved):

dandaman:
Is there an example for how to use this option? 

I'm running into an "Argument list too long"  issue when aggregating dynamic results in a workflow and can't quite wrap my head around how I can generate this file within a workflow. 

A toy example workflow (here the input is known from the beginning, but lets assume it wouldn't be):

```snakemake
from pathlib import Path
from Bio import SeqIO
import polars as pl

inpath="sequences"

rule all:
    input: 
        "tmp/genome_sizes.tsv",

rule get_info:
    input: 
        inpath + "/{sample}.fasta"
    output: 
        temp("tmp/{sample}.tsv")
    run: 
        o=list()
        label = Path(input[0]).stem
        for seq_record in SeqIO.parse(input[0], "fasta"):
            o.append({
                "label": label,
                "seq_label": seq_record.id,
                "length": len(seq_record.seq)
            })
            df = pl.DataFrame(o)
        df.write_csv(output[0], separator="\t", include_header=True)
rule combine:
    input: 
        expand("tmp/{sample}.tsv", sample=glob_wildcards(inpath + "/{sample}.fasta").sample)
    output: 
        "tmp/genome_sizes.tsv"

    run:
        df = pl.concat([pl.read_csv(f, separator="\t") for f in input],how="diagonal_relaxed")
        df.group_by("label").agg(pl.col("length").sum()).sort("label")\
            .write_csv(output[0], separator="\t", include_header=True)
```

PR #932, opened 2021-03-23
Allow params expansion to handle FileNotFoundError (fixes #927)

### LINKED ISSUE #927: Using input lambda in param can cause FileNotFoundError in dry-run
Couldn't see anything like this in previous issues, but encountered this in at least versions 5.32 and 6.0.5.

Dry-runs break when a rule uses the a lambda expression with an input file, **IF** that files does not currently exist. The file is generated before this rule should execute (it is in its input), but obviously the dry-run mechanism bypasses the actual generation. As such, a `FileNotFoundError` is raised. 

Without the dry-run flag, this works correctly as expected, as that rule (and so params) is presumably not evaluated until its input is available.

A simplified version is 

```python
rule all:
    input:
        'final.txt'

rule generate_value:
    output:
        'model.txt'
    shell:
        'echo 1 > {output}'

rule text:
    input:
	    model = 'model.txt'
    output:
        'final.txt'
    params:
	    lambda wildcards, input: [int(i) for i in open(input.model)]
    shell:
	    'echo {params} > {output}'
```

Running with `snakemake -n` breaks with FileNotFoundError

Running with `snakemake -j 1` works correctly.

ASLeonard:
Previously a param value that depended on input that had not been generated previously would fail. Analogous to how resource expansion occurs, I might expect the param expansion to identify the value depends on input not yet available, and display a <TBD> value.

These changes allow for that features by moving the code which handles this in resource expansion to the more general `apply_input_function`. With the FileNotFoundError handling occurring there, the code in the resource expansion could be simplified. 

This was tested on and passed the minimal reproducible example, resolving #927, but may interfere in unexpected ways with other code.

The output of `snakemake -n -p` for a rule with `shell: echo {params}` where the param depends on the input now looks something like 
```
echo <TBD>
```

### DIFF
--- a/snakemake/rules.py
+++ b/snakemake/rules.py
@@ -682,8 +682,8 @@ class Rule:
             value = incomplete_checkpoint_func(e)
             incomplete = True
         except FileNotFoundError as e:
-            # Resources can depend on input files. Since expansion can happen during dryrun,
-            # where input files are not yet present, we need to skip such resources and
+            # Function evaluation can depend on input files. Since expansion can happen during dryrun,
+            # where input files are not yet present, we need to skip such cases and
             # mark them as [TBD].
             if e.filename in aux_params["input"]:
                 # use zero for resource if it cannot yet be determined
--- a/snakemake/rules.py
+++ b/snakemake/rules.py
@@ -681,6 +681,15 @@ class Rule:
         except IncompleteCheckpointException as e:
             value = incomplete_checkpoint_func(e)
             incomplete = True
+        except FileNotFoundError as e:
+            # Resources can depend on input files. Since expansion can happen during dryrun,
+                        # where input files are not yet present, we need to skip such resources and
+                        # mark them as [TBD].
+            if e.filename in aux_params['input']:
+                # use zero for resource if it cannot yet be determined
+                value = TBDInt(0)
+            else:
+                raise e    
         except (Exception, BaseException) as e:
             if raw_exceptions:
                 raise e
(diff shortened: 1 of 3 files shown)


snakemake/rules.py
```diff
@@ -681,6 +681,15 @@ def apply_input_function(
         except IncompleteCheckpointException as e:
             value = incomplete_checkpoint_func(e)
             incomplete = True
+        except FileNotFoundError as e:
+            # Resources can depend on input files. Since expansion can happen during dryrun,
+            # where input files are not yet present, we need to skip such resources and
+            # mark them as [TBD].
+            if e.filename in aux_params["input"]:
+                # use zero for resource if it cannot yet be determined
+                value = TBDInt(0)
```
johanneskoester, line 690: This should then probably be changed into simply a TBD object instead of an Int. Otherwise, it is a big misleading.

snakemake/rules.py
```diff
@@ -681,6 +681,15 @@ def apply_input_function(
         except IncompleteCheckpointException as e:
             value = incomplete_checkpoint_func(e)
             incomplete = True
+        except FileNotFoundError as e:
+            # Resources can depend on input files. Since expansion can happen during dryrun,
```
johanneskoester, line 685:
```suggestion
            # Function evaluation can depend on input files. Since expansion can happen during dryrun,
```

snakemake/rules.py
```diff
@@ -681,6 +681,15 @@ def apply_input_function(
...
+            # where input files are not yet present, we need to skip such resources and
```
johanneskoester, line 686:
```suggestion
            # where input files are not yet present, we need to skip such cases and
```

johanneskoester (changes requested):
This is a good idea. Some comments below.

ASLeonard:
@johanneskoester, these testing action seems to have gotten stuck in some conda/pip loop on the "Setup Snakemake environment" stage for several hours now with the same repeating message. I don't have the ability to kill the workflow action, but these tests are now just burning CPU time.

```
Warning: you have pip-installed dependencies in your environment file, but you do not list pip itself as one of your conda dependencies.  Conda may not use the correct pip to install your packages, and they may end up in the wrong place.  Please add an explicit pip dependency.  I'm adding one for you, but still nagging you.
```

ASLeonard:
Not sure if it was just a temporary issue with the CI or what, but the tests passed fine now without getting stuck on the setup.

epruesse:
ping @johanneskoester - I think this is ready?

ASLeonard:
Any chance of merging this in soon @johanneskoester?

johanneskoester:
Sorry for the delay!

johanneskoester (approved):

PR #936, opened 2021-03-30
fix various bugs in module system that occur only with complex cases

johanneskoester:


dlaehnemann:
Now, it seems to choke on `scattergather: calling=16`, used in `benchmarking.smk` on line 5 with `gather.calling()`, even though this should be correct syntax. Maybe something in the parsing order is off?
https://github.com/snakemake/snakemake/pull/936/checks?check_run_id=2227956056#step:9:378

And for Windows, it joins the remote file paths with the windows folder delimiter `\` instead of the url-one `/`:
https://github.com/snakemake/snakemake/pull/936/checks?check_run_id=2227955981#step:6:409

Issue #883, opened 2021-02-19
Python module not found when submitting jobs to cluster


yztxwd:
**Snakemake version**
snakemake 5.32.2

**Describe the bug**
I used snakemake on a cluster, the rule is:

    _rule mapq_filter:
        input:
            "output/mapped/{sample}-{rep}-{unit}.flag.sort.bam"
        output:
            temp("output/mapped/{sample}-{rep, [^-]+}-{unit, [^.]+}.flag.filtered.bam"),
        params:
            lambda wildcards: (config["filter"]["se"] if is_single_end(**wildcards)
                else config["filter"]["pe"])
        conda:
            "../envs/py3.yaml"
        script:
            "../scripts/reads_filter_smk.py"_

py3.yaml describes an environment with pysam installed, I submitted the job by command

    _snakemake --use-conda --jobs 1 --cluster "qsub -p 1023 -W umask=0007 -W group_list=sam77_collab -l walltime=30:00:00 -l nodes=1:ppn=10:rhel7 -j oe -A sam77_h_g_sc_default -l mem=40gb"  --rerun-incomplete_

It returns an error that it can't find the pysam module.

**Logs**

    _[Thu Feb 18 10:34:57 2021]
    rule mapq_filter:
        input: output/mapped/mES_H3K27ac-rep2-lane1.flag.sort.bam
        output: output/mapped/mES_H3K27ac-rep2-lane1.flag.filtered.bam
        jobid: 0
        wildcards: sample=mES_H3K27ac, rep=rep2, unit=lane1
    
    Activating conda environment: /gpfs/group/sam77/default/lab/jianyu/mES_H3K27ac/.snakemake/conda/45b5e2ae
    Activating conda environment: /gpfs/group/sam77/default/lab/jianyu/mES_H3K27ac/.snakemake/conda/45b5e2ae
    Traceback (most recent call last):
      File "/gpfs/group/sam77/default/lab/jianyu/mES_H3K27ac/.snakemake/scripts/tmp86cj2b0o.reads_filter_smk.py", line 6, in <module>
        import sys, pysam, re
    ModuleNotFoundError: No module named 'pysam'
    [Thu Feb 18 10:35:04 2021]
    Error in rule mapq_filter:
        jobid: 0
        output: output/mapped/mES_H3K27ac-rep2-lane1.flag.filtered.bam
        conda-env: /gpfs/group/sam77/default/lab/jianyu/mES_H3K27ac/.snakemake/conda/45b5e2ae
[...]

PR #938, opened 2021-04-01
Append rather than prepend Snakemake bin path for cluster execution

jaicher:
Fixes unavailble python modules when submitting jobs to cluster (#883)
as observed in some cases using workaround first reported by @yztxwd.

As reported later by @jaicher in the same issue, the Snakemake bin path
needs to be on the PATH in order to ensure that PuLP can find external
solvers (e.g. coincbc) that are used for Snakemake's ILP solver.
However, prepending Snakemake's bin path to the PATH conflicts with
conda activate when:

+ the cluster job already has some (default) conda environment(s)
  activated
+ this default conda environment is different than the one Snakemake is
  being run from

In this case, the call to conda activate replaces the entries in the
PATH coresponding to the previously active environment, leaving
Snakemake's bin path ahead of it. This makes Snakemake's version of
Python the default, leading to missing Python modules when commands are
run in a different environment than intended.

By appending Snakemake's bin path, the solvers are on the path and
should be accessible so long as there are no conflicts on the PATH for
the solvers that are unusable/unacceptable for PuLP.

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -744,11 +744,13 @@ class ClusterExecutor(RealExecutor):
         if self.assume_shared_fs:
             wait_for_files.append(self.tmpdir)
             wait_for_files.extend(job.get_wait_for_files())
-            # Prepend PATH of current python executable to PATH.
+            # Append PATH of current python executable to PATH.
             # This way, we ensure that the snakemake process in the cluster node runs
-            # in the same environment as the current process.
-            # This is necessary in order to find the pulp solver backends (e.g. coincbc).
-            path = "PATH='{}':$PATH".format(os.path.dirname(sys.executable))
+            # with the pulp solver backends (e.g. coincbc) available on the PATH.
+            # Appending ensures that Snakemake's Python cannot supersede Python
+            # from a job's environment (conda activate replaces the first PATH
+            # corresponding to the current environment rather than prepending)
+            path = "PATH=${{PATH}}:'{}'".format(os.path.dirname(sys.executable))
 
         format_p = partial(
             self.format_job_pattern,


snakemake/executors/__init__.py
```diff
@@ -744,11 +744,13 @@ def format_job(self, pattern, job, **kwargs):
...
+            # Append PATH of current python executable to PATH.
             # This way, we ensure that the snakemake process in the cluster node runs
-            # in the same environment as the current process.
-            # This is necessary in order to find the pulp solver backends (e.g. coincbc).
-            path = "PATH='{}':$PATH".format(os.path.dirname(sys.executable))
+            # with the pulp solver backends (e.g. coincbc) available on the PATH.
+            # Appending ensures that Snakemake's Python cannot supersede Python
+            # from a job's environment (conda activate replaces the first PATH
+            # corresponding to the current environment rather than prepending)
+            path = "PATH=${{PATH}}:'{}'".format(os.path.dirname(sys.executable))
```
johanneskoester, line 753: Mhm. Not entirely sure about this. On, e.g., the login node, you start snakemake which submits jobs. Why should these jobs prepend an environment via conda which is something else than the one that you use on the login note to start snakemake?

jaicher: I was thinking of the case when we are using `--use-conda`, and the rule has the `conda:` directive specifying a different conda environment to make. In this case, the current approach (prepending) can lead to the wrong Python being first in the PATH.

For example:
```
1. (qsub batch job starts with activated default conda environment)
    PATH="{sge-binaries}/bin:$HOME/miniconda3/bin:{...}"
2. (Snakemake job script prepends snakemake bin directory)
    PATH="$HOME/miniconda3/envs/snakemake/bin:{sge-binaries}/bin:$HOME/miniconda3/bin:{...}"
3. (Snakemake activates desired conda environment, replacing default conda environment)
    PATH="$HOME/miniconda3/envs/snakemake/bin:{sge-binaries}/bin:{pipeline-dir}/.snakemake/conda/{env-hash}/bin:{...}"

(python runs from $HOME/miniconda3/envs/snakemake/bin, not {pipeline-dir}/.snakemake/conda/{env-hash}/bin...)
```

We want binaries (e.g. Python) from the conda rule directive to take precedence over Python from those in the Snakemake environment.

However, I interpret your comment as rightfully pointing out that we run into the reverse issue if we are not using conda and we use the naive appending solution. Then, to be similar to running locally, we effectively want to use the PATH from where Snakemake was executed. But, if we append, the version of Python we'll end up using will probably be default conda, or worse, the system Python.

A simple solution could be: append if we are using a conda environment, otherwise prepend. What do you think @johanneskoester?

jaicher: @johanneskoester does Snakemake make the assumption that the cluster environment (from `bsub`, `qsub`, `sbatch`, etc.) inherits environment variables from the node on which Snakemake itself runs?

My HPC is transitioning away from SGE (the scheduler that this problem, i.e. #883, seems to always occur on) to slurm. By default Slurm and bsub (the other scheduler I've worked with in the past) copy environment variables, while qsub does not.

The current master should work if:
```
(
 Snakemake is run from the active conda environment
 AND
 the job submission copies the environment variables (in particular those for conda) to the compute node
)
OR
The compute node starts with no conda environment activated
```

If the environment variables assumption is correct, then I suspect a lot of SGE users are breaking it (using older versions of the suggested snakemake profile or from rolling their own -- [the intro to cluster execution](https://snakemake.readthedocs.io/en/stable/executing/cluster.html) uses qsub as an example without the `-V` flag). But, then the solution would be turning this PR into (or starting a new PR to) update documentation to explicitly state this assumption?

johanneskoester:
Thanks a lot for the explanation, now I get it. I fixed this in a different way in commit 3e3a1016: instead of modifying PATH in the entire job, we now explicitly pass the solver PATH to the snakemake spawned in the job. By that, it is not necessary at all to modify PATH.

johanneskoester:
And thanks a lot for digging into this!

johanneskoester:
I'll release a new version soon. If you experience further problems with this, please let me know via the discord chat (that is easier for me to find instead of here).

jaicher:
Wonderful, I was about to close this myself because of the SGE/Torque-specific solution (since this problem only has been reported on those schedulers) that @katiedelange posted (passing in environment variable `CONDA_AUTO_ACTIVATE_BASE=false`) on #883. But your solution should be even better. Thank you!

Unfortunately, I can't test this anymore because the cluster I have access to switched to slurm, but maybe one of the people who have ran into this problem (@katiedelange @yztxwd) could test commit 3e3a101 to confirm that it resolves their issue so that we can close #883 as well?

PR #947, opened 2021-04-08
Paramspace enhancement: Use all dataframe columns when `filename_params == True`

kpj:
This PR allows to set `filename_params=True` when creating a Paramspace object in order to automatically use all columns.

This is useful when immediately passing the dataframe to Paramspace:
```python
import pandas as pd
from snakemake.utils import Paramspace

paramspace = Paramspace(pd.read_csv('config/params.csv'), filename_params=True)
```

### DIFF
--- a/snakemake/utils.py
+++ b/snakemake/utils.py
@@ -618,7 +618,7 @@ class Paramspace:
             self.pattern = "/".join([r"{}"] * len(self.dataframe.columns))
             self.ordered_columns = self.dataframe.columns
         else:
-            if isinstance(filename_params, bool) and filename_params:
+            if isinstance(filename_params, str) and filename_params == "*":
                 filename_params = dataframe.columns
 
             if any((param not in dataframe.columns for param in filename_params)):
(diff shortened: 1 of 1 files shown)


snakemake/utils.py
```diff
@@ -618,6 +618,9 @@ def __init__(self, dataframe, filename_params=None, param_sep="~"):
             self.pattern = "/".join([r"{}"] * len(self.dataframe.columns))
             self.ordered_columns = self.dataframe.columns
         else:
+            if isinstance(filename_params, bool) and filename_params:
```
johanneskoester, line 621: I think it would be more readable if we require a filename_params to be a string `"*"`, instead of true. Or, alternatively, we introduce another keyword argument for this behavior, like `only_filename=True` or so.

johanneskoester (changes requested):

snakemake/utils.py
kpj: Adding a new keyword argument would be very specific but also increase complexity more than making `filename_params` more flexible (I think?).
So I'll make us of `"*"`.

johanneskoester (approved):

PR #948, opened 2021-04-08
Fix wrong conversion of string "False" to boolean True in Paramspace

kpj:
Creating a Paramspace from a dataframe which contains the boolean `False` will erroneously convert `False` to `True` when inferring instance values.

This PR adds a simple fix which checks whether the value is `"False"` and from a boolean column and returns `False`.
While this is not very general, more complicated implementations could lead to unneeded complexity.


johanneskoester:
Not sure I understand the rationale, why not simply rely on the dataframe to be correct when passed over to paramspace? Ensuring that all values and datatypes are correct is up to the developer who loads the dataframe with `pd.read_csv` isn't it?

kpj:
> Not sure I understand the rationale, why not simply rely on the dataframe to be correct when passed over to paramspace? Ensuring that all values and datatypes are correct is up to the developer who loads the dataframe with `pd.read_csv` isn't it?

Absolutely! But this is exactly what is not working.

Consider the following Snakefile:
```python
import pandas as pd
from snakemake.utils import Paramspace

paramspace = Paramspace(pd.DataFrame({"A": [True, False]}))


rule all:
    input:
        expand("results/{params}/foo", params=paramspace.instance_patterns),


rule foo:
    output:
        touch(f"results/{paramspace.wildcard_pattern}/foo"),
    params:
        pm=paramspace.instance,
    run:
        if params.pm["A"].all():
            print("TRUE")
        else:
            print("FALSE")
```

One would expect it to print both `TRUE` and `FALSE`. However, it will print `TRUE` twice.

This is because the parameter instances are created from the wildcard argument, i.e. strings.
Consequently, the line `name: pd.Series([value]).astype(self.dataframe.dtypes[name])` will evaluate to `name: pd.Series(['False']).astype(np.dtype('bool'))`, which in turn evaluates to `True` (because only the empty string is `False`).

In fact, I do also not understand why the parameter values are wrapped in `pd.Series`.

johanneskoester:
Ah, now I get it, thanks!

snakemake/utils.py
```diff
@@ -670,8 +670,16 @@ def instance(self, wildcards):
         """Obtain instance (dataframe row) with the given wildcard values."""
         import pandas as pd
 
+        def convert_value_dtype(name, value):
+            if self.dataframe.dtypes[name] == bool and value == "False":
+                # handle problematic case when boolean False is returned as
+                # boolean True because the string "False" is misinterpreted
+                return pd.Series([False])
```
johanneskoester, line 677:
```suggestion
                return False
```

snakemake/utils.py
```diff
@@ -670,8 +670,16 @@ def instance(self, wildcards):
...
+            else:
+                return pd.Series([value]).astype(self.dataframe.dtypes[name])
```
johanneskoester, line 679:
```suggestion
                return pd.Series([value]).astype(self.dataframe.dtypes[name])[0]
```

johanneskoester (changes requested):
Right, no reason for the series. The [0] is missing.

johanneskoester (approved):

PR #949, opened 2021-04-08
Implement rule-specific `restart_times` directive

### LINKED ISSUE #367: Allow --restart-times to specify rule name to be applied to
For example, jobs which download data via the internet can sometimes (unexpectedly) crash due to network problems. In such cases, restarting them is the only possible fix.

Snakemake has the parameter `--restart-times N` which restarts any failing job `N` times.
It would be great if this parameter would optionally take a rule name to only allow multiple restarts for that particular rule.

It is certainly possible to handle automatic retries within the rule itself, but using Snakemake's `--restart-times` parameter seems like a more elegant way of handling this (in particular with snakemake wrappers).

kpj:
It is possible to call snakemake with `--restart-times` to automatically restart a rule if it crashes (and, e.g., increase resource requirements for the next run). So far this is only possible for all rules together and lacks specificity.

This PR introduces a rule-specific `restart-times` directive which allows to specify the number of restart attempts for a specific rule. It fixes #367.

A (silly) example could look like this:
```python
rule all:
    input:
        "foo",


rule foo:
    output:
        touch("foo"),
    restart_times:
        1
    resources:
        mem_mb=lambda wildcards, attempt: 10 * attempt,
    run:
        print(resources.mem_mb)
        1 / 0
```

### DIFF
--- a/snakemake/__init__.py
+++ b/snakemake/__init__.py
@@ -127,7 +127,7 @@ def snakemake(
     keep_logger=False,
     max_jobs_per_second=None,
     max_status_checks_per_second=100,
-    restart_times=0,
+    workflow_restart_times=0,
     attempt=1,
     verbose=False,
     force_use_threads=False,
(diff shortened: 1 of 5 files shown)


kpj:
At the moment there is an issue where rules which set the `restart_times` directive are executed too often because I do not understand how Snakemake's execution internals work.
The global variable set by the `--restart-times` commandline parameter is reset to `0` for rule executions and only retains its actual value the first time `main()` and thus `exec(compile(..), ..)` is executed.
I cannot figure out how this reset works and cannot apply it to the rule-specific restart times 😞 

I'd be happy to hear how one could solve this.

kpj:
@johanneskoester Do you have any suggestion regarding the implementation? :-)

kpj:
This is now resolved by #1649.

PR #952, opened 2021-04-12
test: module import with `as something_*` prefixing and ruleorder specified

dlaehnemann:
When importing a module with prefixes for the imported rules (`use rule * from something as something_*`), it seems like ruleorder statements (`ruleorder: rule1 > rule2`) are not properly prefixed (statement should become: `ruleorder: something_rule1 > something_rule2`) and throw a respective error of an unknown rule (`rule1` and `rule2` don't exist in the prefixed import).


dlaehnemann:
It seems like this PR cannot reproduce the `ruleorder` problem (maybe it's fixed in the meantime?), but reveals another problem:

For some reason, the `replace_prefix: { "resources": "resources/first/" }` leads to the `rule first_map_reads` to ask for `resources/first/first/genome.fasta.pac` et al. So it seems like prefix-replacement is done twice, but only for `resources/`...?!

johanneskoester (approved):

PR #961, opened 2021-04-16
add module test with script

dlaehnemann:
This is mainly to test adjusting paths to scripts when they are included from a remote Snakefile.

Currently, this first throws a different error, because the respective rule in the included `test_script_py/Snakefile` is unnamed and I assume the `use rule * from with_script as with_script_*` tries to rename the rule but doesn't find a name to append to. But as the original test file is passing, I am assuming that unnamed rules are somehow valid? If not, we would need to first change this test case (and `test_script/Snakefile`, as well), before we can update this to the respective commit and use for testing fixing of the script path adjustment.


johanneskoester:
Unnamed rules are valid, the prefixing needs to handle that, indeed.

johanneskoester:
Good catch!

johanneskoester (approved):

PR #967, opened 2021-04-21
Fix temp() breaks checkpoints (fixes #966)

### LINKED ISSUE #966: Checkpoints with temp() output broken
**Snakemake version**

6.1.2

**Describe the bug**

Checkpoints with outputs marked as `temp()` do not work. 

The result is that `output[0]` is returned instead of the result of the function accessing the checkpoint. (I.e., the placeholder is used and the input function never executed to the end)

**Minimal example**

```python
def get_realfile(wc):
    with checkpoints.cpt.get().output.redirectfile.open() as fd:
        return fd.read().strip()

rule all:
    input:
        get_realfile
    shell:
        "echo {input}"

checkpoint cpt:
   output:
       redirectfile = "myfiles.txt",
       tempfile = temp("tempfile.txt")
   shell: """
   echo realfile.txt > {output.redirectfile};
   touch {output.tempfile} realfile.txt
   """

rule clean:
    shell: """
    rm -f  myfiles.txt tempfile.txt realfile.txt
    """
```

This ***should*** print `realfile.txt`. Instead it prints `myfiles.txt`. If you remove the `temp()` from the tempfile, it will work correctly. 

The cause is `checkpoint.get` raising the `IncompleteCheckpointException` again even though the job has completed because the file flagged `temp` has already been removed:

https://github.com/snakemake/snakemake/blob/8c51c32941e3076b5ccfee98d386256f55e1f6bf/snakemake/checkpoints.py#L33-L35

It should suffice to also check if `"temp" in f.flags`.

epruesse:
Tried to also make this a little more readable / more easily adjusted without getting into formatting issues.

### DIFF
--- a/snakemake/checkpoints.py
+++ b/snakemake/checkpoints.py
@@ -34,7 +34,7 @@ class Checkpoint:
             for iofile in output:
                 if iofile in self.checkpoints.future_output:
                     break
-                if not iofile.exists and not "temp" in iofile.flags:
+                if not iofile.exists and not iofile.is_temp:
                     break
             else:
                 return CheckpointJob(self.rule, output)
--- a/snakemake/checkpoints.py
+++ b/snakemake/checkpoints.py
@@ -30,12 +30,16 @@ class Checkpoint:
             )
 
         output, _ = self.rule.expand_output(wildcards)
-        if self.checkpoints.future_output is None or any(
-            (not f.exists or f in self.checkpoints.future_output) for f in output
-        ):
-            raise IncompleteCheckpointException(self.rule, checkpoint_target(output[0]))
+        if self.checkpoints.future_output is not None:
+            for iofile in output:
+                if iofile in self.checkpoints.future_output:
+                    break
+                if not iofile.exists and not "temp" in iofile.flags:
+                    break
+            else:
+                return CheckpointJob(self.rule, output)
 
-        return CheckpointJob(self.rule, output)
+        raise IncompleteCheckpointException(self.rule, checkpoint_target(output[0]))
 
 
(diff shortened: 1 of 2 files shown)


epruesse:
@johanneskoester - this is ready. Just a few lines to hot fix checkpoints with temp() output.

This is not a full fix. You still cannot access temp files from the checkpoint get() calling rule because the files are already gone. It's not really desirable to allow that, though, as it might trigger inadvertent rerun of the checkpoint. I'd document that checkpoint.get() mustn't access the temp() outputs and possibly add an exception if the code does attempt to do this. Still, at least the behaviour with this patch is "rational" - you get the right file names, not the `output[0]` with the flag stripped and no idea that your checkpoint-calling-function never got executed to the end.

johanneskoester (approved):
Thanks!

johanneskoester:
> @johanneskoester - this is ready. Just a few lines to hot fix checkpoints with temp() output.
> 
> This is not a full fix. You still cannot access temp files from the checkpoint get() calling rule because the files are already gone. It's not really desirable to allow that, though, as it might trigger inadvertent rerun of the checkpoint. I'd document that checkpoint.get() mustn't access the temp() outputs and possibly add an exception if the code does attempt to do this. Still, at least the behaviour with this patch is "rational" - you get the right file names, not the `output[0]` with the flag stripped and no idea that your checkpoint-calling-function never got executed to the end.

I agree. I'll add that to the docs now.

epruesse:
@johanneskoester Thanks for merging quickly! I had been thinking about adding a proxy object to use instead of the NamedList that would raise if a temp file was accessed.

PR #977, opened 2021-05-04
fix: quote jobid passed to status script to support multi-cluster Slurm setup

jdblischak:
I'm a first-time contributor but long-time user, so I wanted to start by saying: Snakemake is awesome! I really appreciate all the effort to develop and maintain this awesome resource.

#### Motivation

I am trying to use `--cluster-status` with a multi-cluster Slurm setup. To check the status of the job, both `sacct` and `scontrol` require the jobid and the name of the cluster. When I use the `sbatch` option `--parsable`, it returns both in the format `jobid;cluster_name`. This entire string gets passed to the custom script I pass to `--cluster-status`, which is great since it provides all the information that the script requires. The problem is that the string isn't quoted, thus the semicolon delimiter is misinterpreted as the end of a line. Thus the shell attempts to execute the name of the cluster. I discovered this behavior with Snakemake 6.2.1.

This happens because the first argument passed to the cluster status script isn't quoted:

https://github.com/snakemake/snakemake/blob/b3c4e687c87c75075393cef842b129dcec70e7f6/snakemake/executors/__init__.py#L1078-L1086

#### Proposed solution

This PR quotes the jobid string passed as the first argument to the cluster status script. I confirmed it works both in a standard Slurm setup (i.e. when `--parsable` returns only the job id) and also when specifying the cluster name with `--clusters`.

### QC

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -1084,7 +1084,7 @@ class GenericClusterExecutor(ClusterExecutor):
                 try:
                     # this command shall return "success", "failed" or "running"
                     ret = subprocess.check_output(
-                        "{statuscmd} {jobid}".format(
+                        "{statuscmd} '{jobid}'".format(
                             jobid=job.jobid, statuscmd=self.statuscmd
                         ),
                         shell=True,
--- a/tests/test_cluster_statusscript_multi/Snakefile.nonstandard
+++ b/tests/test_cluster_statusscript_multi/Snakefile.nonstandard
@@ -1,0 +1,13 @@
+from snakemake import shell
+
+envvars:
+    "TESTVAR"
+
+
+
+rule all:
+	input: 'output.txt'
+
+rule compute:
+	output: 'output.txt'
+	shell: 'touch {output}'
--- a/tests/test_cluster_statusscript_multi/sbatch
+++ b/tests/test_cluster_statusscript_multi/sbatch
@@ -1,0 +1,8 @@
+#!/bin/bash
+echo `date` >> sbatch.log
+tail -n1 $1 >> sbatch.log
+# simulate printing of job id by a random number plus the name
+# of the cluster
+echo "$RANDOM;name-of-cluster"
+cat $1 >> sbatch.log
+sh $1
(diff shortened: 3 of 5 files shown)


jdblischak:
Any interest in this PR? It's a minimal change that shouldn't break existing setups. The GitHub Action jobs won't run until they are first approved by a maintainer.

frankier:
I think the most robust solution is to use [shelex.quote(...) from the standard library](https://docs.python.org/3/library/shlex.html#shlex.quote).

jdblischak:
Thanks for activating the GitHub Actions on my PR!

I took a look at the failing tests. 

The Windows job failed `test_conda` because `test0.out` wasn't exactly as expected.

https://github.com/snakemake/snakemake/runs/4083323515?check_suite_focus=true#step:6:244
https://github.com/snakemake/snakemake/blob/main/tests/test_conda/expected-results/test0.out

The Ubuntu job failed `test_script`.

https://github.com/snakemake/snakemake/runs/4083323380?check_suite_focus=true#step:9:234
https://github.com/snakemake/snakemake/tree/main/tests/test_script

It's not clear to me how my minor change could be causing either of these errors. I'd appreciate any advice. My plan is to rebase my commit in case these issues were recently fixed and merged.

jdblischak:
I rebased my commit onto main 🤞

jdblischak:
I updated this from "feat" to "fix". All the infrastructure to support a multi-cluster Slurm setup already exists. My PR simply fixes the call to the custom cluster status script.

jdblischak:
I created a test case for this change. I modeled it after the test [test_cluster_statusscript](https://github.com/snakemake/snakemake/tree/main/tests/test_cluster_statusscript). Locally, I confirmed that is fails without my edit and passes with my edit. It also passes in the CI run.

jdblischak:
@johanneskoester This PR is ready for review. For more context, this minor edit will support a [multi-cluster Slurm setup](https://github.com/jdblischak/smk-simple-slurm/tree/main/examples/multi-cluster) for my Snakemake profile [smk-simple-slurm](https://github.com/jdblischak/smk-simple-slurm).

jdblischak:
@johanneskoester Friendly reminder that this minor PR is ready for review. It includes a test case, and all the CI jobs are passing. Thanks!

PR #984, opened 2021-05-09
Documentation: fix XRootD remote file example

admorris:
- input of `all` rule is local files, so don't use `XRootD.remote`
- `file_numbers` is the attr `n` of the return value of `glob_wildcards`


johanneskoester (approved):

johanneskoester:
thank you!

Issue #988, opened 2021-05-10
KeyError when ambiguous rule follows a checkpoint


jmeppley:
**Snakemake version**
Tested on the latest version: 6.3.0+2.gec87b97d

**Describe the bug**

Snakemake throws a KeyError while updating jobs after a checkpoint when there are two rules that can produce the same output. 

**Logs**

This is the error output:

```
Traceback (most recent call last):
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.two_outputs/snakemake/snakemake/executors/__init__.py", line 570, in _callback
    callback(job)
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.two_outputs/snakemake/snakemake/scheduler.py", line 544, in _proceed
    job, update_dynamic=update_dynamic
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.two_outputs/snakemake/snakemake/dag.py", line 1341, in finish
    updated_dag = self.update_checkpoint_dependencies(jobs)
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.two_outputs/snakemake/snakemake/dag.py", line 1302, in update_checkpoint_dependencies
    self.replace_job(j, newjob, recursive=False)
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.two_outputs/snakemake/snakemake/dag.py", line 1505, in replace_job
    self.delete_job(job, recursive=recursive)
  File "/mnt/data0/jmeppley/projects/snakemake_bugs/snakemake.two_outputs/snakemake/snakemake/dag.py", line 1464, in delete_job
    del self.dependencies[job_][job]
KeyError: multiple
```

**Minimal example**

Snakefile:
```
n_files = 4

rule all:
    input:
        faa='all.faa',
        fasta='all.fasta'

checkpoint make_files:
    """make a bunch of dummy fasta files"""
    output: 
        files=directory('files')
    run:
        os.makedirs(output.files)
        for i in range(n_files):
            filename = f'{output.files}/f{i}.fasta'
            with open(filename, 'wt') as fasta_out:
                fasta_out.write('>blank\nACTGACTG\n')

def get_cp_files(extension='txt'):
    """ after the checkpoint, get the names of the files with the given extension """
    files_dir = checkpoints.make_files.get().output.files
    files, = glob_wildcards(f'{files_dir}/{{file}}.fasta')
[...]

PR #989, opened 2021-05-10
Fix for issue #988: properly clean up phantom dependencies caused by ruleorder usage

jmeppley:
This is the simple band-aid solution to issue #988. It simply allows for the job to be missing from dependencies and moves on.


johanneskoester (approved):
Thank you! I have modified the solution to properly delete the dangling dependency pointers upon DAG cleanup.

Issue #991, opened 2021-05-11
Snakemake depends on functionality in smart_open 3


Redmar-van-den-Berg:
**Snakemake version**
snakemake-minimal=6.3.0

**Describe the bug**
When installing snakemake from Conda, smart_open version 1.9.0 is installed. However, Snakemake makes use of `smart_open.parse_uri`, which is not available in smart_open 1.9.0.

**Logs**
```
Traceback (most recent call last):
  File "miniconda3/envs/HAMLET/lib/python3.9/site-packages/snakemake/__init__.py", line 593, in snakemake
    workflow.include(
  File "miniconda3/envs/HAMLET/lib/python3.9/site-packages/snakemake/workflow.py", line 1094, in include
    if is_local_file(snakefile):
  File "miniconda3/envs/HAMLET/lib/python3.9/site-packages/snakemake/common.py", line 65, in is_local_file
    from smart_open import parse_uri
ImportError: cannot import name 'parse_uri' from 'smart_open' (miniconda3/envs/HAMLET/lib/python3.9/site-packages/smart_open/__init__.py)
```

**Solution**
Snakemake should depend on smart_open>=3

PR #993, opened 2021-05-11
Set minimum version of smart_open to 3.0

Redmar-van-den-Berg:
This fixes issue #991, which is triggered when installing the latest version of snakemake-minimal from Conda.


johanneskoester (approved):

johanneskoester:
Thank you!

PR #998, opened 2021-05-16
feat: Adding spack as an environment option

vsoch:
This is early work to add support for using spack as an environment! Some important points:

- I refactored the conda.py into a base.py and additional spack.py so shared functionality between spack and conda is there.
- The example tutorial to go along with this can be found at https://github.com/snakemake/snakemake-spack-tutorial
- There are likely many details that I haven't tested, either via unit test or just locally. We should chat about if anything is missing.
- Would it make sense to try and find current workflows that use spack, and see how easy (or hard) it would be to write Snakefiles for?
 - The one feature that spack can't easily support is environment archive, as a spack environment does not currently have an easy way to do this. For the time being, we save a specs.json along the environment, which technically should be useful to reproduce a particular spack install, if the spack.yaml is not appropriate.

Overall - let me know what more work or testing needs to be done, or what other examples I can come up with to better demonstrate or test this use case. I'm hoping that by making it easy to use spack we can attract a more broad set of users at, for example, national labs.

I suspect there will be testing issues with these changes (they are pretty large) so I'll try to address them as they come up.


vsoch:
@johanneskoester I'm seeing an error in a module that the snakemake version is too low:
```python
WorkflowError in line 3 of /tmp/snakemake-pujn5s_2/Snakefile:
Expecting Snakemake version 6.1.1 or higher.
  File "/tmp/snakemake-pujn5s_2/Snakefile", line 3, in <module>
------------------------------ Captured log call -------------------------------
DEBUG    snakemake.logging:logging.py:553 Full Traceback (most recent call last):
  File "/home/runner/work/snakemake/snakemake/snakemake/__init__.py", line 608, in snakemake
    snakefile, overwrite_first_rule=True, print_compilation=print_compilation
  File "/home/runner/work/snakemake/snakemake/snakemake/workflow.py", line 1149, in include
    exec(compile(code, snakefile, "exec"), self.globals)
  File "/tmp/snakemake-pujn5s_2/Snakefile", line 3, in <module>
    min_version("6.1.1")
  File "/home/runner/work/snakemake/snakemake/snakemake/utils.py", line 472, in min_version
    raise WorkflowError("Expecting Snakemake version *** or higher.".format(version))
snakemake.exceptions.WorkflowError: Expecting Snakemake version 6.1.1 or higher.
```
but I suspect something else is going on! The test is test_module_complex -- I tried to reproduce locally and it worked so I haven't been able to. Let me know if you have any insights!

johanneskoester:
> @johanneskoester I'm seeing an error in a module that the snakemake version is too low:
> 
> ```python
> WorkflowError in line 3 of /tmp/snakemake-pujn5s_2/Snakefile:
> Expecting Snakemake version 6.1.1 or higher.
>   File "/tmp/snakemake-pujn5s_2/Snakefile", line 3, in <module>
> ------------------------------ Captured log call -------------------------------
> DEBUG    snakemake.logging:logging.py:553 Full Traceback (most recent call last):
>   File "/home/runner/work/snakemake/snakemake/snakemake/__init__.py", line 608, in snakemake
>     snakefile, overwrite_first_rule=True, print_compilation=print_compilation
>   File "/home/runner/work/snakemake/snakemake/snakemake/workflow.py", line 1149, in include
>     exec(compile(code, snakefile, "exec"), self.globals)
>   File "/tmp/snakemake-pujn5s_2/Snakefile", line 3, in <module>
>     min_version("6.1.1")
>   File "/home/runner/work/snakemake/snakemake/snakemake/utils.py", line 472, in min_version
>     raise WorkflowError("Expecting Snakemake version *** or higher.".format(version))
> snakemake.exceptions.WorkflowError: Expecting Snakemake version 6.1.1 or higher.
> ```
> 
> but I suspect something else is going on! The test is test_module_complex -- I tried to reproduce locally and it worked so I haven't been able to. Let me know if you have any insights!

Hi @vsoch, this was caused by some changes in the CI and our release process. Should work again now if you update with the latest upstream main branch.

vsoch:
@johanneskoester I don't quite understand the title requirements (I tried a topic in brackets and it didn't work) so I'll need a little help here!

So the delay on this PR was getting support for views in spack to be copy/relocated. And that was merged today, but unfortunately we have a new issue that when we relocate, the paths can't go over a certain length:

```bash
==> py-pysam: Successfully installed py-pysam-0.15.3-6ckec7lgzidwysrpnnqgaf4upbaj46wp
  Fetch: 1.11s.  Build: 3m 29.10s.  Total: 3m 30.21s.
[+] /home/vanessa/Desktop/Code/vsoch-spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/py-pysam-0.15.3-6ckec7lgzidwysrpnnqgaf4upbaj46wp
==> Error: New path longer than old path: binary text replacement not possible.
The new path b'/home/vanessa/Desktop/Code/snakemake/tutorial/.snakemake/spack/827d39341264ddaff183ebb496b21eb2/._snakemake_view/dnlpwufag4l4nd5hkjksnior4l5p3pfo' is longer than the old path b'/home/vanessa/Desktop/Code/vsoch-spack/opt/spack/linux-ubuntu20.04-skylake/gcc-9.3.0/libxml2-2.9.10-mpl7mrhnmayionbep5ipqgu23shu3xvh'.
Text replacement in binaries will not work.
Create buildcache from an install path longer than new path.
==> Updating view at /home/vanessa/Desktop/Code/snakemake/tutorial/.snakemake/spack/827d39341264ddaff183ebb496b21eb2/snakemake_view
```
And given the nested .snakemake metadata folder, various levels of spack and the hash, plus the spack installs, and the user likely running this from some nested location on HPC, I don't think it's likely the copy/relocate will work. So what we can't do is create and then easily archive an entire environment for later use. But we still can use spack as the base package manager for running workflows! I've updated this with current main and instructions for how to use it with spack can be found in this tutorial: https://github.com/snakemake/snakemake-spack-tutorial.

@tgamblin @becker33 do you have any clever ideas for the path issue? Even if we can't fix it, I think it's still useful to have spack here, of course it's up to @johanneskoester ultimately. I'm thinking it would be cool to chat with some code teams at the labs that are trying to use spack, and see if snakemake would be useful to them.

tgamblin:
@vsoch: the padded path issue is likely that the environment isn't built with padding, so you can't relocate it to where you want it to be.  If you want to get around that you need to build the env with long paths -- so that we can relocate long (fake) paths to short ones.  That's covered in the mirror tutorial a little bit here: https://spack-tutorial.readthedocs.io/en/latest/tutorial_binary_cache.html.

Basically you need to do the builds and create the binaries with a config option set:

```
$ spack config add "config:install_tree:padded_length:128"
```

Or in YAML:
```yaml
config:
    install_tree:
        padded_length: 128
```

vsoch:
oh neat! I'll try that.

tgamblin:
I haven't dug into how things are configured here but right now, building relocatable binaries with the long path makes the workflow for building a custom environment different from the workflow for installing it again from buildcaches.  i.e., to build binaries that are relocatable like this you have to use the padded path option.  But when you actually install and run your workflow you likely do not want it.  So there may be some lack of transparency there for the first run vs. subsequent runs.  We could probably fix that on the Spack side somehow.

vsoch:
Oh hmm. I think the basic functionality we want is to install (and immediately use) the view, and then to be able to compress it into an archive that could be splot out on a new filesystem for usage (I wasn't thinking of using a build cache proper, just compressing it up) and it sounds like we'd need some kind of custom spack work to add the new path? I guess there isn't any concept of relative paths for a view install? So you are thinking spack wouldn't support what we need yet (or there is another way?)

tgamblin:
If you can guarantee that the workflow runs at the same length or shorter path as the original build, then you can just reinstall it each time. You only need to build in a long path if you expect to have to install to longer paths. So maybe that can be fixed on the snakemake side?

tgamblin:
Oh, and on the view specifically: I *believe* you should be able to reuse it.  At the very least you can reuse it at the same location. If we’re not making the RPATHs relative when we create it, I think we should do that. For shebangs and strings in files though I’m not sure we can guarantee that a relative path would work, so we may need something to relocate the view again. Why aren’t buildcaches good enough though? They’re already relocatable.

vsoch:
@tgamblin I tried a padded length, first 128 and then double that, and I always hit that it's too long.

```
  Fetch: 0.02s.  Build: 3m 36.70s.  Total: 3m 36.72s.
[+] /home/vanessa/Desktop/Code/vsoch-spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_plac/linux-ubuntu20.04-skylake/gcc-9.3.0/py-pysam-0.15.3-6ckec7lgzidwysrpnnqgaf4upbaj46wp
==> Error: New path longer than old path: binary text replacement not possible.
The new path b'/home/vanessa/Desktop/Code/snakemake/tutorial/.snakemake/spack/827d39341264ddaff183ebb496b21eb2/._snakemake_view/cv5j74eolon37t2booswtyubcpwslwtw' is longer than the old path b'/home/vanessa/Desktop/Code/vsoch-spack'.
Text replacement in binaries will not work.
Create buildcache from an install path longer than new path.
==> Updating view at /home/vanessa/Desktop/Code/snakemake/tutorial/.snakemake/spack/827d39341264ddaff183ebb496b21eb2/snakemake_view
```
I don't think this is going to work, unless I need to do this somehow differently? We can't really control where / how a user installs spack and how long the original path is. I sort of wish it would just install to the view and now worry about changing a path from the original.

johanneskoester:
I've just seen that you already have moved that code into your base class, so there should be no such issue here when moving workflows, right?

johanneskoester:
As for the title: I've fixed it. There should be no brackets around the keyword, and the keyword should be one of those allowed in conventional commits. Here, it is a feature addition, encoded by `feat:`in the title.

vsoch:
> I've just seen that you already have moved that code into your base class, so there should be no such issue here when moving workflows, right?

What do you mean when moving workflows?

johanneskoester:
> > I've just seen that you already have moved that code into your base class, so there should be no such issue here when moving workflows, right?
> 
> What do you mean when moving workflows?

With Conda, when an environment is installed, some binaries have to be patched to contain paths to linked libraries. I don't know the details, but sometimes these have to be absolute paths. This is why Conda environments are (after they have been installed) not movable. I interpreted your discussion above such that Spack has the same issue. Maybe I am wrong though.

vsoch:
> With Conda, when an environment is installed, some binaries have to be patched to contain paths to linked libraries. I don't know the details, but sometimes these have to be absolute paths. This is why Conda environments are (after they have been installed) not movable. I interpreted your discussion above such that Spack has the same issue. Maybe I am wrong though.

Yes I think that's spot on - the thing we create (and then would need to save) is moveable, but only given that it's path winds up to be shorter than the original one we fill in to replace. I think there are other tricks we could try, but it should be in a round 2 of work. So to start, I think we should just not allow an archive to be created, and we can further work on this if/when users start to request it (and use it). Is the easiest thing to do to remove move of the archive function logic in spack.py and just print an error message that this isn't supported yet?

johanneskoester:
> > With Conda, when an environment is installed, some binaries have to be patched to contain paths to linked libraries. I don't know the details, but sometimes these have to be absolute paths. This is why Conda environments are (after they have been installed) not movable. I interpreted your discussion above such that Spack has the same issue. Maybe I am wrong though.
> 
> Yes I think that's spot on - the thing we create (and then would need to save) is moveable, but only given that it's path winds up to be shorter than the original one we fill in to replace. I think there are other tricks we could try, but it should be in a round 2 of work. So to start, I think we should just not allow an archive to be created, and we can further work on this if/when users start to request it (and use it). Is the easiest thing to do to remove move of the archive function logic in spack.py and just print an error message that this isn't supported yet?

Sounds ok to me.

tgamblin:
> With Conda, when an environment is installed, some binaries have to be patched to contain paths to linked libraries. I don't know the details, but sometimes these have to be absolute paths. This is why Conda environments are (after they have been installed) not movable.

This is the same for Spack -- Spack binaries, once installed, shouldn't be moved.  Relocation *is* done during the installation, so if you store build caches (binary packages) you can reinstall them each time, and they'll work in the new locations.

I am not sure that this maps to the concept of a cache in snakemake -- basically for a Spack build you likely want a workflow like:

1. Build from source initially (unless there are already public binaries available somewhere)
2. Cache the binary packages from the source build
3. Install the binary packages.
4. Cache the final installation someplace
5. Reuse the cached installation in place

With conda, all packages are binaries, so I think you're really only doing 3-5.  With Spack, you can avoid doing (1) every time you need to move the installation by saving the binary packages from (2) and reinstalling them instead of building from source.  I think the difference is that with Conda, (2) is always some portable binaries in a public channel.  You don't need to save the relocatable binary packages anywhere b/c you did not build them.

> I think we should just not allow an archive to be created, and we can further work on this if/when users start to request it (and use it).

I think this is good. It means that Spack would build from source, cache an installation until it can't be used anymore, then build from source again.  That's probably good for most people.

If you want to avoid the lengthy initial source build after that you could add buildcaches into the mix.

vsoch:
@tgamblin yep I've updated to just do the install. The concept doesn't map cleanly to snakemake, because in these steps:

1. Build from source initially (unless there are already public binaries available somewhere)

We'd stop after that. These are the installs that we would quickly want to do, and then use in a workflow - not need to then create a cache and install again. Perhaps this could be optimized in the future if we are just able to install from publicly available binaries.

tgamblin:
Snakemake seems pretty general so would be interested in whether it's possible to map more complex installation workflows (as above) to its (extremely general) workflow system :).

vsoch:
To have the snakemake DAG used to perform the installs of the packages? Yeah that's a cool idea, I guess we would create a subworkflow sort of thing with spack. But that would mean we'd need to dump out some representation of the dependency graph for snakemake to use, so we don't try to install something that needs something not built yet. It's a cool idea, but shouldn't spack already be optimized to install things in parallel and take this into account (I'm pretty sure that it does). What would be the benefit of dumping out spack metadata, and handing it to snakemake to do? My computer already freezes when installing a lot of spack packages, lol.

tgamblin:
In `spack ci` we already dump out the dependencies as separate jobs for GitLab CI, and the dependencies are there. So the benefit would be that Snakemake could actually build all the spack packages in parallel.  It'd take way less time to build from source, and you could save binaries from the builds to then use in the final environment.

The other benefit of that would be that if you tweaked the env slightly, you could reuse any packages that didn't change from the first to second iteration.  They'd still be there as binaries.

vsoch:
> So the benefit would be that Snakemake could actually build all the spack packages in parallel. It'd take way less time to build from source, and you could save binaries from the builds to then use in the final environment.

Wait, spack doesn't do that?

I do like the idea, I think it would just mean producing a Snakefile from that spack command instead of a gitlab-ci. But I think that's separate from the PR here.

tgamblin:
> Wait, spack doesn't do that?

Spack builds individual packages in parallel (e.g., `make -j16`) but does not build multiple packages at once unless you run multiple spacks.  So, with SLURM, `srun -N 4 -n 32 spack install -j8 pkg` will get you 8 package builds per node across 4 nodes, and each will run 8-way parallel `make`.  That relies on a shared filesystem, though, whereas our GitLab CI integration does not -- it passes dependencies around as artifacts so you could run each build on a completely separate system.

So, yep, the goal would be to set up a snakemake workflow where each node in the workflow could install its dependencies, build a package in parallel, and save its binary to be used as a dependency by future nodes -- like GitLab CI does.

Agree it's not for this PR.

vsoch:
>  That relies on a shared filesystem, though, whereas our GitLab CI integration does not -- it passes dependencies around as artifacts so you could run each build on a completely separate system. I think this is more like snakemake.

Yes that's correct - at least for executors like GLS it's totally separate.

vsoch:
@johanneskoester what else would you like to see here? I like how the structure of deployments is more modular, although that's not the case around the other files (e.g., workflow.py). Should we try to generalize more?

vsoch:
@johanneskoester what else would you like to see here? I like how the structure of deployments is more modular, although that's not the case around the other files (e.g., workflow.py). Should we try to generalize more?

johanneskoester:
> @johanneskoester what else would you like to see here? I like how the structure of deployments is more modular, although that's not the case around the other files (e.g., workflow.py). Should we try to generalize more?

Sorry for the late response. Can you fix the issues raised by sonarcloud? See the respective sonarcloud comment above.

vsoch:
@johanneskoester I attempted to resolve conflicts in the UI - if there are formatting issues I'll need to pull on the command line. I clicked the sonar cloud issues to see, but they had expired I think and we need the re-run.

vsoch:
@johanneskoester 2/3 "security hotspots" are the current implementation hash methods for environments, using md5 hashes. 

image

https://github.com/snakemake/snakemake/blob/4816a58653e466ca94b1482a1d947a856f5381b3/snakemake/deployment/conda.py#L85

and the last (low) is the current implementation, but it's just moved to another file:

image

Are you saying you want me to use a different hash method and change the current? I think md5 is appropriate here because it's just for local user environments, and making the path longer wouldn't be desirable just to address sonarcloud. Let me know how you'd like me to address these!

vsoch:
Ah I found the problem - introduced a bug - will push a fix.

vsoch:
@johanneskoester I'll need to ask for your help on this new one - I'm not familiar with this source cache?
```
WARNING  snakemake.logging:logging.py:545 Building DAG of jobs...
DEBUG    snakemake.logging:logging.py:551 Full Traceback (most recent call last):
  File "/home/runner/work/snakemake/snakemake/snakemake/__init__.py", line 804, in snakemake
    keepmetadata=keep_metadata,
  File "/home/runner/work/snakemake/snakemake/snakemake/workflow.py", line 969, in execute
    dag.create_spack_envs(dryrun=dryrun)
  File "/home/runner/work/snakemake/snakemake/snakemake/dag.py", line 327, in create_spack_envs
    env.create(dryrun)
  File "/home/runner/work/snakemake/snakemake/snakemake/deployment/spack.py", line 123, in create
    env_hash = self.hash
  File "/home/runner/work/snakemake/snakemake/snakemake/deployment/spack.py", line 72, in hash
    md5hash.update(self.content)
  File "/home/runner/work/snakemake/snakemake/snakemake/deployment/base.py", line 48, in content
    self._content = self._get_content()
  File "/home/runner/work/snakemake/snakemake/snakemake/deployment/base.py", line 52, in _get_content
    return self.workflow.sourcecache.open(self.file, "rb").read()
  File "/home/runner/work/snakemake/snakemake/snakemake/sourcecache.py", line 294, in open
    cache_entry = self._cache(source_file)
  File "/home/runner/work/snakemake/snakemake/snakemake/sourcecache.py", line 320, in _cache
    cache_entry = self._cache_entry(source_file)
  File "/home/runner/work/snakemake/snakemake/snakemake/sourcecache.py", line 309, in _cache_entry
    urihash = source_file.get_uri_hash()
AttributeError: '_IOFile' object has no attribute 'get_uri_hash'

DEBUG    snakemake.logging:logging.py:551 unlocking
DEBUG    snakemake.logging:logging.py:551 removing lock
DEBUG    snakemake.logging:logging.py:551 removing lock
DEBUG    snakemake.logging:logging.py:551 removed all locks
=============================== warnings summary ===============================
```

johanneskoester:
Also there are some conflicts now (really sorry about being so late back here).

> @johanneskoester I'll need to ask for your help on this new one - I'm not familiar with this source cache?
> 
> ```
> WARNING  snakemake.logging:logging.py:545 Building DAG of jobs...
> DEBUG    snakemake.logging:logging.py:551 Full Traceback (most recent call last):
>   File "/home/runner/work/snakemake/snakemake/snakemake/__init__.py", line 804, in snakemake
>     keepmetadata=keep_metadata,
>   File "/home/runner/work/snakemake/snakemake/snakemake/workflow.py", line 969, in execute
>     dag.create_spack_envs(dryrun=dryrun)
>   File "/home/runner/work/snakemake/snakemake/snakemake/dag.py", line 327, in create_spack_envs
>     env.create(dryrun)
>   File "/home/runner/work/snakemake/snakemake/snakemake/deployment/spack.py", line 123, in create
>     env_hash = self.hash
>   File "/home/runner/work/snakemake/snakemake/snakemake/deployment/spack.py", line 72, in hash
>     md5hash.update(self.content)
>   File "/home/runner/work/snakemake/snakemake/snakemake/deployment/base.py", line 48, in content
>     self._content = self._get_content()
>   File "/home/runner/work/snakemake/snakemake/snakemake/deployment/base.py", line 52, in _get_content
>     return self.workflow.sourcecache.open(self.file, "rb").read()
>   File "/home/runner/work/snakemake/snakemake/snakemake/sourcecache.py", line 294, in open
>     cache_entry = self._cache(source_file)
>   File "/home/runner/work/snakemake/snakemake/snakemake/sourcecache.py", line 320, in _cache
>     cache_entry = self._cache_entry(source_file)
>   File "/home/runner/work/snakemake/snakemake/snakemake/sourcecache.py", line 309, in _cache_entry
>     urihash = source_file.get_uri_hash()
> AttributeError: '_IOFile' object has no attribute 'get_uri_hash'
> 
> DEBUG    snakemake.logging:logging.py:551 unlocking
> DEBUG    snakemake.logging:logging.py:551 removing lock
> DEBUG    snakemake.logging:logging.py:551 removing lock
> DEBUG    snakemake.logging:logging.py:551 removed all locks
> =============================== warnings summary ===============================
> ```

self.file now needs to be a SourceFile object, see here in the conda specific code:
https://github.com/snakemake/snakemake/blob/main/snakemake/deployment/conda.py#L49

johanneskoester:
> @johanneskoester 2/3 "security hotspots" are the current implementation hash methods for environments, using md5 hashes.
> 
> image
> 
> https://github.com/snakemake/snakemake/blob/4816a58653e466ca94b1482a1d947a856f5381b3/snakemake/deployment/conda.py#L85
> 
> and the last (low) is the current implementation, but it's just moved to another file:
> 
> image
> 
> Are you saying you want me to use a different hash method and change the current? I think md5 is appropriate here because it's just for local user environments, and making the path longer wouldn't be desirable just to address sonarcloud. Let me know how you'd like me to address these!

Ok, good point. Indeed I would like to keep the md5 hashes. I will mark the corresponding errors in sonarcloud as false positives once this is ready for merging.

vsoch:
@johanneskoester all tests are working (minus the sonarcloud) but I want to update the spack test to install a much smaller, mock package - right now just installing jq with spack takes a long time (and it timed out once) and we don't want that for production snakemake pipeline tests!

vsoch:
Ok should be good 👍

johanneskoester:
There are still 12 code smells (see the sonarcloud report) :-).

vsoch:
I'm giving a shot at the last three "cognitive complexity" ones - just taking a bit longer!

vsoch:
okay, no more code smells, and only two security hotspots! As long as my fixes didn't break anything, I (fingers crossed) hope we are good to continue now!

snakemake/deployment/spack.py
```diff
@@ -0,0 +1,232 @@
...
+            md5hash = hashlib.md5()
+            # Include the absolute path of the target env dir into the hash.
+            # By this, moving the working directory around automatically
+            # invalidates all environments. This is necessary, because binaries
+            # in spack environments can contain hardcoded absolute RPATHs.
+            env_dir = os.path.realpath(self._env_dir)
+            md5hash.update(env_dir.encode())
+            md5hash.update(self.content)
+            self._hash = md5hash.hexdigest()
+        return self._hash
```
johanneskoester, line 77: This looks very similar to the conda hash code. Can it be moved into the base class instead of having it in both? Please check all other methods here as well in case I missed something.

snakemake/deployment/spack.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        Create self-contained archive of environment.
+
+        This won't be possible with spack until an environment can be created
+        that uses a copy instead of link (now possible) but ALSO can be
+        relocated. Currently, the spack.yaml files are the best reproducers.
+        """
+        raise WorkflowError("spack does not support environment archive.")
+
+    def create(self, dryrun=False):
```
johanneskoester, line 106: Perhaps this method can share code with the conda implementation as well?

snakemake/deployment/spack.py
```diff
@@ -0,0 +1,232 @@
...
+
+        return env_path
+
+    def __eq__(self, other):
+        if isinstance(other, Env):
+            return self.file == other.file
+        return False
+
+
+class Spack:
```
johanneskoester, line 204: I think this class can also share a base class with the conda implementation.

snakemake/jobs.py
```diff
@@ -339,12 +344,43 @@ def conda_env(self):
     def conda_env_path(self):
         return self.conda_env.path if self.conda_env else None
 
+    @property
+    def spack_env_file(self):
```
johanneskoester, line 348: This looks quite similar to conda_env_file. In total there still seems to be a lot of duplication between the two, sorry that I missed this before. Maybe have some more object oriented solutions for these?

johanneskoester (changes requested):

snakemake/deployment/spack.py
vsoch: I originally left these separate because conda has a container_img to account for (so they are different) but I can add the same variable (empty) to spack and then have some shared code to run.

snakemake/deployment/spack.py
vsoch: Sure I can give a shot - I didn't originally because there are enough differences that trying to separate out into shared code would increase complexity for the developer by quite a bit to jump back and forth.

snakemake/deployment/spack.py
vsoch: oh definitely - I'm seeing a lot of duplication that (assuming I give the ability for the base to have a container and then spack doesn't support that) can simplify the code quite a bit - working now will update soon!

snakemake/jobs.py
vsoch: Yeah - the reason is because it was named "conda_env_file" and used in jobs.py and dag.py. I was worried about renaming the variable and perhaps breaking someone's code that uses snakemake programmatically. But I definitely can rename to make it shared because the classes if that isn't an issue.

vsoch: I'm going to leave this for now - my concern is that this isn't an attribute on the environment class - it's part of the job, and (hypothetically speaking) there is no reason that a job couldn't at some point use two different environments. So maybe we can discuss this further and I'll open a follow up PR if we do a good design - I did a quick grep and it's not used in that many places (but there is the duplication).

I made a LOT of changes and the code is really much better and clean - will push soon to see what I broke! :laughing:

vsoch:
ug, had it working, and the last bit of refactor seemed to break everything:
```
 partially initialized module 'io' has no attribute 'DEFAULT_BUFFER_SIZE'
```
so now I have to start from scratch and see where this happened :sob:

vsoch:
oh my code is smelly again. Why am I always smelly.

vsoch:
okay that should be good - security hotspots are those that were there previously (and just re-located and appearing again). One is the subprocess check command, and the other is for the md5 hash.

snakemake/jobs.py
johanneskoester: So, what we can do however to avoid the redundancy in `Job.*_env_file`and `DAG.create_*_envs` and also `Rule.expand_*_env` is to instead introduce in `Job._env_files: dict`, `Job._envs: dict` and `Rule._env_files: dict`.
Then, all the methods could just be called `expand_env`,`env_file`, and `create_envs` and take a keyword argument `env_type: str=None` which can for now take the values `"spack"` and `"conda"`. This way, there is only one implementation for all these methods. Of course, the implementation can have special cases for certain env types, like the singularity wrapping in the conda case.

vsoch: Gotcha, I'll try this!

vsoch: What about in rules.py when they are properties / have custom setters?

```
@property
    def conda_env(self):
        return self._conda_env

    @conda_env.setter
    def conda_env(self, conda_env):
        self._conda_env = IOFile(conda_env, rule=self)

    @property
    def spack_env(self):
        return self._spack_env

    @spack_env.setter
    def spack_env(self, spack_env):
        self._spack_env = IOFile(spack_env, rule=self)
```

vsoch: I'm going to try:
```
    def env(self, env_type):
        return self._envs.get(env_type)

    @env.setter
    def env(self, env, env_type):
        self._envs[env_type] = IOFile(env, rule=self)
```

vsoch:
We can reconsider this if/when spack has a reliable binary cache - from personal experience I think most users in the bio community are better served using conda. Was fun to try though!

johanneskoester:
@Marco-Masera see this old PR.

Issue #1007, opened 2021-05-21
Environment variables with spaces break execution on clusters when using jobscript


pcm32:
**Snakemake version**

6.2.1

**Describe the bug**

When the  workflow makes use of environment variables (by declaring them in envvars and then using them through os.environ), the workflow execution fails on batch schedulers because the jobscript gets the variable split by the space (it needs quoting).

**Logs**

Look at COVARIATE below

```
/nfs/my-path/.snakemake/tmp.fjbaxuov/snakejob.retrieve_data.3.sh: line 3: CACHE_PATH=/my-path/cache ACCESSIONS=E-MTAB-4395,E-MTAB-4342,E-MTAB-4128,E-MTAB-3826,E-MTAB-3173,E-MTAB-964,E-GEOD-62778,E-GEOD-54272 BATCH=study NEW_ACCESSION=E-CORN-1 COVARIATE=organism part COVARIATE_TYPE=characteristic: No such file or directory
```

**Minimal example**

Run any workflow that uses an env variable that has a space in the value in a cluster environment. On a local execution (no cluster), there i s not issue.

**Additional context**

I suspect that the issue is here: https://github.com/snakemake/snakemake/blob/main/snakemake/executors/__init__.py#L795
although changing that on my side doesn't resolve the problem entirely, but I'm sure that the unquoted values are not helping.

PR #1008, opened 2021-05-21
Protects against spaces in env vars and avoid quoting of all env vars as single entity

pcm32:
Addresses #1007


johanneskoester (changes requested):

johanneskoester:
Closing in favor of a global solution in #1491. Nevertheless, thanks a lot for taking a first stab at this!

PR #1015, opened 2021-05-26
feat: SLURM executor

johanneskoester:
### Description

This PR by @cmeesters adds a dedicated backend for slurm.

### QC


snakemake/resources.py
```diff
@@ -1,5 +1,18 @@
 import re
 
+supported_keys = {
```
jdblischak, line 3: What about "quality of service" (`--qos`)?

snakemake/__init__.py
```diff
@@ -2376,6 +2408,11 @@ def adjust_path(f):
                     file=sys.stderr,
                 )
                 sys.exit(1)
+    if args.slurm:
+        args.use_envmodules = True
+        if args.cores is None:
+            args.cores = 1  # TODO: check - is this ok?
+
```
johanneskoester, line 2415:
```suggestion
```

Not needed altogether:
1. I would not imply envmodules on slurm. There are certainly people who don't use them. Such defaults are a job for a corresponding profile.
2. Cores should not be set here, Snakemake has logic for handling cores = None at a later stage.

snakemake/executors/slurm/slurm_jobstep.py
```diff
@@ -0,0 +1,117 @@
...
+            cluster_config=cluster_config,
+            local_input=local_input,
+            restart_times=restart_times,
+            exec_job=exec_job,
+            assume_shared_fs=False,
+            max_status_checks_per_second=max_status_checks_per_second,
+        )
+        self.mem_per_node = os.getenv("SLURM_MEM_PER_NODE")
+        self.cpus_on_node = os.getenv("SLURM_CPUS_ON_NODE")
+        self.jobid = os.getenv("SLURM_JOB_ID")
```
johanneskoester, line 62: Are these three set internally when invoking the slurm jobstep? Or are those slurm specific env variables that you can always expect? The answer to these questions should be put in here as a comment, for documentation purposes.

snakemake/executors/slurm/slurm_jobstep.py
```diff
@@ -0,0 +1,117 @@
...
+        super()._run(job)
+
+        jobsteps = dict()
+
+        if job.is_group():
+
+            def get_call(level_job, level_id, aux=""):
+                # we need this calculation, because of srun's greediness and
+                # SLURM's limits: it is not able to limit the memory if not per cpu
+                mem_per_cpu = level_job.resources.mem_mb // level_job.threads
```
johanneskoester, line 79: Is this allowed to be zero? If not, consider using `max(..., the_theoretical_minimum)`.

snakemake/executors/slurm/slurm_jobstep.py
```diff
@@ -0,0 +1,117 @@
...
+                    "--exclusive -n 1 {aux} {exec_job}".format(
+                        level_id=level_id,
+                        level_job=level_job,
+                        aux=aux,
+                        jobid=self.jobid,
+                        exec_job=self.format_job(level_job),
+                    )
+                )
+
+            for level in job.toposorted:
```
johanneskoester, line 92: This is quite cool, however it also means that we always wait until one level is completed, right? In principle, this can mean that resources are wasted.

johanneskoester: This entire thing runs on a single node, doesn't it?

johanneskoester: We definitely need to ensure that works with pipe producer and consumer jobs (job.toposorted should return them on the same level, but I think I have not yet ensured that).

johanneskoester, line 92:
```suggestion
            for level in job.toposorted: #TODO ensure that pipe jobs are returned on the same toposorted level.
```

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+        cache=False,
+        local_input=None,
+        restart_times=0,
+        exec_job=None,
+        max_status_checks_per_second=1,
+        cluster_config=None,
+    ):
+        # needs to be set in either case for the submission string
+        if not cores:
+            cores = 1
```
johanneskoester, line 56:
```suggestion
```

Why is this needed? No other executor does that.

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+            quiet=quiet,
+            printshellcmds=printshellcmds,
+            latency_wait=latency_wait,
+            cluster_config=cluster_config,
+            restart_times=restart_times,
+            assume_shared_fs=True,
+            max_status_checks_per_second=max_status_checks_per_second,
+        )
+        self.exec_job += " --slurm-jobstep"
+        print(self.exec_job)
```
johanneskoester, line 76:
```suggestion
```

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+        This method can be overwritten to submit many jobs in a more efficient way than one-by-one.
+
+        Note that in any case, for each job, the callback functions have to be called individually!
+        """
+        # if groupjob:
+        #    attributes of groubjobs (snakemake/jobs.py class GroupJob , _resources
+        #    - see topology
+
+        for job in jobs:
+            # TODO replace by own code
```
johanneskoester, line 113:
```suggestion
```

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+            self.run(
+                job,
+                callback=callback,  # to be executed upon ready jobs
+                submit_callback=submit_callback,  # to be hold for --immediate flag - ignored within this executor
+                error_callback=error_callback,  #
```
johanneskoester, line 118:
```suggestion
                error_callback=error_callback,  # for failed jobs
```

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+            path=path,
+            **kwargs,
+        )
+        try:
+            return format_p(pattern)
+        except KeyError as e:
+            raise WorkflowError(
+                "Error formatting jobscript: {} not found\n"
+                "Make sure that your custom jobscript is up to date.".format(e)
+            )
```
johanneskoester, line 165: This method should just be inherited from the superclass.
```suggestion
```

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+            envvars=envvars,
+            **kwargs,
+        )
+        content = self.format_job(self.jobscript, job, exec_job=exec_job, **kwargs)
+
+        logger.debug("Jobscript:\n{}".format(content))
+        with open(jobscript, "w") as f:
+            print(content, file=f)
+
+        os.chmod(jobscript, os.stat(jobscript).st_mode | stat.S_IXUSR | stat.S_IRUSR)
```
johanneskoester, line 195:
```suggestion
```

Let us just inherit this from the superclass.

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+            )
+            sys.exit(1)
+        if not job.resources.get('walltime_minutes'):
+            logger.warning("No wall time limit is set, setting 'walltime_minutes' to 1.")
+        call += " -t {walltime_minutes}".format(walltime_minutes = job.resources.get('walltime_minutes', default_value=1))
+        if job.resources.get("constraint"):
+            call += " -C {constraint}".format(**job.resources)
+        # TODO: implement when tempfs-resource is defined
+        # if job.resources.get()
+        if job.resources.get("mem_mb_per_cpu"):
```
johanneskoester, line 226: If mem_mb_per_cpu is really supported explicitly here, we also need to support it for local execution. Sorry, I feel like I asked this several times before (getting old I guess): why do we need it again? Couldn't you do the same computation as [here](https://github.com/snakemake/snakemake/pull/1015/files#diff-0039b8713fcb4087ecfa9fcf6462a997d117e39f12268f7a4694444c1725d8c6R79)?

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+        else:
+            logger.warning(
+                "No job memory information ('mem_mb' or 'mem_mb_per_cpu') is given - submitting without. This might or might not work on your cluster."
+            )
+
+        # if ressources == MPI:
+        if job.resources.get("mpi", False):
+            # submit MPI job
+            # mold submissing string, submit according to generic cluster exec
+            pass
```
johanneskoester, line 239: Maybe raise UnimplementedError here?

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+    def get_jobscript(self, job):
+        f = job.format_wildcards(self.jobname, cluster=self.cluster_wildcards(job))
+
+        if os.path.sep in f:
+            raise WorkflowError(
+                "Path separator ({}) found in job name {}. "
+                "This is not supported.".format(os.path.sep, f)
+            )
+
+        return os.path.join(self.tmpdir, f)
```
johanneskoester, line 136:
```suggestion
```

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+                callback=callback,  # to be executed upon ready jobs
+                submit_callback=submit_callback,  # to be hold for --immediate flag - ignored within this executor
+                error_callback=error_callback,  #
+            )
+
+    @property
+    def tmpdir(self):
+        if self._tmpdir is None:
+            self._tmpdir = tempfile.mkdtemp(dir=".snakemake", prefix="tmp.")
+        return os.path.abspath(self._tmpdir)
```
johanneskoester, line 125:
```suggestion
```

inherit

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+        )
+        self.exec_job += " --slurm-jobstep"
+        print(self.exec_job)
+
+    def shutdown(self):
+        """
+        waiting for all jobs to be ready
+        """
+        super().shutdown()
+        # perform additional steps on shutdown if necessary
```
johanneskoester, line 83:
```suggestion
```

inherit

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+                call += " -n 1 -c 1"
+            else:
+                call += " -n 1 -c {threads}".format(**job.resources)
+        else:  # job.is_group:
+            pass
+
+        # as we cannot do 'sbatch ... cd {workflow.workdir_init} && python ...' as snakemake otherwise expects
+        call += " --chdir={workdir_init}".format(
+            workdir_init=self.workflow.workdir_init
+        )
```
johanneskoester, line 255: The `cd ...` happens inside of the jobscript. Hence, this should not be an issue I think, and this additional code could be omitted?

johanneskoester (commented):
Great work so far! Here are my comments.

snakemake/resources.py
cmeesters: Hi,

Please note, this is a PR Johannes opened because I was away and he wanted to comment in a systematic way. Now, that I can proceed testing (and develop further, which all will take some time), quite a few things need to be cleaned.

As to your question: No, the idea is not to mimic all sbatch options. Rather the most common ones to also catch errors and to have a catch-all option for the rest. In that sense, 'constraint' is also a doubtful candidate as I have to admit. Both, 'constraint' and 'qos' are handled differently from plugin to plugin, from cluster to cluster. This is a bit unfortunate, as is may have substantial impact on throughput. In a version 2 of this feature, we need to think about pre-rule injections, such that this can be handled without altering each workflow.

Just my .02€

snakemake/__init__.py
```diff
@@ -1923,6 +1927,32 @@ def get_argument_parser(profile=None):
...
 
+    group_slurm = parser.add_argument_group("SLURM")
+    slurm_mode_group = group_slurm.add_mutually_exclusive_group()
+
+    slurm_mode_group.add_argument(
+        "--slurm",
+        action="store_true",
+        help=(
+            "Execute snakemake rules as SLURM batch jobs according"
+            " to their 'resources' definition. SLRUM resources as "
```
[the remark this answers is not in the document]
cmeesters: and the difference is?

snakemake/__init__.py
```diff
@@ -1923,6 +1927,32 @@ def get_argument_parser(profile=None):
...
+            " 'partition', 'ntasks', 'cpus', etc. need to be defined"
+            " per rule within the 'resources' definition. Note, that"
+            " memory can only be defined as 'mem_mb' or 'mem_mb_per_cpu'"
+            " as analoguous to the SLURM 'mem' and 'mem-per-cpu' flags"
+            " to sbatch, respectively. Here, the unit is always 'MiB'."
```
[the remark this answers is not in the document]
cmeesters: hm, I will do so, but first, please answer/comment this, first: All RAM is bought in MiB, storage sold in MB (or bigger scales). Should a workflow system which intends to allocate RAM not always handle the binary prefix?

snakemake/__init__.py
cmeesters: ad 1: Yes, admitted. On smaller cluster that will be true.
ad 2: OK.

snakemake/executors/slurm/slurm_jobstep.py
cmeesters: No, they are set by SLURM right after the prolog step. Hence, they are indeed always to expect. I will comment accordingly.

snakemake/executors/slurm/slurm_submit.py
cmeesters: Is it?

snakemake/executors/slurm/slurm_submit.py
cmeesters: indeed, here it is obvious

snakemake/executors/slurm/slurm_jobstep.py
johanneskoester: I have now ensured that in PR #1044.

snakemake/resources.py
jdblischak: To make sure I understand: using the built-in Slurm executor implemented in this PR, I wouldn't be able to specify a different value for `--qos` for different rules. Since `qos` is not one of the supported keys, the rules below would trigger an error:

```
rule rule1:
    ...
    resources:
        qos="qos1",
    ...

rule rule2:
    ...
    resources:
        qos="qos2",
    ...

```

johanneskoester: Mhm, I don't see a reason why one would not want to allow that, in particular since it is now possible to define rule resources at the command line (so that the workflow does not get polluted with platform specific stuff). @cmeesters can you explain?

snakemake/__init__.py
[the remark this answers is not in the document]
johanneskoester: I fixed a typo here.

snakemake/__init__.py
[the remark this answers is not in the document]
johanneskoester: MB is more widely known than MiB, hence, defining mem_mb instead of mem_mib in rule resources is more readable to non-technical people.

snakemake/executors/slurm/slurm_submit.py
johanneskoester: Well yes, the superclass has exactly the same method implementation, right?

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,349 @@
...
+            pass
+        # ordinary smp application:
+        elif not job.resources.get("mpi") and not job.is_group():
+            # TODO: this line will become longer
+            # TODO: hence the single command, yet
+            if not job.resources.get("threads"):
+                logger.warning("No value for 'threads' is set, assuming 1.")
+                call += " -n 1 -c 1"
+            else:
+                call += " -n 1 -c {threads}".format(**job.resources)
```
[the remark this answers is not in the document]
johanneskoester: Ok.

snakemake/executors/slurm/slurm_submit.py
johanneskoester: Not sure what files you mean, you'll have to explain further. With the request here I just meant to say that this method is present in the superclass and hence does not need to be redefined here.

snakemake/__init__.py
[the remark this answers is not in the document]
cmeesters: yep, let me re-phrase: Should snakemake not do the conversion _always_, like "dear user, you / we write MB, but really mean MiB"? For, the upper limit of a system, not withstanding OS and service daemons requirements, cannot be reached when submitting with MB. If we do the conversion for this reason in the SLURM executor only, then this is an exception to a rule, which applies everywhere - albeit less explicitly.

Yet, I do understand hesitation at this point, as this would mean a global change to snakemakes established behaviour.

snakemake/__init__.py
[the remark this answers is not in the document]
cmeesters: got it (outch)

snakemake/resources.py
jdblischak: I assume you are doing something fancy with the resources listed in `supported_keys`, i.e. trying to manage the requested resources with the total amount available. I think that's fine to limit this feature to ones that make sense to manage. But I don't see why you can't pass arbitrary resources directory to `sbatch`. If I write a rule definition like:

```
rule fruit:
    resources:
        mem_mb=4000,
        banana="apple",
```

Then this would result in `sbatch --mem=4000 --banana=apple`. Obviously this will cause sbatch to error, but that is on the end user. I don't think Snakemake needs to monitor all ~80 slurm options. You don't want to have to update Snakemake every time Slurm adds a new option.

johanneskoester: Sounds reasonable to me. @cmeesters thoughts?

snakemake/__init__.py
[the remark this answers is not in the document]
johanneskoester: Good point, but then maybe just let us convert automatically. Whenever the `mem_mb` resource is defined, the job automatically adds a `mem_mib` resource with the appropriate conversion.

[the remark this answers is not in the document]
johanneskoester: I will take care of that in a separate PR.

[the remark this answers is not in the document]
johanneskoester: Does that make sense?

snakemake/resources.py
cmeesters: My ideal resource support would:

- allow to pass every SLURM argument, like proposed by @jdblischak .
- have standard ones, which are validated and checked for common typos.
- have standard ones, which have snakemake resemblence and allow for a snakemake->SLURM conversion. E.g. in an smp-case from cpus to cpus-per-task. 

In a first step, we could just drop the `supported_keys` alltogether and follow @jdblischak suggestion to an extreme and later add validators/translators. So, yes that is a jolly good point.

mbhall88: In my fork of the slurm profile I have documented a similar approach to dealing with resources and adding a convenience parser for time as well https://github.com/mbhall88/slurm#rule-specific-resource-configuration. Not sure if this is helpful or not, but thought I'd drop it here in case it helps

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,276 @@
...
+        except KeyError as e:
+            # TODO: make explicit message for account / partition
+            logger.error(
+                "Missing job submission key '{}' for job '{}'.".format(
+                    e.args[0], job.name
+                )
+            )
+            sys.exit(1)
+
+        if not job.resources.get("walltime_minutes"):
```
pvandyken, line 143: For what it's worth, #1218 introduces `--runtime` as a special resource indicating the "walltime" for the job. `--runtime` is also explicitly mentioned in the existing docs on cluster (although even there it's actually a bit inconsistent between `runtime` and `runtime_min`). Not saying we have to use `runtime`, but it would be good to have a standardized term before these things get pushed

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,276 @@
...
+            # TODO: this line will become longer
+            # TODO: hence the single command, yet
+            if job.threads == 0:
+                call += f" -n 1 -c 1 {exec_job}"
+            else:
+                call += f" -n 1 -c {job.threads}"
+        else:
+            ntasks = max(map(len, job.toposorted))
+            threads = sum(j.threads for j in job)
+            call += f" -n {ntasks} -c {threads}"
```
pvandyken, line 184: I'm likely confused here, but it looks like you're adding up threads across all jobs in a group job then supplying that as the number of cpus *per* task. So 50 jobs at one thread each would request 50 cores per task. Whereas I think you'd want 50 tasks with one core per task

snakemake/executors/slurm/slurm_jobstep.py
pvandyken: I wonder if it's possible to use `srun`'s `--dependency` to manage submission dependency, rather than toposort layers. My idea is you would iterate over jobs in toposort order, but keep track of their step id (`<jobid>.<step>` where step increments from 0...). If a job has dependencies according to the DAG, you would communicate that to slurm using the step ids we've been tracking. Then slurm would start each job as soon as each dependency finishes, preventing "one layer at a time" resource wastage.

snakemake/executors/slurm/slurm_submit.py
cmeesters: Yes, you are right: I think that `max()` instead of `sum()` is to be used, here - that would be a price to pay: less synchronization overhead (perhaps even less I/O), but some in-efficiency instead. I could live with that and changed the code accordingly.

However, this still needs testing! Thank you for spotting this glitch!

Due to a public holiday (and a family field trip) I will continue working on it next Monday.

cmeesters:
Hi,

Thanks, @johanneskoester for PR 2eb66fbc53e9e9bad2c5de43566ead81767d1922 , this indeed solved the issue of the lost attribute to the executor class.

However, I think there are still a couple of major issues:
- all test jobs in the github CI are failing because the sbatch command, which is issued by the executor is rejected - even if there are no further resource requirements (I tinkered with the Snakefile). The same rule on "my" cluster is submitted and carried out without issues. We therefore have an issue with the Ansible playbook. And even after reading the playbook I am not sure where the issue is.
- when executing the 2nd rule in `test_slurm_mpi` (the MPI rule) the environment is not loaded (the command is ` snakemake --use-envmodules --slurm -d . -j1` and the log states `Activating environment modules` for rule 1 (the dummy compile-my-code-rule), but not for the 2nd rule. The aim should be to load environment modules prior to each job execution, best _on_ the compute node in job context. That would require to issue the `module purge && module load` via the SlurmJobStepExecutor. As this is currently ultimately done via the shell-class and its `__new__`-method, how should the information be propagated to the executor and why is this even necessary, if the command is run in the first rule?
- after lots of printf-debugging I am confident that the `group` property in the group test is not recognized with `--slurm`. Not sure why.

Any ideas?

Cheers
Christian

johanneskoester:
> Hi,
> 
> Thanks, @johanneskoester for PR [2eb66fb](https://github.com/snakemake/snakemake/commit/2eb66fbc53e9e9bad2c5de43566ead81767d1922) , this indeed solved the issue of the lost attribute to the executor class.
> 
> However, I think there are still a couple of major issues:
> 
> * all test jobs in the github CI are failing because the sbatch command, which is issued by the executor is rejected - even if there are no further resource requirements (I tinkered with the Snakefile). The same rule on "my" cluster is submitted and carried out without issues. We therefore have an issue with the Ansible playbook. And even after reading the playbook I am not sure where the issue is.
> * when executing the 2nd rule in `test_slurm_mpi` (the MPI rule) the environment is not loaded (the command is ` snakemake --use-envmodules --slurm -d . -j1` and the log states `Activating environment modules` for rule 1 (the dummy compile-my-code-rule), but not for the 2nd rule. The aim should be to load environment modules prior to each job execution, best _on_ the compute node in job context. That would require to issue the `module purge && module load` via the SlurmJobStepExecutor. As this is currently ultimately done via the shell-class and its `__new__`-method, how should the information be propagated to the executor and why is this even necessary, if the command is run in the first rule?
> * after lots of printf-debugging I am confident that the `group` property in the group test is not recognized with `--slurm`. Not sure why.
> 
> Any ideas?
> 
> Cheers Christian

To be honest, I have no immediate ideas. All of these things should have worked without doing anything. So I guess something unexpected interferes in the CI and we have to sit down together and debug.

cmeesters:
Just to be clear: The last two issues mentioned in my previous post refer to my local testing. Never got to the point, where those become an issue in the CI, alas.

snakemake/executors/slurm/slurm_submit.py
mbhall88: I also think having specific variable for minute, hours etc is redundant when you can just use this https://github.com/Snakemake-Profiles/slurm#human-friendly-time

cmeesters: Thank you. A user friendly interface certainly is on my/our mind. The resource management, however, is secondary to the PR actually passing the CI in the first place.

Then a little refactoring and polishing will not be an issue. I think something like your proposal certainly should find a place in this code!

cmeesters:
@johanneskoester the code is working. It is polished (not withstanding 2 complexity code smells) and still fails the CI. 

We had the CI failing because the default memory is a 1000 M, which cannot be satisfied. Then I tried submitting with zero memory. However, we get:

```
CRITICAL snakemake.logging:logging.py:553 No SLURM account given, trying to guess.
CRITICAL snakemake.logging:logging.py:553 Unable to guess SLURM account. Trying to proceed without.
CRITICAL snakemake.logging:logging.py:553 No wall time limit is set, setting 'walltime_minutes' to 10.
CRITICAL snakemake.logging:logging.py:553 No job memory information ('mem_mb' or 'mem_mb_per_cpu') is given - submitting without. This might or might not work on your cluster.
DEBUG    snakemake.logging:logging.py:557 Job 3 'snakejob_compile_2' has been submitted
ERROR    snakemake.logging:logging.py:555 sacct process error
ERROR    snakemake.logging:logging.py:555 Command '['sacct', '-P', '-b', '-j', '3', '-n']' returned non-zero exit status 1.
```

Which is rather unpleasant, right? In essence, the fake-SLURM cluster for the CI is non-standard, does not accept (fake) memory requirements nor returns DB inquiries. Suggestions?

cmeesters:
I'm afraid, there is another issue: 

When using the `tmpdir` directive together with a `shadow` directive like:

```
rule <name>:
   shadow: "copy-minimal"
   resources:
      tmpdir=tmpdir="/lustre/scratch/{jobid}/ramdisk".format(jobid=os.environ['SLURM_JOB_ID']),
      ...
```
Here, the environment variable is interpreted upon submission, yet for a non-local rule, a rule which is executed in a SLURM-Job, this env-var obviously does not exist on a head- or login-node, where the orchestrating snakemake instance is invoked.

I am not quite sure how to make the shell-class aware of its executor to delay the evaluation of this env-var up until the job context. @johanneskoester suggestions?

tests/tests.py
```diff
@@ -9,14 +9,33 @@
...
 sys.path.insert(0, os.path.dirname(__file__))
 
 from .common import *
 from .conftest import skip_on_windows, only_on_windows, ON_WINDOWS, needs_strace
 
-from snakemake.resources import GroupResources
+
+@skip_on_windows
+def test_slurm_mpi():
+    run(dpath("test_slurm_mpi"), slurm=True)
```
johanneskoester, line 20: If we need to see the log file of the failing job:
```suggestion
    run(dpath("test_slurm_mpi"), slurm=True, show_failed_logs=True)
```

snakemake/executors/slurm/slurm_submit.py
```diff
@@ -0,0 +1,333 @@
...
+                    logger.error("sacct process error")
+                    logger.error(e)
+                except IndexError as e:
+                    pass
+                # Try getting job with scontrol instead in case sacct is misconfigured
+                if not res:
+                    try:
+                        sctrl_cmd = shlex.split("scontrol -o show job {}".format(jobid))
+                        sctrl_res = subprocess.check_output(sctrl_cmd, encoding="ascii")
+                        logger.debug("The scontrol output is: '{sctrl_res}'")
```
johanneskoester, line 282:
```suggestion
                        logger.debug(f"The scontrol output is: '{sctrl_res}'")
```

tests/Snakefile
```diff
@@ -0,0 +1,33 @@
+localrules: all, clean
```
johanneskoester, line 1: This file has to be moved, it is in the main tests folder.

snakemake/resources.py
cmeesters: oh, I thought I had _already_ written an answer to you @mbhall88 

Well, anyway, I think this is a jolly good idea. In a discussion with @johanneskoester we agreed to add it to the executor class, respectively the resources. Johannes will approach you about this (please keep me in the loop, if you will). One question, though: Your code is covered by MIT, so taking it poses no problem. May I put your name and link to the code and refer to it appropriately?

mbhall88: Absolutely. Take whatever you like.

cmeesters:
Thanks @johanneskoester  - I have been busy (still am, actually) teaching.

Alas, the logs aren't telling us the reason for the compile failure: If I install those libs locally and trigger the compile command, it just works (this is not helpful, I know. ) I understand that openmpi-bin and libopenmpi-dev are available after their installation globally for all tests?

cmeesters:
@johanneskoester 

As fas as I am concerned, this is functional code - not withstanding the fact that there are two major issues left:

- fine-tuning the user interface / resource section
- ensuring that environment modules are loaded only once (to avoid withdrawing snakemake instances which are downstream of the executor triggering the module load command via shell.__new__) from the last snakemake instance

However, the CI keeps failing: lacking a SLURM db we have the fallback to `scontrol`. `scontrol` (and `squeue` likewise) are ignorant to past jobs. When calling these commands for past jobs, the status is inconceivable. For very short jobs, there might even be a race condition: they might finish before the first call to `scontrol`, which is why (very hacky, I know) I implemented a `time.sleep(30)`, because one cannot rely on users always setting `--latency-wait` correctly and waiting an ordinary scheduler cycle in case the fallback is needed does seem ok-ish.

All this is to no avail: the desired output is not detected in the CI, it always fails, the logs are inconclusive. 

Any ideas?

johanneskoester:
Btw. the slurm_mpi testcase works fine on my local system. The files are found.

snakemake/executors/slurm/slurm_jobstep.py
```diff
@@ -111,7 +111,7 @@ def get_call(level_job, aux=""):
             # Note: The '--exlusive' flag is a prevention for triggered job steps within an allocation
             #       to oversubscribe within a given c-group. As we are dealing only with smp software
             #       the '--ntasks' is explicetly set to 1 by '-n1', above.
-            call = f"srun --cpu-bind=q {self.format_job_exec(job)}"
+            call = f"srun --cpu-bind=q --exclusive {self.format_job_exec(job)}"
```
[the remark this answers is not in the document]
cmeesters: Yep, you are right -- comment and code accidentally went to the wrong position. It is applicable to the group job, only. See commit 808b758a1ae4f93930e1fbfc4aa9e2e8898bdfd9

cmeesters:
@johanneskoester you deleted the slurm setup from the github-workflow? is this a mistake, perhaps?

johanneskoester:
That was not meant for this PR, but for the main branch (as the setup there stopped working).

johanneskoester:
It seems as if the job never really runs but is still reported as completed. This is in line with our dummy srun from the step before, which also seems to fail: https://github.com/snakemake/snakemake/actions/runs/3633973818/jobs/6131598569#step:14:84

cmeesters:
> It seems as if the job never really runs but is still reported as completed. This is in line with our dummy srun from the step before, which also seems to fail: https://github.com/snakemake/snakemake/actions/runs/3633973818/jobs/6131598569#step:14:84

Yes, the srun check seems useless, as it is. I am just doing more or less educated guesses, now.

cmeesters:
Actually, the debug information and the error should be printed - I do not see any of it in the logs, ideas?

johanneskoester:
I've now fixed the hello world srun test.

johanneskoester:
And here we finally have the actual error:
```
slurmstepd-localhost: error: execve(): /var/lib/slurm/slurmd/job00002/slurm_script: Permission denied
```

johanneskoester:
Ok, this is the error of the second step:
```
[fv-az370-172:10073] OPAL ERROR: Unreachable in file pmix3x_client.c at line 112
```

Seems related to this: https://github.com/open-mpi/ompi/issues/7025
Do we need to modify our srun statement for the calc rule?

cmeesters:
> Do we need to modify our srun statement for the calc rule?

Yes, I noticed that - I will modify the srun test temporarily to figure out, which mpi support is given, if any. 

Also, the srun-test indicates, we only have 1 CPU and 1 thread. That is a bit unfortunate.

cmeesters:
setting tasks ought to be redundant: this information should be propagated from the slurmd to the slurmstepd (read from sbatch to srun)

cmeesters:
OK, we can now safely conclude: The CI-slurm is not build with MPI-support. But why is there absolutely no output? Trying one thing ...

johanneskoester:
> ```
> tests/test_slurm.py::test_slurm_mpi PASSED                               [ 33%]
> ```
> 
> Oh yeah!
> 
> ```
> tests/test_slurm.py::test_slurm_group_job FAILED                         [ 66%]
> ```
> 
> Oh no!
> 
> Wait:
> 
> ```
>  >       assert os.path.exists(original_snakefile)
> E       AssertionError
> ```
> 
> what does that mean? Why is the snakefile gone?

you had a typo in the name of the test. But indeed, the error message given by the test framework is a bit misleading.

cmeesters:
```
AttributeError: 'GroupJob' object has no attribute 'logfile_suggestion'
```
This is no property introduced (or not introduced) by the SLURM_Executor. 

```
AssertionError: expected successful execution
```
Me too, ;-)

johanneskoester:
> ```
> AttributeError: 'GroupJob' object has no attribute 'logfile_suggestion'
> ```
> 
> This is no property introduced (or not introduced) by the SLURM_Executor.
> 
> ```
> AssertionError: expected successful execution
> ```
> 
> Me too, ;-)

My fault, I introduced it in this PR to get better log file names in slurm. I will fix it now.

cmeesters:
@johanneskoester Nice, nice, nice!

Just, the conversion is to MB, not MiB. I know, the slang is MB, not MiB. But everyone thinks in MiB (with regard to RAM), not MB. This confusion is all just because of the marketing trick of storage vendors, I know. What I don't know, is whether this conversion might add confusion, 'cause SLURM thinks of RAM in MiB (but mentions nothing of the sort anywhere) and everyone else also thinks MiB with regard to RAM, but spells MB. It's a little jinxed!

PR #1017, opened 2021-05-26
feat: support XRootD as a default remote provider

ryuwd:
### Description

It is currently not possible to specify `--default-remote-provider=XRootD` on the command line, and by extension when using a snakemake profile that uses XRootD remote storage. 

### QC


ryuwd:
@chrisburr since you are the original author of the XRootD remote provider, do you have any thoughts on this?

I initially started work on this to see if I could use this for running `snakemake` workflows at the CERN grid site, but have since started writing a remote file provider for Dirac instead. 

Do you think this change would be useful? (and does my meddling with the paths break anything?)

johanneskoester (approved):
To me, this looks good.

johanneskoester:
@chrisburr do you want to give this is final look?

PR #1019, opened 2021-05-27
pass conda base path to cluster jobs

### LINKED ISSUE #1009: Cluster jobs fail claiming mamba is missing, even though mamba is present, and even if conda backend is specified
**Snakemake version**
6.4.0 (not present in 6.3.0)

**Describe the bug**
Cluster jobs fail with an error saying mamba is not found. Mamba works for the workflow up until submission, and the error is still present when using `--conda-frontend conda`. I am using FISH as my main shell, but the rules are running in BASH. Both my `config.fish` and `.bashrc` load conda envs that have mamba.

**Logs**

The following error shows up in the cluster log:

```
Error: mamba package manager is not available. The mamba package manager (https://github.com/mamba-org/mamba) is an extremely fast and robust conda replacement. It is the recommended way of using Snakemake's conda integration. It can be installed with `conda install -n base -c conda-forge mamba.If you still prefer to use conda, you can enforce that by setting `--conda-frontend conda`.
```

**Minimal example**
Mamba works for generating the environments, but it breaks when it tries to run the rules. I'm using the following command: ` snakejob -pj 600 --use-conda`

snakejob looks like this:

```
#!/usr/bin/env bash

set -e

if [ ! -d ".snakejob" ]; then
  mkdir .snakejob
fi

curdir="$(pwd)"
jname="$(basename $curdir)"

bsub=("bsub -K -J $jname:{rule}:{wildcards}"
"-P {cluster.project} -q {cluster.queue}"
"-n {cluster.cores} -R \"span[hosts=1] select[mem>{cluster.mem}]"
"rusage[mem={cluster.mem}]\" -W {cluster.time} -L /bin/bash"
"-oo .snakejob/{rule}:{wildcards}.stdout"
"-eo .snakejob/{rule}:{wildcards}.stderr < ")

snakemake -u cluster.yaml --cluster-sync "${bsub[*]}" \
  --local-cores 4 --max-jobs-per-second 5 "$@"
```

It's basically a script that allows me to use a cluster config file on lsf. I know this behavior is technically deprecated, but we have not been able to implement profiles to our satisfaction.

johanneskoester:
### Description

The conda base path determined on the main Snakemake process is now passed on to jobs (in particular in the cluster). Fixes #1009.

### QC


lailamargaret:
Hi, I'm using snakemake on a HPC with SGE. When I submit snakemake jobs using `--cluster qsub --use-conda`, snakemake still is unable to locate my conda install and I get the error ` The 'conda' command is not available in $PATH`. When I explicitly pass my conda path to snakemake using `qsub -V`, snakemake fails to make the required environments with the packages I need (`conda list` as the shell command in my rule shows only the env that snakemake was called from). Any help would be appreciated! Thanks.

Issue #1028, opened 2021-06-05
Command-line `--touch` option results in FileNotFoundError if a pipeline has outputs marked as 'temp'


iromeo:
**Snakemake version**
6.4.1

**Describe the bug**
Our computation server uses a weakly cleanup policy, so for me is important to `touch` all pipeline results each weak.
I use `--touch` snakemake feature, but it has a bug related to temporary files:

The command below fails if I have outputs marked as temporary: 
`snakemake --jobs 1 --touch --forceall`

**Logs**
```
Building DAG of jobs...
Using shell: /bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job counts:
        count   jobs
        1       all
        1       input
        1       intemediate
        3
Select jobs to execute...

[Sat Jun  5 22:53:44 2021]
rule input:
    output: tmp.txt
    jobid: 2

[Sat Jun  5 22:53:44 2021]
Finished job 2.
1 of 3 steps (33%) done
Select jobs to execute...
Traceback (most recent call last):
  File "/Users/romeo/opt/anaconda3/envs/smk3/lib/python3.9/site-packages/snakemake/__init__.py", line 701, in snakemake
    success = workflow.execute(
  File "/Users/romeo/opt/anaconda3/envs/smk3/lib/python3.9/site-packages/snakemake/workflow.py", line 1036, in execute
    success = scheduler.schedule()
  File "/Users/romeo/opt/anaconda3/envs/smk3/lib/python3.9/site-packages/snakemake/scheduler.py", line 470, in schedule
    run = self.job_selector(needrun)
  File "/Users/romeo/opt/anaconda3/envs/smk3/lib/python3.9/site-packages/snakemake/scheduler.py", line 660, in job_selector_ilp
    sum([size_gb(temp_file) for temp_file in temp_files]), 1
  File "/Users/romeo/opt/anaconda3/envs/smk3/lib/python3.9/site-packages/snakemake/scheduler.py", line 660, in <listcomp>
    sum([size_gb(temp_file) for temp_file in temp_files]), 1
  File "/Users/romeo/opt/anaconda3/envs/smk3/lib/python3.9/site-packages/snakemake/scheduler.py", line 635, in <lambda>
    size_gb = lambda f: f.size / 1e9
  File "/Users/romeo/opt/anaconda3/envs/smk3/lib/python3.9/site-packages/snakemake/io.py", line 242, in wrapper
[...]

PR #1029, opened 2021-06-05
fix: Resolves issue #1028

iromeo:
Fix for the #1028 issue

### Description

Fix for the #1028 issue:
> Command-line --touch option results in FileNotFoundError if a pipeline has outputs marked as 'temp'

### QC


johanneskoester:
Thanks a lot! I have instead fixed the real cause of the issue in the scheduler. However, thanks a lot for the testcase!! Closing in favor of #1094.

PR #1033, opened 2021-06-07
Fix conda support on Windows

melund:
### Description

This PR fixes the issues which prevented conda support on Windows. 

* Fixes a underlying issues with snakemake codebase using `urllib.parse.urlparse` to distinguish between URI's and paths. That doesn't work on windows. 
* Fixes a number of problem with forward vs. backwards slashes, double vs. single quotes etc. 
* Fixes the conda tests so they use a small packages which is also available on Windows. (the 'bwa' is posix only). 

### QC

### DIFF
--- a/snakemake/deployment/conda.py
+++ b/snakemake/deployment/conda.py
@@ -8,7 +8,6 @@ import re
 import subprocess
 import tempfile
 from urllib.request import urlopen
-from urllib.parse import urlparse
 from urllib.error import URLError
 import hashlib
 import shutil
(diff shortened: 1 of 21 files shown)


johanneskoester (changes requested):
We have recently switched to smart_open for these things. Can you rely on that instead?
There is `snakemake.common.is_local_file`, and `smart_open.parse_uri` which should both work on windows as well.

melund:
> We have recently switched to smart_open for these things. Can you rely on that instead?

First time hearing about `smart_open`. Awesome. I can defiantly change  the different places `urlparse` is used. But you better review the changes carefully, since I will be changing some stuff where I don't know the logic.

johanneskoester (approved):

johanneskoester:
Thanks a lot!

PR #1042, opened 2021-06-09
Add new standard resource tmpdir

johanneskoester:
### Description

The tmpdir resource automatically leads to setting the usual tmpdir env variables for shell, script, wrapper, and notebook directives. The default value for the tmpdir resource is the system setting.
Further, --default-resources defaults are now always used.

### QC


johanneskoester:
@melund may I ask you to have a look at this? On windows, I get this error: https://github.com/snakemake/snakemake/pull/1042/checks?check_run_id=2787370703#step:6:314
It seems like the TEMPDIR, TEMP and TMP envvars are not properly set, in contrast to TEMPDIR. Although we set all of them in the same way here: https://github.com/snakemake/snakemake/pull/1042/files#diff-7a75aedad3c9adf7a80ba34778717f85dbd0b0362be0616a8bfac35431e6ad6fR194

melund:
@johanneskoester Yes. I will try to take a look tomorrow if my work permit it. 

There are a lot weird issues with environment variables and subprocess on windows.

melund:
@johanneskoester I think I know what is going on. The 'bash.exe' modifies the `$TMPDIR`, `$TEMP`, and `$TMP` so the make sense for bash,

That is because, `$TMPDIR`, `$TEMP`, and `$TMP` are also windows environment variable which it would otherwise inherit with paths that were unusable. 

But since msys mounts the windows temp directory on `/tmp` the are just changed in bash to just be `/tmp` 

On my machine the running the test gives the following:
```
E                       AssertionError: wrong result produced for file 'test2.txt':
E                       ------found------
E                       foo /tmp/snakemake-hm9002u1/foo /tmp/snakemake-hm9002u1/foo /tmp/snakemake-hm9002u1/foo foo
E
E                       -----expected-----
E                       foo foo foo foo foo
```

It is really puzzling why this is different that the result you linked to CI server.  Maybe it has diffent configuration for the temporary folder. 

But I think the output on my machine makes sense.  The local folder `foo` in current working directory is in reality at `C::/Users/mel/AppData/Local/Temp/snakemake-hm9002u1/foo` which is again mounted on `/tmp/snakemake-hm9002u1/foo`

Very confusing. This maybe a place where tests windows tests should be skipped.

melund:
@johanneskoester Just skip the test on windows for now. I will try to fix it later. 

I am sure the test will work as expected when using "cmd.exe" as the shell. But then the "echo" command works a bit different. 

I also suspect that there is a small issue with the test framework on Windows  so the `shell.executable("bash")` is not reset between tests. That means that almost everything is run with git bash. That is not bad, but I also want some dedicated tests with "cmd.exe".

PR #1046, opened 2021-06-10
fix: avoid "Admin" prompt when using conda on windows

melund:
### Description

This fixes a problem where `--use-conda` can sometimes trigger an ´admin rights` prompt on Windows.

The reason is that some conda packages will try to install start-menu shortcuts, and thus trigger an admin prompt.

There is a `conda create --no-shortcuts` argument.  But, unfortunately, the same doesn't exist for `conda env create`

 See: https://github.com/conda/conda/issues/9474

`conda build` handles the problem by creating environments prefixed with an underscore "_<env-name>", which also cause conda to skip environment creation.

This PR prefixes an underscore to the conda environments created in ".snakemake/conda"

### QC


melund:
For this to work I need help updating the docker image: 

"docker://quay.io/snakemake/containerize-testimage:1.0"

I have updated the docker file, and all that is needed is for someone to make a new version of the test image and upload it.

melund:
For this to work I need help updating the docker image: 

"docker://quay.io/snakemake/containerize-testimage:1.0"

I have updated the docker file, and all that is needed is for someone to make a new version of the test image and upload it.

snakemake/deployment/conda.py
```diff
@@ -125,7 +125,9 @@ def path(self):
         """
         hash = self.hash
         env_dir = self._env_dir
-        get_path = lambda h: os.path.join(env_dir, h)
+        # The underscore prefix cause the conda to skip shortcut installation so
+        # no admin prevleges are needed for windows.
+        get_path = lambda h: os.path.join(env_dir, "_" + h)
```
johanneskoester, line 130: Mhm, I don't like to change a path just to work around some conda behavior. How does this relate to the --no-shortcuts flag below? How will this affect linux? Is this also an issue when using mamba instead of conda for the archive based installation?

melund: @johanneskoester: Yes, I understand why this is a bit ugly.  But it seems like this is only way at the moment, and it is also what `conda-build` does when programmatically creating build and test environments.  

The `--no-shortcuts` flag is only available for `conda create` and not for `conda env create`. 

Maybe Mike Sarahan (@msarahan) can help clarify. Is prepending the environment name with a underscore [still the only](https://github.com/conda/conda/issues/2136#issuecomment-189657796) reliable way to prevent menu installation? 

> Is this also an issue when using mamba instead of conda for the archive based installation?

Mamba uses conda implementation here. So it should be fine. For micro-mamba the `menuinst` feature is beeing added right now: https://github.com/mamba-org/mamba/issues/923#issuecomment-860404947

johanneskoester: Ok, understood. If that is kind of the intended behavior and the underscore is the desired semantic meaning here, then I'm fine with it. Will it also prevent envs to be listed when running `conda env list` (which would be desired because that always annoyed me)?

melund: No. Unfortunately not. But it would be a good idea. I find that annoying as well.

melund: The only thing I could think of is adding something like: 

```
sed -i "/_[^\\]*$/d" ~/.conda/environments.txt
```
in you `.bachrc` to clean you environment listing every time you start a shell. That will remove environments from you listing starting with underscore.  They environments will get re-added once they are activated again. But it is the only workaround I can think of.

snakemake/deployment/conda.py
```diff
@@ -230,18 +230,31 @@ def address(self):
...
+        if self.is_named:
+            return self.name
+        else:
+            hash = self.hash
+            env_dir = self._env_dir
+            get_path = lambda h: os.path.join(env_dir, h)
+            hash_candidates = [
+                hash[:8],
+                hash,
+                hash + "_",  # activate no-shortcuts behavior (so that no admin rights are needed on win)
```
melund: @johanneskoester, Maybe I misunderstand the code here, but I think the environment name should be *prefixes* with an underscore "_". I am not sure that postfix does the same thing.

Issue #1043, opened 2021-06-09
Google Life Sciences executor failing with "read timed out"


johanneskoester:
**Snakemake version**
main branch

**Describe the bug**
The GLS testcase currently fails with a timeout error. Nothing was changed in the implementation or the test:
https://github.com/snakemake/snakemake/runs/2772939881#step:14:6800

@vsoch @moschetti any idea?

**Minimal example**
```
pytest -s -v -x tests/test_google_lifesciences.py
```

vsoch:
Is it consistently failing? It looks like the timeout is 60 seconds (which seems pretty large) so could it just be that the service / API is having trouble? Maybe worth trying later? I checked the Google Cloud status dashboards but they don't seem to be very detailed - either Google is perfect or there really hasn't been a single issue since 2020 :laughing:. @moschetti is there some secret knowledge about up/downtime or changes that we don't know about (and can?)

johanneskoester:
Happens on each CI run since 6 days now.

moschetti:
I don't see anything on any other downtime or relevant status updates.

The [timeout happening on `blob.exists()`](https://github.com/snakemake/snakemake/runs/2772939881#step:14:6777) seems odd as that one shouldn't be an action that takes long. That method does take a [`timeout` or `retry` parameter](https://googleapis.dev/python/storage/latest/blobs.html#google.cloud.storage.blob.Blob.exists) though, so could try increasing timeout or adding retries for error handling. But I don't see any issues on storageapi timeouts generally or libraries that have changed.

vsoch:
That would be good to try first! https://googleapis.dev/python/storage/latest/blobs.html#google.cloud.storage.blob.Blob.exists. It it really takes longer than maybe 2 minutes there is a serious issue. And this is just for the tests? Did anything in the storage for the tests get deleted so it might actually not exist, or is something in the storage exploded so it's taking a long time to find it?

johanneskoester:
The timeout occurs when trying to delete the sourcecode blob during shutdown. I have tried to quickly put all those api calls inside of a retry block, let's see.

johanneskoester:
Fixed with PR #1051.

PR #1051, opened 2021-06-11
Fix google API timeouts in life science backend

johanneskoester:
### Description

This adds retries to google API calls in the life sciences backend.

### QC


johanneskoester:
This should fix issue #1043.

Issue #913, opened 2021-03-13
Rust scripts and notebooks


mbhall88:
**Aim:** Allow snakemake to execute Rust "scripts" in the same way it can execute Python and R scripts. Additionally, allow the use of Rust in notebooks.

A single Rust source file is given to the `script` directive, and if necessary, `Cargo.toml`-like configuration embedded in the script, as in `rust-script`.

[`rust-script`](https://rust-script.org/) somewhat encapsulates the thesis of this feature.

The use of Rust in notebooks is probably best facilitated by [`evcxr`](https://depth-first.com/articles/2020/09/21/interactive-rust-in-a-repl-and-jupyter-notebook-with-evcxr/).

## Usage

A minimal example of how I envisage this working is the following

`Snakefile`

```python
rule snakers:
    output:
        "file.txt",
    params:
        fave_lang="rust",
    log:
        "snakers.log"
    script:
        "scripts/snake.rs"
```

`scripts/snake.rs`

```rust
use std::fs::File;
use std::io::Write;

fn main() {
    let outpath = snakemake.output[0];
    let fave_lang = match snakemake.params.get("fave_lang") {
        Some(s) => s,
        None => "no language",
    };
    let mut outfile = File::open(outpath).expect("Couldn't create outfile");
    write!(&mut outfile, "My favourite language is {}", fave_lang).expect("write fail");
}
```

In terms of how this will be implemented, @johanneskoester has suggested using `rust-script` itself to do this. I have one reservation about this, which is that it adds an external dependency - although we will be relying on either `cargo` or `rustc` anyway.

mbhall88:
Sorry, will leave this open until we have notebooks too

PR #1053, opened 2021-06-14
feat: Add support for rust scripts (enabling directly integrated ad-hoc robust high performance scripting)

tedil:
### Description

This draft PR adds support for using [rust-script](https://rust-script.org/) scripts with snakemake's script directive. It's a proof of concept, see also @mbhall88's notes on rust script support [here](https://hackmd.io/qG4y3JqlTAOHLgxHNdoqDA).

For example, the following script reads from a named input file, appends some string to it and writes it to a positional output file. It also shows how to specify additional dependencies:
```rust
// cargo-deps: maplit="1.0.2"
use std::io::Write;
use maplit::hashmap;

let snakemake = Snakemake::load()?;

let input = std::fs::read(&snakemake.input["hello"])?;
let hello = std::str::from_utf8(&input)?.trim();

let mut file = std::fs::File::create(&snakemake.output["0"])?;
write!(&mut file, "{} {}", hello, "snakemake!")?;

eprintln!("{:?}", hashmap!{"dependencies" => "work"});
```
#### Caveats / Details
- The `snakemake` object has to be obtained using `Snakemake::load()`, which deserializes a pickled version of the (well, at the moment not actually *the* but rather a stripped down version) Snakemake (python) object into a rust Snakemake struct, reading from a (not yet temporary) file. This entails having `serde`, `serde_derive` and `serde-pickle` as default dependencies (which might be a bit much for a small script, but this was the quickest way for me to get started).
- Since heterogeneous collections aren't particularly easy to manage in rust, positional input/output/… arguments are indexed by string aswell, e.g. the first positional input argument can be retrieved as `snakemake.input["0"]`.
- (additional) search paths aren't handled, yet
- rust-script has a template option, but I'm not sure how to leverage that for our use-case
- rust-script has to be installed (`cargo install rust-script`). There's no conda package for that at the time of writing.

### QC


mbhall88:
Thanks for making a start on this @tedil. 

I think for the initial stages the serde dependencies are probably fine until we get something working.

Is there a task you would like me to work on? I'm mindful of not wanting to tread on your toes.

(Also, this will close #913)

tedil:
What still needs to be done:
- [ ] documentation
- [ ] rust-script packaged on conda-forge, [in progress](https://github.com/conda-forge/staged-recipes/pull/15297)
- optional / nice to have / later:
  - [ ] having a dedicated NamedList type which has an API either like `params.positional()[0]` and `params.named()["named"]` or `params.indexed(0)` and `params.named("named")` or `params.get(Index(0))` and `params.get(Name("named"))`.
  - [ ] handle dependencies better (at the moment only single line line comment dependencies are allowed)
  - [ ] think about replacing the serde dependencies with maplit (or nothing, even) and encoding the values ourselves, just as it is done for julia and R. I guess some dicts may have mixed types (especially their values), which makes this rather tedious to work with in rust
  - [ ] use / expose rust-script templates

mbhall88:
Great. There's plenty in that to keep us busy for a little while. Mind if I work "handling dependencies better"? i.e. allowing doc comment Cargo manifests in addition to the single-line dependencies?

Also, I think you might not have added your test Rust script to the PR?

tedil:
> Great. There's plenty in that to keep us busy for a little while. Mind if I work "handling dependencies better"? i.e. allowing doc comment Cargo manifests in addition to the single-line dependencies?

That would be great!

> Also, I think you might not have added your test Rust script to the PR?

Oh, you are right, they got skipped due to .gitignore rules. Fixed now.

tedil:
This NamedList implementation would allow accessing both by index and by name, under the assumption that the order of the dict entries (of the dict that gets pickled in python and unpickled in rust by serde) is stable and corresponds to the index scheme.

```rust
use std::ops::Index;

#[derive(Debug, Deserialize)]
struct NamedList<V>(pub IndexMap<String, V>);

impl<V> Index<usize> for NamedList<V> {
    type Output = V;

    fn index(&self, index: usize) -> &Self::Output {
        self.0
            .get_index(index)
            .unwrap_or_else(|| panic!("index out of bounds: {}", index))
            .1
    }
}

impl<V> Index<&str> for NamedList<V> {
    type Output = V;

    fn index(&self, index: &str) -> &Self::Output {
        self.0
            .get(index)
            .unwrap_or_else(|| panic!("No such key {}", &index))
    }
}
```

This also requires indexmap with its serde feature enabled, which will only be possible once the "better handling of dependencies" checkbox is ticked ;)

johanneskoester:
> This NamedList implementation would allow access both by index and by name, under the assumption that the order of the dict entries (of the dict that gets pickled in python and unpickled in rust by serde) is stable and corresponds to the index scheme.
> 
> ```rust
> use std::ops::Index;
> 
> #[derive(Debug, Deserialize)]
> struct NamedList<V>(pub IndexMap<String, V>);
> 
> impl<V> Index<usize> for NamedList<V> {
>     type Output = V;
> 
>     fn index(&self, index: usize) -> &Self::Output {
>         self.0
>             .get_index(index)
>             .unwrap_or_else(|| panic!("Index out of bounds: {}", index))
>             .1
>     }
> }
> 
> impl<V> Index<&str> for NamedList<V> {
>     type Output = V;
> 
>     fn index(&self, index: &str) -> &Self::Output {
>         self.0
>             .get(index)
>             .unwrap_or_else(|| panic!("No such key {}", &index))
>     }
> }
> ```
> 
> This will also panic for nonexistent keys/indices. But you could always use the inner indexmap instead, e.g. `snakemake.input.0.get(key).map(|value| ...)`
> 
> This also requires indexmap with its serde feature enabled, which will only be possible once the "better handling of dependencies" checkbox is ticked ;)

The NamedList in Snakemake can also return a list of files/items sometimes. This is not properly represented here.

tedil:
> The NamedList in Snakemake can also return a list of files/items sometimes. This is not properly represented here.

Ah, you mean it's not always a map? Bummer. Well we could use a newtype wrapper around `serde_pickle::Value` and provide some convenience functions for that?

johanneskoester:
> > The NamedList in Snakemake can also return a list of files/items sometimes. This is not properly represented here.
> 
> Ah, you mean it's not always a map? Bummer. Well we could use a newtype wrapper around `serde_pickle::Value` and provide some convenience functions for that?
> 
> _edit_: nevermind, we can change lists to maps on the python side (and it's already done I think)

Not sure what you mean. A name pointing to a list of files is important to keep of course. This here might be one option:

```
trait NamedList<K, V> {
    fn get(&self, key: K) -> NamedListValue<V>;
}

enum NamedListValue<V> {
    Single(V),
    Multiple(Vec<V>),
}
```

The other one would be two separate getters `get_single`and `get_multiple`which panic in case of being used with the wrong item.

johanneskoester:
I think the former way is more idiomatic. And one could add helpers to the enum that provide the same convenient panic behavior.

tedil:
> > > The NamedList in Snakemake can also return a list of files/items sometimes. This is not properly represented here.
> > 
> > 
> > Ah, you mean it's not always a map? Bummer. Well we could use a newtype wrapper around `serde_pickle::Value` and provide some convenience functions for that?
> > _edit_: nevermind, we can change lists to maps on the python side (and it's already done I think)
> 
> Not sure what you mean. A name pointing to a list of files is important to keep of course. This here might be one option:
> 
> ```
> trait NamedList<K, V> {
>     fn get(&self, key: K) -> NamedListValue<V>;
> }
> 
> enum NamedListValue<V> {
>     Single(V),
>     Multiple(Vec<V>),
> }
> ```
> 

I thought about an enum solution aswell, but that would make the serde deserialization much more complicated. As it is now, we can just deserialize the dict items we get from python / the pickle file into a struct such as the NamedList one I proposed earlier.

> The other one would be two separate getters `get_single`and `get_multiple`which panic in case of being used with the wrong item
? I don't get it. Are you saying:
a) snakemake.input can be of type list
b) snakemake.input is of type dict but may have (key, value) pairs where value is not a single value?

Case a) (if even possible) is already handled by just pretending its a dict.
Case b) is also not a problem, since the *value* types these maps deserialize to is [`serde_pickle::Value`](https://docs.rs/serde-pickle/0.6.2/serde_pickle/value/enum.Value.html) which is an enum which also has a `List` variant.

mbhall88:
So we can now handle either form of dependency specification (single-line or code block manifest). I added an additional test case so we are testing both forms. The tricky part was just that rust-script requires this *before* the preamble, so I had to scrape the dependency comment out of the original script and insert it above the preamble. Another annoying thing is that rust-script only allows inner doc comments `//!` -  this is annoying because they can't have any empty line between them and the following code. I've [raised and issue](https://github.com/fornwall/rust-script/issues/25) to request more supported doc comments.

I have also restructured the scripts docs a little. I basically just created subheadings for each language as we now support 4 languages :tada: and thought this would make it easier to navigate.

There's still some more docs that need to be added for rust but I might wait until we have nailed down things a little more.

Please feel free to go hard with suggesting changes to my additions etc. Always love to learn better ways of doing things, especially from you two.

snakemake/script.py
```diff
@@ -821,6 +821,293 @@ def execute_script(self, fname, edit=False):
...
+        else:
+            lookup = {
+                (True, True, True): " >> {0} 2>&1",
+                (True, False, True): " >> {0}",
+                (False, True, True): " 2>> {0}",
+                (True, True, False): " > {0} 2>&1",
+                (True, False, False): " > {0}",
+                (False, True, False): " 2> {0}",
+            }
+            log = lookup[(stdout, stderr, append)].format(self.log)
```
mbhall88, line 863: @tedil `self`, `stdout`, `stderr` and `append` don't exist in this context. I didn't want to change anything as I wasn't super clear on the intended functionality here.

tedil: Ah I just copied the PythonScript class for reference, so there are still python specific tidbits in there.

mbhall88: Ok, I'll address this in my next round of work.

mbhall88:
I'm keen to get stuck into another element of the implementation. What are you working on @tedil and what would you consider highest priority for me to get stuck into?  
There's a couple of items from your previous list that I am unclear on
-  log to the correct channels
- consider paths/PATH/syspath etc

Would you be able to elaborate a little on these?

Also, regarding our dependency on serde (and derive feature), I had the original idea of just using some kind of template to generate the snakemake data structure directly from the python one. This is obviously a lot more work, but would allow us to directly encode the correct types without needing to use `Value`. I understand the main roadblock to this is python dictionaries with mixed types? Are there any other problems with this approach you've thought of?  
One potential workaround to the mixed types situation could be, if there are mixed types, everything becomes a `String`?

tedil:
> I'm keen to get stuck into another element of the implementation. What are you working on @tedil and what would you consider highest priority for me to get stuck into?



> There's a couple of items from your previous list that I am unclear on
> 
>     * log to the correct channels
> 
>     * consider paths/PATH/syspath etc
> 

These might also have been python specific. I used the python script functionality implementation as a rough guide, so perhaps these aren't issues we have to deal with in the rust script case. If the channels used for logging are always stdout and stderr, we probably don't have to change anything; but if they are to be redirected (to a file, some logger somewhere else, etc) :shrug: 

> Also, regarding our dependency on serde (and derive feature), I had the original idea of just using some kind of template to generate the snakemake data structure directly from the python one. This is obviously a lot more work, but would allow us to directly encode the correct types without needing to use `Value`. I understand the main roadblock to this is python dictionaries with mixed types? Are there any other problems with this approach you've thought of?

Okay so: I talked with Johannes on Friday: The `input` and `output` values are ever only paths/strings, i.e.: there can be positional values and/or key-value pairs, where values are either single strings or lists of strings. We *could* special case those two. 
However, params etc can be arbitrary python objects. Which is why I just chose to use serde-pickle (or serde_json would *probably* work aswell) so I don't really have to handle that, since I did not want to roll my own code if there's already tested crates that cover most of the work.
If you have a look at the Julia script implementation, you'll find that values are encoded to Julia types, manually (on the python side) but restricted to lists, dicts, strings, ints, floats (at least that's the types I remember off the top of my head). We could do that aswell, but would still have to have some Value type (probably an enum).


> One potential workaround to the mixed types situation could be, if there are mixed types, everything becomes a `String`?

I personally feel this'd be both very inconvenient to work with (having to parse strings yourself, especially inconvenient for nested lists/dicts!) and wrong in the sense that you have the type information of those values at runtime on the python side, just to throw it all away and move the burden of knowing those types to the person writing a rust script.

Also see [NamedList](https://github.com/snakemake/snakemake/pull/1053#issuecomment-864419002) above, which sadly adds another dependency (indexmap), but then having both positional and named values at the same time becomes very easy.

The more I think of it, the better it is to stick with serde for now (we can exchange that with our own code later if needed) but provide convenience traits/wrappers for `serde_pickle::Value`,
so instead of 
```rust
if let Value::F64(some_param) = params["this_is_a_float"] { do_stuff_with(some_param) }
```
it's also possible to do
```rust
let some_param = params["this_is_a_float"].float()?;
```
or
```rust
let some_param: f64 = params.get("this_is_a_float")?;
```

mbhall88:
I agree with everything you've just said @tedil :+1:

mbhall88:
I've just pushed an update with outer line doc comment capabilities (as per https://github.com/fornwall/rust-script/issues/25). In it I've also pinned the minimum rust-script version to the latest version which includes that fix so the build will fail until https://github.com/conda-forge/rust-script-feedstock/pull/1 is merged. I have tested locally and it should pass once the conda package is updated

tedil:
> I've just pushed an update with outer line doc comment capabilities (as per [fornwall/rust-script#25](https://github.com/fornwall/rust-script/issues/25)). In it I've also pinned the minimum rust-script version to the latest version which includes that fix so the build will fail until [conda-forge/rust-script-feedstock#1](https://github.com/conda-forge/rust-script-feedstock/pull/1) is merged. I have tested locally and it should pass once the conda package is updated

Thanks for riasing that issue with rust-script, v0.15 should now be available on conda-forge ;)

mbhall88:
Sadly it seems like there is currently no way to redirect stderr/stdout from inside a rust process that is OS cross-compatible<sup><a href="https://stackoverflow.com/questions/54429389/how-to-redirect-stderr-from-inside-the-process-in-rust">1</a></sup>.  
To that end, I have decided to redirect stderr and stdout to the log (if there is one) during the call to rust-script.   
I also changed the type of log within the snakemake object as it [can also have multiple named files](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#log-files) (like input and output).

tedil:
> Sadly it seems like there is currently no way to redirect stderr/stdout from inside a rust process that is OS cross-compatible[1](https://stackoverflow.com/questions/54429389/how-to-redirect-stderr-from-inside-the-process-in-rust).
> To that end, I have decided to redirect stderr and stdout to the log (if there is one) during the call to rust-script.
> I also changed the type of log within the snakemake object as it [can also have multiple named files](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#log-files) (like input and output).

Okay, too bad, but reasonable "workaround" ;)

Btw, it seems the outer doc test file is missing; I think the directory the script tests are in are excluded by gitignore, so will have to be force added (or we could do some restructuring and refactoring to have them in a different place altogether)

tedil:
@mbhall88 looks like the script file itself is also missing

mbhall88:
> @mbhall88 looks like the script file itself is also missing

:facepalm:

johanneskoester:
> Sadly it seems like there is currently no way to redirect stderr/stdout from inside a rust process that is OS cross-compatible[1](https://stackoverflow.com/questions/54429389/how-to-redirect-stderr-from-inside-the-process-in-rust).
> To that end, I have decided to redirect stderr and stdout to the log (if there is one) during the call to rust-script.
> I also changed the type of log within the snakemake object as it [can also have multiple named files](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#log-files) (like input and output).

The gag crate allows this: https://docs.rs/gag

johanneskoester:
> > I'm keen to get stuck into another element of the implementation. What are you working on @tedil and what would you consider highest priority for me to get stuck into?
> 
> > There's a couple of items from your previous list that I am unclear on
> > ```
> > * log to the correct channels
> > 
> > * consider paths/PATH/syspath etc
> > ```
> 
> These might also have been python specific. I used the python script functionality implementation as a rough guide, so perhaps these aren't issues we have to deal with in the rust script case. If the channels used for logging are always stdout and stderr, we probably don't have to change anything; but if they are to be redirected (to a file, some logger somewhere else, etc) 🤷
> 
> > Also, regarding our dependency on serde (and derive feature), I had the original idea of just using some kind of template to generate the snakemake data structure directly from the python one. This is obviously a lot more work, but would allow us to directly encode the correct types without needing to use `Value`. I understand the main roadblock to this is python dictionaries with mixed types? Are there any other problems with this approach you've thought of?
> 
> Okay so: I talked with Johannes on Friday: The `input` and `output` values are ever only paths/strings, i.e.: there can be positional values and/or key-value pairs, where values are either single strings or lists of strings. We _could_ special case those two.
> (Also see [NamedList](https://github.com/snakemake/snakemake/pull/1053#issuecomment-864419002) above, which sadly adds another dependency (indexmap), but then having both positional and named values at the same time becomes very easy.)
> 
> However, params etc can be arbitrary python objects. Which is why I just chose to use serde-pickle (or serde_json would _probably_ work aswell) so I don't really have to handle that, since I did not want to roll my own code if there's already tested crates that cover most of the work.
> If you have a look at the Julia script implementation, you'll find that values are encoded to Julia types, manually (on the python side) but restricted to lists, dicts, strings, ints, floats (at least that's the types I remember off the top of my head). We could do that aswell, but would still have to have some Value type (probably an enum).
> 
> > One potential workaround to the mixed types situation could be, if there are mixed types, everything becomes a `String`?
> 
> I personally feel this'd be both very inconvenient to work with (having to parse strings yourself, especially inconvenient for nested lists/dicts!) and wrong in the sense that you have the type information of those values at runtime on the python side, just to throw it all away and move the burden of knowing those types to the person writing a rust script.
> 
> The more I think of it, the better it is to stick with serde for now (we can exchange that with our own code later if needed) but provide convenience traits/wrappers for `serde_pickle::Value`,
> so instead of
> 
> ```rust
> if let Value::F64(some_param) = params["this_is_a_float"] { do_stuff_with(some_param) }
> ```
> 
> it's also possible to do
> 
> ```rust
> let some_param = params["this_is_a_float"].float()?;
> ```
> 
> or
> 
> ```rust
> let some_param: f64 = params.get("this_is_a_float")?;
> ```

Mhm, changing the params conversion at a later stage does not sound like a good idea to me, because it will be a breaking change. I think I'd prefer to have conversion (e.g. float -> f64, etc.) within Snakemake as we have for the other languages, because that will be more convenient on the rust side.

docs/snakefiles/rules.rst
```diff
@@ -713,6 +699,98 @@ In the R Markdown file you can insert output from a R command, and access variab
...
+            "path/to/script.rs"
+
+The ability to execute Rust scripts is facilitated by |rust-script|_. As such, the
+script must be a valid ``rust-script`` script.
+
+In the Rust script, an instance of a ``Snakemake`` struct can be obtained via
+
+.. code-block:: rust
+
+    let snakemake = Snakemake::load()?;
```
johanneskoester, line 747: Why is this needed? Can't it be done in the preamble?

tedil:
> Mhm, changing the params conversion at a later stage does not sound like a good idea to me, because it will be a breaking change. I think I'd prefer to have conversion (e.g. float -> f64, etc.) within Snakemake as we have for the other languages, because that will be more convenient on the rust side.

The "conversion" already happens in Snakemake when pickling. There's *no* `String -> f64` conversion on the rust side. It's just a question of how to make the enum variants most conveniently accessible.

docs/snakefiles/rules.rst
tedil: You can have scripts both with and without an explicit main fn. In the latter case, it sure could be done in the preamble, in the former case it's not that easy in general. There probably is a nice way to solve this, I'm just not sure yet what that'd look like.
*If* we ditched serde (be it -pickle or -json) and provided our own conversion to some custom types (purely on the snakemake side), we could probably make the snakemake instance be constant / a singleton / whatever.

johanneskoester: But you should be able to use lazy_static for this: https://docs.rs/lazy_static
Should work in both cases.

johanneskoester:
> > Mhm, changing the params conversion at a later stage does not sound like a good idea to me, because it will be a breaking change. I think I'd prefer to have conversion (e.g. float -> f64, etc.) within Snakemake as we have for the other languages, because that will be more convenient on the rust side.
> 
> The "conversion" already happens in Snakemake when pickling. There's _no_ `String -> f64` conversion on the rust side. It's just a question of how to make the enum variants most conveniently accessible.

Right, if we stick to the map-like pickled structure, this is unavoidable. It would be more convenient to generate a struct from scratch though, right? Something like

```
struct Params {
    a: f64
    b: String
}

Snakemake {
    ...
    params: Params { a: 0.1, b: "foo".to_owned() }
}
```

Then, one could simply write

```
snakemake.params.a
```

for accessing, analogous to the Python case.

tedil:
> > > Mhm, changing the params conversion at a later stage does not sound like a good idea to me, because it will be a breaking change. I think I'd prefer to have conversion (e.g. float -> f64, etc.) within Snakemake as we have for the other languages, because that will be more convenient on the rust side.
> > 
> > 
> > The "conversion" already happens in Snakemake when pickling. There's _no_ `String -> f64` conversion on the rust side. It's just a question of how to make the enum variants most conveniently accessible.
> 
> Right, if we stick to the map-like pickled structure, this is unavoidable. It would be more convenient to generate a struct from scratch though, right? Something like
> 
> ```
> struct Params {
>     a: f64
>     b: String
> }
> 
> Snakemake {
>     ...
>     params: Params,
> }
> 
> lazy_static! {
>     static ref snakemake: Snakemake = Snakemake {
>         ...
>         params: Params { a: 0.1, b: "foo".to_owned() }
>     };
> }
> ```
> 
> Then, one could simply write
> 
> ```
> snakemake.params.a
> ```
> 
> for accessing, analogous to the Python case.

Yes, that should work, let's do that.

tedil:
After some searching, I found [json_typegen](https://github.com/evestera/json_typegen) which can generate rust structs from json representation. I've hence ditched serde-pickle in favour of serde_json + json_typegen, such that fields can be accessed by name and values aren't wrapped in a Value enum but have a primitive or string type.

Snakemake named lists are converted to `dict(keys=keys, values=values, named_param_1=42, …, some_other_named_param="bar")`, such that we can generate implementations which allow both indexing by index and name, as well as iteration over the values.
E.g.:

```rust
let input = &snakemake.input;
assert_eq!(input[0], "foo");
assert_eq!(input.some_named_input, "foo");
assert_eq!(input["some_named_input"], "foo");
for value in input {
    assert_eq!(value, "foo");
}
```

tedil:
The current version now has iterator and index implementations in input/output/wildcards *for positional arguments only*. Everything else *must* be accessed by field access.

As for redirecting stdout with gag: For scripts without a main fn (i.e. those that are implicitly enclosed in a main fn), we can just add redirects to the preamble, for scripts with an explicit main fn, we'd have to inject that at the start of the main fn.
Not sure how easy it is to write some robust piece of code that finds that (since `fn main() { ... }`, `fn main() -> Result<Whatever, SomeErrorType>` or even `#[derive(paw)] fn main …` are all valid entrypoints).
Also, what do we redirect, both stdout and stderr? Where do we redirect them?

johanneskoester:
> The current version now has iterator and index implementations in input/output/wildcards _for positional arguments only_. Everything else _must_ be accessed by field access.

good

 
> As for redirecting stdout with gag: For scripts without a main fn (i.e. those that are implicitly enclosed in a main fn), we can just add redirects to the preamble, for scripts with an explicit main fn, we'd have to inject that at the start of the main fn.
> Not sure how easy it is to write some robust piece of code that finds that (since `fn main() { ... }`, `fn main() -> Result<Whatever, SomeErrorType>` or even `#[derive(paw)] fn main …` are all valid entrypoints).
> Also, what do we redirect, both stdout and stderr? Where do we redirect them?

For the other script types, we do not have automatic redirects, but the script author has to do it. I would think that is reasonable for rust as well (also helps with transparency when just reading). Hence, you do not need to detect these cases. Just a simple helper function that could be used like `snakemake.redirect_stderr(snakemake.log[0])` would suffice.

tedil:
I updated the preamble to use fully qualified names/paths, to avoid confusion. The docs also got some updates.
@mbhall88 the rust-outer-doc-line test fails at the moment, could you look into that?

mbhall88:
Sorry for the radio silence

I agree that it is probably best to leave it to the user to do what they like with the error stream - I don't know why I was redirecting the rust-script streams to the log file, that seems like a terrible idea in hindsight.

> The current version now has iterator and index implementations in input/output/wildcards for positional arguments only. Everything else must be accessed by field access.

The log should also be included. I've added it in https://github.com/snakemake/snakemake/pull/1053/commits/fca4aaf5071d71ac817c6334b84f40ec8b719ce4 and am testing it in one of the rust test scripts.

One thing that is bugging me is a compiler warning

```
warning: static variable `snakemake` should have an upper case name
   --> tmpg1ff5cx8.test-manifest.rs:122:16
    |
122 |     static ref snakemake: Snakemake = {
    |                ^^^^^^^^^ help: convert the identifier to upper case: `SNAKEMAKE`
    |
    = note: `#[warn(non_upper_case_globals)]` on by default

warning: 1 warning emitted
```

We have an allow attribute on the preceding line so I wonder if this is a rust bug?

tedil:
> Sorry for the radio silence
> 

No harm done ;)

> I agree that it is probably best to leave it to the user to do what they like with the error stream - I don't know why I was redirecting the rust-script streams to the log file, that seems like a terrible idea in hindsight.
> 
> > The current version now has iterator and index implementations in input/output/wildcards for positional arguments only. Everything else must be accessed by field access.
> 
> The log should also be included. I've added it in [fca4aaf](https://github.com/snakemake/snakemake/commit/fca4aaf5071d71ac817c6334b84f40ec8b719ce4) and am testing it in one of the rust test scripts.

Good point, thanks!

> One thing that is bugging me is a compiler warning
> 
> ```
> warning: static variable `snakemake` should have an upper case name
>    --> tmpg1ff5cx8.test-manifest.rs:122:16
>     |
> 122 |     static ref snakemake: Snakemake = {
>     |                ^^^^^^^^^ help: convert the identifier to upper case: `SNAKEMAKE`
>     |
>     = note: `#[warn(non_upper_case_globals)]` on by default
> 
> warning: 1 warning emitted
> ```
> 
> We have an allow attribute on the preceding line so I wonder if this is a rust bug?

Yes, it's indeed [a bug](https://github.com/rust-lang-nursery/lazy-static.rs/issues/153), I think I have linked that bug in the source as well, so once it's resolved, we can remove that link.

tedil:
Testing fails because of some github actions missing, I guess that is due to merging the changes from main into this branch?

(We might also want to modularize script.py into `script/julia.py`, `script/r.py` etc. But that's out of scope for this PR.)

mbhall88:
What's left to do now? Is it just docs? I'm happy to tackle that in the coming days? Are there any special requests for things to add to the docs?

tedil:
> What's left to do now? Is it just docs? I'm happy to tackle that in the coming days? Are there any special requests for things to add to the docs?

I *think* there's actually not much more to do now; if you could have a look at the docs, that would be great!

mbhall88:
Docs look great to me. 

The only two things for future would be the rust-script templates and also adding Rust support in jupyter notebooks (as mentioned in https://github.com/snakemake/snakemake/issues/913#issue-830816514 probably best done with `evcxr`)

Also, CI has been failing due to

```
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/_actions/GoogleCloudPlatform/github-actions/master/setup-gcloud'. Did you forget to run actions/checkout before running your local action?
```

mbhall88 (approved):

snakemake/script.py
```diff
@@ -823,6 +870,371 @@ def execute_script(self, fname, edit=False):
...
+        )
+
+    def get_preamble(self):
+        wrapper_path = self.path[7:] if self.path.startswith("file://") else self.path
+        # preamble_addendum = (
+        #     "__real_file__ = __file__; __file__ = {file_override};".format(
+        #         file_override=repr(os.path.realpath(wrapper_path))
+        #     )
+        # )
+        preamble_addendum = ""
```
tedil, line 1082: Can we remove this altogether?

johanneskoester (approved):
Awesome work @mbhall88 and @tedil! This is a massive step forward! For the first time, we have integrated ad-hoc high performance scripting in a workflow management system.

PR #1054, opened 2021-06-17
Groupjobs gls

sposadac:
### Description

Allow using Google Life Science preemptible rules and job grouping, by checking if any of the rules in a given group has been flagged for execution using preemptible instances. 

### QC


docs/snakefiles/rules.rst
```diff
@@ -1331,6 +1331,8 @@ However, if we would add ``group: "mygroup"`` to rule ``c``, all jobs would end
 Alternatively, groups can be defined via the command line interface.
 This enables to almost arbitrarily partition the DAG, e.g. in order to safe network traffic, see :ref:`here <job_grouping>`.
 
+For execution on the cloud using Google Life Science API and preemptible instances, note than if any of the rules in a group is included in the subset of rules to be executed using preembtible instances (with command line option ``--preemptible-rules``), the group will be also scheduled to be executed on a preemptible instance.
```
johanneskoester, line 1334: Mhm, I wonder if that is really the desired behavior? Shouldn't it rather be: all jobs must be preemptible for the group to be preemptible as well?

sposadac: I don't have a strong opinion, my only thought went into cost optimization. If one were to mark a rule as "preemptible" because being somehow expensive, I would like to make sure is executed in a preemptible instance.

That said, I wouldn't mind to replace the `any` by the `all` condition. Let me know if it is still your preference

johanneskoester: THinking about it, we could also fail with a WorkflowError in case not all jobs are preemptible, and request people to set this homogeneously for all jobs in the group, pointing to the corresponding section in the docs.

johanneskoester: I think this is best, because peoples credits depend on that behavior, so we'd rather not imply something that will cost more money than expected.

sposadac: It sounds reasonable to me!. I implemented the few edits.

johanneskoester (approved):

Issue #1062, opened 2021-06-25
FTPRemoteProvider complains about ftps:// url even when ftp:// is specified


lparsons:
**Snakemake version**

Bug appears starting in version 6.4.0 and exists in 6.4.1, 6.5.0, and 6.5.1

**Describe the bug**
Attempting to download a file using the `FTPRemoteProvider` results in an error.

**Logs**
```
SyntaxError:
encrypt_data_channel=False cannot be used with a ftps:// url
  File "/home/lparsons/Documents/projects/MitoRiboSeq/code/mito_readphasing_metagene.snakefile", line 16, in <module>
  File "/home/lparsons/Documents/projects/MitoRiboSeq/code/rules/genome.smk", line 15, in <module>
  File "/home/lparsons/miniconda3/envs/MitoRiboSeq/lib/python3.8/site-packages/snakemake/remote/FTP.py", line 70, in remote
```

**Minimal example**
```
from snakemake.remote.FTP import RemoteProvider as FTPRemoteProvider    
FTP = FTPRemoteProvider()    
    
rule all:    
    input:    
        "genome/Homo_sapiens.GRCh38.dna_sm.primary_assembly.fa.gz"    
    
rule get_fasta_ftp:    
    input:    
        FTP.remote("ftp://ftp.ensembl.org/pub/release-100/fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna_sm.primary_assembly.fa.gz", keep_local=True)    
    output:    
        "genome/Homo_sapiens.GRCh38.dna_sm.primary_assembly.fa.gz"    
    shell:    
        "mv {input:q} {output:q}"    
```

**Additional context**

johanneskoester:
@jmeppley this seems to be related to one of your commits (30877ff10). Could you have a look?

jmeppley:
Yes. That's definitely from my PR. I can't tell if it's entirely my fault, or if I just did a bad job fixing a previous bug. Either way, I think it's a simple fix.

I have a fix [here](https://github.com/jmeppley/snakemake/tree/issue_1062) that seems to work.

I'll submit a PR as soon as I have the test set up.

jmeppley:
@johanneskoester It looks like it passed all the checks and can be merged

johanneskoester:
Fixed by PR #1068!

PR #1068, opened 2021-06-29
Bugfix for Issue #1062: correct a typo in parameter checks

jmeppley:
### Description

Fixed a typo in the parameter checks for remote.FTP that seems to be causing issue #1062

### QC

### DIFF
--- a/tests/test_github_issue1062/Snakefile
+++ b/tests/test_github_issue1062/Snakefile
@@ -1,0 +1,18 @@
+from snakemake.remote.FTP import RemoteProvider as FTPRemoteProvider    
+FTP = FTPRemoteProvider()    
+    
+rule all:    
+    input:    
+        "robots.txt"
+
+FTP_URL = \
+    "ftp://ftp.ensembl.org/pub/release-100/fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna_sm.primary_assembly.fa.gz"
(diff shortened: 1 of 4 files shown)


tests/test_github_issue1062/Snakefile
```diff
@@ -0,0 +1,18 @@
+from snakemake.remote.FTP import RemoteProvider as FTPRemoteProvider    
+FTP = FTPRemoteProvider()    
+    
+rule all:    
+    input:    
+        "robots.txt"
+
+FTP_URL = \
+    "ftp://ftp.ensembl.org/pub/release-100/fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna_sm.primary_assembly.fa.gz"
```
lparsons, line 9: Thanks for tracking this down and getting a fix in so quickly. Just a note that the file here is a bit large, you could try something like `ftp://ftp.ensembl.org/pub/release-100/fasta/homo_sapiens/dna/README` instead.

jmeppley: Nice catch. I meant to use the robots.tx file. It's smaller and less likely to be removed. I was just switching back and forth between it and the file from the bug report...and I checked in the wrong one.

johanneskoester (approved):
Awesome, thank you!

Issue #1073, opened 2021-07-02
.jobfinished file Permission denied, NFS latency?


jeskowagner:
**Update 20210705** implemented a possible fix in PR #1078 
**Update 20210715** the proposed fix did not permanently remove the issue. I have since altered my workflow to have longer-running jobs and therefore have no encountered the issue again. 

Hi and thanks for developing snakemake, which is incredibly useful!

I am encountering an issue related to the `.jobfinished` files and possibly linked to filesystem latency in an HPC environment. When running a pipeline with ~100 parallel jobs on a cluster using slurm, this error pops up every few thousand jobs, particularly with short jobs (<10 s).

snakemake version: 6.5.1
slurm version: 20.11.7
filesystem: nfs4
OS: CentOS 7.6.1810
[Relevant excerpt of log here](https://github.com/snakemake/snakemake/files/6763275/log.txt)

This error is thrown in the console and snakemake freezes, with no new jobs being submitted nor the task quitting (I kill it manually in this case). Restarting the pipeline overcomes the problem, and new jobs are being submitted just fine. The job causing the issue ran fine, with the output being the expected. Before killing the snakemake task, I manually checked the `.jobfinished` file and permissions are `rw-r--r--`, which makes me think this might be a filesystem latency issue. 

I have been unable to reproduce the issue with a minimal example unfortunately.

Expected behavior:
Option A: snakemake quitting and passing the Permission denied error.
Option B: snakemake respecting the `--latency-wait` parameter also for `.jobfinished` files.

jeskowagner:
Due to a change in my workflow I no longer have such short running jobs and will therefore not be able to reproduce the issue anytime soon. If it does re-occur with a different pipeline I will comment anew on this issue.

PR #1078, opened 2021-07-05
Fix NFS latency causing denied permissions (#1073)

jeskowagner:
### Description
Changed how `.jobfinished`, `.jobscript` and `.jobfailed` files are removed. On NFS systems with considerable latency, in edge cases these files would show as existing but not writable and could not be removed (see #1073). This PR addresses that by wrapping the removal of the files in a `try/except` that forces write permissions if the removal fails.
I do not think this behavior is currently covered by a test. Likely it should not be anyway, given that it is a difficult-to-reproduce error. However, in the environment I worked in, this error occurred once every ~1500 steps in short running steps (5-15 s) and caused considerable headache. With this fix, 10,000 steps of the same pipeline ran through without issue. 

### QC

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -1098,32 +1098,28 @@ class GenericClusterExecutor(ClusterExecutor):
         else:
 
             def job_status(job):
-                if os.path.exists(active_job.jobfinished):
-                    try:
-                        os.remove(active_job.jobfinished)
-                    except PermissionError:
-                        os.chmod(active_job.jobfinished, stat.S_IWRITE)
-                        os.remove(active_job.jobfinished)
-                    try:
-                        os.remove(active_job.jobscript)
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):
Thank you so much, good catch!

jeskowagner:
Unfortunately the issue popped back up even with this proposed change. I will investigate further and report back, but will depend on the condition to occur as I am still lacking a reproducible example.

jeskowagner:
As mentioned in #1073 I have not been able to reproduce this issue again and likely will not be able to anytime soon. But since the fix in this PR did not completely solve the issue I would recommend against merging the changes. My apologies I could not be of more help.

johanneskoester:
Ok, thank you!

johanneskoester:
Closing this for now.

PR #1111, opened 2021-07-28
Load config files using anyconfig

kpj:
### Description

This PR replaces Snakemake's custom (JSON, YAML) config parsing with the [anyconfig](https://github.com/ssato/python-anyconfig) module.

### QC

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ setup(
         "requests",
         "ratelimiter",
         "pyyaml",
+        "anyconfig",
         "configargparse",
         "appdirs",
         "datrie",
--- a/snakemake/io.py
+++ b/snakemake/io.py
@@ -1577,7 +1577,7 @@ class Log(Namedlist):
 
 def _load_configfile(configpath_or_obj, filetype="Config"):
     "Tries to load a configfile first as JSON, then as YAML, into a dict."
-    import yaml
+    import anyconfig
 
     if isinstance(configpath_or_obj, str) or isinstance(configpath_or_obj, Path):
         obj = open(configpath_or_obj, encoding="utf-8")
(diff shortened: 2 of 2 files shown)


DrYak:
@johanneskoester : do you have any idea why [even setting up the Linux environment](https://github.com/snakemake/snakemake/pull/1111/checks?check_run_id=3183313089) (before even running the tests themselves) fails in the GitHub actions?

DrYak:
*Note:* In the current form, there is still a caveat when loading INI-style python configparser files (it's not anyconfig specific, it's a general caveat with python configparser's format) compared to the "standard" YAML/JSON used in Snakemake:
 - by default configparser's entries are case-insensitive (and converted to lowercase internally), whereas JSON/YAML are case-sensitive.
 - configparser's stores everything as a strings and puts the onus of converting types to the program (a bit similar to strictyaml), whereas YAML can auto-guess types, and JSON has type as part of its syntax.
 
The way we currently handle this [in our pipeline is to validate the structure against a schema](https://github.com/cbg-ethz/V-pipe/blob/fbe013aba1f95969914751339b249b5a7f381ed5/rules/common.smk#L40).

What would be needed to perfect this pull request is that:
 - from the snakemake side of things, we give a possibility to pass a schema to the loader.
 - from the anyconfig side of things, we should send a PR to ssato for handling similar scheme-based INI validations.
 
 But if the user only uses strings and lower-case entries, the current PR works as-is, including config file merging (well as long as PR #1126 is also added ;-) ) and overriding.
And also adds "for-free" support for a bunch of other standard formats such as XML for people who might want those.

kpj:

cademirch:
Snakemake has evolved substantially since this was opened. If you'd like to revive this work, a new PR against the current main branch would be the best path forward.

Thank you for your patience and for contributing to the project.

Issue #1106, opened 2021-07-25
AmbiguousRuleException when there is no ambiguity


dkuzminov:
**Snakemake version**
6.5.3

**Describe the bug**
Snakemake may throw an AmbiguousRuleException when there is no ambiguity. This bug depends on the processing order of rules.

**Minimal example**
To reproduce the bug the file *non_existing_file.txt* shall not be present while the file *optional_input.txt* must exist.
```python
rule all:
    input: "target.txt"

rule from_non_existing_file:
    input: "non_existing_file.txt"
    output: "target.txt"

rule broad_input:
    output: "target.txt"

rule narrow_input:
    input: "optional_input.txt"
    output: "target.txt"

ruleorder: narrow_input > broad_input
```
If this script is being run in the `--dry-run` mode multiple times, most of the times it would correctly resolve the ambiguities:
```
Job counts:
        count   jobs
        1       all
        1       narrow_input
        2
```
In 1/6 runs however it would throw an exception:
```
AmbiguousRuleException:
Rules narrow_input and broad_input are ambiguous for the file target.txt.
Consider starting rule output with a unique prefix, constrain your wildcards, or use the ruleorder directive.
Wildcards:
        narrow_input:
        broad_input:
Expected input files:
        narrow_input: optional_input.txt
        broad_input: Expected output files:
        narrow_input: target.txt
        broad_input: target.txt
```

**Additional context**
In the example above there are 3 rules that claim *target.txt* as their output. One rule (`from_non_existing_file`) should be eliminated because of missing input. For the other two there is an explicit `ruleorder` defined, and the `narrow_input` shall be preferred if the file *optional_input.txt* exists. However this behaviour is not stable and in approx. 1/6 runs Snakemake forgets of the explicit `ruleorder` defined.

sebschmi:
Seems to be the same as I reported already some time ago :/ #1023

dkuzminov:
> Seems to be the same as I reported already some time ago :/ #1023

Actually I've found the problem and have a fix. Waiting for the permission from my employer to make a PR.

PR #1124, opened 2021-08-06
fix: AmbiguousRuleException bug caused by weak ordering of rules

dkuzminov:
Link to the issue: https://github.com/snakemake/snakemake/issues/1106
Bug description on StackOverflow: https://stackoverflow.com/questions/68514884/ambiguousruleexception-when-there-is-no-ambiguity

### Description

The AmbiguousRuleException is being thrown whenever there is no ambiguity. This behaviour is not deterministic, and there is no simple way to predict whether the invocation of the pipeline would produce this error.

The actual problem is incorrect algorithm of rule selection out of the list of possible producers. This list is being sorted, but there is no strict order of the rules unless explicitly specified by `ruleorder`. The rule that has no explicit `ruleorder` is neither greater nor smaller than other rules, and the result of sorting such a list becomes incorrect. Consider three rules: smallest, neutral, greatest. There is an explicit `ruleorder: greatest > smallest`. That means that:
```
__lt__(smallest, greatest) == True
__lt__(greatest, smallest) == False
__lt__(smallest, neutral) == __lt__(neutral, smallest) == __lt__(greatest, neutral) == __lt__(neutral, greatest) == False
```

That the result of `sorted([smallest, neutral, greatest])` would be `[smallest, neutral, greatest]` while the result of `sorted([greatest, neutral, smallest])` would be `[greatest, neutral, smallest]`, which is obviously incorrect.

This commit addresses the issue and changes the logic of producer selection removing the dependency on the result of incorrect sorting.

### QC
The changes cannot be verified with a deterministic test case because the issue depends on the random order of rules processing. Anyway, the change shall not break other test cases.

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -731,9 +731,8 @@ class DAG:
             visited = set()
         if known_producers is None:
             known_producers = dict()
-        producer = None
+        producers = []
         exceptions = list()
-        jobs = sorted(jobs, reverse=not self.ignore_ambiguity)
         cycles = list()
 
         for job in jobs:
@@ -756,12 +755,7 @@ class DAG:
                 )
                 # TODO this might fail if a rule discarded here is needed
                 # elsewhere
-                if producer:
-                    if job < producer or self.ignore_ambiguity:
-                        break
-                    elif producer is not None:
-                        raise AmbiguousRuleException(file, job, producer)
-                producer = job
+                producers.append(job)
             except (
                 MissingInputException,
                 CyclicGraphException,
@@ -783,7 +777,7 @@ class DAG:
                     if file
                     else "",
                 )
-        if producer is None:
+        if not producers:
             if cycles:
                 job = cycles[0]
                 raise CyclicGraphException(job.rule, file, rule=job.rule)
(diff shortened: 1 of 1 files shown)


snakemake/dag.py
```diff
@@ -783,29 +777,35 @@ def update(
...
-                    exception=WorkflowError(*exceptions),
-                )
-            )
 
         n = len(self.dependencies)
         if progress and n % 1000 == 0 and n and self._progress != n:
             logger.info("Processed {} potential jobs.".format(n))
             self._progress = n
 
+        producers.sort(reverse=True)
```
johanneskoester, line 794: So far I could follow, but can't we run into the described sorting issue here again?

dkuzminov: No, we cannot run into the same problem.
What we actually need is the producer that strictly dominates other possible producers. That means that it shall be comparable to any other producer and shall be larger than each one. If such rule exists, it will get to the first position in the list (line 795). If there is no such rule, then we have an ambiguity (actually we may have multiple ambiguous rules). The lines 796-798 are needed just to produce a meaningful error message (not to complain on two rules that are not actually ambiguous).

dkuzminov: Even simpler would be to describe the issue in different words. Current code sorts first and then eliminates non-producing rules. My solution eliminates non-producing first and sorts after that. If the first case the order is not predictable and can be broken by "spoilers": the rules that would be eliminated later. In my solution the spoilers are removed first, and then sorting gives us some guarantees. First guarantee: if the unambiguous rule exists, it is on the first position. Second guarantee: whenever there is an ambiguity, the item on the first position shall be non-compairable with at least one rule on another position.

johanneskoester: Ah, right, that makes sense.

johanneskoester (approved):

PR #1126, opened 2021-08-09
fix: Use 'snakemake.utils.update_config' instead of 'dict.update'

kpj:
### Description

When merging configs, Snakemake uses `dict.update` instead of `snakemake.utils.update_config` which is not recursive.
Nested config options can thus disappear unexpectedly.

### QC


kpj:

johanneskoester:
Soemthing fails on windows now. Do you have a clue?

johanneskoester:
Thanks a lot, finally done :-).

kpj:
Wonderful! Thanks a lot for the debugging as well :-)

PR #1127, opened 2021-08-09
fix: Ga4gh tes bugfixes

svedziok:
### Description

This PR contains two bugfixes
- Adds missing py-tes package, which is required for GA4GH job submission
- Resolve bug with remote log or benchmarkfiles

### QC

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,7 @@ setup(
         "filelock",
         "stopit",
         "tabulate",
+        "py-tes",
     ],
     extras_require={
         "reports": ["jinja2", "networkx", "pygments", "pygraphviz"],
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,6 @@ setup(
         "filelock",
         "stopit",
         "tabulate",
-        "py-tes",
     ],
     extras_require={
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -76,6 +76,7 @@
         "filelock",
         "stopit",
         "tabulate",
+        "py-tes",
```
johanneskoester, line 79: I don't think this needs to be a required dependency. But the docs should probably say that py-tes needs to be installed.

johanneskoester: And the import statement [here](https://github.com/snakemake/snakemake/pull/1127/files#diff-a1865e2681ce5d0b775d62c336fe1ac260c6855d91715e5bb033a2babaae830dR40) should be wrapped in a try-except with a `WorkflowError("Unable to import Python package tes. TES backend requires py-tes to be installed. Please install py-tes, e.g. via Conda or Pip.")`

johanneskoester (changes requested):

setup.py
svedziok: Ok, I added a try-except statement in the __init__ function. I will also a note at the documentation

johanneskoester (approved):

johanneskoester:
Thank you!

Issue #759, opened 2020-11-20
snakemake main job hangs indefinitely; no new jobs submitted on slurm in newer versions


Kelzor:
**Snakemake version**
Version ≥26.0 (and possibly other newer versions)

**Describe the bug**

I am running a large snakemake pipeline with ~90k steps/jobs through a slurm cluster with the following submission commands:

`snakemake --snakefile Snakefile -j 50 --use-conda --keep-target-files --keep-going --rerun-incomplete --latency-wait 30 --cluster "sbatch -A keblevin --mem=32768 -n 1 -c 8 -t 3:00:00 -e /scratch/keblevin/11_19_20_megapipe_MTBC_comp_data/slurm_outputs/slurm.%j.err -o /scratch/keblevin/11_19_20_megapipe_MTBC_comp_data/slurm_outputs/slurm.%j.out"
`

New jobs slowly stop being submitted such that after 30 minutes to an hour only 30 jobs are maintained in the queue, then 20, then 10, etc. After about 1-2 hours or 800-1500 jobs, new jobs are no longer submitted at all, and the main job hangs indefinitely. The slurm.err log has no errors and also supports the observation that the main job is just hanging without submitting new jobs. Here are the last 25 lines of a slurm.err, which remains un-updated until the job is manually cancelled. In this log, the main job ran over night ~8 hours without submitting any new jobs. I cancelled it the following morning.

**Logs**

Removing temporary output file output/New_Zealand_BLENHEIM_2000_3990/New_Zealand_BLENHEIM_2000_3990.sam.
[Tue Nov 17 23:32:05 2020]
Finished job 42963.
1080 of 86656 steps (1%) done
Removing temporary output file output/Germany_2012_2110/Germany_2012_2110-modern.1.sai.
Removing temporary output file output/Germany_2012_2110/Germany_2012_2110-modern.2.sai.
Removing temporary output file output/Germany_2012_2110/Germany_2012_2110-modern.1.trimmed.fq.
Removing temporary output file output/Germany_2012_2110/Germany_2012_2110-modern.2.trimmed.fq.
[Tue Nov 17 23:35:55 2020]
Finished job 30728.
1081 of 86656 steps (1%) done

[Tue Nov 17 23:36:03 2020]
rule sam_to_bam:
    input: output/Germany_2012_2110/Germany_2012_2110.sam
    output: output/Germany_2012_2110/Germany_2012_2110.bam
    jobid: 30727
    wildcards: sample=Germany_2012_2110

[...]

PR #1142, opened 2021-08-20
fix: improved error handling for cluster status scripts and smarter job selector choice in case of cluster submission (use greedy for single jobs).

johanneskoester:
### Description

See title. Could help with issue #759.

### QC


johanneskoester:
@vsoch do you have any idea why the GLS test suddenly fails here? I did not expect my changes to have any side effects on GLS.

vsoch:
The last step in the GLS test it looks like is now using this greedy selector, so that's the change:

> DEBUG    snakemake.logging:logging.py:553 Using greedy selector because only single job has to be scheduled.

This one here:

And IIRC this step was added to make sure that we can handle the directory creation:

https://github.com/snakemake/snakemake/blob/7f54c391f2821655ed168bcdafad6d07b96fcec7/tests/test_google_lifesciences/Snakefile#L39

I think it probably has to do with the fact that in this new function we try to calculate a size for the input directory, which won't have one / possibly hasn't been created yet.

https://github.com/snakemake/snakemake/blob/3530f328f1af572187a8e9a2db64a8191d3a31a8/snakemake/scheduler.py#L897

Which ultimately gets us here:

https://github.com/snakemake/snakemake/blob/3530f328f1af572187a8e9a2db64a8191d3a31a8/snakemake/jobs.py#L394

and then calls this size_local function on it:

https://github.com/snakemake/snakemake/blob/3530f328f1af572187a8e9a2db64a8191d3a31a8/snakemake/remote/GS.py#L220

So ultimately it's this call that is wrong and isn't aware of directories perhaps?

https://github.com/snakemake/snakemake/blob/cd5c58da5da3aff5692e7186244d67d638008d17/snakemake/io.py#L574

If I remember, we had to add a bunch of new logic to handle directories, and I think it would need to be added in here somewhere so we essentially don't stat something that is an input but doesn't exist yet.

That's my best guest without interacting with it directory - I would add directory handling, meaning some check using https://github.com/snakemake/snakemake/blob/3530f328f1af572187a8e9a2db64a8191d3a31a8/snakemake/io.py#L372. It looks like there is other logic in that file for that case in other places!

mbhall88:
I'm not sure if it's caused by these changes or not, but I've noticed recently that snakemake's status checking for cluster jobs is very delayed.

For example, in the LSF profile, if we can't get the job status using `bjobs -o stat <jobid>` - because it completed a while ago - then we go looking in the log file. I've noticed with a recent pipeline that has ~29k jobs that it spends a lot of time submitting jobs and then after a while, it tries to go back and check the status of the earlier jobs and they've now dropped out of the bjobs history.

I hope this makes sense? Basically I'm asking the there is some kind of new priority between submitting and status-checking?

SichongP:
Dos bhist work? https://www.ibm.com/support/pages/finding-historical-job-information-old-completed-jobs

mbhall88:
Sort of. It looks at the last 100 lsb.events files. Which means potentially opening 100 files. This is much slower than our current solution of looking in the job's cluster log file. Anway, that's not really the point of my question. I was just wanting to know if the submission/status priority has changed.

PR #1147, opened 2021-08-23
perf: improve job selection performance in case of potential ambiguity that is resolved by comprehensive ruleorder statements.

johanneskoester:
### QC

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -734,6 +734,11 @@ class DAG:
         producers = []
         exceptions = list()
         cycles = list()
+        
+        # check if all potential producers are strictly ordered
+        jobs = sorted(jobs, reverse=True)
+        primary_job = jobs[0]
+        is_strictly_ordered = all(primary_job > job for job in jobs)
 
         for job in jobs:
             logger.dag_debug(dict(status="candidate", job=job))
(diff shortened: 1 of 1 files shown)


johanneskoester:
@dkuzminov you probably want to have a look at this?

snakemake/dag.py
```diff
@@ -735,6 +735,11 @@ def update(
         exceptions = list()
         cycles = list()
 
+        # check if all potential producers are strictly ordered
+        jobs = sorted(jobs, reverse=True)
+        primary_job = jobs[0]
+        is_strictly_ordered = all(primary_job > job for job in jobs)
```
dkuzminov, line 741: This is not correct, as the first element shall be compared to the rest of the elements:
```python
is_strictly_ordered = all(primary_job > job for job in jobs[1:])
```

dkuzminov (changes requested):
Overall this improvement is incorrect. Assume there is a job `primary` that dominates all other jobs. Next, imagine that this rule is disregarded later because of missing input. We would get an incorrect `is_strictly_ordered` value because we know nothing about the ordering of the rest of the jobs. We would select the first producer even if it is ambiguous.

snakemake/dag.py
johanneskoester: good catch!

johanneskoester:
> Overall this improvement is incorrect. Assume there is a job `primary` that dominates all other jobs. Next, imagine that this rule is disregarded later because of missing input. We would get an incorrect `is_strictly_ordered` value because we know nothing about the ordering of the rest of the jobs. We would select the first producer (other than `primary`) even if it is ambiguous.

Right, I missed that case indeed. I have worked a bit on it. Still not 100% validated in my mind, but I think this is more correct (although it is 20pm here now, so I might miss something again ;-).

johanneskoester:
I think it is fine.

Issue #1152, opened 2021-08-25
Full copy shadow rules


sebschmi:
**Is your feature request related to a problem? Please describe.**
I run a large pipeline on a cluster. Some of my rules are normal computations, while others are benchmarks. I want to benchmark the performance of my tools as a whole, including file read and write times. However this requires me to use the local file system of a cluster node for that rule, as otherwise the variability in performance of the network file system will spoil my benchmark.

Therefore, I would like to copy the input files of a rule to the local disk before execution, and upon success copy the output and log files back to the network file system, deleting them from the local disk.

This should also be toggleable, such that I can run the workflow without copying for testing/running on my local machine/etc.

**Describe the solution you'd like**
The shadow feature appears to be a perfect candidate for this. The directive `shadow: "copy"` could cause a rule to behave as I described.

Then I could set the flag `--shadow-directory` to the local disk of the node I want to benchmark on. This path is only valid on that node though, so that should not cause an error. Otherwise, maybe a `shadow-directory: "..."` directive would be useful to set the path per rule.

**Describe alternatives you've considered**

 * I could copy the files manually in the shell script of each rule, and make it conditional based on if I am executing on a cluster or not. However that is tedious and error-prone.
 * I could just run the whole pipeline on the one node I use for benchmarking, but then I would not be able to use different nodes for the non-benchmark computations.

PR #1155, opened 2021-08-26
feat: Add `shadow: "copy-minimal"` directive

sebschmi:
"copy-minimal" is similar to "minimal", but it creates copies instead of symlinks of the files. This makes it possible to use local hard disks of a node in a cluster.

### Description

See #1152 for motivation and further explanation.

I implemented the feature and added tests and documentation.
The test is a simple copy of the `test_shadow` test, removing all non-minimal parts and changing the minimal parts to copy-minimal, as well as testing if the files are physical copies instead of symlinks in the rules.
In the documentation I simply searched for "shadow" and added the description about the new feature at the necessary places.

### QC


DrYak:
Big thanks for this feature, I've dreamed for such a feature. That will *so much* help our use case! (currently I have a jury-rigged rsync-based solution ).

PR #1160, opened 2021-08-29
Add quotes to tmp dir in wait-for-file options

fdchevalier:
Hello Snakemake team,

### Description

**Context**: Job submission to SGE cluster using `qsub` from a folder with a path containing a white space.

**Problem**: First batch of jobs is successfully submitted but snakemake issue an error regarding missing files. Below is the typical error issued when submitting a job from `$HOME/test test/':
```
Waiting at most 10 seconds for missing files.
Missing files after 10 seconds:
test/.snakemake/tmp.6tq2tga_
```
The error is due to the absence of quotes around the tmp directory path when generating the job script. The typical line for the the `wait-for-files` options is as follows:
```bash
--wait-for-files /XXXX/test test/.snakemake/tmp.6tq2tga_ test test2 --latency-wait 10 \
```
but should be:
```bash
--wait-for-files '/XXXX/test test/.snakemake/tmp.6tq2tga_' test test2 --latency-wait 10 \
```

**Solution**: Add quotes when appending the path to the `wait_for_files` list.

#### Comments

The solution has been tested in cases with white space and and with special character (&).

I am not fluent enough in Python so feel free to edit the code if needed.

#### Minimal reproducible example

In case this needs to be reproduced.
```bash
# Create working directory
mkdir "test test"
cd "test test"

# Create status directory for qsub and output directory
mkdir status
mkdir output

# Create test file
touch test

# Create snakefile
cat  << EOF > snakefile
rule test_rule:
    input:
        "test"
    output:
        "output/a.txt"
    shell:
        'sleep 10 && echo h=$(hostname) "{input}" > "{output}"'
EOF

# Submit job
snakemake --cluster "qsub -V -cwd -o status -j y" --jobs 20 -w 10

# Print status
cat status/*
```

### QC

Thank you for considering this patch.

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -769,7 +769,7 @@ class ClusterExecutor(RealExecutor):
         wait_for_files = []
         scheduler_solver_path = ""
         if self.assume_shared_fs:
-            wait_for_files.append(self.tmpdir)
+            wait_for_files.append("'" + self.tmpdir + "'")
             wait_for_files.extend(job.get_wait_for_files())
             # Prepend PATH of current python executable to PATH.
             # This way, we ensure that the snakemake process in the cluster node runs


fdchevalier:
I noticed that the bug was also affecting files with white spaces. So I came with a better solution to quote every item of the `wait_for_files` list.

I also tested the behavior of Snakemake when having more than 20 elements in the `wait_for_lists` which creates a file with the list of elements: in that case, the content of the file worked as expected (no need of quotes around each element) but the path to the file required quotes. I modified the code accordingly.

snakemake/executors/__init__.py
```diff
@@ -788,11 +788,11 @@ def format_job(self, pattern, job, **kwargs):
 
             waitfiles_parameter = format(
                 "--wait-for-files-file {wait_for_files_file}",
-                wait_for_files_file=wait_for_files_file,
+                wait_for_files_file=('"' + wait_for_files_file + '"'),
```
johanneskoester, line 791: For quoting a string, you can just use `repr`, which will also automatically take care of escaping quotes that are already occuring within the strings.

snakemake/executors/__init__.py
```diff
@@ -788,11 +788,11 @@ def format_job(self, pattern, job, **kwargs):
...
             )
         else:
             waitfiles_parameter = format(
-                "--wait-for-files {wait_for_files}", wait_for_files=wait_for_files
+                "--wait-for-files {wait_for_files}", wait_for_files=(" ".join(['"{0}"'.format(x) for x in wait_for_files])),
```
johanneskoester, line 795: Same as above.

johanneskoester (changes requested):
Thanks a lot! Very good idea. Python allows to make the changed code a bit simpler, see below.

snakemake/executors/__init__.py
johanneskoester: You need this: `wait_for_files=[repr(f) for f in wait_for_files]`, i.e., apply it to each item of the list, not the entire list. Otherwise you get the representation of the entire list as a single string, this is not what you want.

fdchevalier: Hi Johannes,

Thank you  for your guidance and patience. I definitely learned something. I apply the `repr()` on both section and tested the new version. Works like a charm. PR is now updated.

Thank you again,

Fred

snakemake/executors/__init__.py
fdchevalier: Added, see below. Thank you.

snakemake/executors/__init__.py
johanneskoester: Great! Now it just needs a final black reformatting :-).

fdchevalier: Done! (did not know about black reformatting but this is pretty handy. Thanks.)

PR #1181, opened 2021-09-17
Bugfix/wms monitor arg ints

Tobias-Ternent:
### Description

Fixing an issue I was experiencing when passing in wms-monitor-arg values, it was casting everything to integers, and I want to be able to pass in other things like Strings. So just removing usage of the parse_resources() in the WMS logger.

### QC

### DIFF
--- a/snakemake/logging.py
+++ b/snakemake/logging.py
@@ -131,7 +131,7 @@ class WMSLogger:
         from snakemake.resources import parse_resources
 
         self.address = address or "http:127.0.0.1:5000"
-        self.args = parse_resources(args) or []
+        self.args = args or []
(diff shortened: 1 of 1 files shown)


fgypas:
Thanks a lot @Tobias-Ternent . I just tested the changes and it works fine for me.

@vsoch I think it would be great to have your input and make sure that we no breaking changes are introduced for [snakeface](https://github.com/snakemake/snakeface).

vsoch (approved):
I think this should be ok! If there is an issue with snakeface it will pop up and we can address at that time.

fgypas:
@johanneskoester It would be great if you can include this fix for the next release.

Issue #1093, opened 2021-07-16
Specifying a path starting with "/" in --default-resources tmpdir=<path> creates a syntax error in checkpoints


RomainFeron:
**Snakemake version**
6.5.3

**Describe the bug**
When using the command line argument `--default-resources` to specify a tmp directory starting with "/", a Syntax Error occurs for the first `checkpoints.xxx.get()` line.

**Logs**

*Using a non-/ tmpdir:*

```bash
(snakemake) ➜  snakemake snakemake -j1 --forceall --default-resources tmpdir=test 
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job                   count    min threads    max threads
------------------  -------  -------------  -------------
example_checkpoint        1              1              1
total                     1              1              1

Select jobs to execute...

[Fri Jul 16 11:09:17 2021]
checkpoint example_checkpoint:
    output: 1.txt, 2.txt, 3.txt
    jobid: 0
    resources: mem_mb=1000, disk_mb=1000, tmpdir=test
Downstream jobs will be updated after completion.

[Fri Jul 16 11:09:17 2021]
Finished job 0.
1 of 1 steps (100%) done
Complete log: /home/romain/work/tmp/snakemake/.snakemake/log/2021-07-16T110917.108029.snakemake.log
```

*Using a / tmpdir:*

```bash
(snakemake) ➜  snakemake snakemake -j1 --forceall --default-resources tmpdir=/home/romain/test
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
InputFunctionException in line 2 of /home/romain/work/tmp/snakemake/Snakefile:
Error:
  SyntaxError: invalid syntax (<string>, line 1)
Wildcards:

Traceback:
```

**Minimal example**
*Code used to generate the logs above:*

```python
checkpoint example_checkpoint:
    '''
    '''
    output:
        '1.txt',
        '2.txt',
        '3.txt'
    shell:
        'touch {output}'


def example_input(wildcards):
    '''
    '''
    rule_input = checkpoints.example_checkpoint.get(**wildcards).output[0] 
    return rule_input


rule example_rule:
    '''
    '''
    input:
        example_input
    output:
        'test.txt'
    shell:
        'cat {input} > {output}'
```

[...]

PR #1192, opened 2021-09-28
Add more informative errors when evaluation of `--default-resources` fails

DonFreed:
### Description

`--default-resources` values are evaluated with `eval()` and some default resources strings can fail if not properly quoted (e.g. see #1093). This PR adds a more informative error message for `--default-resources` arguments that raise an exception.

### QC

### DIFF
--- a/snakemake/resources.py
+++ b/snakemake/resources.py
@@ -1,6 +1,8 @@
 import re
 import tempfile
 
+from snakemake.logging import logger
+
 
 class DefaultResources:
     defaults = {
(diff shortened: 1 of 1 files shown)


DonFreed:
I'm not sure why the tests took so long here?

snakemake/resources.py
```diff
@@ -58,6 +60,19 @@ def callable(wildcards, input, attempt, threads, rulename):
                     # Triggers for string arguments like n1-standard-4
                     except NameError:
                         return val
+                    except Exception as e:
+                        if not (
+                            isinstance(e, FileNotFoundError) and e.filename in input
+                        ):
+                            # Missing input files are handled by the caller
+                            logger.error(
```
johanneskoester, line 68: Good idea. Instead of logging manually here, can you please raise a `WorkflowError` with that message? Snakemake will then handle the display properly.

johanneskoester (changes requested):

snakemake/resources.py
DonFreed: Thanks! The most recent commit will raise a `WorkflowError` instead of logging the error message.

DonFreed:
It looks like there may have been some DNS issues at Quay during testing, https://status.quay.io/incidents/bt8yqm82069z

Issue #574, opened 2020-08-18
default remote prefix duplicated for checkpoint output on cloud


aryarm:
**Snakemake version**
5.22.1

**Describe the bug**
Using the `--default-remote-prefix` parameter will cause `MissingInputException` errors to arise for the outputs of checkpoint rules.

**Steps to Reproduce**
Follow the [Google Life Sciences Executor Tutorial](https://snakemake.readthedocs.io/en/stable/executor_tutorial/google_lifesciences.html) but convert the `bwa_map` rule into a checkpoint like this
```
diff --git Snakefile Snakefile
index 68974d1..d519d4f 100644
--- Snakefile
+++ Snakefile
@@ -4,7 +4,7 @@ rule all:
     input:
         "plots/quals.svg"

-rule bwa_map:
+checkpoint bwa_map:
     input:
         fastq="samples/{sample}.fastq",
         idx=multiext("genome.fa", ".amb", ".ann", ".bwt", ".pac", ".sa")
@@ -19,7 +19,7 @@ rule bwa_map:

 rule samtools_sort:
     input:
-        "mapped_reads/{sample}.bam"
+        lambda wildcards: checkpoints.bwa_map.get(sample=wildcards.sample).output[0]
     output:
         "sorted_reads/{sample}.bam"
     conda:
```
and then you should see the following when running the pipeline:
```
Building DAG of jobs...
MissingInputException in line 20 of Snakefile:
Missing input files for rule samtools_sort:
snakemake-testing-data/snakemake-testing-data/mapped_reads/A.bam
```
Notice how `snakemake-testing-data` appears prepended twice?

**Bug Hypothesis**
I think `rules.apply_default_remote()` is being applied to the output more than once. It might help to check whether `incomplete` is true [on this line](https://github.com/snakemake/snakemake/blob/7481db14d93b73d6c0aeb8e42f3dc510019c48e4/snakemake/rules.py#L721) before executing `rules.apply_default_remote()`?

It might be possible that this is a bug that extends beyond the life sciences executor (to other cloud environments), but I haven't tested that yet.

vsoch:
[...]

PR #1199, opened 2021-09-29
test: issue with duplicated prefix for checkpoints on cloud

aryarm:
Implements a test to reproduce issue #574: checkpoints in cloud execution mode.

### QC


aryarm:
@johanneskoester , I'm not really sure why the new test passed 😕 

It didn't pass when I tried it locally...

tests/test_google_lifesciences.py
```diff
@@ -90,3 +90,22 @@ def test_touch_remote_prefix():
...
+    create_google_storage(bucket_name)
+    storage_prefix = "test_cloud_checkpoints_issue574"
+    workdir = dpath("test_cloud_checkpoints_issue574")
+    try:
+        run(
+            workdir,
+            use_conda=True,
+            default_remote_prefix="%s/%s" % (bucket_name, storage_prefix),
+            google_lifesciences=True,
+            google_lifesciences_cache=True,
```
aryarm, line 108: ah, I think this should be changed based on the other functions in this file? That might be a mistake on my part.
```suggestion
            google_lifesciences_cache=False,
```

johanneskoester: Yes, this should be False indeed.

aryarm:
hmm the checks seem to be passing, still :(
But on second thought, it doesn't look like the GLS tests are actually getting executed in the first place.

Maybe it's because the PR needs the `update-container-image` tag as described in https://github.com/snakemake/snakemake/pull/691?

johanneskoester:
The problem is that you are on a fork. forks don't have access to all the necessary secrets. Let me do the following: you become part of the Snakemake organization, then you can re-issue your PR from within the original repo, not as a fork. Would that be ok for you?

aryarm:
Sure, that sounds good to me. As a long time user of Snakemake, I really value having the opportunity to help contribute to it!

johanneskoester:
Awesome, you have an invitation now.

aryarm:
closing in favor of #1294

PR #1203, opened 2021-10-02
fix: Repair MREs from #823

epruesse:
### Description

see #823 

Three MREs:

- [x] Checkpoints with `temp()` output lead to incomplete DAG execution
- [ ] Checkpints with `pipe()` lead to incomplete DAG execution
- [ ] Checkpoint target rules removed before 100% message displayed

### QC


epruesse:
@johanneskoester Snakemake exits when no more runnable or running jobs exist. One indicator of the 823 issue is that there still are jobs that should be run left. I'm checking for this here, printing an error and exiting with 1. You may want to change the issue url to point to an issue of your own, where you've got the top post to inform about the bug. I was able to fix the MREs posted in the issue, but I feel like this isn't the last we've heard of this.

PR #1206, opened 2021-10-07
fix: Only --bind Snakemake module into container when we're working with a Python script

frankier:
### Description

By default Snakemake binds its Python module into the container, but this is only needed for Python script mode.

This is problem if Snakemake is already running inside a Singularity container, as I am experimenting with at https://github.com/frankier/singslurm2 then the Snakemake module is not available on the host when executing.

This patch makes it so the bind is only made when we actually need to run a Python script.

Long term, I would favor a situation where the `snakemake` namespace available to Python scripts was more like a data file, which can somehow be shimmed directly into the script, like the current RustScript implementation --- assuming this is possible. This would seem to allow more flexibility for combining the Singularity implementation with other stuff.


edwinyyyu:
Why is bind needed for Python scripts? Is it only necessary for importing logger from snakemake.logging?

Issue #872, opened 2021-02-11
Bundle many small jobs into one larger job submission


fischuu:
**Is your feature request related to a problem? Please describe.**
Depending on the definition of the rules, it might happen that Snakemake creates huge amount of jobs submitted to a cluster

**Describe the solution you'd like**
In order to reduce the amount of jobs created by Snakemake, it would be really nice if Snakemake could check the resource allocations from different rules, combines them then into a single script (with more resources) and submits them together as one job. 

For example, if subgraphs of the DAG would have a combined runtime less than some parameter, these jobs would be submitted together as one script in right order with the accumulated resource allocations.

**Describe alternatives you've considered**
Currently I put commands with short runtime together into one rule, but it would make a workflow more readable if each task would have it's own rule.

ASLeonard:
That sounds quite similar to the [job grouping feature](https://snakemake.readthedocs.io/en/stable/executing/grouping.html). It may not be as nuanced as you want, but it is a pretty straightforward way to bundle together many small jobs with some internal handling of pooled resources.

fischuu:
Thanks, I was thinking first the same that the group feature could be used for it, but isn't it so that with the group-option one can only control which jobs will be submitted to the cluster at the same time? But they would still have their own resource allocations, slurm-jobid, etc, isn't it? I tried to find some examples for group, if it can be used also to merge rules into single jobs, but couldn't find any.

But maybe I misunderstood how group can be used, do you know if there is some webpage with more detailed use-cases of the job-grouping feature?

ASLeonard:
From experience, if you run 10 tasks in the same group, setting `--group-components mygroup=5` would launch 2 unique jobs, `--group-components mygroup=10` would only launch a single unique job. The resources get combined, so if each task needs 1 core, the examples request 5 and 10 cores respectively. 

There is room for improvement here imo, as if I have 100 small jobs all with the same resource needs, I'd prefer `group` to just concatenate them, and still only request 1x job worth of resources.

fischuu:
Great, thanks a lot, at least that is then something to play around and familiarize myself with it. I totally misunderstood this part of the manual and the use of `group`. Also, even if there is space for improvement, I think it will make a difference in my workflows, as I currently have 1.000+ samples with 30.000+ jobs, so if `group` could bring it down to 5.000-10.000 jobs that would be already something!

jdblischak:
[...]

Issue #109, opened 2019-11-14
Resources with cluster jobs


troycomi:
**Snakemake version**
5.7.4+20.gab943e6

**Describe the bug**
When grouping jobs with resources defined, the resources.RESOURCE is taken from the max of the jobs.  However for time it should be the sum of the resource.

**Minimal example**
Using the qsub in tests/test_group_jobs with the following Snakefile:
```
samples = [1,2,3,4,5]

rule all:
    input:
        "test.out"

rule a:
    output:
        "a/{sample}.out"
    group: 0
    resources:
        mem=20,
        time=30
    threads: 2
    shell:
        "touch {output}"

rule b:
    input:
        "a/{sample}.out"
    output:
        "b/{sample}.out"
    group: 0
    resources:
        mem=10,
        time=20
    threads: 1
    shell:
        "touch {output}"

rule c:
    input:
        expand("b/{sample}.out", sample=samples)
    output:
        "test.out"
    group: 1
    resources:
        mem=30,
        time=10
    threads: 3
    shell:
        "touch {output}"

```
Running `snakemake --cluster "echo mem is {resources.mem} ; ./qsub" --jobs 1` produces the following log entry:
```
group job 0 (jobs in lexicogr. order):

    [Thu Nov 14 11:25:05 2019]
    rule a:
        output: a/3.out
        jobid: 9
        wildcards: sample=3
        threads: 2
        resources: mem=20, time=30

    [Thu Nov 14 11:25:05 2019]
    rule b:
        input: a/3.out
        output: b/3.out
        jobid: 4
        wildcards: sample=3
        resources: mem=10, time=20

Submitted group job f0282ae1-97a7-51a9-86f3-7ee58771d003 with external jobid 'mem is 20 threads are 2 time is 30'.
```
Taking the max makes sense for threads and memory, but the time should sum.  I.e. I would expect the last line to be `'mem is 20 threads are 2 time is 50'`

**Additional context**
I'm also wondering how this system deals with a case of:
```
rule a:
  output: {sample}.txt
  resource: mem=100
  threads: 1
  group: 0

rule b:
  input: expand({sample}.txt, sample=range(20))
  resource: mem=100
  threads: 10
  group: 0
```
Will it submit 100 MB with 10 cores, causing 10 instances of rule a to run at once (and use 1 GB)?

troycomi:
I made the title more general because I found another corner case (which is unfortunately important to me).  Taking the above and using a function for resources:
```
[...]

PR #1218, opened 2021-10-13
feat: Group resources

pvandyken:
### Overview
Enhances group job resource calculation by fitting its component jobs into user-defined constraints. Jobs that could normally be run in parallel but do not fit within constraints will be run in series. Resources of parallel jobs are summed, resources of jobs in series are determined via `max()`. An exception is given to `runtime`, which is calculated via `max()` for parallel jobs, and via `sum()` for jobs in series. 

Corrects resource calculation of pipe groups, allowing them to be treated as their own contained groups even when buried inside another user-defined group. 

### Issue solved
Currently, Snakemake handles resource requests for group jobs unintelligently. While it's able to sort which jobs can be run concurrently and which must be run in sequence (via `toposort()`), it makes no attempt to fit these jobs within user-provided constraints (via the `--resources` or `--cores` flags). Thus, it assumes that any jobs that can be run in parallel will be run in parallel. In some situations, especially when `--group-components` is used and there are many parallel jobs, this leads to absurdly high resource requests. The only current work-around is setting absolute limits in a custom job-script, a clunky and imprecise approach, or carefully massaging resource requirements within the Snakefile, which hampers modularity.

In addition, Snakemake currently has no conception of runtime. It treats runtime like any other resource, summing it across jobs in parallel, and taking the max across jobs in series. This is physically unrealistic and necessitates further massaging of resources within the Snakefile. Ideally, each rule should be able to specify how long it will individually take.

### Fix
This PR solves this by allowing Snakemake to determine when jobs running in parallel would exceed constraints set via `--resources` or `--cores` and stacking parallel jobs that would exceed these constraints so that they run in series. 

Each sibling group generated by `toposort()`, referred to in the code as a `block`, is sorted into `layers`. Each `layer` is guaranteed to fit within whatever resource constraints are set by the user. A basic sorting algorithm was implemented to fit as many jobs into each layer as possible, while keeping the runtime of the jobs within each layer relatively consistent. The total runtime for each layer is set as the max runtime of its constituent jobs. Other resources within each layer are calculated via summation. To get the total runtime for the entire group, the runtime of each layer in each block is summed. Other resources are calculated via the max across all layers. 

#### Cores Flag
To set group-wide constraints for cores, both the `--cores` and `--max-threads` CLI arguments presented themselves as options. `--cores` was chosen to keep it's behaviour consistent with its local action: constraining cores across a number of jobs. The current documentation says that, in cluster mode, `--cores` limits the total number of cores available to all jobs. It's not clear whether this referred to every job across the entire Snakemake workflow or every job submitted to a cluster, but either way, that functionality does not seem to be implemented. The only purpose appears to be supplying a value to `workflow.cores` for use in the Snakefile. The changes made here bolster this use by ensuring the value of `--cores` will be applied in the local context of a cluster. The documentation of `--cores` has been updated to reflect this. The documentation of `--max-threads` has also been updated to clarify that it will only act on individual jobs, not on groups. Thus, `--max-threads` is still useful in a cluster context to limit the number of `threads` requested by any individual rule.

#### Pipe groups
Resource calculation for pipe groups has been repaired and made compatible with the new group resource calculation. Previous code within the `finalize()` method in `GroupJob` was documented as placing jobs connected by Pipes into the same toposort level so that they could be run simultaneously, however, the code did not actually achieve this. This behaviour has now been implemented, in part by adding a `pipe_group` property to the `Job` class. This property tracks the specific pipe group a job belongs to (via the `CandidateGroup` class). Thus, a `Job` can belong to both a User defined group and, within that, a Pipe group. `Job`s that are not part of a user defined group will still have their `group` property assigned to the auto-generated `CandidateGroup`. 

For group resource calculation, the resources requested by each job in the pipe group will be first merged into one "job". Here, all resources including runtime are summed, except for `_cores`, for which the `max()` is taken across the jobs.

### Previous Discussion
I found two issues discussing a similar point. #109 discussed some of the difficulties describe above, particularly surrounding the treatment of runtime. This PR address that issue. Some of the resulting discussion pointed out that the application of groups can be very context dependent, with @bnprks mentioning the use of pipes. This PR is compatible with named pipes. It also does not limit "boutique" usage of groups any more than previously (see Backward Compatibility below). Future work may allow even more fine-grained control over group resources, but that is not the issue of this PR.

Another discussion of the issue occurred in #872. Here @ASLeonard mentioned an ability to specify resources just once for an arbitrary number of instances of a rule. @jdblischak relates how currently attempting anything like that results in 100s of cores being requested. While this PR does not directly allow that behaviour, the same effect could be easily achieved by specifying a very short runtime for the rule, and allowing snakemake to stack all the job instances within the core restraints provided. Memory is only summed across jobs actually running in parallel, so memory can be easily specified. 

### Tests
Tests have been written covering "typical" use of the functionality, including a single resource contraint, a constraint on a resource never requested by a job, parallel jobs requesting contrasting resources, the use of runtime, and jobs in series. Tests also cover the use of `--max-threads`, the use of constraints requiring the "stacking" of parallel jobs, groups with pipes, and groups with pipes with resources too constrained to fit the pipe job, resulting in an error.

The tests were written with an additional dependency: [pytest-mock](https://pypi.org/project/pytest-mock/), used to help evaluate if resources were calculated as expected.

### Backward Compatibility
Currently, Snakemake requires the user to manually constrain group submissions via methods described above. The new implementation of resource calculation has no effect if resources already fit within user constraints. Thus, this part of the update should not break any existing workflows. 

The handling of pipe groups is now slightly different. Previously, if a group contained a pipe, every job in the group would be treated as belonging to the pipe. In other words, every job in the group would be merged together via summation, regardless of whether that job was involved in the pipe. This behaviour has been changed such that only jobs actually involved with the pipe will be treated that way. Users taking advantage of the old behaviour may find their workflows requesting far fewer resources, since not every job will be summed to get the final resource total. 

I do not believe there are any other compatibility concerns.

### Outstanding Issues
- I took the liberty of defining `runtime` as special. If another name would be more acceptable I'm open to changing it. Regardless, not every user may use `runtime` as the job time resource. While migrating their code may be a reasonable expectation to take advantage of new features, it may also be acceptable to define a few aliases, although this would break with current practices of special resources. Importantly, users may desire the old behaviour of summing runtime across parallel jobs. To achieve this, they can simply change runtime to something similar, like `runtime_min`.
- As I don't use pipe groups myself, it was not entirely obvious to me how the resources within a pipe group should be combined. I based my decisions off of discussion found in #109. Feedback on whether this choice was appropriate would be appreciated.
- I have not tested this explicitly on Cloud executors as I don't have access to those types of environments. None of the relevant tests already written have failed, but I would appreciate feedback on any further considerations to be made.
- I was unsure why, in the existing code, when grouping multiple sibling groups together into the final resource calculation, a check is made for `self.dag.workflow.run_local`. I adapted the new code to retain that check and run, I believe, in the same way, but more context on its purpose would help ensure I didn't introduce bugs.

### QC


johanneskoester:
Awesome and very valuable work! Thanks a lot. I will need a day to dive into this, which I'll try to do as soon as possible.

pvandyken:
I just discovered an important limitation of this PR that needs to be addressed before pulling it into the main:

I used `--resources` to define the constraints on any one cluster, however, Snakemake uses that same argument to define constraints for the **global** process. In other words, Snakemake tries to keep every job scheduled within those constraints, even across multiple nodes. So if even one job is scheduled that fills those constraints, no other jobs will be scheduled. 

The challenge is that this behaviour of Snakemake is perfectly reasonable when running on a single node. It doesn't make any sense in cluster mode, where jobs can be freely scheduled without any heed to "global resources" (so long as `--max-jobs` is not broken). However, some semblance of the behaviour must be kept in cluster mode for local jobs. which are still subject to the constraints of the master process's node.

I'm going to look into this to see how easy it is to correct. Another potential solution is to invent a new command line arg: `--max-resources` or something, but I don't care to add yet another argument and the difference between it and `--resources` will be inherently confusing.

Let me know if you have any thoughts or suggestions.

pvandyken:
I addressed the above problem by adding a new `@property` to both `GroupJob` and `Job`: `local_resources`. This evaluates to `resources` if `is_local` is `True`, but if `is_local` is `False`, it includes only `"_nodes"` and `"_cores"`. I updated the `Scheduler` to use `local_resources` instead of `resources`. This way, the scheduler will not take submitted job resources into account (except for total nodes and cores, which are the only actually relevant values), but will still evaluate resource consumption for locally running jobs.

One thing that's a bit unsatisfactory, the `local_resources` code in `GroupJob` and `Job` is exactly the same. I could have potentially put the code in `AbstractJob`, but being abstract, it didn't seem the right place. It also didn't have the `resources @property`, and I didn't want to take too many liberties with refactoring. If you (@johanneskoester ) have any alternative ideas or suggestions, let me know.

On the maintenance front, I REBASED the PR to account for the latest pushes to main.

fischuu:
I would like to suggest another feature for the group job resource calculation/allocation. 

Would it be possible to take the `group` name as it is used in the various grouped rules and then just allow for a manual resource allocation in the server_config.yaml file to give the user the flexibility to adjust the resources? Somethings like this:

```
rule rule1:
    [...]
    group: "group1"
    [...]

rule rule2:
    [...]
    group: "group1"
    [...]
```

and then we could have a `server_config.yaml` file like this:

```
__default__:
    time: 00:30:00
    mem: 2000

rule1:
    time: 01:00:00
    mem: 6000

rule2:
    time: 00:05:00
    mem: 9000

group1:
    time: 01:00:00
    mem: 9000
```

so if there is a resource allocation that matches a group name, then those resources are used for the group job instead of doing some automatic adjustments?!

pvandyken:
@fischuu, I've certainly thought of such functionality, and I think it would ultimately be the way to go for the most amount of control. You could create a new configuration file to contain the information, or could define a new root level keyword `group` within the `Snakefile`/`.smk` files. I think it goes beyond the scope of this PR, but it's something I'd like to see in the future.

I understand hesitancy behind automatic adjustments; I also like my software to be transparent. My attempt here was to simply recreate constraints that would be set by internal scheduler onto cluster submission. If you were to take any group job and define and submit it manually along with `--resource` specifications, the Snakemake scheduler would automatically stack and organize the component jobs such that they fit within the constraints. The goal here was to do the same. All resources are treated equally as "width" constraints, with the one exception of `runtime`, which is additive across sequential layers of job. However, I believe this designation is sensible and almost always the desired behaviour. So I don't think the functionality added is any more complicated than the local job scheduler. Actually, I think it makes cluster execution more intuitive. The current documentation for `--resources` and `--max-threads`/`--cores` already suggest they constrain the requested resources, when in fact, they don't.

Ultimately, if you don't want jobs automatically adjusted, you can manually design them so that there is no possibility of them overrunning the constraints. And this is, in fact, how you must design jobs currently, as jobs with too many resources will be rejected by the cluster scheduler (e.g. SLURM).

pvandyken:
On another note, it turns out Snakemake does not include any `--resource` flag when scheduling a job on the cluster. There was never any need before. It's necessary to prevent cluster jobs from overconsuming resources. So that's something I'll need to work out before merging this.

SichongP:
> I addressed the above problem by adding a new @property to both GroupJob and Job: local_resources. This evaluates to resources if is_local is True, but if is_local is False, it includes only "_nodes" and "_cores". I updated the Scheduler to use local_resources instead of resources. This way, the scheduler will not take submitted job resources into account (except for total nodes and cores, which are the only actually relevant values), but will still evaluate resource consumption for locally running jobs.  

I'm not entirely sure how this changes snakemake behavior on HPC but I'd like to point out that even in cluster mode, restraining total RAM consumption across all nodes may be a legitimate need -- particularly for reasons related to accounting and resource sharing under a single paying account. One may also want to limit IO if they are not utilizing /scratch directory attached to local nodes. Thus it would break a lot of people's workflows if Snakemake no longer respects global resource restraints on HPC

pvandyken:
Thanks @SichongP, that's a really good point. I can see both global and cluster specific resource constraints as necessary, and since `--resource` currently (intentionally or not) refers to global, backwards compatibility would demand it remain so.

I still see node specific resource constraints as a necessary feature. The most direct route to include it without breaking compatibility would be to add another command line argument (perhaps `--cluster-resources`, fitting nicely with the other cluster args). @SichongP, any thoughts on another solution?

SichongP:
> The most direct route to include it without breaking compatibility would be to add another command line argument (perhaps --cluster-resources, fitting nicely with the other cluster args).

I think this makes most sense. Node configs are largely dependent on how each specific cluster is set up (different HPC facilities will have different max cpus and RAM on a single node. Some will have different queues for "big memory" nodes, etc). So realistically this should be something that a user specifies on CLI or in their profile.

Instead of `--cluster-resources`, maybe call it `--group-resources` to avoid confusion with existing `--resources` flag? It also makes it clearer that such constraints apply only to a group of jobs.

Also, if a user sets up job groups without specifying `--group-resources`, it may be worth issuing a warning?

johanneskoester:
> > The most direct route to include it without breaking compatibility would be to add another command line argument (perhaps --cluster-resources, fitting nicely with the other cluster args).
> 
> I think this makes most sense. Node configs are largely dependent on how each specific cluster is set up (different HPC facilities will have different max cpus and RAM on a single node. Some will have different queues for "big memory" nodes, etc). So realistically this should be something that a user specifies on CLI or in their profile.
> 
> Instead of `--cluster-resources`, maybe call it `--group-resources` to avoid confusion with existing `--resources` flag? It also makes it clearer that such constraints apply only to a group of jobs.
> 
> Also, if a user sets up job groups without specifying `--group-resources`, it may be worth issuing a warning?

Thanks folks for continuously working on this. Indeed, --resources are meant to be global. On a cluster you would not use them to constrain e.g. memory, but rather something like API requests to some web service (if your workflow has something like that). Instead of having a separate argument for groupwise resource constraints, we could provide a syntax within the workflow for defining which resources are interpreted globally or node-local:

```
node_specific_resources:
    "my_fancy_resource"
```
By default, our standard resources `_cores`, `mem_mb` and `disk_mb` would be considered node_specific, while everything else would be considered global. This way, people can just use the normal --resources argument for configuring the cluster execution in 99% of the cases. And if they have a custom resource that needs to behave as being a local constraint for within group parallelism, they just need above directive.

johanneskoester:
Good point, who knows what people want. Let us instead allow the following:
```
resources:
    my_fancy_resource="local"
```
Then, in case somebody wants to overwrite the default (local) for e.g. disk_mb, one can say

```
resources:
    disk_mb="global"
```

By default, all resources are global, except `_cores`, `mem_mb`, and `disk_mb`.

docs/snakefiles/rules.rst
```diff
@@ -343,10 +343,18 @@ Resources are always meant to be specified as total per job, not by thread (i.e.
...
 Apart from making Snakemake aware of hybrid-computing architectures (e.g. with a limited number of additional devices like GPUs) this allows us to control scheduling in various ways, e.g. to limit IO-heavy jobs by assigning an artificial IO-resource to them and limiting it via the ``--resources`` flag.
 Resources must be ``int`` or ``str`` values. Note that you are free to choose any names for the given resources.
 
+When submitting :ref:`group jobs <job_grouping>` to the cluster, Snakemake calculates how many resources to request by first determining which component jobs can be run in parallel, and which must be run in series.
+For most resources, such as ``mem_mb`` or ``threads``, a sum will be taken across each parallel layer.
+The layer requiring the most resource (i.e. ``max()``) will determine the final amount requested.
+The only exception is ``runtime``.
+For it, ``max()`` will be used within each layer, then the total amount of time across all layers will be summed.
+If resource constraints are provided (via ``-resources`` or ``--cores``) Snakemake will prevent group jobs from requesting more than the constraint.
```
ElDeveloper, line 356: Just caught this while reading through this stupendous PR.

```suggestion
If resource constraints are provided (via ``--resources`` or ``--cores``) Snakemake will prevent group jobs from requesting more than the constraint.
```

pvandyken:
> Good point, who knows what people want. Let us instead allow the following:
> 
> ```
> resource_types:
>     my_fancy_resource="local"
> ```
> 
> Then, in case somebody wants to overwrite the default (local) for e.g. disk_mb, one can say
> 
> ```
> resource_types:
>     disk_mb="global"
> ```
> 
> By default, all resources are global, except `_cores`, `mem_mb`, and `disk_mb`.

After thinking some more about this, I'd propose making this a cli flag rather than a workflow directive. A "resource_type" indicates something about the capacity of the computer(s) running the job. For instance, my cluster account has a global limit on memory usage, vs my system just has a memory cap on individual nodes. Those questions aren't workflow specific, so for maximum portability, I suggest we embed this at the configuration layer, rather than the app layer.
So the example would be similar, but in the `config.yaml`:

```yaml
#config.yaml
resource_types:
  - disk_mb=global
```

Regarding name, may I suggest `--resource-scope` as a more descriptive choice? @johanneskoester, any thoughts?

As a side note: it might be nice to allow resource configuration at both the global and nodal level. I have some thoughts on how to do that, but I think that's best left as the subject of another issue.

pvandyken:
> Mhm, I am suprised by this. You mean your cluster requires you to have mem_mb globally constrained, over all nodes??

Ah, not at all, apologies. I see where I was unclear. I was merely using `mem_mb` as a hypothetical example. Although a real life example of a global `mem_mb` limit was alluded to in another post above:

> I'm not entirely sure how this changes snakemake behavior on HPC but I'd like to point out that even in cluster mode, restraining total RAM consumption across all nodes may be a legitimate need -- particularly for reasons related to accounting and resource sharing under a single paying account.

Generally, my point was that a workflow developer may have a rule that consumes resources `foo`, and the developer may assume that `foo` capacity always scales with the number of clusters and accordingly set it to local, only for a user to consume the pipeline in a unique environment where `foo` capacity is globally fixed. The only way for this user to adjust the resource scope would be to change the `Snakefile` which, in some packaging schemes, may not be trivial. On the other hand, if resource scope is set by configuration (either CLI or, more preferably, a `profile.yml`), users can set these resources however they wish.

I do see utility in a `Snakefile` directive version of `resource_scope`, however. It allows a way to set sensible defaults for resources we haven't thought of. Just reading through the slurm reference, there's a number of numerical resources that, if used, would almost certainly be "local" (e.g. gpus, other less relevant like `ntasks`, etc). Without a `resource_scope` directive, users would be essentially required to specify these scopes in their private config, which would be inconvenient.

Perhaps the best solution is doing both: making `resource_scope` a directive overrideable by a config. But perhaps that's making it too complicated.

In any case, let me know what you think, and I'll go with that.

johanneskoester:
> > Mhm, I am suprised by this. You mean your cluster requires you to have mem_mb globally constrained, over all nodes??
> 
> Ah, not at all, apologies. I see where I was unclear. I was merely using `mem_mb` as a hypothetical example. Although a real life example of a global `mem_mb` limit was alluded to in another post above:
> 
> > I'm not entirely sure how this changes snakemake behavior on HPC but I'd like to point out that even in cluster mode, restraining total RAM consumption across all nodes may be a legitimate need -- particularly for reasons related to accounting and resource sharing under a single paying account.
> 
> Generally, my point was that a workflow developer may have a rule that consumes resources `foo`, and the developer may assume that `foo` capacity always scales with the number of clusters and accordingly set it to local, only for a user to consume the pipeline in a unique environment where `foo` capacity is globally fixed. The only way for this user to adjust the resource scope would be to change the `Snakefile` which, in some packaging schemes, may not be trivial. On the other hand, if resource scope is set by configuration (either CLI or, more preferably, a `profile.yml`), users can set these resources however they wish.
> 
> I do see utility in a `Snakefile` directive version of `resource_scope`, however. It allows a way to set sensible defaults for resources we haven't thought of. Just reading through the slurm reference, there's a number of numerical resources that, if used, would almost certainly be "local" (e.g. gpus, other less relevant like `ntasks`, etc). Without a `resource_scope` directive, users would be essentially required to specify these scopes in their private config, which would be inconvenient.
> 
> Perhaps the best solution is doing both: making `resource_scope` a directive overrideable by a config. But perhaps that's making it too complicated.
> 
> In any case, let me know what you think, and I'll go with that.

Ah, that makes sense! I agree. Let's do both. A `resource_scope` directive for sensible defaults in the Snakefile, and the ability to set scopes via the CLI as `--set-resource-scope`, similar to what we have now for other things like `--set-scatter`, `--set-resources` and `--set-threads`.

johanneskoester:
Btw. great addition, and of course this will make you a co-author of the Snakemake rolling paper (I am planning to update it during the summer).

docs/snakefiles/rules.rst
pvandyken: Thanks for catching this!

pvandyken:
Alrighty, `resource-scopes` has been implemented and everything appears to be working. I've added tests for the new features and updated the docs. Unless there's any other general comments or changes, I'd say this is ready for a full review and, pending final approval, merging in.

cmeesters:
Sorry, for the lack of progress in the SLURM Executor (#1015 ), yet I did not realize this thread until the ping. I hope that we can have a merge, soon-ish.

> Mhm, I am suprised by this. You mean your cluster requires you to have mem_mb globally constrained, over all nodes??
I would have really thought that resources scopes (good name!) are indeed independent of the underlying cluster.

> By default, our standard resources _cores, mem_mb and disk_mb would be considered node_specific

Yet, I need to clarify: In a cluster context (e.g. SLURM) resources are job, rather than node specific. Flags like `--mem` or `--mem-per-cpu` (see `man sbatch`) are SLURM job specific. A group job, will see 1 snakemake instance instructing the slurmstepd with multiple `srun`s. 

That's actually a neat feature of SLURM: by grouping a scatter-gather, using `sbcast` to ensure file consistency in (reference) input files across nodes, a multi-node job with smp software (the usual in bioninformatics) can be achieved without causing I/O issues.

Upon submission `--mem` would be applied on all nodes to be requested. `--mem-per-cpu` would be multiplied with the number of tasks (potentially many in a group job) and cpus per task - just to give an example.

I hope we are able to consider all of this as generically as possible.

pvandyken:
@cmeesters thanks for getting back. I've felt slurm execution could be handled better, and I'm glad some progress is being made in #1015.

So this PR was made with the current capabilities of snakemake in mind. I use my own variation of the [slurm profile](https://github.com/Snakemake-Profiles/slurm) from Snakemake-Profiles, which just takes the group job and `sbatch` submits it to run as a single job step. No `srun` involved. That's not the best way to interact with slurm, obviously, but before #1015 using `srun` in a meaningful way is not really possible.

That means 1 node per group job. And as far as I know, that's really the limit for snakemake right now: you have a bunch of processes in a box (the computer) of certain dimensions (number of cores, amount of memory, the runtime, etc) and you try to pack the processes in as efficiently as possible. For group jobs, we're just opening up a new box (e.g. the slurm job submitted onto a single node with one task). All this PR really does is take the capacity of the node and the size of your processes and calculate how long they will take to run.

So to properly treat slurm tasks, with memory per task and cpus per task, things need to be approached differently. I'll leave a comment at the specific place, but I noticed in `SlurmScheduler`, you just sum across threads for all jobs, and provide that to the `--cpus-per-task` param, which doesn't seem right (although slurm easily confuses me, so I very well could be wrong!). I'd think you want `--cpus-per-task` set to the number of threads requested by an individual job, then set `--ntasks` to the number of jobs which will run at one time (as you currently do). But this becomes hard to calculate when you submit a heterogenous group (e.g. some jobs have  1 thread, others have 32). As far as I know, a similar problem will occur with `--mem-mb-per-cpu` in group jobs, as it also will get summed across each toposort level. (But again, I could be wrong).

I'm not really sure what the best approach is, and I'm not super comfortable with the fine details of slurm submission. I don't think, however, that this PR creates any additional problems for the SLURM executor. As discussed in the intro post, it should be fully backward compatible, and only has any effect if you craft group jobs with resource needs higher than specified in `--resources`, which currently would cause a scheduler error.

One thing that may be useful for group jobs is an additional `average_resources` parameter, that gives the average needs of each child job? That would make it easier to set `-per-task` entities.

pvandyken:
Sorry, the above post is a little long. The short version is that I don't know that snakemake is able handle resources per task within group jobs, at least not currently.

snakemake/jobs.py
```diff
@@ -1192,24 +1218,79 @@ def merge(self, other):
...
+            dependencies = {}
+            for job in self.jobs:
+                if job.pipe_group in pipe_dependencies:
+                    dependencies[job] = pipe_dependencies[job.pipe_group]
+                else:
+                    dependencies[job] = self.dag.dependencies[job]
+
+            self.toposorted = [*self.toposort(self.jobs, dependencies)]
+
+    def toposort(self, group, dependencies):
```
johanneskoester, line 1256:
```suggestion
    def _toposort_jobs(self, jobs, dependencies):
```

snakemake/jobs.py
```diff
@@ -1192,24 +1218,79 @@ def merge(self, other):
...
+        pipe_groups = defaultdict(set)
+        for sibling in self.toposorted:
+            for job in sibling:
+                if job.pipe_group is None:
+                    yield job
+                    continue
+                pipe_groups[job.pipe_group].add(job)
+
+            for group in pipe_groups.values():
+                yield from chain(*self.toposort(group, self.dag.dependencies))
```
johanneskoester, line 1294: I think this should rather be done in finalize by resorting each level in the toposorted that is stored there with the strategy you use here. This is be better from a performance perspective.

pvandyken: Good call. I actually didn't realize until recently that the `toposort` implementation was moved to the `dag` module, but I still had mine in `jobs`. I'll have a commit soon that unifies all the code under the `dag` version

snakemake/executors/__init__.py
```diff
@@ -344,6 +344,7 @@ def get_job_args(self, job, **kwargs):
                 format_cli_arg("--cores", kwargs.get("cores", self.cores)),
                 format_cli_arg("--attempt", job.attempt),
                 format_cli_arg("--force-use-threads", not job.is_group()),
+                self.get_resource_declarations(job),
```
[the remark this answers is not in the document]
pvandyken: Yep, can definitely do that. Resource scopes don't affect local executions, and a submitted job counts as local, so there wouldn't have been a bug, but that could change one day

snakemake/jobs.py
johanneskoester: DAG has toposort as well, indeed. But it happens on a global scale there, not within a group. Nevertheless, it might indeed be a good idea to unify the two. Maybe put both into a new module under snakemake.common.

pvandyken: I've update jobs so that it call the dag toposort in the same way it currently does on the master branch. If you prefer I move it, I can do that too

pvandyken:
I've addressed the comments above

pvandyken:
Also just updated the docs on resource scopes removing references to nodes. A local resource scope is actually local to a specific job submission, not a node, as a group may extend over multiple nodes. (@cmeesters, this may actually clarify your comment above)

pvandyken:
I fixed the error from the last test run: it turns out booleans subclass integers in python, so when calling a subprocess, snakemake was sending along boolean resources in the `--resources` param. But they got printed as strings (e.g. `False`), resulting in an error. So I added a separate check to make sure we don't send booleans.

snakemake/executors/__init__.py
```diff
@@ -109,6 +109,22 @@ def get_default_resources_args(self, default_resources=None):
...
+            "--set-resource-scopes", self.workflow.overwrite_resource_scopes
+        )
+
+    def get_resource_declarations(self, job):
+        resources = [
+            f"{resource}={value}"
+            for resource, value in job.resources.items()
+            if isinstance(value, int)
+            # need to check bool seperately because bool is a subclass of int
+            and not isinstance(value, bool)
```
johanneskoester, line 123: Ok, so isn't this too late? And in particular just silently skipping them does not sound right. I think bools as resources should cause an error beforehand, where they are specified. I.e. we should not allow them to be passed to the `snakemake()` API, and also not allow them to be present after resource expansion in `Rule.expand_resources`.

pvandyken: Was just looking more into this, I made a mistake above: `parse_resources()`, and by extension `--resources` only allows `int`, not `int` or `str`, which is why currently all the `str` resources are being filtered out in `get_resource_declarations()`. 

I think passing in booleans to `--resources` will have unintended consequences on resource reckoning in the sub-workflow. Because `True` looks just like the integer `1`, snakemake will suddenly limit jobs requesting this boolean resource to one at a time, which would not be the user's intent. For example, in the test above, if 20 instances of `rule a` were being run on cluster as part of a group job with connected components, only one instance of `rule a` would run at a time because `shouldfail == True == 1`, and in the subprocess, `--resources shouldfail=true == 1`. But this limitation is not intended.

Note that this doesn't apply to `--set-resources` and `--default-resources`, which currently allow `str`, `int`, and even python expressions (for default-resources), but neither of those flags are touched by anything in this PR. 

So I would say the status quo is probably best. One other formulation could be:
```python
if isinstance(value, int)
and str(value).isdigit()
```
(or adapted as in [this SO](https://stackoverflow.com/questions/1265665/how-can-i-check-if-a-string-represents-an-int-without-using-try-except#1265696)). This would remove the specific mention of the bool class while ensuring that only "stringafiable" ints are passed on as resource limits.

johanneskoester: > So I would say the status quo is probably best. One other formulation could be:
> 
> ```python
> if isinstance(value, int)
> and str(value).isdigit()
> ```
> 
> (or adapted as in [this SO](https://stackoverflow.com/questions/1265665/how-can-i-check-if-a-string-represents-an-int-without-using-try-except#1265696)). This would remove the specific mention of the bool class while ensuring that only "stringafiable" ints are passed on as resource limits.

Good points. I agree, only passing real ints as limits makes most sense. The proposed solution looks good to me.

docs/snakefiles/rules.rst
```diff
@@ -1662,6 +1710,10 @@ It is possible to combine explicit group definition as above with pipe outputs.
 Thereby, pipe jobs can live within, or (automatically) extend existing groups.
 However, the two jobs connected by a pipe may not exist in conflicting groups.
 
+As with other groups, Snakemake will automatically calculate the required resources for the group job (see :ref:`resources <snakefiles-resources>`.
+The way resources are combined differs, though.
+Most resources, including ``runtime``, are summed accross all jobs involved in the pipe.
+Threads, however, is set by the job requesting the highest number of threads (``max()``).
```
johanneskoester, line 1716: Sorry, now I am a bit confused (maybe we discussed this before, sorry). Why this behavior? Say A pipes into B, shouldn't the runtime of both still be the maximum of the two? Also, for the threads, don't we need the sum of their threads?

snakemake/__init__.py
```diff
@@ -1122,8 +1147,9 @@ def get_argument_parser(profile=None):
             "Use at most N CPU cores/jobs in parallel. "
             "If N is omitted or 'all', the limit is set to the number of "
             "available CPU cores. "
-            "In case of cluster/cloud execution, this argument sets the number of "
-            "total cores used over all jobs (made available to rules via workflow.cores)."
+            "In case of cluster/cloud execution, this argument sets the maximum number "
+            "of cores requested from the cluster or cloud scheduler. "
+            "This number is available to rules via workflow.cores."
```
johanneskoester, line 1152: Let us add a link to your nice explanation about the scopes in the docs here.

snakemake/__init__.py
```diff
@@ -1157,11 +1183,13 @@ def get_argument_parser(profile=None):
...
             "Define additional resources that shall constrain the scheduling "
-            "analogously to threads (see above). A resource is defined as "
+            "analogously to --cores (see above). A resource is defined as "
             "a name and an integer value. E.g. --resources mem_mb=1000. Rules can "
             "use resources by defining the resource keyword, e.g. "
             "resources: mem_mb=600. If now two rules require 600 of the resource "
-            "'mem_mb' they won't be run in parallel by the scheduler."
+            "'mem_mb' they won't be run in parallel by the scheduler. In "
+            "cluster/cloud mode, this argument will also constrain the amount of "
+            "resources requested from the server."
```
johanneskoester, line 1192: Let us add a link to your nice explanation about the scopes in the docs here.

snakemake/__init__.py
```diff
@@ -1198,6 +1228,19 @@ def get_argument_parser(profile=None):
...
+        metavar="RESOURCE=[global|local]",
+        nargs="+",
+        help="Overwrite resource scopes. A scope determines how a constraint is "
+        "reckoned in cluster execution. With RESOURCE=local, a constraint applied to "
+        "RESOURCE using --resources will be considered the limit for each group "
+        "submission. With RESOURCE=global, the constraint will apply across all groups "
+        "cumulatively. By default, only `mem_mb` and `disk_mb` are considered local, "
+        "all other resources are global. This may be modified in the snakefile using "
+        "the `resource_scopes:` directive. Note that number of threads, specified via "
+        "--cores, is always considered local.",
```
johanneskoester, line 1242: Let us add a link to your nice explanation about the scopes in the docs here.

snakemake/executors/__init__.py
```diff
@@ -466,6 +490,9 @@ def get_python_executable(self):
     def get_envvar_declarations(self):
         return ""
 
+    def get_resources_declarations(self, job):
+        return ""
```
johanneskoester, line 494: Let's add a comment here that explains why the CPU executor should not pass resource declarations to spawned run-directive jobs.

johanneskoester: Are you sure that it should not? I currently don't see why. I think it actually should.

johanneskoester (changes requested):
Excellent! I think I have just some final things below. Please feel free to directly contact me via discord in case I am too slow in seeing your answers. I would like to merge this ASAP.

docs/snakefiles/rules.rst
pvandyken: Truthfully, I don't remember exactly why I set it up this way. I've never used the piping function myself; I know I was unsure how to do it, and I was hoping for some guidance.

I'll go ahead and change it to as you say. That makes things much simpler anyway: pipe jobs will behave like any other job.

snakemake/executors/__init__.py
pvandyken: Can't remember why I did that, but I'll remove it and see if it passes the tests

snakemake/__init__.py
pvandyken: How should I add the link? I can do a :ref:`tag` but that won't render in the CLI help (or at least I don't think it will). Or I can do a hard link, but then if the section moves the link won't update. Is there a better way?

pvandyken:
Okay, I did a bunch of rebasing to clean up the early git history. I have a backup in case this causes any problems.

I went ahead and did some refactoring on a bunch of the code. The changes are listed in the commits following "Ensure resources passed on to subcommands are digits", but here's a summary:

## Refactor Group Resource calculation

* I moved all of the code involved in group resource calculation to a new class `GroupResources` in the `resources` module
* `runtime` is no longer hardcoded into the logic, but is extracted into a couple of variables: `additive_resources` and `sortby`, which together cover the various special treatments of "runtime"

The idea with both changes is to make resource calculation easier to maintain and modify in the future. For instance, if future work wanted to add aliases for "runtime", or allow users to define their own aliases, or indeed any other additive resource, it should be much easier to implement. The structure of the `GroupResources` class should also make it easier to add new resource calculation algorithms. The current approach lives in a classmethod called `basic_layered`, but other methods could be easily defined in the future.

## Refactor resource scopes

* Move all involved code to a new `ResourceScopes` class in the `resources` module, implemented as a `UserDict`

The previous implementation was rather messy, but this cleans it up a lot by consolidating all the code into one space and reducing complexity.

* Use a new scope "excluded" to exclude "runtime" from being passed on to snakemake child processes on clusters.

This is in the same spirit as above: trying to not have `runtime` hard coded into business logic. This way, future resources could potentially receive the "excluded" scope and have the same treatment. I purposefully did not document "excluded" anywhere public, as I don't indent it to be used by end users. If someone finds a valid reason to make it available in the future, perhaps it can be documented in a future PR.

snakemake/__init__.py
johanneskoester: I think all we can do is a full URL to the stable docs page. Better than nothing.

johanneskoester:
> Okay, I did a bunch of rebasing to clean up the early git history. I have a backup in case this causes any problems.

Thanks. For the future: I always squash merge, so this is not super important. Anyway, it makes this PR (to which one easily gets from the squash commit) cleaner, so the cleanup is still valuable (just not mandatory).

johanneskoester (approved):

pvandyken:
Sounds good. I'll just figure out why the one test is failing.

pvandyken:
Okay, I *hopefully* fixed the broken test (tbd on completion). It had to do with switching the pipe group calculation to `sum` from `max`. gh actions has only two cores, so in the subsnakemake process, the pipe job was now requesting 4 cores instead of 2. The fix was to patch over `RealExecutor.get_job_args` and have it run the subsnakemake with `--cores 6` instead of `--cores all`. Hacky, but the test should now work in any environment.

Also added one small final refactor to GroupResources and wrote an extended docstring explaining the rationale of the algorithm, intended for future developers.

I thought it would be good to clarify what snakemake version the new docs section on `Resources and Remote Execution` takes affect. I put `v7.10`; let me know if the actual version this code ends up in will be different.

johanneskoester:
Wow, so happy for this to land in main! Thanks a lot! This is epic!

Issue #1153, opened 2021-08-25
SameFileError when running snakemake --gernerate-unit-tests


lenaschneehas:
Dear all,

**Snakemake version**
I'm using the snakemake version 6.6.0

**Describe the bug**
I encounter the following error when I try to run a unit test on the our pipeline:

Tries to copy the file from the original location to the same locations instead of into units folder.

**Logs**
`Building DAG of jobs...` 
` Generating unit tests for each rule...` 
` Generating unit test for rule all: .tests/unit/test_all.py.` 
` Generating unit test for rule report_qc: .tests/unit/test_report_qc.py.` 
` Traceback (most recent call last):` 
  ` File "/home/user/anaconda3/envs/testenv/lib/python3.8/site-packages/snakemake/__init__.py", line 699, in snakemake` 
  `   success = workflow.execute(` 
  ` File "/home/user/anaconda3/envs/testenv/lib/python3.8/site-packages/snakemake/workflow.py", line 857, in execute` 
`     unit_tests.generate(` 
`   File "/home/user/anaconda3/envs/testenv/lib/python3.8/site-packages/snakemake/unit_tests/__init__.py", line 89,  in generate` 
   `  copy_files(job.input, "data")` 
 `  File "/home/user/anaconda3/envs/testenv/lib/python3.8/site-packages/snakemake/unit_tests/__init__.py", line 83,  in copy_files` 
   `  shutil.copy(f, target)` 
  ` File "/home/user/anaconda3/envs/testenv/lib/python3.8/shutil.py", line 418, in copy` 
    ` copyfile(src, dst, follow_symlinks=follow_symlinks)` 
  ` File "/home/user/anaconda3/envs/testenv/lib/python3.8/shutil.py", line 244, in copyfile` 
    ` raise SameFileError("{!r} and {!r} are the same file".format(src, dst))` 
` shutil.SameFileError: PosixPath('/home/common/data/output/projects/intermediate_output/merge_clusterings.rds')`  ` and '/home/data/output/projects/intermediate_output/merge_clusterings.rds' are the same file` 

**Minimal example**
`snakemake --profile /PATH_TO_CONFIG/config/profile --conda-frontend conda --generate-unit-tests 
`
**Additional context**
- I changed the common.py file according to this:
   https://github.com/nikostr/dna-seq-deepvariant-glnexus-variant-calling/blob/e6a945b7a0a3ac006c68a5291e28f7bc8fbafdad/.tests/unit/common.py#L29-L36
[...]

PR #1220, opened 2021-10-15
fix: SameFileError #1153

stydodt:
### Description

Pull request to fix bug #1153.

When relative paths were provided to rules with the `--generate-unit-tests` flag the target path and the tests were successfully created. However, when full paths were used, the automatic unittest generation failed. This was caused by behavior of the Path.parent() property. As `path` and `f.parent` are matching this leads to `f` and `target` matching and the SameFileError raised by `shutils`.

#### Behaviour of `Path.parent`:
```python
>>> from pathlib import Path
>>> f = Path("/dummy/test/path")
>>> f.parent
PosixPath('/dummy/test')
>>> path = Path("/dummy/test/path/")
>>> rulename = "test"
>>> content_type = "expected"
>>> target = path / rulename / content_type / f.parent
>>> target
PosixPath('/dummy/test')
```

Removing `f.parent` creates a successful, matching folder structure with both relative and full paths and allows unittests to be generated.

### QC


stydodt:
Sorry for the delay - I did not get a notification from Github about activity. Thanks for running the pipeline. I must have set my local environment incorrectly as I didn't spot this when I ran the tests - I will investigate.

stydodt:
Thanks for finishing and merging this @johanneskoester ! Getting the tests working was proving challenging given the time constraints and being new to the codebase.

Issue #1253, opened 2021-11-10
Editing 'notebooks' with relative paths in Snakefile causes TypeError


bhjolly:
**Snakemake version**
6.10.0
mamba create -n snakemake -c conda-forge bioconda::snakemake=6.10

**Describe the bug**
When using --edit-notebook with a Snakefile with a relative notebook path, a TypeError is raised when Snakemake tries to turn that into an absolute path. It looks like the 'basedir' variable type has changed but snakemake/notebook.py was not updated to reflect this.

**Logs**
```
Traceback (most recent call last):
  File "/Users/benjolly/miniconda3/envs/snakemake/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 593, in _callback
    raise ex
  File "/Users/benjolly/miniconda3/envs/snakemake/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/benjolly/miniconda3/envs/snakemake/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 579, in cached_or_run
    run_func(*args)
  File "/Users/benjolly/miniconda3/envs/snakemake/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 2461, in run_wrapper
    raise ex
  File "/Users/benjolly/miniconda3/envs/snakemake/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 2418, in run_wrapper
    run(
  File "/Users/benjolly/Documents/arsa-beech/arsa-beech-analysis/workflow/Snakefile", line 253, in __rule_seed_corr
  File "/Users/benjolly/miniconda3/envs/snakemake/lib/python3.9/site-packages/snakemake/notebook.py", line 233, in notebook
    local_path = os.path.join(basedir, path)
  File "/Users/benjolly/miniconda3/envs/snakemake/lib/python3.9/posixpath.py", line 76, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not LocalSourceFile
```

**Minimal example**
```
rule all:
    input:
        "../results/graph.eps"
rule graph:
    output:
        "../results/graph.eps"
    notebook:
        "./notebooks/test.py.ipynb"
```
then
`snakemake --cores 1 --edit-notebook ../results/graph.eps`
(obviously need a working Jupyter environment as well)

**Additional context**

PR #1255, opened 2021-11-10
fix: Issue #1253 (problems editing Jupyter Notebooks)

bhjolly:
### Description
Snakemake 6.10 was unable to open Jupyter Notebooks for editing as the use of LocalSourceFile objects for file paths was not supported in notebook.py or script.py (affecting Notebooks). This PR addresses that issue (logged as GH Issue #1253  )

### QC

### DIFF
--- a/snakemake/notebook.py
+++ b/snakemake/notebook.py
@@ -42,12 +42,12 @@ class JupyterNotebook(ScriptBase):
 
         nb["cells"].append(nbformat.v4.new_code_cell("# start coding here"))
 
-        os.makedirs(os.path.dirname(str(self.local_path)), exist_ok=True)
+        os.makedirs(os.path.dirname(self.local_path), exist_ok=True)
 
-        with open(str(self.local_path), "wb") as out:
+        with open(self.local_path, "wb") as out:
             out.write(nbformat.writes(nb).encode())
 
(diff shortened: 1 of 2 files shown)


snakemake/notebook.py
```diff
@@ -42,12 +42,12 @@ def draft(self, listen):
 
         nb["cells"].append(nbformat.v4.new_code_cell("# start coding here"))
 
-        os.makedirs(os.path.dirname(self.local_path), exist_ok=True)
+        os.makedirs(os.path.dirname(str(self.local_path)), exist_ok=True)
```
johanneskoester, line 45: Can you use `self.local_path.get_path_or_uri()` here? That is more idiomatic with the new API.

johanneskoester: And in all str() occurrences below as well.

snakemake/script.py
```diff
@@ -393,7 +393,7 @@ def evaluate(self, edit=False):
 
     @property
     def local_path(self):
-        path = self.path[7:]
+        path = str(self.path)#[7:]
```
johanneskoester, line 396: Here, I'd rather rely on the SourceFile methods for checking whether the path is absolute or not and so on.

johanneskoester (changes requested):

Issue #1257, opened 2021-11-12
Job failure with shadow directive if log files not created


jaicher:
**Snakemake version**: v6.10.0

**Describe the bug**

For any job that specifies a logfile: if I am not using the shadow directive, it is okay if my job fails to make a logfile itself (because Snakemake creates an empty one at the final location for me, it looks like). But, if I use the shadow directive, I must explicitly make sure that a logfile is created in the specified location, otherwise, my job fails with `FileNotFoundError` in `snakemake.dag.unshadow_output(...)` because the logfile does not exist in the shadow directory, so it cannot move it to its final desired location. I think that if it succeeds without the shadow directive, it should succeed with the shadow directive.

**Minimal example**

As is, this fails. from `FileNotFoundError` in `unshadow_output(...)`. Comment out the shadow directive, and it works.
```
# Snakefile
rule fails_if_shadow_enabled:
    shadow: "minimal"  # comment out and it succeeds
    output: touch("results/fails_if_shadow_enabled.done")
    log: "results/log/fails_if_shadow_enabled.log"
```

This seems like it'd be a pretty easy fix: either automatically touch the logfile in the shadow directory rather than the original location, or ignore FileNotFoundError for logfiles.

johanneskoester:
Fixed with PR #1258.

PR #1258, opened 2021-11-12
fix: issue #1257 (missing logfile failure when using shadow directory)

jaicher:
### Description

+ Test `test_shadowed_log` that demonstrates the failure case with issue #1257
+ If logfile(s) were not created by job, Snakemake now creates empty logfile(s) in directory they would have been created in (i.e. shadow directory, if it was enabled) instead of its final destination. This allows `snakemake.dag.unshadow_output(...)` to successfully copy the logfile from the shadow directory to its final destination.

### QC


johanneskoester (approved):
Thanks a lot, I have slightly modified your fix to also take care of remote files in the right way.

PR #1276, opened 2021-11-24
add LockException

Maarten-vd-Sande:
### Description

Stub for this PR. Let me know if this is something you will/won't support :smile: 

I'm using the API, and I am trying to catch the case where snakemake can not lock the directory. However since no exception is called it gets hard to catch that.

### QC

### DIFF
--- a/snakemake/exceptions.py
+++ b/snakemake/exceptions.py
@@ -532,3 +532,17 @@ class IncompleteCheckpointException(Exception):
 
 class CacheMissException(Exception):
     pass
+
+
+class LockException(WorkflowError):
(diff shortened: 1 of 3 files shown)


johanneskoester:
Sure, no objections. Thanks for looking into that!

Maarten-vd-Sande:
@johanneskoester I'm not sure where I should add the test. I'm pretty sure the `testapi.py` file is wrong, but I couldn't really figure out where else it would make sense

johanneskoester (approved):

johanneskoester (approved):

PR #1296, opened 2021-12-09
feat: enable the registration of new executors from external packages.

johanneskoester:
### Description

This will allow to register executors as plugins, thereby further opening up the development towards external teams.

### QC


johanneskoester:
Implemented in Snakemake 8.

PR #1300, opened 2021-12-13
feat: Support conda environment definitions to be passed as function pointers, similar to input, params, and resources

kpj:
### Description

This PR allows to specify conda environments using job parameters.

### QC


kpj:
Do you have a suggestion on how to make this string formatting more general (how is this done in other cases?) and how to support function arguments as well?

snakemake/jobs.py
```diff
@@ -317,6 +317,9 @@ def conda_env_file(self):
         if self._conda_env_file is None:
             expanded_env = self.rule.expand_conda_env(self.wildcards_dict)
             if expanded_env is not None:
+                # insert parameters
+                expanded_env = expanded_env.format(params=self.params)
```
johanneskoester, line 321: I think this rather should happen in rule.expand_conda_env, which could take params as an argument, similar to how rule.expand_params takes resources etc. as additional arguments.

johanneskoester (changes requested):
Thanks, interesting idea! See below for a better place to handle this.

snakemake/jobs.py
kpj: Thanks a lot for the suggestion! I updated the code accordingly.

johanneskoester (changes requested):
I must say now that I think about it I am a bit unsure about the approach here. It kind of breaks with the way files are formatted in Snakemake. So far, we just had `path/to/{somename}.yaml` where `{somename}` was a wildcard name. For stuff that is more complex we had functions. In case of e.g. resources definition these functions can already now take `params` as an additional argument. I think I would prefer to allow the same mechanism for conda envs instead of introducing something new that solely works like this for conda.

johanneskoester:
Yes, exactly. The upside is that the mechanism would be exactly the same as with input, params and resources.

kpj:
I agree, that would be a great solution!
Would this be quick for you to implement or could you give me a pointer which part of the code to adapt in order to support this functionality?

snakemake/rules.py
```diff
@@ -1124,25 +1118,54 @@ def expand_group(self, wildcards):
...
+            conda_env, _ = self.apply_input_function(
+                self._conda_env, wildcards, params, input
             )
 
-        if conda_env is not None:
-            conda_env.check()
+            if is_conda_env_file(conda_env):
+                self._conda_env = CondaEnvFileSpec(conda_env, rule=self)
+            else:
+                self._conda_env = CondaEnvNameSpec(conda_env)
```
johanneskoester, line 1136: We can't set the objects _conda_env here as this would overwrite what was defined for the rule (e.g. the funciton pointer). But the pointer is needed for subsequent calls to expand_conda_env from other jobs of the same rule. Instead, I suggest to remove this block.

snakemake/rules.py
```diff
@@ -1124,25 +1118,54 @@ def expand_group(self, wildcards):
...
+                and is_local_file(conda_env)
+                and not os.path.isabs(conda_env)
+            ):
+                conda_env = self.basedir.join(conda_env).get_path_or_uri()
 
-        return conda_env
+            if is_conda_env_file(conda_env):
+                self._conda_env = CondaEnvFileSpec(conda_env, rule=self)
+            else:
+                self._conda_env = CondaEnvNameSpec(conda_env)
```
johanneskoester, line 1149: Move this outside of the else and do it for both the callable and the str case. And then, assign the specs to conda_env, not self._conda_env for the same reason as above.

snakemake/rules.py
```diff
@@ -1124,25 +1118,54 @@ def expand_group(self, wildcards):
...
+
+            try:
+                self._conda_env = (
```
johanneskoester, line 1152:
```suggestion
                conda_env = (
```

snakemake/rules.py
```diff
@@ -1124,25 +1118,54 @@ def expand_group(self, wildcards):
...
+                    self._conda_env.apply_wildcards(wildcards)
```
johanneskoester, line 1153:
```suggestion
                    conda_env.apply_wildcards(wildcards)
```

snakemake/rules.py
```diff
@@ -1124,25 +1118,54 @@ def expand_group(self, wildcards):
...
+                    if self._conda_env
```
johanneskoester, line 1154:
```suggestion
                    if conda_env
```

snakemake/rules.py
```diff
@@ -1124,25 +1118,54 @@ def expand_group(self, wildcards):
...
+                )
+            except WildcardError as e:
+                raise WildcardError(
+                    "Wildcards in conda environment file cannot be "
+                    "determined from output files:",
+                    str(e),
+                    rule=self,
+                )
+
+        if self._conda_env is not None:
```
johanneskoester, line 1165:
```suggestion
        if conda_env is not None:
```

snakemake/rules.py
```diff
@@ -1124,25 +1118,54 @@ def expand_group(self, wildcards):
...
+            self._conda_env.check()
```
johanneskoester, line 1166:
```suggestion
           conda_env.check()
```

snakemake/rules.py
```diff
@@ -1124,25 +1118,54 @@ def expand_group(self, wildcards):
...
+
+        return self._conda_env
```
johanneskoester, line 1168:
```suggestion
        return conda_env
```

snakemake/rules.py
johanneskoester: Further, this wildcard application has to be performed before the inference of the spec from above.

johanneskoester (changes requested):

snakemake/rules.py
```diff
@@ -1124,20 +1118,40 @@ def expand_group(self, wildcards):
...
+            is_conda_env_file,
+            CondaEnvFileSpec,
+            CondaEnvNameSpec,
+        )
+
+        if callable(self._conda_env):
+            conda_env, _ = self.apply_input_function(
+                self._conda_env, wildcards=wildcards, params=params, input=input
             )
+        else:
```
johanneskoester, line 1132: This does not need to be in an else, but instead done always, also if it was a callable.

johanneskoester (changes requested):

snakemake/rules.py
```diff
@@ -1124,13 +1118,28 @@ def expand_group(self, wildcards):
...
+    def expand_conda_env(self, wildcards, params=None, input=None):
+        from snakemake.deployment.conda import (
+            is_conda_env_file,
+            CondaEnvFileSpec,
+            CondaEnvNameSpec,
+        )
+
+        if callable(self._conda_env):
+            conda_env, _ = self.apply_input_function(
+                self._conda_env, wildcards=wildcards, params=params, input=input
```
johanneskoester, line 1130:
```suggestion
        conda_env = self._conda_env
        if callable(conda_env):
            conda_env, _ = self.apply_input_function(
                conda_env, wildcards=wildcards, params=params, input=input
```

johanneskoester (changes requested):

johanneskoester (approved):

snakemake/rules.py
```diff
@@ -1124,13 +1115,34 @@ def expand_group(self, wildcards):
...
+                conda_env, wildcards=wildcards, params=params, input=input
             )
+
+        if (
+            conda_env is not None
+            and is_conda_env_file(conda_env)
+            and is_local_file(conda_env)
+            and not os.path.isabs(conda_env)
+        ):
+            conda_env = self.basedir.join(conda_env).get_path_or_uri()
```
johanneskoester, line 1139:
```suggestion
        if isinstance(conda_env, SourceFile):
            conda_env = conda_env.get_path_or_uri()
        elif (
            conda_env is not None
            and is_conda_env_file(conda_env)
            and is_local_file(conda_env)
            and not os.path.isabs(conda_env)
        ):
            conda_env = self.basedir.join(conda_env).get_path_or_uri()
```

johanneskoester (changes requested):

johanneskoester (approved):

johanneskoester (approved):

PR #1325, opened 2022-01-11
feat: support for post deploy scripts

johanneskoester:
### QC

### DIFF
--- a/docs/snakefiles/deployment.rst
+++ b/docs/snakefiles/deployment.rst
@@ -267,9 +267,9 @@ Conda deployment also works well for offline or air-gapped environments. Running
 Providing post-deployment scripts
 -------------------------
 
-With Snakemake 9.14 post-deployment shell-scripts can be provided to perform additional adjustments of a conda environment.
+From Snakemake 6.14 onwards post-deployment shell-scripts can be provided to perform additional adjustments of a conda environment.
 This might be helpful in case a conda package is missing components or requires further configuration for execution.
-post-deployment scripts must be placed next to their corresponding environment-file and require the suffix ``.post-deploy.sh``.
+Post-deployment scripts must be placed next to their corresponding environment-file and require the suffix ``.post-deploy.sh``, e.g.:
 
 .. code-block:: python
 
@@ -280,9 +280,15 @@ post-deployment scripts must be placed next to their corresponding environment-f
             "results.tsv"
         conda:
             "envs/interproscan.yaml"
-            "envs/interproscan.post-deploy.sh
         shell:
             "interproscan.sh -i {input} -f tsv -o {output}"
+
+.. code-block:: none
+
+    ├── Snakefile
+    └── envs
+        ├── interproscan.yaml
+        └── interproscan.post-deploy.sh
 
 The path of the conda environment can be accessed within the script via ``$CONDA_PREFIX``.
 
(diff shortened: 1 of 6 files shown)


snakemake/deployment/conda.py
```diff
@@ -365,6 +366,28 @@ def create(self, dryrun=False):
...
+                                post_deploy_script.relative_to(os.getcwd())
+                            )
+                        )
+                        shell.check_output(
+                            conda.shellcmd(
+                                env_path, "sh {}".format(post_deploy_script)
+                            ),
+                            stderr=subprocess.STDOUT,
+                        )
+
```
johanneskoester, line 390: Move this below the cleanups and outside of the if-else block (because we also need it in the if case). Further, move it into a separate method that is called from here, to avoid some code complexity in this already lengthy method.

FelixMoelder:
This should be ready now.
Mind that I mentioned snakemake version 6.14. So in case the final release will be a different version we should change that.

docs/snakefiles/deployment.rst
```diff
@@ -262,6 +262,30 @@ Conda deployment also works well for offline or air-gapped environments. Running
 
 .. _singularity:
 
+
+-------------------------
+Providing post-deployment scripts
+-------------------------
+
+With Snakemake 9.14 post-deployment shell-scripts can be provided to perform additional adjustments of a conda environment.
```
johanneskoester, line 270:
```suggestion
From Snakemake 6.14 onwards, post-deployment shell-scripts can be provided to perform additional adjustments of a conda environment.
```

docs/snakefiles/deployment.rst
```diff
@@ -262,6 +262,30 @@ Conda deployment also works well for offline or air-gapped environments. Running
...
+This might be helpful in case a conda package is missing components or requires further configuration for execution.
+post-deployment scripts must be placed next to their corresponding environment-file and require the suffix ``.post-deploy.sh``.
```
johanneskoester, line 272:
```suggestion
Post-deployment scripts must be placed next to their corresponding environment-file and require the suffix ``.post-deploy.sh``.
```

docs/snakefiles/deployment.rst
```diff
@@ -262,6 +262,30 @@ Conda deployment also works well for offline or air-gapped environments. Running
...
+.. code-block:: python
+
+    rule NAME:
+        input:
+            "seqs.fastq"
+        output:
+            "results.tsv"
+        conda:
+            "envs/interproscan.yaml"
+            "envs/interproscan.post-deploy.sh
```
johanneskoester, line 283:
```suggestion
```
The plan was to have them in the same folder, but there is no need to mention them in the conda env definition.

snakemake/deployment/conda.py
```diff
@@ -211,6 +212,29 @@ def create_archive(self):
             raise e
         return env_archive
 
+    def execute_deployment_script(self, env_file):
+        """Execute post-deployment script if present"""
+        from snakemake.shell import shell
+
+        post_deploy_script = Path(env_file).with_suffix(".post-deploy.sh")
```
johanneskoester, line 219: Having thought about this, it is actually not that easy. env_file can be a remote URL. We need to handle the two cases in line 246 below, and also store the post deploy script locally in case it is present.

docs/snakefiles/deployment.rst
FelixMoelder: True. My example is wrong. I'll update the documentation.

snakemake/deployment/conda.py
FelixMoelder: This should be done now.

Issue #1331, opened 2022-01-13
groupjob fails when downstream of checkpoint (re-evaluation of DAG bug?)


Maarten-vd-Sande:
**Snakemake version**
6.12.3, seems to be introduced in version 6.5.1 by commit 4dbb7ad353424d790b49b1f5b672c326efa4819c

**Describe the bug**

I have a checkpoint trimming, and a follow up align and sort groupjob. When I run this with 2 cores, all is well. However when I run it with 4 cores, what I think happens is: trimming 1 finishes -> re-evaluate DAG -> align+sort -> trimming 2 finishes -> re-evaluate DAG -> CRASH because align+sort output already exists but rule hasn't finished yet. 

**Minimal example**

Running this with **2 cores works**, but running with **four cores causes a crash**!! Note that it isn't guaranteed to happen, so you might need to re-run this once or twice..

```
rule all:
    input:
        [
         "output/aligned_and_sort/1.txt",
         "output/aligned_and_sort/2.txt",
        ]


checkpoint trimming:
    output:
        "output/trimmed/{sample}.txt"
    shell:
        "touch {output}; sleep 1"


rule align:
    input:
        "output/trimmed/{sample}.txt"
    output:
        pipe("output/aligned/{sample}.txt")
    shell:
        "touch {output}; sleep 1"


rule sort:
    input:
        "output/aligned/{sample}.txt"
    output:
        "output/aligned_and_sort/{sample}.txt"
    shell:
        "touch {output}; sleep 1"
```

**output**

```
snakemake --cores 4
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 4
Rules claiming more threads will be scaled down.
Job stats:
job         count    min threads    max threads
--------  -------  -------------  -------------
align           2              1              1
all             1              1              1
sort            2              1              1
trimming        2              1              1
total           7              1              1

Select jobs to execute...

[Wed Jan 19 09:34:38 2022]
checkpoint trimming:
    output: output/trimmed/1.txt
    jobid: 3
    wildcards: sample=1
    resources: tmpdir=/tmp
Downstream jobs will be updated after completion.


[Wed Jan 19 09:34:38 2022]
checkpoint trimming:
    output: output/trimmed/2.txt
    jobid: 6
    wildcards: sample=2
    resources: tmpdir=/tmp
Downstream jobs will be updated after completion.

[Wed Jan 19 09:34:39 2022]
Finished job 3.
1 of 7 steps (14%) done
[...]

PR #1332, opened 2022-01-13
fix: bug in pipe group handling that led to multiple assignments of the same group id to different groups; bug that accidentally added already running groups of the list of ready jobs (issue #1331)

Maarten-vd-Sande:
### Description

The failing tests of issue [1331](https://github.com/snakemake/snakemake/issues/1331)

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1455,6 +1455,7 @@ def test_modules_ruledeps_inheritance():
     run(dpath("test_modules_ruledeps_inheritance"))
 
 
+@skip_on_windows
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):
Thank you so much for providing the testcase. It took me half a day and was really a pain, but I managed to fix the bug!

iamh2o:
Awesome-  I will give it a test drive soon.  Will save me from a pretty gnarly patched production deployment I've been working with :-)

thank you--
john

PR #1340, opened 2022-01-20
feat: specify conda environments via their name

johanneskoester:
### Description

This replaces #494 with an object oriented solution.

### TODO

* [x] implement `CondaEnvNameSpec`
* [x] add test cases

### QC


snakemake/deployment/conda.py
```diff
@@ -583,24 +635,98 @@ def bin_path(self):
...
+    def is_file(self):
+        return True
+
+
+class CondaEnvNameSpec(CondaEnvSpec):
+    def __init__(self, name: str):
+        self.name = name
+
+    def apply_wildcards(self, wildcards):
+        return CondaEnvNameSpec(apply_wildcards(self.name, wildcards))
```
nh13, line 715: This is really nice to support wildcards in the conda environment name. Do you want to mention that briefly in the docs?

snakemake/jobs.py
```diff
@@ -316,34 +316,30 @@ def reset_params_and_resources(self):
...
-    @property
-    def conda_env_path(self):
-        return self.conda_env.path if self.conda_env else None
-
     def archive_conda_env(self):
         """Archive a conda environment into a custom local channel."""
-        if self.conda_env_file:
+        if self.conda_env_spec:
+            if self.conda_env.is_named:
+                raise WorkflowError(
```
nh13, line 338: I suppose that the names conda environment could be exported for reproducibility later?

nh13:
Thank you for making this feature, even though it’s not 100% aligned with the goal of reproducibility. I appreciate the pragmatism! 

And my apologies as I would not have been able to figure this myself without more explicit help understanding the code, but it will help my understanding for any future contributions.

snakemake/jobs.py
johanneskoester: Yes, that should be possible (have to store them by name in the archive then, but I'll have a look.

johanneskoester: Thought about this again. I don't think we can support that, because named environments could be modified in various ways that we cannot capture with the archiving process, at least not at the moment (it just extracts all conda packages). Of course, in theory one could try to come up with another way of archiving such an environment. But it will at least be hard to find a proper solution that is guaranteed to always work with any kind of modification.

johanneskoester: Thought about this again. I don't think we can support that, because named environments could be modified in various ways that we cannot capture with the archiving process, at least not at the moment (it just extracts all conda packages). Of course, in theory one could try to come up with another way of archiving such an environment. But it will at least be hard to find a proper solution that is guaranteed to always work with any kind of modification.

johanneskoester: Open to separate PRs here for the future of course.

nh13: I think you're right in the limitation.  If we run `conda export` it captures only conda packages and other python packages (eg. installed by `pip`), not any other binaries that may be moved into the environment (how insidious).  But that may be good enough?  On the other hand, there are platform specific differences such that what's exported depends on the OS anyhow, so it's not fully reproducible without a container anyhow.

nh13:
Thank you!!!

09oulin:
@johanneskoester @nh13 
Hello,

I am new to snakemake, and find it a excellent way to deal with redundant tasks.

But I couldn't find suitable guidance for using kubernetes. I want to know how to install and use snakemake in kubernetes, including environment building(install executor in the snakemake environment?), set the kubernetes config file, and submit task to background...

Besides, if the rules in my workflow need to be run in different environments, how could I do? So far, I build the needed environments with snakemake environment in the same pod, but I have no idea how to define it in my snakemake file? Could you please give some examples?

Thank you for your time.

Lin Ou

PR #1341, opened 2022-01-20
fix: remove raise that limits using --config with dicts

lpla:
### Description

Snakemake use through other Python scripts using its API class 
```
>>> import snakemake
>>> snakemake.snakemake(snakefile='Snakefile', config={"this_option": "does_not_break", "test": {'this_dict':'will_break'}})

  File "<stdin>", line 1, in <module>
  File "/home/lpla/bitextorment/lib/python3.8/site-packages/snakemake/__init__.py", line 508, in snakemake
    config_args = unparse_config(config)
  File "/home/lpla/bitextorment/lib/python3.8/site-packages/snakemake/__init__.py", line 960, in unparse_config
    raise ValueError("config may only be a flat dict")
ValueError: config may only be a flat dict
```
allows setting any kind of parameters but dictionaries. The code this PR removes was triggering an exception when there is no issue in using dictionaries as a setting inside the Python dictionary given to the Snakemake class through `config`. 

A simple test named `test_dicts_in_config()`, similar to the code above, is added to `tests/testapi.py` file to reproduce this issue. Test could be even simpler, as the Snakefile can be completely empty to reproduce the issue, but the one I wrote shows a rule using dicts through `config` working without the `raise`.

### QC


johanneskoester:
Thanks! Actually a test is of the CLI possible. This works via the shellcmd argument of the test run function, see this example: https://github.com/snakemake/snakemake/blob/main/tests/tests.py#L267

lpla:
Hi. Sorry about the initial version of the pull request. I thought we were running Snakemake through a Bash script, but it was a Python one, so we were using the Snakemake API. I updated the description and added a simple test. Let me know if this style is correct, or if I could do anything else to fix it better or cleaner.

lpla:
Also, I saw you edited the title as "feat:". Isn't this a bug?

Pbdas:
Just want to second this request. I ran into this limitation yesterday; as it is right now I'd have to choose between changing the config values I want to alter to be flat, or something janky like having a template config file in which I do string replacement. Passing a (nested) python dict is so much cleaner. Hope to see it in the next version so I can use it in production!

Pbdas:
Actually, there is a somewhat dirty hack to make this behaviour work correctly, by specifying some valid, or a meaningless dummy "config_args" value when calling snakemake from python:
```Python
snakemake.snakemake(
  ...
  # dict with just the values to be updated, properly nested as the config expects it
  config=config_overrrides,
  # some dummy config arg, as would normally be passed on the command line via --config-args
  config_args=["dummy=0"]
  )
```

This is because the throwing function `unparse_config` is only called if there is a `config` specified, but no `config_args`, see [here](https://github.com/snakemake/snakemake/blob/main/snakemake/__init__.py#L507)

PR #1355, opened 2022-01-27
Introduction of remote_prefix remotes

bcouturi:
### Description

The idea of this PR is to add a "remote_prefix" in RemoteObject, thus making sure we can have different paths for the local file and the remote one e.g. with

`
from snakemake.remote.XRootD import RemoteProvider as XRootDRemoteProvider
XRootD = XRootDRemoteProvider(stay_on_remote=False,  keep_local=True, 
                              remote_prefix="root://server//eos/path/")

rule test1:
    input:
        "input.txt"
    output:
        XRootD.remote("output.txt")
    shell:
        "echo 'Output is: {output}.'; cat {input} > {output}"
`

The local output would be 'output.txt' but the remote file would be 'root://server//eos/path/output.txt'

### QC


bcouturi:
Hi,
The AbstractRemoteProvider has been updated indeed, and the remote_file of the AbstractRemoteObject has been changed to add the remote_prefix after the protocol.

Issue #1367, opened 2022-02-02
Allow duplicate rule outputs if they are empty?


vinisalazar:
Hi,

**Is your feature request related to a problem? Please describe.**
I have a rule in my workflow that has four outputs. However, I want two of these outputs to only be generated if a value in config is set to true. The way I achieved this by doing this:

```python
rule my_rule:
    [...]
    output:
        optional_output_one: "file1.txt" if config["optional_output_one"] else (),
        optional_output_two: "file2.txt" if config["optional_output_two"] else (),
```

However, I get the following error when linting my workflow:
```
Duplicate output file pattern in rule my_rule. First two duplicate for entries optional_output_one and optional_output_two
```

**Describe the solution you'd like**
I modified the [following line](https://github.com/snakemake/snakemake/blob/8c5c5e8261ae12fc40c3c293045dea6af7df0e01/snakemake/rules.py#L447) in function `check_output_duplicates` of `rules.py`:
```python
if value in seen:
```
to
```python
if value and value in seen:
```
and it fixed the problem. This disables the check if it's an empty value for the output file.

**Describe alternatives you've considered**
If you have any other suggestions or patterns on how to implement this logic that is better than what I've done, that would be very helpful.

**Additional context**
I'm not entirely sure if this change would introduce undesired or unexpected behaviour; however, if you agree with it, I'd appreciate the opportunity of contributing to Snakemake.

Thank you for any assistance you can provide.

Best,
Vini

PR #1368, opened 2022-02-02
fix: don't raise WorkflowError when entry is empty

vinisalazar:
Hi, this is in relation to #1367.

Summary of changes:
  - Modify function 'check_output_duplicates' to check for empty value
  - Complete docstring
  - Add period at the end of raise message

### Description

Fixes the aforementioned issue. I ran pytest (skipped `test_tibanna` which was failing due to lack of credentials) and all tests passed.

### QC


johanneskoester (approved):
Thanks, good catch!

PR #1373, opened 2022-02-03
fix: CondaEnvFileSpec lose the rule object to create new IOFile object

ddesvillechabrol:
### Description

Hello !
This PR is a bugfix for the bug that broke the tests for snakemake-wrappers like this one https://github.com/snakemake/snakemake-wrappers/pull/449.
The implementation with `CondaEnvFileSpec` seems to lose the rule object needed to create new IOFile object.
I am not sure but my bugfix may lead to memory leak.
Another solution is to pass the rule object when the `apply_wildcards` is called.
Cheers
Dimitri

### QC


johanneskoester:
Thanks a lot, also for pointing me to the issue! Closing in favor of a slightly different fix (as you already suggest) in PR #1377.

Issue #1372, opened 2022-02-03
LocalGitFile never opend due to variable passed as string


Smeds:
I have been trying to use repos on my local computer/servers when importing snakemake modules without success and believe I have found a possible issue in the source code.

in the sourcecache.py, snakemake files are opened at line 340 by passing a string object to _open

https://github.com/snakemake/snakemake/blob/5fabffbb4af8e9e122677e5adeaebf2d6bd0eeb3/snakemake/sourcecache.py#L339-L341

but it looks a bit like the code expects a LocalGitFile (line 348) or a SourceFile (line 357). If I remove get_path_or_uri() from line 340 I'm able to successfully open the Snakefile in my module, however it fails when trying to run include statements in the newly opened Snake file. 

Example: fetching module from github which seem to work
```
from snakemake.utils import min_version
min_version("6.0")

configfile: "config/config.yaml"

module dna_seq:
    snakefile: github("snakemake-workflows/dna-seq-gatk-variant-calling", path="workflow/Snakefile", tag="main")
    config: config

use rule * from dna_seq

# easily extend the workflow
rule plot_vafs:
    input:
        "filtered/all.vcf.gz"
    output:
        "results/plots/vafs.svg"
    notebook:
        "notebooks/plot-vafs.py.ipynb"
```

modifying the source I'm able to open the snakefile but not call include in the Snakefile

```
from snakemake.utils import min_version
min_version("6.0")

configfile: "config/config.yaml"

module dna_seq:
    snakefile: sourcecache.LocalGitFile("file:///beegfs-storage/..../dna-seq-gatk-variant-calling", path="workflow/Snakefile", tag="main")
    config: config

use rule * from dna_seq

# easily extend the workflow
rule plot_vafs:
    input:
        "filtered/all.vcf.gz"
    output:
        "results/plots/vafs.svg"
    notebook:
        "notebooks/plot-vafs.py.ipynb"
```

error
```
TypeError in line 1 of git+/beegfs-storage/..../dna-seq-gatk-variant-calling/workflow/Snakefile@main:
__init__() missing 1 required positional argument: 'path'
  File "/beegfs-storage/..../Snakefile.test.local", line 11, in <module>
  File "git+/beegfs-storage/..../dna-seq-gatk-variant-calling/workflow/Snakefile@main", line 1, in <module>
```

johanneskoester:
Fixed with PR #1376.

Issue #1308, opened 2021-12-20
Unable to use LocalGitFile


Smeds:
https://github.com/snakemake/snakemake/blob/1349254de57ced0466bfe1e98f0f7c2a1b247d2d/snakemake/sourcecache.py#L328-L353

I have successfully been using sourcecache.GithubFile to be able to import snakemake module located on github but not been able to use GitLocalFile to import local repos, which is needed on our cluster that have no access to internet.

And it seems to me that  def _open(self, path_or_uri, mode) (line 335) in sourcecache will never enter that part that fetch a local repo. Since the code in function def _do_cache(self, source_file, cache_entry) doesn't provide _open with SourceFile, it instead provides a string generated by get_path_or_uri()

Smeds:
I have tried to update the code so that it makes a valid check of a LocalGilFile. But now it fails  later with a 

```
TypeError in line 7 of ....
__init__() missing 1 required positional argument: 'path

```

Smeds:
If I try the example in the docs with a repo on github everything works fine
```
from snakemake.utils import min_version
min_version("6.0")

configfile: "config/config.yaml"

module dna_seq:
    snakefile: github("snakemake-workflows/dna-seq-gatk-variant-calling", path="workflow/Snakefile", tag="main")
    config: config

use rule * from dna_seq

# easily extend the workflow
rule plot_vafs:
    input:
        "filtered/all.vcf.gz"
    output:
        "results/plots/vafs.svg"
    notebook:
        "notebooks/plot-vafs.py.ipynb"
```

but if I clone the repo and update the file to this
```
from snakemake.utils import min_version
min_version("6.0")

configfile: "config/config.yaml"

module dna_seq:
    snakefile: sourcecache.LocalGitFile("/home/patrik/test/dna-seq-gatk-variant-calling", path="workflow/Snakefile", tag="main")
    config: config

use rule * from dna_seq

# easily extend the workflow
rule plot_vafs:
    input:
        "filtered/all.vcf.gz"
    output:
        "results/plots/vafs.svg"
    notebook:
        "notebooks/plot-vafs.py.ipynb"
```

I get the following error:
```
WorkflowError in line 12 of /home/patrik/test/Snakefile:
Failed to open source file /home/patrik/test/git+/home/patrik/test/dna-seq-gatk-variant-calling/workflow/Snakefile@main
[...]

PR #1376, opened 2022-02-04
feat: add gitfile option to make it possible to use local git repos when importing modules

Smeds:
Feat: add gitfile option to make it possible to use local git repos when importing modules.

Should make it possible to include module like this:
```
module dna_seq:
    snakefile: gitfile("file:///beegfs-storage/LOCAL_PATH/dna-seq-gatk-variant-calling", path="workflow/Snakefile", tag="main")
    config: config
```

solves issues:
- #1372 
- #1308


snakemake/sourcecache.py
```diff
@@ -349,7 +365,7 @@ def _open(self, path_or_uri, mode):
             import git
 
             return io.BytesIO(
-                git.Repo(path_or_uri.repo_path)
+                git.Repo(path_or_uri.repo_path.replace("file:/",""))
```
johanneskoester, line 368: Maybe rather have this replace as a method in the LocalGitFile class?

snakemake/sourcecache.py
```diff
@@ -266,9 +281,10 @@ def infer_source_file(path_or_uri, basedir: SourceFile = None):
         if not os.path.isabs(path_or_uri) and basedir is not None:
             return basedir.join(path_or_uri)
         return LocalSourceFile(path_or_uri)
+
     if path_or_uri.startswith("git+file:"):
         root_path, file_path, ref = split_git_path(path_or_uri)
-        return LocalGitFile(root_path, file_path, ref=ref)
+        return LocalGitFile("file:/" + root_path, file_path, ref=ref)
```
johanneskoester, line 287: Why adding this prefix here? The class name already says that it is a local thing?

snakemake/sourcecache.py
```diff
@@ -251,6 +262,10 @@ def infer_source_file(path_or_uri, basedir: SourceFile = None):
             return path_or_uri
         else:
             path_or_uri = path_or_uri.get_path_or_uri()
+    if isinstance(basedir, LocalGitFile) and isinstance(path_or_uri, str):
+        local_base = os.path.dirname(basedir.path).lstrip("/")
```
johanneskoester, line 266: The SourceFile class has a get_basedir method.

johanneskoester: It also has a join method. In fact, I am quite sure that no special handling for a local git file should be necessary. Otherwise, we should rather fix it in the class hierarchy

snakemake/sourcecache.py
```diff
@@ -337,7 +353,7 @@ def _cache(self, source_file):
 
     def _do_cache(self, source_file, cache_entry):
         # open from origin
-        with self._open(source_file.get_path_or_uri(), "rb") as source, open(
+        with self._open(source_file, "rb") as source, open(
```
johanneskoester, line 356: Good catch!

snakemake/sourcecache.py
```diff
@@ -126,13 +126,24 @@ def __init__(
         self.repo_path = repo_path
         self.path = path
 
+    def get_basedir(self):
+        root_path, file_path, ref = split_git_path(self.get_path_or_uri())
+        file_name = "{}/{}".format(root_path, file_path).replace(root_path, "")
+        return LocalGitFile("file:/{}".format(root_path), file_name, ref=ref)
```
johanneskoester, line 132: What's wrong with the get_basedir method of the SourceFile class?

snakemake/sourcecache.py
```diff
@@ -126,13 +126,24 @@ def __init__(
...
+
     def get_path_or_uri(self):
         return "git+{}/{}@{}".format(self.repo_path, self.path, self.ref)
 
     def join(self, path):
+        if isinstance(path, LocalSourceFile):
+            path = path.get_path_or_uri()
```
johanneskoester, line 139: Seems cumbersome to me to have such special handling in a subclass like this. Other classes assume that path is a string. Why do you think this is needed?

snakemake/sourcecache.py
```diff
@@ -126,13 +126,24 @@ def __init__(
...
+        local_path = os.path.dirname(self.path).lstrip("/")
+        while path.startswith("../"):
+            path = path.lstrip("../")
+            local_path = os.path.dirname(local_path)
```
johanneskoester, line 143: Can you add some comments explaining why this is needed?

johanneskoester (changes requested):
Thanks a lot. Some comments below. Apart from that, can you add some test case(s)?

snakemake/sourcecache.py
Smeds: I tried to modify the code and ended up with having problems at

```
  File "/home/patsm159/workspace/snakemake/snakemake/utils.py", line 63, in validate
    schemafile = workflow.current_basedir.join(schemafile)
```

would it be ok to change 

`schemafile = workflow.current_basedir.join(schemafile)`

to

`schemafile = workflow.current_basedir.join(schemafile.get_path_or_uri())`

snakemake/sourcecache.py
Smeds: refactored and removed

snakemake/sourcecache.py
Smeds: refactored and removed

Smeds: refactored and removed

snakemake/sourcecache.py
Smeds: refactored

snakemake/sourcecache.py
Smeds: I have update the code

snakemake/sourcecache.py
Smeds: we have a common.smk with the following format for schemas: 

https://github.com/snakemake-workflows/rna-seq-star-deseq2/blob/dfd0b9cb8c1ef8607505aef8baf885632227e8ac/workflow/rules/common.smk#L8-L10

which git.show didn't like since it generates a path looking like this 
`project/workflow/rules/../schemas/samples.schema.yaml"`

the code has been refactored and are now using os.path.normpath

which will change

`project/workflow/rules/../schemas/samples.schema.yaml"`

to 

`project/workflow/schemas/samples.schema.yaml"`

Smeds:
> Thanks a lot. Some comments below. Apart from that, can you add some test case(s)?

I have updated the code now.

I could add small git repo in the tests folder with commits that a workflow will have to use.

Smeds:
> Thanks a lot. Some comments below. Apart from that, can you add some test case(s)?

I have updated the code now.I could add small git repo in the tests folder with commits that a workflow will have to use.

I could add small git repo in the tests folder with commits that a workflow will have to use.

Smeds:
I need some thoughts of how to handle tests for this pull-request. I have create "module" use to test the new functionality:
1 - A successfull use of gitfile
2 - A case where a rule is missing
3 - A case where a schema is missing

But to be able to run the test I would have to include a .git folder with the module, which git isn't alllowing me to do. 

The only solution I can think of now is to compress the folder and decompress it before running the different tests.

@johanneskoester  Do you have any good suggestion of how to handle this?

Smeds:
Could git Submodules be a acceptable solution?

Smeds:
> Could git Submodules be a acceptable solution?

Didn't work

Smeds:
> )

I have added test cases now

snakemake/sourcecache.py
```diff
@@ -148,41 +135,37 @@ def __init__(
...
             commit=self.commit,
-            ref=self.ref,
         )
 
     def is_persistently_cacheable(self):
         return False
 
     def get_filename(self):
-        return posixpath.basename(self.path)
+        return os.path.basename(self.path)
```
johanneskoester, line 163: What is the reason to not use posixpath here?

snakemake/sourcecache.py
```diff
@@ -306,12 +292,7 @@ def infer_source_file(path_or_uri, basedir: SourceFile = None):
             return basedir.join(path_or_uri)
         return LocalSourceFile(path_or_uri)
     if path_or_uri.startswith("git+file:"):
-        try:
-            root_path, file_path, ref = split_git_path(path_or_uri)
-        except Exception as e:
-            raise WorkflowError(
-                f"Failed to read source {path_or_uri} from git repo.", e
-            )
+        root_path, file_path, ref = split_git_path(path_or_uri)
```
johanneskoester, line 295: Why removing the exception handling here?

snakemake/sourcecache.py
```diff
@@ -346,7 +327,7 @@ def runtime_cache_path(self):
 
     def open(self, source_file, mode="r"):
         cache_entry = self._cache(source_file)
-        return self._open(LocalSourceFile(cache_entry), mode)
+        return self._open(cache_entry, mode)
```
johanneskoester, line 330: Please revert this to what we had before. The LocalSourceFile for the cache entry is intentional.

snakemake/sourcecache.py
```diff
@@ -397,34 +378,20 @@ def _do_cache(self, source_file, cache_entry):
...
 
-        if isinstance(source_file, LocalGitFile):
+        if isinstance(path_or_uri, LocalGitFile):
             import git
 
             return io.BytesIO(
-                git.Repo(source_file.repo_path)
-                .git.show("{}:{}".format(source_file.ref, source_file.path))
+                git.Repo(path_or_uri.repo_path)
+                .git.show("{}:{}".format(path_or_uri.ref, path_or_uri.path))
```
johanneskoester, line 389: I believe these changes are not needed anymore. I had already refactored the main branch based on ideas from your PR before.

snakemake/sourcecache.py
```diff
@@ -397,34 +378,20 @@ def _do_cache(self, source_file, cache_entry):
...
-                backoff=2,
-                logger=logger,
-            )
-
-    def _open(self, source_file, mode):
+    def _open(self, path_or_uri, mode):
         from smart_open import open
 
-        if isinstance(source_file, LocalGitFile):
+        if isinstance(path_or_uri, LocalGitFile):
```
johanneskoester, line 384: I believe these changes are not needed anymore. I had already refactored the main branch based on ideas from your PR before.

snakemake/sourcecache.py
```diff
@@ -397,34 +378,20 @@ def _do_cache(self, source_file, cache_entry):
...
-                self._open,
-                [source_file, mode],
-                tries=3,
-                delay=3,
-                backoff=2,
-                logger=logger,
-            )
-
-    def _open(self, source_file, mode):
+    def _open(self, path_or_uri, mode):
```
johanneskoester, line 381: I believe these changes are not needed anymore. I had already refactored the main branch based on ideas from your PR before.

snakemake/sourcecache.py
```diff
@@ -397,34 +378,20 @@ def _do_cache(self, source_file, cache_entry):
...
-                git.Repo(source_file.repo_path)
-                .git.show("{}:{}".format(source_file.ref, source_file.path))
+                git.Repo(path_or_uri.repo_path)
+                .git.show("{}:{}".format(path_or_uri.ref, path_or_uri.path))
                 .encode()
             )
 
-        path_or_uri = source_file.get_path_or_uri()
+        if isinstance(path_or_uri, SourceFile):
+            path_or_uri = path_or_uri.get_path_or_uri()
```
johanneskoester, line 394: I believe these changes are not needed anymore. I had already refactored the main branch based on ideas from your PR before.

snakemake/sourcecache.py
```diff
@@ -286,10 +265,17 @@ def get_path_or_uri(self):
...
+        if (
+            basedir is None
+            or isinstance(path_or_uri, HostingProviderFile)
+            or isinstance(path_or_uri, LocalGitFile)
+        ):
             return path_or_uri
         else:
             path_or_uri = path_or_uri.get_path_or_uri()
+    if isinstance(basedir, LocalGitFile):
+        return basedir.join(path_or_uri)
```
johanneskoester, line 277: I think this is not needed. We take care of it in the block below already.

johanneskoester (changes requested):

snakemake/sourcecache.py
Smeds: old habit from my side! Are there any benefits of using posixpath compared to os.path?

snakemake/sourcecache.py
Smeds: I have readded the check

snakemake/sourcecache.py
Smeds: reverted to the old version

snakemake/sourcecache.py
Smeds: I have removed the changes

snakemake/sourcecache.py
Smeds: I have removed the changes

snakemake/sourcecache.py
johanneskoester: It always uses /, whereas path uses \ on windows (which is not what we want for urls).

johanneskoester (changes requested):
There are still some changes which

Smeds:
@johanneskoester  would there be any point of supporting relative path when importing a module using gitfile? At the moment I believe the could requires an absolute path.

Smeds:
> 

done

Smeds:
@johanneskoester  it should now allow for relative paths

johanneskoester (approved):

Issue #570, opened 2020-08-17
Canceling jobs through Snakemake profiles?


holtgrewe:
**Is your feature request related to a problem? Please describe.**

- When running cluster jobs through DRMAA, they are nicely canceled when Snakemake is stopped with Ctrl-C. This is not the case for the Slurm profile.

**Describe the solution you'd like**

- Similarly to when running DRMAA cluster jobs, I'd like to have child jobs being killed (e.g., via `scancel`).

**Describe alternatives you've considered**

- I don't think there are alternatives. I have not seen any mention of "canceling" in the Slurm or SURF profile.

**Additional context**

- Maybe I'm missing something?

percyfal:
Hi @holtgrewe, I have just been reviewing a related bug report for the slurm profile (referenced above). Much of the solutions you request cannot be implemented in the profile itself, but would require modifications to the job executors, maybe even adding a specific SlurmExecutor; I believe there is an ongoing effort to refactor the executors. `scancelling` jobs should work, but whereas metadata for failed jobs is removed, incomplete files are not; see [Persistence.cleanup](https://github.com/snakemake/snakemake/blob/07f73006dcee1e6ce6c271355fff83705f188de4/snakemake/persistence.py#L201). Adding a `job.cleanup()` statement here seems to fix that particular issue, but there may be a reason for it's absence that I'm unaware of.

holtgrewe:
Hi @percyfal thanks for your reply.

IMO it's borderline cheeky to mark `--cluster`/`--drmaa` support in Snakemake as deprecated without the profiles offering equivalent functionality. Well, the good is the enemy of the better. I'd dearly like to see profile support to reach the same level of features as `--drmaa` as using `--time=7-00` is much nicer than saying `--time=168:00:00`...

PR #1395, opened 2022-02-14
feat: Adding --cluster-cancel and --cluster-cancel-nargs

holtgrewe:
### Description

This PR adds the optional arguments `--cluster-cancel` and `--cluster-mcancel`. The user can specify commands for terminating jobs by their job ID, either one at a time (`--cluster-cancel`) or multiple at once (`--cluster-mcancel`). Testing this was somewhat tricky because you have to simulate senting `SIGINT` to the job so I ended extending `tests/common.py` a bit.

### QC

* [x] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of 

Addresses: #570


snakemake/executors/__init__.py
```diff
@@ -976,8 +980,38 @@ def __init__(
...
+                chunks = list(_chunks(jobids, max_mcancel))
+            else:
+                cmd = self.cancelcmd
+                chunks = list(_chunks(jobids, 1))
+            # Go through the chunks and cancel the jobs.
+            for chunk in chunks:
+                try:
+                    subprocess.Popen([cmd] + chunk, shell=False)
+                except:
+                    # This is common - logging a warning would probably confuse the user.
```
johanneskoester, line 1007: True, but we should collect the number of failed cancel commands, and issue a warning at the end that tells the user how many of them failed (if that number is greater than 0).

tests/common.py
```diff
@@ -156,16 +159,24 @@ def run(
...
-            )
-            success = True
+            if sigint_after is None:
+                subprocess.check_output(
+                    shellcmd, cwd=path if no_tmpdir else tmpdir, shell=True
+                )
+                success = True
+            else:
+                with subprocess.Popen(
+                    shellcmd.split(" "), cwd=path if no_tmpdir else tmpdir
```
johanneskoester, line 169: I think the idiomatic way is to use `shutil.split`.

johanneskoester (changes requested):
Thanks a lot, very nice work! The Snakefile of your testcase is missing. Apart from that, a few requests below.

snakemake/executors/__init__.py
holtgrewe: Agreed, this was here before but will be fixed.

snakemake/executors/__init__.py
```diff
@@ -976,8 +980,38 @@ def __init__(
             )
 
     def cancel(self):
-        logger.info("Will exit after finishing currently running jobs.")
-        self.shutdown()
+        # maximum number of jobs to cancel at once
+        max_mcancel = 1000
```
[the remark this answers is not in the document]
holtgrewe: Good point, will add.

holtgrewe:
Sorry for the many pushes but I had a hard time uncovering why the CI was failing and it was running locally (as so often, timing issues because of fewer cores in CI than in dev). I'm positive that this is ready for re-review :eyeglasses: :mag: :microscope:

holtgrewe:
:tada:

johanneskoester (approved):
Excellent, awesome and clean work, as always!

PR #1397, opened 2022-02-15
feat: cluster sidecar

holtgrewe:
### Description

This PR adds a new argument `--cluster-sidecar` argument, as documented. This is necessary to implement a REST server as as a caching layer for `--cluster-status` queries, for example. See https://github.com/Snakemake-Profiles/slurm/issues/81 for an example use case.

**based on #1395, should be rebased once that is merged**

### QC


johanneskoester:
PR #1395 is merged, I guess you can rebase this now.

johanneskoester (approved):

PR #1414, opened 2022-02-21
feat: define grouplocal jobs [WIP]

johanneskoester:
### Description

```python
rule a:
    output:
        "foo.txt"
    grouplocal: True # repeat a job from this rule in each instance of the defined group
    group: "a"
    shell:
        ...

```
### QC


johanneskoester:
Closing in favor of #1418

PR #1451, opened 2022-03-02
fix: work around segfault with >100 jobs in google life sciences backend

cademirch:
### Description

PR to fix segfault issue with `--google-lifesciences` with larger (at least 100 jobs) workflows. #1444 
### QC


vsoch:
@moschetti would you mind taking a look at this and we can talk about what might be best practice? E.g., importing and defining inline vs. defining globally vs limiting the scope vs making it customizable, etc.

snakemake/executors/google_lifesciences.py
```diff
@@ -897,24 +899,27 @@ def _retry_request(self, request, timeout=2, attempts=3):
         attempts: remaining attempts, throw error when hit 0
         """
         import googleapiclient
-
+        import google.auth
+        credentials, project_id = google.auth.default(scopes=["https://www.googleapis.com/auth/cloud-platform"])
```
johanneskoester, line 903: @vsoch do we authenticate this way in the other places as well? Or can there maybe be an auth object for the entire life science executor object instead creating a special one here in the retry?

johanneskoester (changes requested):

snakemake/executors/google_lifesciences.py
vsoch: We go about it different ways, but I suspect under the hood since both are using the default application credentials (exported to the environment as `GOOGLE_APPLICATION_CREDENTIALS`) the methods are similar. We use the discovery API clients to authenticate clients that are attached to the entire class, e.g., here https://github.com/snakemake/snakemake/blob/417f40deb06a01d836c4ec632fc3b8f67672a403/snakemake/executors/google_lifesciences.py#L136-L165. And if you look into the source code, the googleapiclient discovery has that same "google.auth.default" function [here](https://github.com/googleapis/google-api-python-client/blob/9639c0709897ab4f56e4fb0e0fe78f2c339e59fe/googleapiclient/_auth.py#L51-L66) which will call that same auth function via [here](https://github.com/googleapis/google-api-python-client/blob/9639c0709897ab4f56e4fb0e0fe78f2c339e59fe/googleapiclient/discovery.py#L575-L579). Notably build takes an http object [here](https://github.com/googleapis/google-api-python-client/blob/9639c0709897ab4f56e4fb0e0fe78f2c339e59fe/googleapiclient/discovery.py#L179) and given that we don't provide one and it's generated,I think it might be saved on the returned service objects [here](https://github.com/googleapis/google-api-python-client/blob/9639c0709897ab4f56e4fb0e0fe78f2c339e59fe/googleapiclient/discovery.py#L1345) so I think what I would do is compare the object there (likely directly or its credentials) to the one created above. If they are comparable, arguably we could grab this object from one of the service objects generated from the discovery build API. The discovery build APIs are going to be different to have scopes for their respective endpoints, vs the one above which is all of google cloud. We could also try the reverse - creating one credential object and then seeing if we can pass to the services, although we'd want to double check how the scopes line up. And if neither of those ideas work, we could minimally have this call in the same function so they are in the same place!

vsoch: The request object is generated ultimately _from_ one of those services - either directly or as a result of doing like pipelines.run() so I'm curious why the request object isn't coming with its own http already (or maybe some of them are but not consistently?) E.g., pipelines.run() here is the one that was originally giving us trouble: https://github.com/snakemake/snakemake/blob/bf5fdfe2a0547bc20873ee4afcbd438928368c02/snakemake/executors/google_lifesciences.py#L831 and maybe @moschetti knows what using these endpoints and generated from a discovery endpoint, e.g, [here](https://cloud.google.com/life-sciences/docs/reference/rest/v2beta/projects.locations.pipelines/run) doesn't generate a request with a comparable http object that we are generating later in retry request?

snakemake/executors/google_lifesciences.py
```diff
@@ -897,24 +899,27 @@ def _retry_request(self, request, timeout=2, attempts=3):
...
+        if http is None:
+            http = google_auth_httplib2.AuthorizedHttp(credentials=credentials, http=httplib2.Http())
```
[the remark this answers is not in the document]
cademirch: Yes, that was my rationale - I'll add a comment.

snakemake/executors/google_lifesciences.py
cademirch: In my first attempt I tried passing the credentials created here: https://github.com/snakemake/snakemake/blob/417f40deb06a01d836c4ec632fc3b8f67672a403/snakemake/executors/google_lifesciences.py#L150
To `google_auth_httplib2.AuthorizedHttp` in `_retry_request`, but it did not like those credentials giving this error: `AttributeError: '_JWTAccessCredentials' object has no attribute 'before_request'` which led me to this [issue](https://github.com/googleapis/google-cloud-python/issues/3281), where I found to use `google.auth` for the credentials.

cademirch:
So I pushed some new commits using [google's recommended way](https://googleapis.github.io/google-api-python-client/docs/thread_safety.html) to build the service apis. Seems to work so far on the test case. Still testing on a more complex workflow.

Sorry if the commits are a bit messy.

vsoch:
No need to apologize, you're doing great.

johanneskoester (approved):
Awesome work @cademirch! So, am I getting it right that basically using google.auth overall solves the segfault issue then?

johanneskoester:
Let us know once your larger tests pass. I will wait with the merging until then.

johanneskoester:
> @johanneskoester
> 
> > Awesome work @cademirch! So, am I getting it right that basically using google.auth overall solves the segfault issue then?
> 
> Yes mostly. Though I believe using google_auth_httplib2 for building the requests is the more critical part of the solution.
> 
> > Let us know once your larger tests pass. I will wait with the merging until then.
> 
> My larger workflow ran overnight without error, though it did end prematurely:
> 
> ```
> Finished job 572.
> 162 of 269 steps (60%) done
> Complete log: /mnt/ccgp_workflow/.snakemake/log/2022-03-03T210937.628868.snakemake.log
> ```
> 
> I don't think this premature ending is related to this issue though, I've seen this behavior before when running on cloud execution mode.

Mhm, that premature ending is kind of worrying as well of course. I have never seen anything like that. Can you file a separate issue for that or is it already open?

cademirch:
> > @johanneskoester
> > > Awesome work @cademirch! So, am I getting it right that basically using google.auth overall solves the segfault issue then?
> > 
> > 
> > Yes mostly. Though I believe using google_auth_httplib2 for building the requests is the more critical part of the solution.
> > > Let us know once your larger tests pass. I will wait with the merging until then.
> > 
> > 
> > My larger workflow ran overnight without error, though it did end prematurely:
> > ```
> > Finished job 572.
> > 162 of 269 steps (60%) done
> > Complete log: /mnt/ccgp_workflow/.snakemake/log/2022-03-03T210937.628868.snakemake.log
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't think this premature ending is related to this issue though, I've seen this behavior before when running on cloud execution mode.
> 
> Mhm, that premature ending is kind of worrying as well of course. I have never seen anything like that. Can you file a separate issue for that or is it already open?

Yeah I have only seen it a few times. I think it could be related to `--until` which I did use in testing my large workflow with this PR. I'll look into it and open an issue.

PR #1462, opened 2022-03-07
feat: trigger rerun when new files are yield by an input function

cclienti:
### Description

Added the "missing" input flag to trigger rerun when new files are yield by an input function. Such flag is forwarded to all elements yield by the input function.

### QC

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -996,7 +996,12 @@ class DAG:
                 output_mintime_ = output_mintime.get(job)
                 if output_mintime_:
                     updated_input = [
-                        f for f in job.input if f.exists and f.is_newer(output_mintime_)
+                        f
+                        for f in job.input
+                        if (
+                            (f.exists and f.is_newer(output_mintime_))
+                            or (not f.exists and is_flagged(f, "missing"))
+                        )
                     ]
                     reason.updated_input.update(updated_input)
             if noinitreason and reason:
--- a/docs/project_info/faq.rst
+++ b/docs/project_info/faq.rst
@@ -48,7 +48,7 @@ For debugging such cases, Snakemake provides the command line flag ``--debug-dag
 
 In addition, it is advisable to check whether certain intermediate files would be created by targetting them individually via the command line.
 
-Finally, it is possible to constrain the rules that are considered for DAG creating via ``--allowed-rules``. 
+Finally, it is possible to constrain the rules that are considered for DAG creating via ``--allowed-rules``.
 This way, you can easily check rule by rule if it does what you expect.
 However, note that ``--allowed-rules`` is only meant for debugging.
 A workflow should always work fine without it.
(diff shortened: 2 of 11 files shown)


cclienti:
I think that the proposed PR cannot work properly when using "temp" flags.

snakemake/dag.py
```diff
@@ -996,7 +996,9 @@ def update_needrun(job):
                 output_mintime_ = output_mintime.get(job)
                 if output_mintime_:
                     updated_input = [
-                        f for f in job.input if f.exists and f.is_newer(output_mintime_)
+                        f
+                        for f in job.input
+                        if (f.exists and f.is_newer(output_mintime_)) or (not f.exists)
```
cclienti, line 1001: @johanneskoester

In order to handle new missing files returned by an input function (after a first successful workflow run) I tried to change the condition by adding `or (not f.exists)` but this is not a valid option because of temporary output.

I'm currently trying to replace the condition (not yet pushed) `if (f.exists and f.is_newer(output_mintime_))` by something using a new flag like "missing": `if (f.exists and f.is_newer(output_mintime_)) or (not f.exists and is_flagged(f, "missing")`.

So I added the new flag "missing" in the code but I discovered that flags are not "forwarded" to input function elements during wildcards resolution.  I don't know how to do that and if you think that it would be a good idea.

cclienti: I've just pushed a new version where the new "missing" flag is applied correctly to all element yield by the input function.

johanneskoester:
Hi, thanks a lot for the proposal. However, I think Snakemake has a mechanism for that purpose already, which is --list-input-changes. The latest versions of Snakemake already warn you in such cases at the end of a dryrun and suggest a command to trigger a rerun. Hence, there is not even manual flagging needed as you have implemented it here. What do you think?

cclienti:
Thank you for the response. I agree and I tried the proposed mechanism, but sometimes the command line length is too long because of thousands new files. Moreover the DAG is analyzed two times which can be long for some large workflows (>100k nodes).

So, in such case I was forced to create a snakemake-update python code that parses the arguments (using the internal API) to add the "list-input-changes" option and run snakemake to get the target python list. Then the snakemake can run a second times with the python list of changes. A same class solution would be to pass to snakemake command line the targets using a file, but the two run mechanism would be the same.

I was thinking that adding a flag would be nicer solution for such use case (only one snakemake run). But maybe I did not see the big picture or implemented things correctly.

johanneskoester:
I agree about the wasting of runtime. There is anyway the plan to implement these auxilliary change detections as an arg like `--rerun-changes params input code`. The mechanism would be basically the same as now, but no recomputation would be needed because it can happen directly before the needrun determination in the DAG. Would that be a solution to your problem?

johanneskoester:
See also here: https://github.com/snakemake/snakemake/pull/1107

So, maybe it is time to finally implement that now. I'll try to come to that ASAP.

cclienti:
> I agree about the wasting of runtime. There is anyway the plan to implement these auxilliary change detections as an arg like `--rerun-changes params input code`. The mechanism would be basically the same as now, but no recomputation would be needed because it can happen directly before the needrun determination in the DAG. Would that be a solution to your problem?

Yes this is also a good solution 👍

cclienti:
Thank you for mentioning me in the release note 7.8.0. You just made a little error on my nick, this is cclienti, not cclienty :D

PR #1464, opened 2022-03-08
fix: implement lock-free source file caching. This avoids hangs on network file systems like NFS.

johanneskoester:
### QC


johanneskoester:
cc @alethomas @thomasbtf

Issue #1463, opened 2022-03-07
Do no create conda environments within cluster jobs by default


sebschmi:
**Snakemake version**
7.1.0

**Describe the bug**
It seems that in cluster mode, only those conda environments required for local jobs are created locally. The others are created in the cluster jobs. But if there is more than one cluster job with the same conda environment, the jobs try to create the environment at the same time and fail. This happens on a shared distributed file system.

**Logs**
This is an example log of a cluster job that creates an environment simultaneously with other jobs:

```
Building DAG of jobs...
Creating conda environment config/conda-biopython-env.yml...
Downloading and installing remote packages.
CreateCondaEnvironmentException:
Could not create conda environment from /proj/sebschmi/git/practical-omnitigs/config/conda-biopython-env.yml:
warning  libmamba Cannot lock '/proj/sebschmi/miniforge3/envs/practical-omnitigs/pkgs/cache/497deca9.solv'
    Waiting for other mamba process to finish
warning  libmamba Cannot lock '/proj/sebschmi/miniforge3/envs/practical-omnitigs/pkgs/cache/09cdf8bf.solv'
    Waiting for other mamba process to finish
warning  libmamba Cannot lock '/proj/sebschmi/miniforge3/envs/practical-omnitigs/pkgs/cache/ffeee55f.solv'
    Waiting for other mamba process to finish
warning  libmamba Cannot lock '/proj/sebschmi/miniforge3/envs/practical-omnitigs/pkgs/cache/2a957770.solv'
    Waiting for other mamba process to finish
warning  libmamba Cannot lock '/proj/sebschmi/miniforge3/envs/practical-omnitigs/pkgs/cache/47929eba.solv'
    Waiting for other mamba process to finish
  Package               Version  Build                Channel                    Size
───────────────────────────────────────────────────────────────────────────────────────
  Install:
───────────────────────────────────────────────────────────────────────────────────────

  + _libgcc_mutex           0.1  conda_forge          conda-forge/linux-64     Cached
  + _openmp_mutex           4.5  1_gnu                conda-forge/linux-64     Cached
  + biopython              1.79  py310h6acc77f_1      conda-forge/linux-64     Cached
  + bzip2                 1.0.8  h7f98852_4           conda-forge/linux-64     Cached
  + ca-certificates   2021.10.8  ha878542_0           conda-forge/linux-64     Cached
  + ld_impl_linux-64     2.36.1  hea4e1c9_2           conda-forge/linux-64     Cached
[...]

PR #1472, opened 2022-03-11
fix: always deploy conda envs in main process when assuming a shared file system (fixes issue #1463)

johanneskoester:
### QC


johanneskoester:
cc @dlaehnemann

PR #1501, opened 2022-03-19
Tibanna

SooLee:
### Description

This one fixes tibanna disk_mb error.

### QC

* [v ] The PR contains a test case for the changes or the changes are already covered by an existing test case.
* [v ] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake).


SooLee:
job.resources no longer has default disk_mb.

johanneskoester:
Thanks! This is fixed slightly differently in PR #1491. Btw. in that PR I noticed that the tibanna test case actually never executed remotely (because the tibanna arg was missing). That is fixed as well now, so that I hope to see less breakages in the future.

SooLee:
I see. Thanks a lot!

Issue #910, opened 2021-03-12
Python 3.5 compatibility - Paths


mbhall88:
**Snakemake version**
6.0.5

**Describe the bug**

Snakemake isn't currently compatible with python version 3.5. Whilst I had thought #898 had solved this problem, it turns out there is more to it. [`open()`](https://docs.python.org/3/library/functions.html#open) only added support for `os.PathLike` objects in python 3.6. So all uses of `pathlib.Path` objects in `open()` calls are causing the version incompatibility. I list a couple of examples, but these aren't all of the cases.

https://github.com/snakemake/snakemake/blob/f55801bfabfb916f77fa4ec8f9eb0667e5c8a92e/snakemake/persistence.py#L113

https://github.com/snakemake/snakemake/blob/f55801bfabfb916f77fa4ec8f9eb0667e5c8a92e/snakemake/unit_tests/__init__.py#L50

Would you like me to try and fix all of these and add python 3.5 to the CI - or do you want to remove python 3.5 support?

johanneskoester:
Closing as suggested by @jdblischak since Snakemake requires Python 3.7 now.

PR #1509, opened 2022-03-22
fix: update for minimum of Python 3.7

jdblischak:
### Description

**tl;dr** As of fd5daaeff070f9987dba411a0f5262c533a2f666, snakemake requires Python 3.7. This PR updates any reference to previous versions

I came to the GitHub repo to determine the minimum required Python version to run Snakemake, and I quickly got confused. The README displays Python 3.5:

Screen Shot 2022-03-22 at 2 48 08 PM

But `setup.py` clearly requires Python 3.7+

https://github.com/snakemake/snakemake/blob/5f7170e095a961ce11ac87467a20879dc4f220f4/setup.py#L14-L16

I was going to simply update the metadata

https://github.com/snakemake/snakemake/blob/5f7170e095a961ce11ac87467a20879dc4f220f4/setup.py#L109

But I saw in commit 0357911cc919a1289004bad9144e75d2880665c7 that the minimum required version is specified in multiple other places. This PR is my attempt to update the code and tests to remove any reference to Python versions < 3.7.

A few other notes:

* In searching old Issues, I found #910. Now that Python 3.5 is no longer supported, I think it can be closed
* I didn't update any of the tutorials in the documentation since technically a user could install older versions of Snakemake and Python

### QC

I don't think this PR requires additional tests or documentation. Please let me know if there is more you'd like me to do.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,7 +106,7 @@ setup(
         "Intended Audience :: Science/Research",
         "License :: OSI Approved :: MIT License",
         "Natural Language :: English",
-        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.7",
         "Topic :: Scientific/Engineering :: Bio-Informatics",
     ],
 )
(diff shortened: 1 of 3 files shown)


johanneskoester (approved):

johanneskoester:
Thanks a lot!

Issue #1460, opened 2022-03-07
Existing intermediate outputs are mishandled when reading from Google Cloud Storage


CowanCS1:
**Snakemake version**
7.1, 7.0.1

**Describe the bug**
When an existing, intermediate output is present in the GS remote location it is not recognized as being present, possibly on purpose due to versioning considerations. When the rule that produces the file is run, it correctly creates the files and uploads them to the remote but then returns a `MissingOutputException`, not recognizes the files are present.

This minimal example is a partial reproduction of the issue. I've seen cases where the pipeline is stuck in an infinite loop or otherwise enters a state where it can be run repeatedly, correctly completing and uploading the outputs each time, without the outputs ever being recognized as present. This minimal example is partial in that it generates the bug on the first run, but then correctly recognizes the intermediates on the second run. I believe the error produced here also underlies the more severe cases.

The error appears to be restricted to intermediate rules which have a list of outputs. In all cases I've seen, the error can be circumvented by manually deleting the outputs 

**Logs**

**Minimal example**
Instructions
- Create the environment and Snakefile
- Create the remote storage location
- Run snakemake with the command below
- The pipeline should run cleanly
- Delete the output files `test.txt`, `blob.txt` and `pretest.txt` from the remote, leaving the file `preblob.txt`
- Run the pipeline to generate the `MisingOutputException`. 
- Go to GCS and note two important things, `preblob.txt` has been overwritten so it is not missing in any sense, and despite the script reporting that `pretest.txt` was removed as an output of a failed job it is actually still present. So two cases in which operations performed on GS files are causing the system to lose track of the files.
- Running the pipeline a third time will complete the pipeline

Snakefile
```
bucket = 'cowan_snakemake'
rule all:
    input:["blob.txt", "test.txt"]
rule intermediate:
    input:["preblob.txt", "pretest.txt"]
    output:["blob.txt", "test.txt"]
    shell:"""
          cp {bucket}/preblob.txt {bucket}/blob.txt
          cp {bucket}/pretest.txt {bucket}/test.txt
          """
rule create:
    output:["preblob.txt", "pretest.txt"]
    shell:'''
          echo "test file" > {bucket}/preblob.txt
          echo "test file" > {bucket}/pretest.txt
          '''
```

Command to execute
```
[...]

PR #1531, opened 2022-03-29
chore: add testcase for issue #1460

johanneskoester:
### QC

### DIFF
--- a/tests/test_google_lifesciences.py
+++ b/tests/test_google_lifesciences.py
@@ -20,7 +20,7 @@ google_credentials = pytest.mark.skipif(
 )
 
 
-def cleanup_google_storage(prefix, bucket_name="snakemake-testing"):
+def cleanup_google_storage(prefix, bucket_name="snakemake-testing", restrict_to=None):
     """Given a storage prefix and a bucket, recursively delete files there
     This is intended to run after testing to ensure that
     the bucket is cleaned up.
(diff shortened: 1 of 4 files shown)


tests/test_google_lifesciences.py
```diff
@@ -34,9 +34,14 @@ def cleanup_google_storage(prefix, bucket_name="snakemake-testing"):
     blobs = bucket.list_blobs(prefix="source")
     for blob in blobs:
         blob.delete()
-    # Using API we get an exception about bucket deletion
-    shell("gsutil -m rm -r gs://{bucket.name}/* || true")
-    bucket.delete()
+    blobs = bucket.list_blobs(prefix=prefix)
+    for blob in blobs:
+        if restrict_to is None or f"{bucket_name}/{blob.name}" in restrict_to:
```
vsoch, line 39: Should this be if `restrict_to` is not None AND the bucket path is in restrict_to? As is now, the default restrict_to being None doesn't make sense - we would always delete files.  Is that intended? And what is this variable saying? If so, I find the naming of the variable confusing.

vsoch: I would rename to something with `keep_` which is more logical to see if keep is None, then we want to remove everything. But then I still find the logic kind of confusing.

vsoch (commented):
okay haven't taken a look at the code to figure out the bug yet but will soon - this is just some feedback for the variable naming of this new test to reproduce the issue!

tests/test_google_lifesciences.py
vsoch: Could we add that as a comment somewhere?

PR #1541, opened 2022-03-30
fix: issue1460 intermediate files existing (or not)

vsoch:
### Description

Just testing different ideas for #1460 - feel free to ignore!

### QC

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,6 +116,15 @@ jobs:
           echo CONTAINER_IMAGE=snakemake/snakemake:$GITHUB_SHA >> $GITHUB_ENV
 
       # TODO reactivate in April (we have no free resources left this month)
+      - name: Test Google Life Sciences Executor
+        if: env.GCP_AVAILABLE
+        run: |
+          # activate conda env
+          export PATH="/usr/share/miniconda/bin:$PATH"
+          source activate snakemake
+          pytest -s -v -x tests/test_google_lifesciences.py
+
+      # TODO reactivate in April (we have no free resources left this month)
       - name: Test Kubernetes execution
         if: env.GCP_AVAILABLE
         env:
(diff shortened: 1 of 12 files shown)


vsoch:
okay doing a little testing - I've gotten down to the io.py function to wait for files, and looking at the files and attributes, according to this, one of them does not exist:

image

indeed it does exist.

image

Indeed it does come down to checking if a remote object exists, e.g., self.blob.exists() erroneously returns False:

```
Type 'copyright', 'credits' or 'license' for more information
IPython 7.23.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: self.blob
Out[1]: <Blob: snakemake-testing-d924es45, test_github_issue1460/preblob.txt, 1648855265000563>

In [2]: self.blob.exists()
Out[2]: False
```
but there it is!

image

self.blob.reload() does not work

```python
self.blob.reload()
...
NotFound: 404 GET https://storage.googleapis.com/storage/v1/b/snakemake-testing-d924es45/o/test_github_issue1460%2Fpreblob.txt?generation=1648855265000563&projection=noAcl&prettyPrint=false: No such object: snakemake-testing-d924es45/test_github_issue1460/preblob.txt
```
But indeed if we "ask again" for the key, we see it exists:

```
In [11]: self.bucket.get_blob(self.key)
Out[11]: <Blob: snakemake-testing-d924es45, test_github_issue1460/preblob.txt, 1648856699412397>

In [12]: blob = self.bucket.get_blob(self.key)

In [13]: blob.exists()
Out[13]: True
```
so that's the change I did for the PR here - looks like Google tests just pass, let's wait to see the other ones too. Going to have dinner, back later!

johanneskoester:
Awesome work @vsoch, thanks a lot!

johanneskoester (approved):

PR #1551, opened 2022-03-31
Add remote plugin system

hwalinga:
### Description

I made the groundworks for the plugin system.

This all works and is tested (manually).

This can be merged. It doesn't break any current functionality. (Submodules are still loaded.)

It all works, there is an example plugin here: hwalinga/snakemake-plugin-remote-tar

The functionality of the example is currently untested.

# How it works:

For plugin loading we utilize the flexibility of the Python import system by
using the importlib and pkgutil modules to find the plugin distributions.
These modules are then loaded and set onto the
`globals()` variable so that everybody importing from `snakemake.remote` can
find them. This functionality has been put into `snakemake/common/plugin.py`
so that other components that need a plugin infrastructure can utilize that as well.

All abstract base class are moved to snakemake/common/remote.py to prevent any
circular dependencies for the plugins and using the ABC system from python
we can ensure that these classes implement all the needed methods. There
is still some better design and refactoring needed for these.

A plugin should be distributed as a pip package starting with the prefix
`snakemake-plugin-remote-`. The name of the installed module will be used
to set on the `snakemake.remote` namespace. If you want to set a name that might
conflict with other globally installed modules you can set `snakemake_submodule_name`
on your module. Your module should expose a bunch of variables such as `__version__`
and `__author__`. The RemoteProvider and RemoteObject it exposes should be derived
from one of the abstract classes now moved to `snakemake/remote/common.py`.

The AUTO loader will also look into the plugins and some of its functionality has
been broken into parts and moved the `plugin.py` package as it also utilizes
some of the `importlib` and `pkgutil` standard library.

# What we did:

There is a snakemake/common/plugin.py with some common functions regarding plugin loading.

We moved the remote abstract classes to snakemake/common/remote.py to prevent circular dependencies.

The snakemake/remote/__init__.py now exposes the common classes, loads the plugins, and sets up AUTO (somewhat refactored).

We created a snakemake/common/future.py for future Python features not yet in 3.7.

We added a raise_ to the snakemake/utils.py for some more elegant check dictionary creation.

# Considerations

However, we believe there need some work to be done before releasing this to the world.
We would at least suggest a bit of a refactor and redesign.

## Coupling RemoteObject and _IOFile

There is an annoying coupling between RemoteObject and _IOFile.

It is not too bad too refactor this, it is mostly annoying because it is confusing when you work with RemoteObject.

As we believe it is now, there is an object  _iofile that holds the string to the location of the remote file (in ._file), however that same string is also the location to the local file. That is very confusing. A RemoteObject should just hold a string to its remote file only (and updated whenever that changes). Because well, the remote file might have a different name than the local file. The signatures for _download/_upload should than be acordingly: `_dowload(to_local_file: _IOFile)` and `_upload(from_local_file: _IOFile)`. Because well, maybe the local file really needs a different name (because the remote file has spaces you don't want, or the remote file has a : that is an illegal character for file on Windows).

## ABC

__metaclass__ = ABCMeta is Python2 syntax, it does nothing in Python3. That said, a lot of classes derived from the abstract classes fail to implement all abstractclasses. Seems some def's can be default raise NotImplementedError, and some def's are just not needed at all.

## Recent change

There is been a very recent change to the RemoteObject interface. (The change even introduces some bugs, (just some wrong imports, but it still needs to be fixed)). Maybe the codebase is just not yet stable for this?

# Refactor

We can follow up with such a refactor, but we thought we give you an update on the work.

# Other remarks

(you can ignore those, just some confusment from me, also just a personal preference for coding style too we guess, but if you want me to refactor this, just say it!):
    * A bit too often the @propery is abused, it is confusing. RemoteObject.list should just be a function. You can also see that upon implementing glob_wildards someone got confused and used a list comprehension. Also all the is_... property on the _IOFile should be a function IMO.
    * Flag system is a bit confusing. We would get rid of it. We think the flags attribute should just be its own class, maybe a dataclasses.dataclass, maybe a types.SimpleNamespace?
    * `def IOFile` is confusing. Functions should be lowercase. Functions as constructors should be classmethods. And the _IOFile is definitely not private despite the underscore prefix (underscore prefixes indicate private).

### QC


johanneskoester:
Thanks a lot!
Regarding your considerations:

### "Coupling RemoteObject and _IOFile"
yes, what you describe is a good idea for solving this issue which has also bugged me. Do you want to give it a try in a separate PR, that would be merged before this one?

### ABC
yes agreed. A fixing PR would be very welcome.

### Recent change
Mhm, which change do you mean, and can you point me to the specific bugs? I can take care of fixing them so that you can concentrate on the other stuff.

### Other remarks
1. Agreed on some of the properties. But that is nothing pressing unless abuse is completely clear (for the is_ cases, it is not imho, for `list` definitely).
2. Yes, one could use a class for flags. But that is not an urgent change. Let us opt for not changing too much at a time. One can always rewrite flags at some point in the future.
3. Yes, IOFile/_IOFile is a bit of a mess and a hack and has lots of historical stuff in it. For the sake of stability, let us keep it like that for now however.

snakemake/common/plugin.py
```diff
@@ -0,0 +1,132 @@
...
+    """Iterator to find all plugins by their prefix of the package name."""
+    plugins = []
+    for finder, name, ispkg in pkgutil.iter_modules():
+        # Check if the *package name* of the module
+        # start with the correct prefix.
+        try:
+            packages = cached_packages_distributions[name]
+        except KeyError:
+            continue
+        if packages[0].startswith(prefix):
```
johanneskoester, line 30: What if there is more than one item in packages? Can that happen? What would that mean?

johanneskoester: Would it make sense to throw an error if there is more than one?

snakemake/common/plugin.py
```diff
@@ -0,0 +1,132 @@
...
+        plugin_module,
+        alternative_mod_name or "snakemake_submodule_name",
+        plugin_name,
+    )
+    globals_dict[snakemake_submodule_name] = plugin_module
+    globals_dict[snakemake_submodule_name.lower()] = plugin_module
+    return True
+
+
+def internal_submodules(paths):
```
johanneskoester, line 111: This function is a bit unclear to me.

snakemake/common/plugin.py
```diff
@@ -0,0 +1,132 @@
...
+        module = importlib.util.module_from_spec(spec)
+
+        try:
+            sys.modules[module_name] = module
+            spec.loader.exec_module(module)
+        except Exception as e:
+            logger.debug(f"Autoloading {module_name} failed: {e}")
+            continue
+
+        submodules(module)
```
johanneskoester, line 130: isn't submodules defined as a list above?

snakemake/common/plugin.py
johanneskoester: What is it supposed to do?

johanneskoester (changes requested):
Thanks a lot. Looks good at first sight. Two additional things: please format with black and check out the linter warnings from sonar cloud.

snakemake/common/plugin.py
hwalinga: yes, it should be submodules.append

that's a bug. fixed no.

snakemake/common/plugin.py
hwalinga: If you install a package, the package name does not have to be equal to the module that the package installs. You can install multiple packages that provide a module with the same name. So packages_distributions lists all packages that provide that module. Which is always one, but in a weird situation, it can be more. But if there are more, the first one is always used. Also by `import` (I tested that). But yes, it would make sense to throw an error. I will add that.

PR #1562, opened 2022-04-04
docs: checkpoint documentation

gregdenay:
### Description

This is an attempt at slightly rewritting the documentation of the `checkpoint` keyword to make its use easier to understand.
Adding one very simple example of creating an arbitrary number of files before deirect aggregation, on which the second existing example is then built.
Requirement to use `directory` vs. filenames as output in this case is made clearer.

Feedback welcome!

### QC


docs/snakefiles/rules.rst
```diff
@@ -1865,8 +1867,61 @@ Instead, the output file will be opened, and depending on its contents either ``
...
+  checkpoints.clustering.get(sample=wildcards.sample).output[0]
+
+we use the shorthand
+
+.. code-block:: python
+
+  checkpoints.clustering.get(**wildcards).output[0]
+
+which automatically unpacks the wildcards as keyword arguments (this is standard python argument unpacking).
+If the checkpoint has not yet been executed, accessing ``checkpoints.clustering.get(**wildcards)`` ensure that Snakemake records the checkpoint as a direct dependency of the rule ``aggregate``.
```
johanneskoester, line 1918:
```suggestion
If the checkpoint has not yet been executed, accessing ``checkpoints.clustering.get(**wildcards)`` ensures that Snakemake records the checkpoint as a direct dependency of the rule ``aggregate``.
```

johanneskoester (approved):

Issue #1615, opened 2022-04-28
Set-threads parameter does not work in combination with kubernetes


LarsStegemanGT:
**Snakemake version**
7.5.0
**Describe the bug**
When using the parameter --set-threads in combination with the --kubernetes parameter, the command given to the pod will not execute and give an error. 
The parsing of the CLI argument --set-threads is done correctly on the initial snakemake execution. However when giving this argument to the kubernetes cluster to be executed on a node, it will be in a different format and cause an error.

**Logs**
The logs in the pod:
``Invalid threads definition: entries have to be defined as RULE=THREADS pairs (with THREADS being a positive integer). Unparseable value: '{a:'.
``

**Minimal example**
_Snakefile_
```
rule a: 
   run: 
      print("run a")
```

Running this with the command:
`snakemake --kubernetes -j 1`

will work normally and give correct results.
However the following will give an error on the created pod (not on the current machine but on cluster)
`snakemake --kubernetes -j 1 --set-threads 'a=0'`

Using _kubectl describe_ you can see that the following argument is given to the pod:
`` Args:
      -c
      cp -rf /source/. . && python -m snakemake --snakefile 'Snakefile' 'a' --allowed-rules 'a' --cores 'all' --attempt 1 --force-use-threads   --force --keep-target-files --keep-remote --max-inventory-time 0 --nocolor --notemp --no-hooks --nolock --ignore-incomplete --skip-script-cleanup  --conda-frontend 'mamba' --wrapper-prefix 'https://github.com/snakemake/snakemake-wrappers/raw/' --set-threads {'a': 0} --latency-wait 5 --scheduler 'greedy' --default-remote-prefix 'test' --default-remote-provider 'AzBlob' --default-resources 'mem_mb=max(2*input.size_mb, 1000)' 'disk_mb=max(2*input.size_mb, 1000)' 'tmpdir=system_tmpdir' --mode 2
``

_--set-threads 'a=0'_ is now given as _--set-threads {'a': 0}_

**Additional context**
I have a fix for this issue. I will create a PR after.

PR #1617, opened 2022-04-28
fix: issue 1615 - Switch formatting condition for dictionary

LarsStegemanGT:
### Description
Fixes issue #1615 

A dictionary will return true when using _not_iterable_ from snakemake.io. The elif condition is never reached/true in the original code. Swapping them will cause dictionaries to be formatted as expected by snakemake.
### QC

I'm not sure how to test this, or if i break any other functionality by changing this.


johanneskoester (approved):

johanneskoester:
Thank you so much, good catch!

PR #1625, opened 2022-05-02
test: test to evaluate pipelines that using modules that them self use modules.

Smeds:
Tests used to import a module that contain a module
### QC


Hocnonsense:
This should be fixed by #3714

Issue #1602, opened 2022-04-26
conditional temp output


sneuensc:
**Is your feature request related to a problem? Please describe.**
It would be great to have a conditional temp() for the output. I am developing a pipeline which allows to skip modules. The skipping is done using symlinks (ln -s). This works great if all output is kept. However, my files are too large, so I need to clean up all intermediate files, but the output of the last step which is referred to with a symlink in the final rule (ln -s foo.bam final.bam).  The function temp() is great, however, does not allow to make it conditional, i.e. just keep the output of the last step.

**Describe the solution you'd like**
Ideal would be to have (simple) functions allowed in the output, e.g.:
```
rule foo:
  output:
    if FINAL_STEP == "foo":
      "foo.bam"
    else:
      temp("foo.bam")
```

**Describe alternatives you've considered**
My current implementation uses an if else around the entire rule definition (doubling the rule) which is ugly, difficult to read and thus dangerous, but luckily works, e.g.:
```
if FINAL_STEP == "foo":
  rule foo:
    output:
      "foo.bam"
    ...
else:
  rule foo:
    output:
      temp("foo.bam")
    ...
```

gshiba:
I too would love to have a feature like this.

Idea: Can the `temp` function accept a callable that returns a boolean, with the signature `lambda wildcards:...` (similar to input functions)?

```
FINAL_STEP = ... 

rule foo:
    output: temp("{name}.{ext}", lambda w: FINAL_STEP != "foo")
```

Other examples:

```
rule x:
    # Delete if prefix == "foo"
    output: temp("{prefix}.{ext}", lambda w: w.prefix == "foo")

rule y:
    # Delete if my flag is passed in from CLI like: snakemake [...] -C my_temp_flag=2
    output: temp("file", lambda w: int(config.get("my_temp_flag", 0)) > 2)
```

sneuensc:
Closing this in favour of https://github.com/snakemake/snakemake/pull/1969 which can achieve the same goals with more generalisation and less argument clutter.

PR #1650, opened 2022-05-11
feat: Allow "temp" files to be conditionally removed by user-set parameter (extending --nt)

ASLeonard:
### Description
This PR changes the handling of temp files and the `--nt` parameter. It isn't directly related to #1602, but I believe it may be able to solve it.

The motivation is currently snakemake can delete all temporary files or none. As stated in the `--nt` parameter, if you are only running part of a pipeline but want to keep temp files that would be used later, you have to switch off all temporary file removal. So you end up in two situations
  - keep unwanted `temp` flagged files
  - comment out the `temp` flag on files you do want to keep for now

With this implementation, the default behaviour can be maintained
  - all `temp` flagged files are deleted
  - using `--nt` keeps all temp files
while also extending to allow
  - deleting certain `temp` flagged files based on a value set at runtime/config
  
This would work in the snakefile as 
```
rule test1:
    output:
        temp("foo.txt",threshold=5)

rule test2:
    output:
        temp("bar.txt",threshold=2)
```

and running with 
```
snakemake --nt
```
would delete all temporary files while
```
snakemake --nt 4
```
would only delete bar.txt (threshold 2<4) and keep foo.txt (threshold 5>4).

This easily allows users to configure which temp files should be deleted in partial runs, debug runs, release runs etc. There are some files that are 100% temporary and I would never like to keep while some are worth keeping while I flesh out the pipeline, which can be resolved here.

### DRAFT

The main issues here are
  - hijacking the `--nt` parameter, so the name may be misleading now. I was avoiding introducing another parameter, but may be the cleanest solution.
  - determining the logic to be backward compatible. In this current implementation, `notemp` should be set to inf, so may just need to make thresholds implicitly negative or something else to make sure `--nt` behaves as expected.

### QC


ASLeonard:
Actually it may be more in keeping with the meaning of `--notemp` to reverse the logic, and *not delete* files which have a threshold set below the user provided value. 

So
```
rule test1:
    output:
        temp("foo.txt",threshold=1)

rule test2:
    output:
        temp("bar.txt",threshold=2).
```
And then
```
snakemake --nt
```
would keep all
```
snakemake --nt 1
```
would keep foo.txt, and 
```
snakemake --nt 2
```
would keep both foo.txt and bar.txt

ASLeonard:
Can extend this to also include `--delete-temp-output` to have a user set level. The main issue is now `--notemp` (and maybe later `--delete-temp-output`) want to take an argument, so previous snakemake calls like 
```python
snakemake --quiet --notemp rule_of_choice
```
try to convert "rule_of_choice" to a float for notemp, so would need to be changed into
```python
snakemake --quiet --notemp -- rule_of_choice
```
Not a difficult change, but was hoping to avoid too many syntax changes. Might be better to just add a new parameter called "--delete-temp-level" or something...

ASLeonard:
May be overcomplicating, but it was also possible to extend `--nt` as a parameter accepting 0 or more arguments. The logic would then be as follows
  - empty case: `--nt -- <other args>`  would implicitly set `notemp` to negative infinity and thus ignore all temp directives
  - single value case: `--nt 5` would ignore temp directives for anything with threshold equal to or above the given value
  - multiple value case: `--nt 3 5 10` would ignore temp directives for with the specific threshold values (would retain if threshold is 3, 5, or 10, would delete if threshold is 2, 4, 6, 7, 11, etc)

Does get complicated, but allows users as much granularity as they are willing to set.

cbp44:
I think this is a very useful feature to add. I often find myself commenting out `temp()` declarations during workflow development, and have forgotten to comment them back in for production.

I thought I would add my input on your current implementation and reply to some of your comments.

1. I vote to not hijack the `--nt` parameter for new use.

   This would prevent backward compatibility of workflows. Anyone who uses `--nt` in their scripts with Snakemake would need to update them in order to upgrade Snakemake to a new version. Since most people don't read release notes when updating software, they would likely be unaware of the change to `--nt`. 

    For this reason, it is better to add a separate `--nt-threshold` parameter.
    
2. Same thing for `--delete-temp-output`. I know this is not implemented in the current PR, but if added, I think it would be better to just add a `--delete-temp-output-threshold` parameter.

3. In response to [your comment about overcomplicating it](https://github.com/snakemake/snakemake/pull/1650#issuecomment-1144556956), I think that _is_ overly complicated. Just the threshold is sufficient for users to have the flexibility that this feature is meant to provide, namely, deleting only certain temporary files when running a workflow.

### Possible Concerns

- **Scalability** - This code adds logic comparing `<=` between the threshold file flag of temporary files and a user speciefied threshold value `--nt`. It would run this logic for every single temporary file in a workflow to see if it meets the threshold for deletion. If a workflow has millions of temporary files, this might be noticable speed-wise.

- **Infinite Loops** - You defined and stored the `--nt` threshold as a float. Comparing floats can be imprecise sometimes and even lead to infinite loops ([see here](https://stackoverflow.com/questions/61284254/infinite-while-loop-when-checking-floats-for-equality)).

  It is sufficient for this feature to use an integer for the threshold parameter, so it would be better to just define it like that. This avoids that problem.

sneuensc:
Looking forward for this feature. It will extend the usage of snakemake. Thank you for implementing it!

gshiba:
Thinking out loud here... can the `temp` function accept:

- a regex, which, if it `re.match`es the pattern, will/not keep the file?
- a callable that returns a boolean, with the signature `lambda wildcards:...` (similar to input functions)?

```
rule a:
    input: ...
    output: temp("{name}.{ext}, lambda w: w.ext == "bam")  # mark as temp if ext == "bam"
```

ASLeonard:
Closing this in favour of #1969 which can achieve the same goals with more generalisation and less argument clutter.

PR #1651, opened 2022-05-11
feat: add flag `ensure` that allows to annotate that certain output files should be non-empty or agree with a given checksum

johanneskoester:
### QC


johanneskoester:
cc @alethomas

PR #1663, opened 2022-05-17
feat: automatically rerun jobs if parameters, code, input file set, or software stack changed (thanks to @cclienty and @timtroendle). This also increases performance of DAG building by handling job "needrun" updates level wise, while avoiding to perform a full check for those jobs that are already downstream of a job that has been determined to require a rerun.

johanneskoester:
### QC


hermidalc:
This feature does not completely work with checkpoints, there is a major issue or regression #1818

hermidalc:
I've tested the #1818 simple example with 7.15.1 (latest avail on bioconda) and still it re-runs all the rules downstream of a `checkpoint` that just ran and created all the files.

Issue #1267, opened 2021-11-19
Benchmarking + container returns wrong values.


b1999:
**Snakemake version**
6.10.0

**Describe the bug**
Benchmark returns wrong values whenever a tool is executed in a docker container. As seen in the example below, it took
Freebayes 1.5 hour to finish with only 0.01 CPU time. 

**Executing snakemake**
~/.local/bin/snakemake --cores 32 --use-singularity --latency-wait 5 --snakefile Snakefile.smk 

**Benchmark file created when running Freebayes**
s                 h:m:s     max_rss   max_vms   max_uss   max_pss   io_in    io_out  mean_load   cpu_time
5674.7210  1:34:34  33.81       578.17       34.08         34.09        0.00     0.00     0.00             0.01

**Rule example**
rule run_freebayes:
	input: 
		genome_fa = "resources/homo_sapiens_genome.fa",
		alignment = "results/{sample}/{aligner}/output/output_aligner_sorted.bam"
	output: "results/{sample}/freebayes/output/{aligner}_variants.vcf"
	message: "Calling variants using Freebayes on alignment file generated by {wildcards.aligner}."
	benchmark: "results/{sample}/benchmark/freebayes/output/benchmark_freebayes_{aligner}.txt"
	log: "results/{sample}/log/freebayes/output/freebayes_{aligner}.log"
	threads: 32
	container: "docker://davidyuyuan/freebayes"
	shell: "freebayes -f {input.genome_fa} -b {input.alignment} -v {output} 2> {log}"

Smeds:
I have seen similar problems when using singularity and believe I have pinpointed the problem. The problem ariase for me when the benchmark script tries to read memory information using memory_full_info from a process named starter-suid which my run doesn't have permission to read.

https://github.com/snakemake/snakemake/blob/26e57d69fc320adc972967a8046c5163b455456c/snakemake/benchmark.py#L230-L275

Reading the process result in a AccessDenied event which will be caught by  "except psutils.Error as e" line (274), resulting in a return which will then leads to not reading information about the other processes.

I can solve it by either running my snakemake jobs as root, not a good solution.

or wrapping to "meminfo = proc.memory_full_info()" with a try statement that gives a continue event if a error is caught.

PR #1671, opened 2022-05-19
fix: handle benmark bug that arise with singularity

Smeds:
### Description

Running snakemake with singularity results in (nearly) empty benchmark statistics. benchmark tries to read a process with name "starter-suid" which raise a AccessDenied event, resulting in that the other processes belonging to the snakemake process aren't read. Wrapping the proc.memory_full_info() in a try-statement that fires a continue event enables the other processes to be parsed.

This should help solve issue: #1267 

### QC


johanneskoester (approved):
Thanks!

snakemake/benchmark.py
```diff
@@ -237,7 +237,10 @@ def _update_record(self):
                             this_time - self.bench_record.prev_time
                         )
 
-                    meminfo = proc.memory_full_info()
+                    try:
+                        meminfo = proc.memory_full_info()
+                    except psutil.Error:
```
johanneskoester, line 242: May I ask you to add your explanation from the PR text as a rationale here? So that we know in the future why we catch and continue here.

johanneskoester (changes requested):

snakemake/benchmark.py
Smeds: done

johanneskoester: Also, any idea about the reason for this starter-suid thing showing up? proc should not point to a surrounding process... Why does that happen?

johanneskoester: Yes, setting to NA seems like the most transparent way to me, and also display a warning via `logger.warning()`.

johanneskoester (changes requested):

Smeds:
@johanneskoester  I have updated the commit with more handling for events when no benchmark can be collected. When debbuing is turned on and if any process cannot be processed information about which process that can and cannot be processed will be printed.

snakemake/benchmark.py
Smeds: New commit that should handle this have been pushed.

johanneskoester (approved):

Issue #1674, opened 2022-05-22
Snakemake conda bug- command: conda env export --name


eeko-kon:
**Snakemake version**
6.15.5

**Conda version** 
conda 4.12.0

**Describe the bug**
When I specify the conda env name (`.snakemake/conda/envname`) in a rule, the env is activated but then the job crashes when conda tries to export env name:

```
raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'conda env export --name '.snakemake/conda/envname'' returned non-zero exit status 1.
```

eeko-kon:
I have first created the environment:

`conda env create --prefix ".snakemake/conda/envname" -f "workflow/envs/requirements.yaml" `

Example rule:

```
rule example:
    input:
        "results/Interim/input.tsv" 
    output:
        "results/Interim/output.tsv"
    conda:
        ".snakemake/conda/envname"
    shell:
    """
    mv {input} {output}
    """
```

eeko-kon:
If I don't create the env using a prefix it should work, but I want it in my .snakemake/conda path. I guess this is more of a conda limitation regarding default paths, but if there is smth you can add in the snakemake code to avoid the error, that would be great.

54mu:
Facing the same issue, manually applying the changes of #1708 fixes it until it gets merged

Jerry-is-a-mouse:
@eeko-kon  Hello, how do you deal with this issue finally? I also faced with this problem

eeko-kon:
Hi @Jerry-is-a-mouse this should be fixed with the new snakemake release. Try updating and re-running! Otherwise #1708 has a workaround!

mfansler:
I can confirm that upgrading from 8.15 to 8.27 resolved this issue.

PR #1708, opened 2022-06-09
fix: processing in conda yaml and url bug

ShogoAkiyama:
### Description
If conda is url path, `self.is_named` will be true, so this process will result in an error like the following issue.
So I modified the process.

issue: #1674
```
conda env export --name 'URL'
```

fixed like below
```
conda env export --prefix 'URL'
```


ShogoAkiyama:
issue #1998

migrau:
This works https://github.com/snakemake/snakemake/pull/1708/files
Thanks, @ShogoAkiyama

AroneyS:
@johanneskoester Is there something blocking this pull-request?

Unaimend:
Also encountered this problem today

johanneskoester:
I believe this is already fixed in a different way in the current stable snakemake.

PR #1711, opened 2022-06-10
fix: conda python interpreter path on Windows

yeyuan98:
### Description

`PythonScript._is_python_env` was not working properly as of v7.8.2 on Windows thereby preventing python scripts from executing with python interpreters from rule-based conda environments.

The bug should be resulting from inconsistencies when conda creates Python environments on Windows versus Unix-like systems. On Windows, the python interpreter will not exist as `<conda_env_root>/bin/python` but as `<conda_env_root>/python.exe`. Please refer to Anaconda documentation [here](https://docs.anaconda.com/anaconda/user-guide/tasks/integration/python-path/).

Added a new test case *test_conda_python_script* in tests folder and added the test to *tests/tests.py*. However, it seems that `nose` is unmaintained and I could not run the nosetests in my local computer.

### QC


snakemake/script.py
```diff
@@ -1,1369 +1,1374 @@
-__author__ = "Johannes Köster"
```
johanneskoester, line 1: Mhm, this gigantic change in this file is fishy... What is going on here?

johanneskoester: Maybe windows linebreaks?

johanneskoester (changes requested):

snakemake/script.py
yeyuan98: Thanks a lot for your explanation.. Sorry, new git user here. I think the problem is I pushed from a Windows machine and did not set hard line break settings. Git forces CRLF and replace LF on Windows by default.

Fixed the line break problem. Hopefully it will work!

johanneskoester (approved):

yeyuan98:
The previous commit got an issue with the `black` formatting guideline. Reformatted `tests.py` with `black`.

PR #1717, opened 2022-06-14
feat: make it possible to exclude rules that will be imported when using 'use rule' statement

Smeds:
### Description

When combining multiple modules to create a pipeline we often get problems with ambiguous rules exception, we have generic rules for sorting, index, etc. Which can be solved with ruleorder, but for more complex pipelines this can be rather troublesome.
And to specifiy all rules, instead of using *, when using multiple rules from a module can result in lines that are hard to read and maintan. For the cases where one would like to exclude one or two rules and use the remaining rules it would be rather neat to have an exclude option.

This pull-request give the option of excluding rules, ex:
```
use rule * from test exclude b, d
```

### QC


johanneskoester:
May I ask you to format with black?

Smeds:
> Awesome work and a great usability improvement!

Thanks!

Smeds:
> May I ask you to format with black?

Of course!

Smeds:
> > May I ask you to format with black?
> 
> Of course!

@johanneskoester  done

PR #1725, opened 2022-06-18
chore: add pyproject.toml to use setuptools features

mwtoews:
### Description

This PR introduces a `pyproject.toml` file described by [PEP 518](https://peps.python.org/pep-0518/), and uses PyPA's [build](https://github.com/pypa/build) to check building the package (sdist and wheel files).

In addition, lots from `setup.py` is removed or simplified:

- E.g. first few lines are not needed for Python 3, which is UTF-8 by default and has a print function
- Dunders `__author__` etc are not normally used in this file
- Remove `setuptools` check, since this is a build requirement specified in the correct place (`pyproject.toml`)
- Remove Python version check, since this is done with setuptools' `python_requires`
- Note that lots of the static metadata from `setup.py` can be moved to either `setup.cfg` or `pyproject.toml`, but not done in this PR

### QC

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -31,7 +31,8 @@ jobs:
         if: ${{ steps.release.outputs.release_created }}
         run: |
           python -m pip install --upgrade pip
-          python setup.py sdist
+          pip install build
+          python -m build
 
       - name: Publish to PyPI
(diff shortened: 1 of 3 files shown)


johanneskoester (approved):

johanneskoester:
Thank you!

Issue #1342, opened 2022-01-20
JSONDecodeError while handling job success on LSF


omarelgarwany:
**Snakemake version**
6.12.3

**Describe the bug**
This problem happens when running a snakemake pipeline on LSF. The pipeline exits immediately and it is not reported as a job failure. 

**Logs**
```
Traceback (most recent call last):
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/__init__.py", line 783, in snakemake
    keepmetadata=keep_metadata,
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/workflow.py", line 1073, in execute
    success = self.scheduler.schedule()
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/scheduler.py", line 440, in schedule
    self._finish_jobs()
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/scheduler.py", line 529, in _finish_jobs
    self.get_executor(job).handle_job_success(job)
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/executors/__init__.py", line 871, in handle_job_success
    job, upload_remote=False, handle_log=False, handle_touch=False
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/executors/__init__.py", line 272, in handle_job_success
    keep_metadata=self.keepmetadata,
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/jobs.py", line 1033, in postprocess
    self, keep_metadata=keep_metadata
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/persistence.py", line 239, in finished
    starttime = self._read_record(self._metadata_path, f).get(
  File "/home/anaconda3/envs/gtex/lib/python3.7/site-packages/snakemake/persistence.py", line 414, in _read_record_uncached
    return json.load(f)
  File "/home/anaconda3/envs/gtex/lib/python3.7/json/__init__.py", line 296, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/home/anaconda3/envs/gtex/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/home/anaconda3/envs/gtex/lib/python3.7/json/decoder.py", line 337, in decode
[...]

PR #1730, opened 2022-06-20
fix: Adding file system based locks for meta data (#1342)

holtgrewe:
### Description

The locks are based on the fasteners package and will be enabled
automatically in cluster mode. API users can explicitely enable and
disable them and CLI users can disable them explicitely. This prevents
multiple processes from writing to the same metadata file and thus
damaging the file as described by multiple cluster execution users in
issue #1342.

**Note** and question to @johanneskoester: It might have been possible to use a single lock file which reduces the number of small files in the metadata directory. I don't have a sufficient insight in the `Persistence` code so this would be up to you. We would only require a small change to use a static path for the lock but the path to the file inside of the `.snakemake` directory must be well chosen.

### QC

### DIFF
--- a/docs/tutorial/additional_features.rst
+++ b/docs/tutorial/additional_features.rst
@@ -292,6 +292,15 @@ The line will subsequently be available in the calls to ``--cluster``, ``--clust
 In the case of a REST server, you can use this to return the port that the server is listening on and credentials.
 When the Snakemake process terminates, the sidecar process will be terminated as well.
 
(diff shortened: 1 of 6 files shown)


holtgrewe:
Github actions are not working at the moment (Azure?), things to fail to download.

johanneskoester:
See discussion in #1342. I think I would favor #1745.

holtgrewe:
I agree, #1745 is better.

PR #1749, opened 2022-06-29
fix: ensures that REncoder also checks for numpy.bool_ in encode_value

### LINKED ISSUE #1045: Using Boolean values with Paramspace Expansion fails with R scripts
**Snakemake version**
6.4.1

**Describe the bug**
When using Paramspace functions where True or False is used, scripts using R code will fail if when attempting to set a parameter to TRUE with "Unsupported value for conversion into R. True".

**Logs**
If applicable, any terminal output to help explain your problem.

```

rule simulate:
    output: results/simulations/alpha~False/beta~0.0/gamma~3.9.tsv
    jobid: 4
    wildcards: alpha=False, beta=0.0, gamma=3.9

[Thu Jun 10 11:28:29 2021]
Finished job 4.
1 of 5 steps (20%) done
Select jobs to execute...

[Thu Jun 10 11:28:29 2021]
rule simulate:
    output: results/simulations/alpha~True/beta~0.1/gamma~0.99.tsv
    jobid: 2
    wildcards: alpha=True, beta=0.1, gamma=0.99

[Thu Jun 10 11:28:29 2021]
Error in rule simulate:
    jobid: 2
    output: results/simulations/alpha~True/beta~0.1/gamma~0.99.tsv

RuleException:
ValueError in line 26 of /Users//Documents/snakemake_bug/Snakefile:
Unsupported value for conversion into R: True
  File "/usr/local/Caskroom/miniconda/base/envs/snakemake/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 2357, in run_wrapper
  File "/Users//Documents/snakemake_bug/Snakefile", line 26, in __rule_simulate
  File "/usr/local/Caskroom/miniconda/base/envs/snakemake/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 574, in _callback
  File "/usr/local/Caskroom/miniconda/base/envs/snakemake/lib/python3.9/concurrent/futures/thread.py", line 52, in run
  File "/usr/local/Caskroom/miniconda/base/envs/snakemake/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 560, in cached_or_run
  File "/usr/local/Caskroom/miniconda/base/envs/snakemake/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 2390, in run_wrapper
Complete log: /Users//Documents/snakemake_bug/.snakemake/log/2021-06-10T112827.881561.snakemake.log
```

**Minimal example**
Use Parameter space expansion with True values and then attempt to use it with an R script. 

```python
rule simulate:
    output:
        # format a wildcard pattern like "alpha~{alpha}/beta~{beta}/gamma~{gamma}"
        # into a file path, with alpha, beta, gamma being the columns of the data frame
        f"results/simulations/{paramspace.wildcard_pattern}.tsv"
    params:
        # automatically translate the wildcard values into an instance of the param space
        # in the form of a dict (here: {"alpha": ..., "beta": ..., "gamma": ...})
        simulation=paramspace.instance
    script:
        "simulate.R"

```

**Additional context**
This bug didn't occur previously but i don't remember which snakemake versions i was using before.

Christian-Heyer:
### Description

Fixes #1045 which was caused when `np.bool_` type was used for boolean values and was slipping through the instance checking and raising an error. 

### QC

* [x ] The PR contains a test case for the changes or the changes are already covered by an existing test case.

### DIFF
--- a/snakemake/script.py
+++ b/snakemake/script.py
@@ -205,7 +205,7 @@ class REncoder:
                     return str(value)
                 elif isinstance(value, np.bool_):
                     return "true" if value else "false"
- 
+
             except ImportError:
                 pass
         raise ValueError("Unsupported value for conversion into R: {}".format(value))
--- a/snakemake/script.py
+++ b/snakemake/script.py
@@ -180,6 +180,8 @@ class REncoder:
 
     @classmethod
     def encode_value(cls, value):
+        import numpy as np
+
         if value is None:
             return "NULL"
         elif isinstance(value, str):
@@ -188,7 +190,7 @@ class REncoder:
             return repr(str(value))
         elif isinstance(value, dict):
             return cls.encode_dict(value)
-        elif isinstance(value, bool):
+        elif isinstance(value, bool) or isinstance(value, np.bool_):
             return "TRUE" if value else "FALSE"
         elif isinstance(value, int) or isinstance(value, float):
             return str(value)
@@ -198,7 +200,6 @@ class REncoder:
         else:
             # Try to convert from numpy if numpy is present
             try:
-                import numpy as np
 
                 if isinstance(value, np.number):
                     return str(value)
--- a/snakemake/script.py
+++ b/snakemake/script.py
@@ -264,6 +264,8 @@ class JuliaEncoder:
 
                 if isinstance(value, np.number):
                     return str(value)
+                elif isinstance(value, np.bool_):
+                    return "true" if value else "false"
             except ImportError:
                 pass
         raise ValueError(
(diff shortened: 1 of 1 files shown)


snakemake/script.py
```diff
@@ -202,6 +203,9 @@ def encode_value(cls, value):
 
                 if isinstance(value, np.number):
                     return str(value)
+                elif isinstance(value, np.bool_):
+                    return "true" if value else "false"
```
johanneskoester, line 207:
```suggestion
                    return "TRUE" if value else "FALSE"
```

johanneskoester (changes requested):

johanneskoester (approved):

PR #1761, opened 2022-07-08
feat: Add token auth to GitLab/GitHub hosting providers

### LINKED ISSUE #1301: Use modules from private github repository
I would like to use modules from private github repositories. I very much like to option to specify 'imported' modules using a commit tag, as it implements a kinda source version specification of the external modules without having to include the source code itself. But, my repos' are private. Currently, files from github are source from raw URLs which does not allow authentication of any kind.

Ability to 'import' modules from private github repos.  
I don't know in detail what a good solution would be. One possibility is to make use of [github's API](https://docs.github.com/en/rest/reference/repos), possibly through [pyGithub](https://pygithub.readthedocs.io/en/latest/introduction.html)

I'm thinking of having local git mirrors of private github repos and sourcing from this but that would require managing local copies.

**Additional context**  
https://stackoverflow.com/q/70275063/4768122

tdido:
### Description

This PR adds ~~an optional "token" argument~~ the ability to get a token from environment variables to the `gitlab()` and `github()` code hosting provider. This allows for the usage of modules stored in private repositories.

### QC

Closes #1301


docs/snakefiles/modularization.rst
```diff
@@ -301,6 +301,12 @@ For the latter, it is also possible to specify an alternative host, e.g.
 
     gitlab("owner/repo", path="workflow/Snakefile", tag="v1.0.0", host="somecustomgitlab.org")
 
+A private token can be specified to access private repositories:
+
+.. code-block:: python
+
+    github("owner/repo", path="workflow/Snakefile", tag="v1.0.0", token="xxx_ccluNyxdHXXXXX9H6nvC873n3lLG")
+    gitlab("owner/repo", path="workflow/Snakefile", tag="v1.0.0", host="somecustomgitlab.org", token="xxxxx-1fhwBFYCDh-QbB")
```
johanneskoester, line 309:
```suggestion
    gitlab("owner/repo", path="workflow/Snakefile", tag="v1.0.0", host="somecustomgitlab.org", token="xxxxx-1fhwBFYCDh-QbB")
    
Naturally, in reality you should query these tokens from environment variables (using ``os.environ``) instead of hardcoding them into the source file.
```

johanneskoester (changes requested):

tdido:
@johanneskoester seems like conda env files are still fetched without the token. I've tested hardcoding it in the conda deployment file [1] and it works. Any pointers on the best way to make the token available there?

[1] https://github.com/tdido/snakemake/blob/435913d63cc42ac7b9e0cf438673e339aa25d055/snakemake/deployment/conda.py#L843

johanneskoester:
A cli arg for the token is not an option IMHO, because then the token will be inside of the bash history, which is a security risk. Can you explain what exactly fails when just fetching the token from os.environ?

tdido:
Right, yes, that makes sense. The failure I was referring to was the one when adding the argument to the gitlab() function, where it doesn't pass the token to the conda environments it fetches.

I'll change the cli arg to os.environ instead, should be straightforward.

tdido:
@johanneskoester I've switched to using an env variable for the token and it seems to be working fine. Please let me know whether the implementation looks OK.

tdido (commented):
Some requested changes not needed anymore with env vars approach, not sure how to mark as solved...

snakemake/workflow.py
```diff
@@ -1197,6 +1198,9 @@ def include(
         basedir = self.current_basedir if self.included_stack else None
         snakefile = infer_source_file(snakefile, basedir)
 
+        if "HOSTING_PROVIDER_TOKEN" in os.environ and isinstance(snakefile, HostingProviderFile):
+            snakefile.set_token(os.environ["HOSTING_PROVIDER_TOKEN"])
```
johanneskoester, line 1202: Instead of using set_token here, I'd rather always check for the token in the constructor of GithubFile and GitlabFile. This way, one does not even need to modify the usage of them in the codebase. Also, I would rather use service specific tokens. Usually, people use the env vars `GITHUB_TOKEN` and `GITLAB_TOKEN`.

snakemake/deployment/conda.py
```diff
@@ -874,9 +874,20 @@ def check(self):
...
+            if "HOSTING_PROVIDER_TOKEN" in os.environ:
+                token=os.environ["HOSTING_PROVIDER_TOKEN"]
+                if "github" in filepath:
+                    return filepath.replace("raw",f":{token}@raw")
+                elif "gitlab" in filepath:
+                    return filepath + f"&private_token={token}"
+                else:
+                    raise WorkflowError(
+                        f"Access to private repositories is only available for github.com and gitlab.org"
+                    )
```
johanneskoester, line 887: This "code duplication" I don't really understand. If the conda env path is coming from a github or gitlab resource, it should already contain the token from the implementation in GithubFile and GitlabFile.

johanneskoester (changes requested):

snakemake/workflow.py
tdido: Oh yeah, I didn't realise that this wasn't needed anymore if not fetching an argument from the workflow. That's much better.

snakemake/deployment/conda.py
tdido: That's a leftover from back when I was struggling to figure this out. Removed.

johanneskoester (approved):

johanneskoester:
Can you please run black on your changes?

johanneskoester (changes requested):
Almost forgot: could you also add a sentence about the use of the two env vars to the respective section in the docs? ([this one](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html#code-hosting-providers))

Issue #1474, opened 2022-03-11
A `node_temp()` modifier for temporary files on node-local storage?


DrYak:
**Is your feature request related to a problem? Please describe.**

According to [official Snakemake documentations](https://snakemake.readthedocs.io/en/stable/project_info/faq.html#how-can-i-make-use-of-node-local-storage-when-running-cluster-jobs):
> There is currently no way of doing this in Snakemake, but a possible workaround involves the shadow directive and setting the --shadow-prefix flag to e.g. /scratch.

Shadow is indeed one possible solution, but...

**Describe the solution you'd like**

I would like Snakemake to introduce some kind of `node_temp()` modifier.

If you squint at it, a lot of the necessary work already exists in Snakemake:

 - `pipe()` modifier already specifies special files that:
   - make mandatory to run both the producing rule and the consuming rule within the same group on the same node.
      - the difference is that `pipe()` requires both rule to be run in parallel (and considers sum of required resources).
      - whereas a putative `node_temp()` modifier would simply require both rules to run one after the other (and considers the max of required resources), just like any other `group: ` rules' property.
   - expects the file to only be alive for the duration of the group and not outlive it/not exist outside of it.
 - `temp()`modifier already specifies files that:
   - will not be kept long term
   - will be cleaned up after all the rules finish.
   - the difference being (due to the scratch directory being node-local):
     - the main Snakemake process that dispatches to the cluster should not wait for the files to appear (if files aren't visible by the end of `--latency-wait`, do not consider a failure).
     - the responsibility of cleaning this up shouldn't be on the shoulders of the main Snakemake (the one that submits jobs to the cluster), but on those of the job's Snakemake (the instance that is started at the main point of the cluster job and which runs all the rules that are members of the group) as the later one has visibility to the node-local storage.
 - the resource `tmpdir` already takes care of passing/setting up a temp directory, so it should be possible to directly use it as a prefix for `node_temp()` (instead of requiring the use to add one manually like [in the case of shared scratch](https://github.com/cbg-ethz/V-pipe/blob/67e3e7d214840748d67175ab67f061edce1fe980/workflow/rules/common.smk#L621))

**Describe alternatives you've considered**

 - `shadow:` _does_ cover some possible use cases (uses within a single rule), but not all cases (e.g.: passing temporary files between rules).
 - `temp()` will not work, because node-local storage only exists during the execution of group, not outside.
[...]

PR #1765, opened 2022-07-11
fix: Improved handling of missing output files in group job postprocessing, accounting for temporary files.

pvandyken:
When I run group jobs on a cluster, I'll often write temporary files to the node-specific local scratch folder. I'll label them `temp(...)` and the sub-snakemake instance will dutifully delete the files when no longer needed. Even without that deletion, the files would disappear with the node local scratch as soon as the group job finishes.

The problem occurs when the parent snakemake instance re-assumes control. It checks for all the outputs from the group job, finds the temp files missing, and exits with an error. But these temp files are no longer needed anywhere in the workflow, and indeed, Snakemake's immediate next step if it were to find the temp files would be to delete them.

This PR filters the group job outputs before the existence checks, removing any that are:
- Not needed anywhere else in the workflow
- Not target files

In particular, only files that would otherwise be immediately removed are excluded from the existence check (at least, that's the idea).

The idea is to remove a small frustration without causing any breaking or generally noticeable changes. I've no test yet, but I'll write one if the concept here is approved.

### QC

## Summary by CodeRabbit

- **New Features**
	- Enhanced output checking mechanism to selectively ignore specific output files.
	- Improved handling of missing output files in job postprocessing, accounting for temporary files.

- **Bug Fixes**
	- Refined logic for determining if files are needed by pending jobs, improving accuracy in output management.


DrYak:
This looks nice because it also solves a feature I would have liked to have: #1474 

It solves the whole "_files that only live for the duration of the group running on the node_" part of the problem.

pvandyken:
One clarification: sub-snakemake instances on the cluster actually do **not** automatically delete the temp files (they get called with `--notemp`). This behaviour is fine, it just means this PR more specifically deals with temp files automatically deleted by the cluster off of local scratch. I don't think this will have any effect on temp files not saved on local scratch (i.e. they should be deleted as normal after the group job finishes), but I'll do some more specific investigations on this at some point.

DrYak:
@pvandyken : Yes, I meant that we could start also storing our on node-local scratch, completely aleviating all problems of file that don't even live beyond group job.

We're currently testing this on our production server. (:notes: Living on the edge!  :guitar: :man_singer:)

pvandyken:
@DrYak Ah, no worries, I was clarifying myself because I made a mistake in the initial post. I wanted to correct it for the record.

john-hen:
@DrYak Actually, what we're currently testing on our production server (with a "hot patch", as it hasn't been merged yet), is #1779.

johanneskoester (approved):

pvandyken:
@johanneskoester thanks for merging!

PR #1778, opened 2022-07-25
feat: print reason summary in case of dryrun

johanneskoester:
### QC

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -2356,7 +2356,7 @@ class DAG:
         """Print summary of execution reasons."""
         reasons = defaultdict(set)
         for job in self.needrun_jobs():
-            reasons[self.reason(job)].append(job.rule.name)
+            reasons[str(self.reason(job))].append(job.rule.name)
         if reasons:
             msg = "reasons:"
             for reason, rules in sorted(reasons.items()):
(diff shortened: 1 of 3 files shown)


snakemake/dag.py
```diff
@@ -2352,6 +2352,21 @@ def edge(a, b):
                 ],
             )
 
+    def print_reasons(self):
+        """Print summary of execution reasons."""
+        reasons = defaultdict(set)
+        for job in self.needrun_jobs():
+            reasons[self.reason(job)].append(job.rule.name)
```
johanneskoester, line 2359:
```suggestion
            reasons[str(self.reason(job))].append(job.rule.name)
```

snakemake/dag.py
```diff
@@ -2352,6 +2352,21 @@ def edge(a, b):
                 ],
             )
 
+    def print_reasons(self):
+        """Print summary of execution reasons."""
+        reasons = defaultdict(set)
+        for job in self.needrun_jobs():
+            reasons[str(self.reason(job))].append(job.rule.name)
```
johanneskoester, line 2359:
```suggestion
            reasons[str(self.reason(job))].add(job.rule.name)
```

PR #1800, opened 2022-08-08
fix: remote-azblob-sasToken-Authorization

### LINKED ISSUE #1794: Snakemake Uploading output to azure blob storage with sas token
**Snakemake version**
7.12.0

**Describe the bug**
I'm trying to use snakemake.remote.AzBlob to work with input & output from/to Azure blob storage. It works fine when I pass the account key as credentials. However, when I use sas token (generated for specific container) with all permissions (rawcl) I can download the inputs and run the script, but it's failing in uploading to the remote directory. I have tested the sas token and I can write to storage account with other tools.

**Logs**

`Uploading to remote: snakemake/mapped_reads_sas_ud/A.bam HttpResponseError: This request is not authorized to perform this operation.`

**Minimal example**

```
from snakemake.remote.AzBlob import RemoteProvider as AzureRemoteProvider
AS = AzureRemoteProvider(account_url= "https://{accountname}.blob.core.windows.net?{sas-token}")

rule bwa_map:
    input:
        AS.remote("snakemake/data/genome.fa",keep_local=True),
        AS.remote("snakemake/data/samples/A.fastq")
    output:
        AS.remote("snakemake/mapped_reads_sas_ud/A.bam")
    shell:
        "bwa mem {input} | samtools view -Sb - > {output}"
```

**Additional context**

I think the root cause is the following: 
In AzBlob function, line 226, it tries to create the container. This is failing, because the sas token is not authorized to create a container. The exception handling is only considering existence of the container, and is not handling the permission error. 
I spotted the following in my storage account logs: 
`https://{accountname}.blob.core.windows.net:443/snakemake?restype=container...`
which is the indicator of creating the snakemake container. This failed with StatusText: AuthorizationFailure.

RezaMadi:
### Description

In case SAS tokens are generated in the level of a container, uploading the outputs to Azure blob is going to fail on the creation of container. This fix will pass the authorization exception for such cases. This fixes #1794 

### QC

- [x] The PR contains a test case for the changes or the changes are already covered by an existing test case.


RezaMadi:
Dear @johanneskoester, I appreciate it if you could review this PR.

snakemake/remote/AzBlob.py
```diff
@@ -226,6 +226,8 @@ def upload_to_azure_storage(
             container_client.create_container()
         except azure.core.exceptions.ResourceExistsError:
             pass
+        except azure.core.exceptions.ClientAuthenticationError:
```
johanneskoester, line 229: This needs an explaining comment. Why is the authentication error not an issue here. To me it means that the container has not been created, which would be a problem, right?

johanneskoester (changes requested):

snakemake/remote/AzBlob.py
RezaMadi: A SAS token can be created in the level of a container. Unlike a SAS token that is created in the level of the entire storage account, this SAS token is not allowed to perform the create_container operation. 
So, even though the container exists, and the create_container should return the ResourceExistsError, since by this SAS token it's not allowed to perform the operation, the ClientAuthenticationError error will be raised beforehand. Thus, the whole upload fails. 
Does it clarify the situation?

johanneskoester: Ok, understood. And it is impossible that the container does not exist and this exception is still raised? If not, maybe one could check for existence in the exception handling?

johanneskoester: I might miss your answer (I get lots of github notifications), please feel free to contact me via discord.

RezaMadi: Thanks Johannes for the feedback. I thought of checking for the existence of the container, and only try to create it if it doesn't exist. I guess it would cover your concern to some extent. I cannot think of any scenario where your point will be raised.

It will nevertheless still fail for the scenario that I have described on checking the existence of the container. I cannot find any other alternative than passing this exception. 
I have issued a new pull request as I realized the ClientAuthenticationError is not going to capture the error correctly (looks strange). In the new PR I added the condition for checking whether container exists.

johanneskoester (changes requested):

snakemake/remote/AzBlob.py
```diff
@@ -222,10 +222,21 @@ def upload_to_azure_storage(
...
+            if (e.status_code==403):
+                pass
+        
+        
         try:
             container_client.create_container()
         except azure.core.exceptions.ResourceExistsError:
             pass
+        except azure.core.exceptions.ClientAuthenticationError:
+            pass
```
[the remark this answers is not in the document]
johanneskoester: And also please format with black. Thanks a lot!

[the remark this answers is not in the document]
RezaMadi: Hi Johannes,

I realized there are leftovers. Thanks for pointing this out in your comment item 2. It's now corrected. 
I have also added a comment for future reference.

johanneskoester (approved):

PR #1804, opened 2022-08-10
fix: remove superfluous comma causing TypeError in conda-frontend error message

mobilegenome:
### Description
When running snakemake with micromamba a `CreateCondaEnvironmentException` should be raised (should it really? that is a different story... ;)

However, the code `msg += ("lorem ipsum",) tries to concatenate a str and tuple and causes a TypeError. 
Removing the trailing comma fixes this issue and the error message is successfully created and raised. 

### QC

I don't think these are relevant for this simple fix. 


johanneskoester (approved):
Thanks!

johanneskoester:
And yes, micromamba support would be nice! Any PRs welcome!

PR #1810, opened 2022-08-12
Adding flux executor

vsoch:

### Description

This will add the Flux Executor to snakemake. Flux is a very programmatic job manager (maybe that's too limited a term) but it can be run in the context of HPC or cloud! I've written a containerized tutorial so reviewers can test it out. Also pinging @twrs and @grondo to take a look.

@johanneskoester since this requires a new executor backend, let me know how you'd like to test. We do have a container so we could try a strategy that uses that.

### QC


snakemake/executors/flux.py
```diff
@@ -0,0 +1,266 @@
...
+        for key in self.envvars:
+            try:
+                envvars[key] = os.environ[key]
+            except Exception:
+                continue
+        return envvars
+
+    def format_job_exec(self, job):
+        """
+        We don't want a python -m to run.
```
johanneskoester, line 158: Why not? Is there any downside? If this is really needed to be different, rather make this specific part a separate overwritable method in the AbstractExecutor, instead of copying the entire method from there.

snakemake/executors/flux.py
```diff
@@ -0,0 +1,266 @@
...
+
+        # This needs to be called after the line above
+        self._prepare_job_formatter()
+
+        # Prepare job resourcces
+        self._set_job_resources(job)
+
+        # The entire snakemake command to run, etc
+        command = self.format_job_exec(job)
+        nodes = job.resources.get("_nodes")
```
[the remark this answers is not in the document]
johanneskoester: MPI?

snakemake/executors/flux.py
```diff
@@ -0,0 +1,266 @@
...
+            with self.lock:
+                if not self.wait:
+                    return
+                active_jobs = self.active_jobs
+                self.active_jobs = list()
+                still_running = list()
+
+            # Loop through active jobs and act on status
+            for j in active_jobs:
+                logger.info("Checking status for job {}".format(j.jobid))
```
johanneskoester, line 237:
```suggestion
                logger.debug(f"Checking status for job {j.jobid}")
```

snakemake/executors/flux.py
```diff
@@ -0,0 +1,266 @@
...
+
+        # The entire snakemake command to run, etc
+        command = self.format_job_exec(job)
+        nodes = job.resources.get("_nodes")
+        logger.debug(command)
+
+        # Generate the flux job
+        fluxjob = JobspecV1.from_command(
+            command=shlex.split(command),
+            num_nodes=nodes,
```
johanneskoester, line 200: Don't we need to pass additional resources to flux? mem, disk, etc.?

johanneskoester (changes requested):
Awesome, flux looks great! I am eager to try it on our local cluster here. I have some comments below.

snakemake/executors/flux.py
```diff
@@ -0,0 +1,266 @@
...
+        # Attach variables for easy access
+        self.quiet = quiet
+        self.workdir = os.path.realpath(os.path.dirname(self.workflow.persistence.path))
+        self.envvars = list(self.workflow.envvars) or []
+
+        # Quit early if we can't access the flux api
+        self._init_flux()
+
+    def get_default_resources_args(self, default_resources=None):
+        assert default_resources is None
```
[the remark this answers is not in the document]
vsoch: An artifact from GLS - it isn't even used (and I'll remove it).

snakemake/executors/flux.py
```diff
@@ -0,0 +1,266 @@
...
+        Given a particular job, generate the resources that it needs,
+        including default regions and the virtual machine configuration
+        """
+        mem_mb = job.resources.get("mem_mb", 15360)
+
+        # Update default resources
+        self.default_resources = DefaultResources(
+            from_other=self.workflow.default_resources
+        )
+        self.default_resources.set_resource("mem_mb", mem_mb)
```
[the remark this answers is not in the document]
vsoch: They probably don't need to - I'll remove this.

snakemake/executors/flux.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def get_snakefile(self):
+        assert os.path.exists(self.workflow.main_snakefile)
+        return self.workflow.main_snakefile.replace(self.workdir, "").strip(os.sep)
```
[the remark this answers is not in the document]
vsoch: This is another artifact from GLS!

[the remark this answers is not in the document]
vsoch: But why wouldn't the Snakefile at least have the prefix of the workdir? :thinking:

snakemake/executors/flux.py
```diff
@@ -0,0 +1,266 @@
...
+        """
+        Loop through envvars (keys to host environment) and add
+        any that are requested for the runtime environment
+        """
+        envvars = dict(os.environ)
+        for key in self.envvars:
+            try:
+                envvars[key] = os.environ[key]
+            except Exception:
+                continue
```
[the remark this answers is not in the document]
vsoch: This was recommended by a flux developer - I suspect if there is a non zero exit code it raises an exception, possibly this one: https://github.com/flux-framework/flux-core/blob/0fa713c19b1f51b76d9f74063657b9fd54f8e60f/src/bindings/python/flux/job/executor.py#L248-L257.

We could try changing to RuntimeError to be more specific - I'm not practiced enough with flux to know if it can throw other kinds of exceptions.

snakemake/executors/flux.py
vsoch: I'm not sure, but in my example the python -m didn't work, but running snakemake did. I think the issue I was running into is having a different flux install (provided in the container) with the one that works with snakemake (via miniconda). I couldn't get them working from the same one - so python -m won't work (the python does not know about snakemake as it's not installed there) but calling the executable directly does. This is likely an artifact of the container I'm using (already built by the flux crew) but if I added python -m it won't work.

snakemake/executors/flux.py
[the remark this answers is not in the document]
vsoch: I'm actually not sure! I'm not familiar enough with flux to be able to answer, but I'm happy to start with one node and we can tweak if someone opens an issue.

snakemake/executors/flux.py
vsoch: Very likely! Here are the options:

https://github.com/flux-framework/flux-core/blob/0fa713c19b1f51b76d9f74063657b9fd54f8e60f/src/bindings/python/flux/job/Jobspec.py#L568-L571

Which should I try?

docs/executor_tutorial/flux.rst
```diff
@@ -0,0 +1,146 @@
...
+.. code-block:: console
+
+    $ ls
+    Snakefile  snakemake
+
+Here is how to run the workflow:
+
+.. code:: console
+
+    $ snakemake --flux --cores=1
```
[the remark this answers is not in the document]
vsoch: Just testing this now - yes we can do jobs! I've strangely only worked on cloud executors so I wasn't as familiar with jobs as I should be (but I remember it now).

snakemake/executors/flux.py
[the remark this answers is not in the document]
vsoch: okay changed to RuntimeError to reflect what I see in the flux code.

snakemake/executors/flux.py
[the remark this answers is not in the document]
vsoch: yep confirmed we can do MPI! So okay to leave as is?

snakemake/executors/flux.py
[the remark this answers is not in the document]
johanneskoester: You can do `snakemake --snakefile /some/path/Snkaefile --directory /some/completely/disjoint/path`

snakemake/executors/flux.py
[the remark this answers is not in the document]
johanneskoester: Mhm, I still don't see how there could possibly a runtime error when accessing os.environ.

snakemake/executors/flux.py
johanneskoester: Mhm, I think we should try to understand that. So, how is that with flux. Do jobs always run inside of a container, or only in case of cloud mode?

snakemake/executors/flux.py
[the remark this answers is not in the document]
johanneskoester: I am still seeking for a proper way to reflect MPI in Snakemake rules. This will then however most likely not be via nodes resource, but rather some mpi specific keyword. Hence, I'd not make use of nodes here and keep this to the default in flux, which is likely 1, correct?

snakemake/executors/flux.py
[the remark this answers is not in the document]
vsoch: Sorry - the RuntimeError is here https://github.com/snakemake/snakemake/pull/1810/files#diff-b201d1534feeee9b245773a64a6e98ab170f984d4a7ddd0def068448ceaf3273R232 when we get a future result (I associated the comment with the wrong spot). For the above, I originally had the environment empty (and think maybe it should still be that way?) as opposed to starting with the entire os.environ - I forgot to ask about that. For a cluster-like executor should we always be including the user environment, or only a subset of variables requested (e.g., if we did:

```
        envvars = {}
        for key in self.envvars:
            try:
                envvars[key] = os.environ[key]
            except Exception:
                continue
```
that's where the exception could be needed - because perhaps something requested in self.envvars isn't acutally in the environment. Buit maybe I'm misunderstanding the expected content of self.envvars? Sorry it's been a while!

snakemake/executors/flux.py
vsoch: Flux isn't typically in a container I don't think - this is just the dummy example that I'm testing/developing with! Ping @trws and @grondo I think he can add more insight (I haven't yet used flux outside of this single container demo environment).

snakemake/executors/flux.py
johanneskoester: I think we should let flux know about mem_mb, disk_mb, and runtime, see here: https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#standard-resources

snakemake/executors/flux.py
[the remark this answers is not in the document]
vsoch: gotcha!

snakemake/executors/flux.py
grondo: Flux is a resource manager similar to Slurm, PBS, Torque, etc in that it is designed to run across a set of resources, as on a cluster for example. However, it can also be started as a job within one of those resource managers (including itself), to give full RM capabilities to the job. It can also be started standalone on a node, VM, or container to help manage those resources.

I guess to answer more specifically, Flux is not typically used inside a container, but nothing prevents the user from doing that.

snakemake/executors/flux.py
vsoch: @johanneskoester - I checked with @grondo and out of those three, flux only supports runtime, and I've added this as a note to the docs (and that the default is 0 seconds, and should be provided in seconds).

snakemake/executors/flux.py
vsoch: @grondo another question - how does flux handle environment? E.g., should we be grabbing the user environment to provide to snakemake, or start with an empty set and add only those envar names requested? I did the first because I guessed that we would want to preserve most about the user path /home, etc., but maybe that's the HPC use case and not for Flux?

grondo: Flux is more like an HPC resource manager/scheduler,  and most of the command line utilities copy the current environment into the job environment. I'm not familiar with snakemake, but I would assume that would be preferred here as well?

trws: By default flux propagates the entire environment captured where `flux run` or similar was invoked, because that tends to be what a user wants, but it provides a variety of ways to cur that down or override it as well.  It's situational enough that I would think we'd probably want to do what a snakemake user would expect, whatever that is.

To answer the earlier question, flux doesn't run jobs in containers, though that could be supported.  It sounds like there were two python environments installed, and the one with flux bindings is different from the one with snakemake installed.  We could look at making it easier to install the flux binding components into the same environment, or snakemake into the one flux is using to get around the `python -m` problem.

vsoch: @trws that would be fantastic - is this something I could help with? I don't remember the exact issues, but I couldn't get snakemake installed into the default Python in the container.

trws: Actually, you probably could, we should chat about that separately.  It would take some re-packaging of the flux python binding component so it can build separately from the autotools-based build of flux to make it convenient. That's something we have several customers asking for, so if you're up for it that would be really easy to support.

snakemake/executors/flux.py
trws: Our model supports all of the above, but our interface is currently limited.  Worth noting that will likely change.

snakemake/executors/flux.py
vsoch: Yep - would love to help with that! Will ping you tomorrow (and leave me a message in chat if you want to show me stuff sooner).

snakemake/executors/flux.py
vsoch: Maybe I can help here too. <3

snakemake/executors/flux.py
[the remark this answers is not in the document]
johanneskoester: Snakemake ensures that everything in self.envvars is actually present in os.environ before the executor is invoked. Hence, you can safely assume that no error happens. And if so, we actually want to see that exception because that would be a bug in Snakemake. Hence, all you need should be
```
envvars = {os.environ[key] for key in self.envvars}
```

johanneskoester:
> @johanneskoester since this requires a new executor backend, let me know how you'd like to test. We do have a container so we could try a strategy that uses that.

Would be nice to set up a flux instance in some github action before the actual snakemake tests.

vsoch:
I'm going to need the wisdom of the flux folks here - @grondo when you have time can you share what a `flux start start.sh` script might look like? @trws mentioned this could be something to try so it doesn't go into interactive mode! So far my GHA tricks to allow it haven't worked.

snakemake/executors/flux.py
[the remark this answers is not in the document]
vsoch: I've changed this to provide flux the entire user env - there are errors finding snakemake, etc. without that.

vsoch:
woooo @johanneskoester 38th time is the charm!! I got the flux workflow running in the test!! :partying_face:

vsoch:
Sorry for using all the CI cycles :laughing:

johanneskoester:
Nice!

johanneskoester:
So, @vsoch anything still missing here or is it ready for a final review?

vsoch:
Ping @grondo and @trws see above - is this ready or anything missing?

examples/flux/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM fluxrm/flux-sched:focal
+# docker build -t flux-snake .
+# docker run -it flux-snake
+USER root
+ENV PATH=/opt/conda:/bin:$PATH
+RUN pip install git+https://github.com/vsoch/snakemake@add/flux-executor
```
jameshcorbett, line 6: Should this be pointing at the real repo and not your clone? Or were you thinking you would make a PR to change this line once this PR is merged?

snakemake/executors/flux.py
```diff
@@ -0,0 +1,242 @@
...
+        for field in self.jobs_formatter.fields:
+            if field not in fields2attrs:
+                self.job_attrs.update((field,))
+            else:
+                self.job_attrs.update(fields2attrs[field])
+
+    def _init_flux(self):
+        """
+        Ensure we can import flux in the environment.
+        """
```
jameshcorbett, line 98: It looks like this function is only used once, I think it might be marginally clearer if the body of this function was added to `__init__`. Plus I think it's Python best practice to initialize all instance attributes (e.g. `self._fexecutor`) in `__init__`.

snakemake/executors/flux.py
```diff
@@ -0,0 +1,242 @@
...
+        if not flux:
+            sys.exit(
+                "Cannot import flux. Is a cluster available to you with Python bindings?"
+            )
```
jameshcorbett, line 102: I don't know about `snakemake` but usually `sys.exit` is used (if at all) in extreme scenarios. I think it might be more 'polite' to throw an `ImportError`, either directly or by running `import flux`.

snakemake/executors/flux.py
```diff
@@ -0,0 +1,242 @@
...
+            ]
+        )
+
+    def run(self, job, callback=None, submit_callback=None, error_callback=None):
+        """
+        Submit a job to flux.
+        """
+        super()._run(job)
+
+        from flux.job import JobspecV1
```
jameshcorbett, line 160: Is this import necessary, given that `__init__` throws an error if `flux.job` can't be imported? Is `run()` ever invoked as a free function rather than an instance method?

snakemake/executors/flux.py
```diff
@@ -0,0 +1,242 @@
...
+
+        # A duration of zero (the default) means unlimited
+        fluxjob.duration = job.resources.get("runtime", 0)
+
+        # Ensure the cwd is the snakemake working directory
+        fluxjob.cwd = self.workdir
+        fluxjob.environment = dict(os.environ)
+        flux_future = self._fexecutor.submit(fluxjob)
+
+        # Waiting for the jobid is a small performance penalty, same as calling flux.job.submit
```
jameshcorbett, line 184: I think this is my old comment, meant more for you (in case you cared) than to actually go in snakemake

jameshcorbett (approved):
LGTM, sorry it took me so long to get around to this. A few minor comments below, feel free to ignore if you think they aren't worthwhile.

examples/flux/Dockerfile
vsoch: I had this here just so a tester of the PR could quickly use and test - I can either change this just before merge, or right after (the first is probably better). @johanneskoester if you want to tell me right before you merge (and we are sure others don't need to review) I can quickly update!

snakemake/executors/flux.py
vsoch: This function was originally callable and would do the import there, and that design was so a developer could create the class without importing flux (which maybe doesn't work and they are debugging). But with this changed design I agree it doesn't make sense to have separate - I think it might just be my tendency to write things into small functions when possible. I'll push a change to update this.

snakemake/executors/flux.py
vsoch: Agree - it should raise a workflow error - fixing shortly!

snakemake/executors/flux.py
vsoch: I remember there being something buggy going on where if I didn't have this line, it wouldn't work. So for now I think it's safer to be conservative and leave, if that's OK!

snakemake/executors/flux.py
vsoch: haha but it's useful developer info! I loved your comments. OK to keep? I will definitely remove if that is your preference, or if you'd like me to reword it.

examples/flux/Dockerfile
vsoch: Ok this is fixed! If anyone still wants to test, they can tweak to use the branch here.

snakemake/executors/flux.py
vsoch: This is fixed.

snakemake/executors/flux.py
vsoch: This is updated to a WorkflowError - I think it could also be an ImportError, but it does reflect an issue with the workflow running.

vsoch:
Sonarbot be like "stop pushing @vsoch I was already running your tests!"

snakemake/executors/flux.py
jameshcorbett: Not sure I understand your last comment. But regarding your first comment, if you _just_ removed the `from flux.job import JobspecV1` line I would expect it to fail, you would also need to change the `JobspecV1(...)` calls to `flux.job.JobspecV1(...)` calls. In any case not a big deal though.

vsoch: Yep wrong thread (sorry!) I don't remember the issue, but I know I put it there like that for a specific reason.

snakemake/executors/flux.py
vsoch: This is updated to a WorkflowError - I think it could also be an ImportError, but it does reflect an issue with the workflow running. (posting in correct spot)

vsoch:
I think Windows was a transient error - triggered a re-run to test that.

vsoch:
@johanneskoester I think this is ready to go, @trws said yesterday it looked good.

vsoch:
okay @jameshcorbett you were right that we don't need the import there - I'm not sure if it was something about the job formatted, but it's gone now so moving to the top works!

I also removed a huge chunk of code that we needed before futures (but no longer need). I've tested the updates locally and all is working.

vsoch:
oh no snakemake why you doing these random failures!! I'll try a re-run a bit later.

vsoch:
@johanneskoester we are updated and green!!!

snakemake/__init__.py
```diff
@@ -402,7 +404,7 @@ def snakemake(
         assume_shared_fs = False
         default_remote_provider = "GS"
         default_remote_prefix = default_remote_prefix.rstrip("/")
-    if kubernetes:
+    if kubernetes or flux:
         assume_shared_fs = False
```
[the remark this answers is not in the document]
vsoch: It _can_ be liked a cluster executor but doesn't necessarily run in that context.

snakemake/executors/flux.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+
+        # Attach variables for easy access
+        self.workdir = os.path.realpath(os.path.dirname(self.workflow.persistence.path))
+        self.envvars = list(self.workflow.envvars) or []
+
+        # Quit early if we can't access the flux api
+        if not flux:
+            raise WorkflowError(
+                "Cannot import flux. Is a cluster available to you with Python bindings?"
```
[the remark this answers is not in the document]
vsoch: The way Flux is designed now, you usually get the core software installed with the Python bindings - and we don't yet have a "go off and install flux Python" on your own sort of thing. So the expected thing is that they would need to get access to a cluster, or ask their HPC admin. I can put a link to the GitHub repository or docs if you like, but I'm not sure it's going to be useful to most users. But I'm not confident in my answer - ping @trws are there cases where users are able/willing to install their own Flux?

snakemake/executors/flux.py
```diff
@@ -0,0 +1,236 @@
...
+            )
+        self._fexecutor = flux.job.FluxExecutor()
+
+    def _prepare_job_formatter(self):
```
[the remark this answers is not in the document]
vsoch: I wasn't confident that some of those discovered attributes couldn't change before creating the FluxExecutor and then using it, so I was trying to mimic how I see it done in Flux and other tools - having the job formatter called directly before submitting the job.

snakemake/__init__.py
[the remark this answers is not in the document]
vsoch: My thinking here is we should wait for users to try it out, and report an issue (if that case is ever hit) and then we can add a flag that switches between the two cases.

snakemake/executors/flux.py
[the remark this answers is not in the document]
grondo: FYI - Since Flux can run under other resource managers or very easily standalone with `flux start` it is much more likely users might install their own copy (or use the spack package) than a traditional RM.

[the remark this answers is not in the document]
vsoch: Hmm so maybe we should direct them to an install page. @grondo which one would you suggest?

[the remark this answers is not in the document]
grondo: If the current cluster is running Flux and the bindings can't be imported then most of the Flux commands won't work so users will already be contacting their sysadmins. I'd just point users to the homepage on error as suggested.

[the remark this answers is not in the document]
vsoch: okay added! https://github.com/snakemake/snakemake/pull/1810/commits/92d78e827cf6753669a6f41211f5b0e01f4f3a00

snakemake/__init__.py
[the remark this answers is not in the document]
johanneskoester: ok

snakemake/executors/flux.py
johanneskoester: Was there any progress with this? I still would like to get rid of this repetition of superclass code.

.github/workflows/test-flux.yaml
```diff
@@ -0,0 +1,42 @@
...
+          export PATH=/opt/conda:/bin:$PATH
+          # Install snakemake
+          pip install .
+
+      - name: Start Flux and Test Workflow
+        run: |
+          su fluxuser
+          cd examples/flux
+          which snakemake
+          flux start snakemake --flux --jobs=1
```
johanneskoester, line 42: Can you explain this? I would have expected that flux is started with `flux start`, and then snakemake is started with `snakemake --flux` to submit jobs to flux.

johanneskoester (changes requested):

.github/workflows/test-flux.yaml
vsoch: If you do `flux start` on it's own, it essentially opens up an interactive shell (and then you can go as you please to interact with flux). To run in actions (or to do in one command) if you provide a script (or other binary to run) to flux start, it runs that (in the context of flux) so it will function as expected. If we do flux start (newline) followed by snakemake run in actions, I suspect it will hang when it hits the interactive shell, or GitHub actions will kill the shell and then the snakemake command will fail.

But to be clear, this is different from when you interact with flux on a cluster where it is installed - you typically don't need to run flux start because you already have access. The exception would be if you have, for example, a SLURM cluster and you install flux on your own to use under an allocation, you would again be calling flux start. And if you wanted something to run programatically, you'd probably do `flux start <command>` as we do here.

Conceptually (and I might not be the best one to explain this) Flux start is creating a flux instance https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-start.html with one or more brokers. I can do this on my own as a user that installed my own Flux (or in a container I just created, akin to the action) or there can already be a resource with brokers running I have access to. I think in the future we could have support for some kind of background flux start, but for now this is the strategy I've been using in CI setups running flux.

vsoch:
> Was there any progress with this? I still would like to get rid of this repetition of superclass code.

@johanneskoester I don't feel super strongly about this - if you just want it moved to the init I'd be happy to do that!

vsoch:
Oh wait, this was different:

```python
    def format_job_exec(self, job):
        """
        We don't want a python -m to run.
        """
        prefix = self.get_job_exec_prefix(job)
        if prefix:
            prefix += " &&"
        suffix = self.get_job_exec_suffix(job)
        if suffix:
            suffix = f"&& {suffix}"
        return join_cli_args(
            [
                prefix,
                self.get_envvar_declarations(),
                "snakemake",
                format_cli_arg("--snakefile", self.get_snakefile()),
                self.get_job_args(job),
                self.general_args,
                suffix,
            ]
        )
```
I couldn't get this to work without this function because we can't be sure that the Flux Python is installed to the same python install as snakemake. With the way Flux is installed now, it's more likely to be installed to a system python (and the user will be using their own Python environment). If we do python -m then (from my testing) the job is going to fail. By just calling snakemake directly, we don't require the python where the flux is installed to be the same as where the snakemake binary is.

I tried to get this working the other way and failed multiple times.

vsoch:
> I don't really get this. So there is the flux python bindings, which apparently you can install in the same env as snakemake (and have to, since otherwise your executor code would not even work). But the job which is executed by flux, is just called from the environment where flux itself (like their core system, whatever) is installed into? And there is no way to tell flux about the execution env the job shall run in? Or configure that somehow?

The only way I could get this to work (in the CI and the container) is to install snakemake to the same python as flux, having them separate never worked. I'll prepare you an example that shows the error (or proves me wrong and it works) and then maybe you can make a suggestion as to what to try next.

> So, is the problem indeed just the test setup with the container? Then, we could probably just install flux into the github actions, thereby more closely mimicking a real setup?

I've never tried this, but I'd advocate to use the container so 1. a user can reproduce, and 2. we can always be sure to use the latest flux container provided (and not fall out of sync / require updating two things).

More ideally we'd have a GitHub action that installs flux (akin to the container) but I haven't made that yet.

vsoch:
I think this might be an issue with the Python install - let me test something out and see if it resolves the issue - I forgot that these containers have python3 (but not python). Be back in a bit!

vsoch:
Okay just pushed a change without it - I think the underlying issue was not having python exposed (and I didn't interpret the error message correctly the first time). I'm also thinking it would be hugely useful to have native flux install actions - going to work on that soon in case we want to use here instead of the container!

.github/workflows/test-flux.yaml
johanneskoester: Ok, but still. To me, this one: https://flux-framework.readthedocs.io/en/latest/quickstart.html#starting-a-flux-instance
looks like we could as well do
```
flux start --test-size=4
snakemake --flux --jobs 1
```
in order to have a mode of execution that is closer to reality. Or am I still not understanding this correctly?

vsoch: That command starts an interactive shell - it will work interactively but not in the CI.

vsoch: Would you like me to try it out just for a sanity check and to satisfy your curiosity? Happy to do that - and hey maybe I'm wrong!

vsoch: Here you go: https://github.com/snakemake/snakemake/actions/runs/3177693238/jobs/5178416860 let me know when you've seen it and I'll change it back.

grondo:
```
flux start --test-size=4
snakemake --flux --jobs 1
```

`flux start --test-size=N` starts a test flux instance with N brokers and sets up an environment such that that instance can be used and then spawns a command (by default an interactive shell) with that environment (e.g. like `/usr/bin/env`). Without running _under_ `flux start`, the subsequent `snakemake` would have no way to connect to Flux.

It works this way so we can run many tests in parallel, each under its own `flux start`.

johanneskoester: Ok, understood. I think that might need some clarification in your docs. Maybe I was also just blind.

johanneskoester (approved):

vsoch:
Thanks @johanneskoester ! And apologies if the docs aren't clear - it's good feedback indeed.

trws:
Not that I know of, if it covers what snakemake needs I'm good.

Issue #1811, opened 2022-08-15
Prebuild conda env does not activate in "script:" rule


Jahysama:
**7.12.1**

**Bug**
Prebuild conda env won't activate if  python `script:` is used in rule.

**Logs**
```
Traceback (most recent call last):
  File "/home/kosar/.snakemake/scripts/tmp2b81whnx.test.py", line 5, in <module>
    import luigi
ModuleNotFoundError: No module named 'luigi'
[Mon Aug 15 14:36:04 2022]
Error in rule test:
    jobid: 1
    output: output.txt
    conda-env: test

RuleException:
CalledProcessErrorin line 13 of /home/kosar/Snakefile:
Command 'set -euo pipefail;  /home/kosar/miniconda3/envs/snakemake/bin/python /home/kosar/.snakemake/scripts/tmp2b81whnx.test.py' returned non-zero exit status 1.
  File "/home/kosar/Snakefile", line 13, in __rule_test
  File "/home/kosar/miniconda3/envs/snakemake/lib/python3.9/concurrent/futures/thread.py", line 58, in run
Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: .snakemake/log/2022-08-15T143603.846897.snakemake.log

```

**Minimal example**
Prebuild conda env does not activate in `script:` rule. A simple example files to reproduce:
```
rule all:
    input:
        o='output.txt'
        
rule test:
    input:
        i='input.txt'
    output:
        o='output.txt'
    conda:
        'test'
    script:
        'test.py'
```
test.py:
```
import luigi

with open('output.txt', 'w') as file:
	file.write("something")
```
You also need to create a conda env named `test` with luigi in it (I'm using luigi, but it can be any package not installed in your base env):
```
conda create --name test python=3.9
conda activate test
pip install luigi
```

**Additional context**
I'm considering using argparse with python scripts inside `shell:` to get rid of this error. It works with `.yaml/.yml` defined envs in `conda:`. I'm using prebuild envs, because i need to speed up my initial pipeline working time.

moritzschaefer:
Can we close this (fixed by 7.30.1 #1812/#2332)?

tgelafr-btx:
I continue to have this problem even with Snakemake version 7.30.1, using the same minimal example provided above. For me, though, it also doesn't work from a yaml.

Janzulene:
[...]

PR #1812, opened 2022-08-15
fix: conda env inside script

Jahysama:
### Description

Small fix which enables to run conda envs inside `script:`. Because path to conda was not absolute, python executable also wasn't defined correctly. This fixes the issue described in [#1811](https://github.com/snakemake/snakemake/issues/1811).

### QC


dlaehnemann:
Thanks for digging into this!

Would you feel comfortable adding your minimal example from #1811 as a test case? This would ensure that (i) this problem is fixed and (ii) no future changes break this, again.

Such a test case would be a folder here:
https://github.com/snakemake/snakemake/tree/main/tests

And an entry here:
https://github.com/snakemake/snakemake/blob/main/tests/tests.py

Existing test cases are always a good starting point for creating a new one.

Jahysama:
Thank you for response!
I will add new test case as soon as i can :)

Jahysama:
Finished tests, @dlaehnemann could you please take a look at my PR again, thank you!

dlaehnemann:
As I'm not a code owner, I cannot provide the final code review. But I'll point @johanneskoester here.

dlaehnemann:
@Jahysama : It somehow doesn't find the dummy package on Windows. I have no clue about Windows and Python on Windows, and didn't see anything obvious, where it might go wrong. Could you check if you have any idea?

Also, I'm not a code owner, so I cannot provide the final code review. But I'll point @johanneskoester here.

Jahysama:
@dlaehnemann I think it's the way pip is called on windows; instead of doing it with `pip install -e dummy_package`, it should be done like `python -m pip install -e dummy_package`. It happens because there is no `pip` by default in windows `PATH`.
Source: https://stackoverflow.com/questions/23832586/how-to-run-pip-on-windows. I already applied this fix.

dlaehnemann:
Perfect, thanks for the quick fix. Fingers crossed that the test now also completes on windows.

snakemake/script.py
```diff
@@ -611,9 +611,9 @@ def write_script(self, preamble, fd):
 
     def _is_python_env(self):
         if self.conda_env is not None and ON_WINDOWS:
-            prefix = self.conda_env
+            prefix = os.path.join(self.conda_base_path, "envs", self.conda_env)
         elif self.conda_env is not None:
-            prefix = os.path.join(self.conda_env, "bin")
+            prefix = os.path.join(self.conda_base_path, "envs", self.conda_env, "bin")
```
johanneskoester, line 616: Conda envs are not always named. If they are not named (which is actually the recommended default), self.conda_env is already the correct path. Hence, you need to check here whether it is named or not. I wonder how this actually worked in the test cases apart from the prebuilt one. Maybe first try self.conda_env alone, and if that fails, try prefixing it with the conda base path?

johanneskoester (changes requested):

snakemake/script.py
Jahysama: Sorry for such a long delay.  i have added `if` statements to detect if  `self.conda_env` is a correct path and to add `self.conda_base_path` if necessary.

Jahysama:
@dlaehnemann Hello, David! Sorry for bothering you, but could you please run execution of remaining tests, thank you in advance!

dlaehnemann:
Is it intended that the script `touch.py` is not included in the changes?

Jahysama:
Sorry about it, just added it to my fork, forgot to add it to git

dlaehnemann:
Ah, and then `input.txt` should probably not be included, as it should be created by that rule, right?

Jahysama:
Yes, it is not included, because it will be created by the first Snakefile rule. Although `input.txt` is included only in `expected-results` folder.

dlaehnemann:
Ah, I see. Thanks for the explanation.

I just merged the latest main branch status into this branch, because this is required for merging anyways. So if you add further changes to this branch, make sure to first `git pull` this merge commit.

Jahysama:
@dlaehnemann Hello again, David, and a happy new 2023! I tried a different approach: installing a dummy package before the snakemake run. Could you please execute the tests again, fingers crossed. Thank you in advance!

moritzschaefer:
@Jahysama @johanneskoester I see that after so much great efforts (just read through #1811 and the comments here), the PR has stalled. 

As I am also affected by this bug now, I would be super greatful if you could revisit and finalize this.

Jahysama:
It seems like new version of `eido` package was released causing error below. [Args for `validate_project()` were changed](https://github.com/pepkit/eido/blob/266d78850617e9f65f8bf7651b6a88c80d97529a/eido/validation.py#L61), it doesn't seem to use `exclude_case` arg in [`validate_object`](https://github.com/pepkit/eido/blob/dea59d64c14eeb32b236c25d3fee90702aa2584f/eido/validation.py#L26C38-L26C50) in older version anyway, instead it got replaced with `sample_name_colname`, which is not directly required in `validate_project()`. I think it is safe to remove `exclude_case` arg from [`snakemake/workflow.py`](https://github.com/snakemake/snakemake/blob/f1b8342757fc5554eb7a9ed5d393c376e9089c3c/snakemake/workflow.py#L1332C57-L1332C69)
```
  File "/home/runner/work/snakemake/snakemake/snakemake/__init__.py", line 654, in snakemake
    workflow.include(
  File "/home/runner/work/snakemake/snakemake/snakemake/workflow.py", line 1231, in include
    exec(compile(code, snakefile.get_path_or_uri(), "exec"), self.globals)
  File "/tmp/snakemake-f7zdwzdv/Snakefile", line 8, in <module>
    expand("{sample}.txt", sample=pep.sample_table["sample_name"])
  File "/home/runner/work/snakemake/snakemake/snakemake/workflow.py", line 1331, in pepschema
    eido.validate_project(
TypeError: validate_project() got an unexpected keyword argument 'exclude_case'
```

johanneskoester (approved):

johanneskoester:
Thanks a lot!

Issue #1178, opened 2021-09-16
GenericClusterExecutor is always using all cores


veprbl:
**Snakemake version**
6.7.0

**Describe the bug**

Both `SynchronousClusterExecutor` and `GenericClusterExecutor` are initialized with the `cores` parameter set to `None`
https://github.com/snakemake/snakemake/blob/26373e66fbaba04fa076aebb24ad869f41f228b9/snakemake/scheduler.py#L202

The interpretation of that value is to run on all cores
https://github.com/snakemake/snakemake/blob/26373e66fbaba04fa076aebb24ad869f41f228b9/snakemake/executors/__init__.py#L707

In practice this means that a group job with 1000 constituents will launch some jobs that each spawns up to a 1000 processes.

**Logs**

**Minimal example**

Do something like
```
snakemake --cluster foo --jobs 100000 --cores 1 --local-cores 1 --groups some_workflow=group0 --group-components group0=1000
```

**Additional context**

caspargross:
This bug is urgent since it makes cluster execution of group jobs impossible for me.
Group Rules consume more threads then provided by the SGE job and then are killed.

veprbl:
The group rules seem to not work as documented, so you might want to not use them anyways.

dkoppstein:
Hi, I ran into this bug while running the [smk-simple-slurm](https://github.com/jdblischak/smk-simple-slurm) profile on SLURM, it appears this happens even without setting group. 

ssh into node where snakemake is running: 

```sh
ps -efAL | grep koppstein | grep snakemake | wc -l
130
```

Then there are 130 threads which all look like so: 

```sh
[...]

PR #1820, opened 2022-08-19
fix: set number of cores for GenericClusterExecutor and DRMAAExecutor to cores rather than None (i.e. all)

dkoppstein:
set number of cores for GenericClusterExecutor and DRMAAExecutor to cores rather than None (which defaults to all)

### Description

This pull request passes the --cores parameter to the GenericClusterExecutor and DRMAAExecutor to address #1178. This behavior makes them consistent with the TibannaExecutor and GoogleLifeSciencesExecutor, although it is left to @johanneskoester to decide whether this is the correct behavior in general. Please note that the `--cores=N` spawned snakemake processes on the node continue to be set to `--cores all`, which I'm also not sure is intended.  

### QC

### DIFF
--- a/snakemake/scheduler.py
+++ b/snakemake/scheduler.py
@@ -199,7 +199,7 @@ class JobScheduler:
                 self._executor = constructor(
                     workflow,
                     dag,
-                    None,
+                    cores,
                     submitcmd=(cluster or cluster_sync),
                     cluster_config=cluster_config,
                     jobname=jobname,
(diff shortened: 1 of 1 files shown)


dkoppstein:
Edit, this PR upon further testing does not fully solve the problem. All cores are still requested.

johanneskoester:
This should already work as intended in Snakemake 8 with the plugins.

PR #1834, opened 2022-08-27
remote: catch errors in AUTO provider list

kdm9:
### Description

The AUTO remote imports and initialises all remote providers, however
some remote providers assume they will only be imported or instantiated
if their dependencies are installed and/or their configuration is set within the environment.
This leads to a situation where *all* remote dependencies and *all*
remote configuration must be installed/configured before using the AUTO
provider.

This patch avoids this issue by skipping remote providers that raise
exceptions when either imported or instantiated.

### QC

* [x]  ~The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behaviour or functionalities of Snakemake).~


johanneskoester (approved):
Thanks, good catch!

Issue #1856, opened 2022-09-13
k8s: support fractional CPU requests


kdm9:
Kubernetes will reserve a small fraction of each available node's CPU for kube-system pods that keep the cluster running. This means that  (for example) on an 8 core node, 7900m CPUs (milli-cpus, i.e. 7.9 cpus) are requestable by snakemake jobs. This interacts poorly with Snakemake's view of threads/cpus coming only in integer units: either you ask for 7 CPUs, and waste 0.9 CPUs per node, or 8, in which case nothing will run as 8 > 7.9 so all nodes have insufficient CPU available.

The corresponding PR changes this, by always requesting CPUs in units of milli-CPUs. It adds a CLI flag, `--k8s-cpu-scalar`, to configure the factor by which each job's `{threads}` value will be scaled down. For the example above, we could give `snakemake ... --k8s-cpu-scalar=0.98`, and each 8-core job would request only 7840m CPUs, allowing it to be scheduled on all 8-core nodes. 

Note that this *only* applies to the resource request to kubernetes itself. All other uses of `{threads}` see the typical integral value, including the Snakemake scheduler and the value substituted in the job script itself (e.g. `bwa mem -t {threads}` would still become `bwa mem -t 8`)

I'm not sure if this is the best approach to the problem, but I've used this in testing for a week or so now, and it has functioned well in all testing so far.

PR #1857, opened 2022-09-13
feat: k8s: add --k8s-cpu-scalar

kdm9:
### Description

see #1856. In short, this adds the ability to scale CPU requests to most efficiently use the CPUs of underlying nodes of a k8s cluster, of which typically a small fraction are reserved for system pods.

Open questions for maintainers:

- Is this a reasonable approach?
- Does this need to be extended to other resource requests (i.e. memory)?

### QC

(these aren't done yet, I'll do them once a maintainer approves of the design of this solution)

* [ ] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behaviour or functionalities of Snakemake).


snakemake/executors/__init__.py
```diff
@@ -1913,7 +1915,7 @@ def run(self, job, callback=None, submit_callback=None, error_callback=None):
         # request resources
         container.resources = kubernetes.client.V1ResourceRequirements()
         container.resources.requests = {}
-        container.resources.requests["cpu"] = job.resources["_cores"]
+        container.resources.requests["cpu"] = "{}m".format(int(job.resources["_cores"] * self.k8s_cpu_scalar * 1000))
```
kdm9, line 1918: Can job.resources["_cores"] ever be non-numeric? e.g. a function? Or have any functions been evaluated before it gets to this point?

johanneskoester: No, it is always an integer.

snakemake/__init__.py
```diff
@@ -2286,6 +2290,20 @@ def get_argument_parser(profile=None):
...
+        type=float,
+        help="K8s reserves some proportion of available CPUs for its own use. "
+        "So, where an underlying node may have 8 CPUs, only e.g. 7600 milliCPUs "
+        "are allocatable to k8s pods (i.e. snakemake jobs). As 8 > 7.6, k8s can't "
+        "find a node with enough CPU resource to run such jobs. This argument acts "
+        "as a global scalar on each job's CPU request, so that e.g. a job whose "
+        "rule definition asks for 8 CPUs will request 7600m CPUs from k8s, "
+        "allowing it to utilise one entire node. N.B: the job itself would still "
+        "see the original value, i.e. as the value substituted in {threads}.",
+    )
```
johanneskoester, line 2306: Very good catch! Can this fraction be somehow queried from kubernetes API instead of requiring the user to set it? Or alternatively, can we set it to a reasonable default that will work at least on all k8s clusters?

kdm9: So I believe there is a way to query how much CPU resource is assignable per node, but that is a "live" metric (basically N - `${currently_assigned_mCPUs}`) which will change depending on how many pods are running on the node. During execution, those pods will be a mix of system pods and snakemake jobs, so we don't want to use that number. It would also vary a little between nodes, and we don't know which node this job will be scheduled on.

So, I think it is best to set a global default rather than query k8s. I don't think there is a single value across all K8S deployments, but at least in my experience 0.95 would be a reasonably safe default. The other option would be that we always subtract e.g. 500mCPUs (0.5 of a CPU) from all nodes, hoping that that is enough to catch all system pods, but that seems like a bad idea as we would waste (proportionally) more with smaller nodes and still not be guaranteed to solve the original problem.

snakemake/executors/__init__.py
kdm9: great! thanks

snakemake/__init__.py
johanneskoester: Ok, I trust your expertise here, and we go for the 0.95 as a default with this parameter.
```suggestion
    group_kubernetes.add_argument(
        "--k8s-cpu-scalar",
        metavar="FLOAT",
        default=0.95,
        type=float,
        help="K8s reserves some proportion of available CPUs for its own use. "
        "So, where an underlying node may have 8 CPUs, only e.g. 7600 milliCPUs "
        "are allocatable to k8s pods (i.e. snakemake jobs). As 8 > 7.6, k8s can't "
        "find a node with enough CPU resource to run such jobs. This argument acts "
        "as a global scalar on each job's CPU request, so that e.g. a job whose "
        "rule definition asks for 8 CPUs will request 7600m CPUs from k8s, "
        "allowing it to utilise one entire node. N.B: the job itself would still "
        "see the original value, i.e. as the value substituted in {threads}.",
    )
```

tests/test_kubernetes.py
```diff
@@ -48,7 +48,9 @@ def run(self, test="test_kubernetes", **kwargs):
                     default_remote_provider="GS",
                     default_remote_prefix=self.bucket_name,
                     no_tmpdir=True,
-                    default_resources=DefaultResources(["mem_mb=1000", "disk_mb=50"]), # ensure that we don't get charged too much
+                    default_resources=DefaultResources(
+                        ["mem_mb=1000", "disk_mb=50"]
+                    ),  # ensure that we don't get charged too much
```
kdm9, line 53: this is from black's formatting

johanneskoester (approved):

PR #1858, opened 2022-09-13
fix: request disk_mb resource from k8s

kdm9:
### Description

k8s supports adding ephemeral storage (e.g. node-local SSDs) to the working directory of the container. However, the size of this working directory isn't configured. This PR maps the `disk_mb` resource to the `ephemeral-storage` k8s Allocatable.

As an aside, while reviewing this PR, I think I have found a small issue: during the container init code, we are setting the `container.volume.mounts` property twice with single-entry lists, rather than once with a list of two mounts as their API seems to want. Is my reading correct, and this is a bug? If so, I'll quickly fix it.  https://github.com/snakemake/snakemake/blob/a1e49b6f290daeb2012575ab3a85ca72cfe42747/snakemake/executors/__init__.py#L1855-L1860

### QC

* [x] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake). **My (@kdm9) reading of the docs implied this already happened, but I can add a more explicit statement if need be**

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -1911,6 +1911,7 @@ class KubernetesExecutor(ClusterExecutor):
             container.env.append(envvar)
 
         # request resources
+        logger.debug(f"job resources:  {dict(job.resources)}")
         container.resources = kubernetes.client.V1ResourceRequirements()
         container.resources.requests = {}
         container.resources.requests["cpu"] = job.resources["_cores"]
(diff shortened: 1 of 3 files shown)


kdm9:
@johanneskoester is there a way to trigger the k8s tests on github? They seem to be skipped (I assume as I'm not a maintainer and the PR isn't approved yet?)

johanneskoester (approved):

johanneskoester:
Indeed, they only work for people with access to the repo, because of required tokens (they run on GCE). I always wanted to use something like minikube directly inside of github actions, but I did not yet find the time to implement that (maybe there is even a prebuilt action for that). However, I think that the changes are minimal enough to risk just merging them. If main branch fails afterwards, I will just contact you via this PR. For the future, I think you can also enable in your github settings that the upstream repo is allowed to commit into your PR. This way I could trigger a test run with my permissions.

Regarding the volume mounts, I think you are right! A PR would be very welcome!

kdm9:
Looks like I did break the tests, but not directly: the `gcloud` commands relating to local ssds seem to have changed between the version on github actions and what I'm using. I'll make a PR tomorrow.

johanneskoester:
OK, thanks.

PR #1860, opened 2022-09-14
fix: delay evaluation of tmpdir to actual job execution, and not submission. This way, tmpdir can be dependent on the node context.

johanneskoester:
### Description

see title

### QC


cmeesters:
@johanneskoester I am really sorry, but I have quite some teaching obligations this week (and too many open tasks in this context). Will resume working on this PR next week.

PR #1861, opened 2022-09-14
feat: allow for human friendly resource definitions (e.g. mem="5GB", runtime="1d")

johanneskoester:
### QC


johanneskoester:
cc @cmeesters

cmeesters:
Thank you, looks great. I have not yet found the time to scrutinize everything, but stumbled across the out commented group-job lines in the slurm_jobstep-executor. The comment there is one we had already. So, is it really meant this way? Or did this change creep in accidentally?

cmeesters:
Or perhaps I overlooked "This branch is out-of-date with the base branch" and the merge with main needs to performed, first?

johanneskoester:
> Thank you, looks great. I have not yet found the time to scrutinize everything, but stumbled across the out commented group-job lines in the slurm_jobstep-executor. The comment there is one we had already. So, is it really meant this way? Or did this change creep in accidentally?

See my comment in issue #2060. I think this PR is unrelated.

PR #1864, opened 2022-09-16
fix: consistently use text output in conda shell commands and various little fixes for failing test cases due to conda package changes

johanneskoester:
### QC


vsoch:
ah this must be the fix for my failing test! Watching and I'll re-run when it goes through. Thanks @johanneskoester !

PR #1870, opened 2022-09-21
fix: allocation of local ssds in k8s tests

kdm9:
### Description

Fixes a bug in tests introduced in #1858 

### QC


kdm9:
@johanneskoester this is a fix for the issue in #1858

kdm9:
ignore the failure, it's because there's only one PR and it wants you to squash merge so the commit gets the PR title. Code testing passes fine

PR #1876, opened 2022-09-26
fix: reduce resource requirements for kubernetes tests

johanneskoester:
### QC


johanneskoester:
@kdm9 the kubernetes tests on the main branch have a timeout now. Maybe this is due to the large disk size that is requested there. I am trying to reduce it now.

Issue #1881, opened 2022-09-29
Errors in run directive are not raised and lead to obscure errors. 


SilasK:
**Snakemake version**

7.14.2

python 3.10.6

**Describe the bug**

Errors in run directive are not raised and lead to obscure errors. 

Where does the logging statements go?
Where does the Exception go?

**Minimal example**

Snakefile:
```
rule run_with_error:
    output:
        touch("flag")
    run:
        print("before error")
        logger.info("logging")
        raise Exception("An error occured, but I don't see the erro rmessage")

```

**Logs**

```
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job               count    min threads    max threads
--------------  -------  -------------  -------------
run_with_error        1              1              1
total                 1              1              1

Select jobs to execute...

[Thu Sep 29 09:45:40 2022]
rule run_with_error:
    output: flag
    jobid: 0
    reason: Missing output files: flag
    resources: tmpdir=/tmp

before error
Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: .snakemake/log/2022-09-29T094540.574530.snakemake.log

```

**Additional context**

the snakemake log file. doesn't contain more info. even the print stement is not there.

cgr71ii:
Duplicate of https://github.com/snakemake/snakemake/issues/1698

PR #1883, opened 2022-09-30
fix: Add back logging of run directives

### LINKED ISSUE #1698: exception tracebacks in `run:` blocks are sent to /dev/null, making rules hard to debug
**Snakemake version**

7.8.1

**Describe the bug**

When calling python code in a `run:` block, an uncaught exception will cause the job to fail. As it should. However, the stack trace isn't captured so its not possible to track down the cause of the exception!

**Minimal example**

```
$ cat Snakefile
rule all:
    input:
        "test-output"

rule test:
    output:
        "test-output"
    run:
        raise Exception("Hello? Is anybody out there?")
```

```
$ snakemake -c1
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Job stats:
job      count    min threads    max threads
-----  -------  -------------  -------------
all          1              1              1
test         1              1              1
total        2              1              1

Select jobs to execute...

[Fri Jun  3 15:40:47 2022]
rule test:
    output: test-output
    jobid: 1
    reason: Missing output files: test-output
    resources: tmpdir=/tmp

Complete log: .snakemake/log/2022-06-03T154046.915828.snakemake.log
```
```
$ cat .snakemake/log/2022-06-03T154046.915828.snakemake.log
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Job stats:
job      count    min threads    max threads
-----  -------  -------------  -------------
all          1              1              1
test         1              1              1
total        2              1              1

Select jobs to execute...

[Fri Jun  3 15:40:47 2022]
rule test:
    output: test-output
    jobid: 1
    reason: Missing output files: test-output
    resources: tmpdir=/tmp

Complete log: .snakemake/log/2022-06-03T154046.915828.snakemake.log
```

SilasK:
### Description

Fix #1698 exception tracebacks in run: blocks are sent to /dev/null, making rules hard to debug

### QC

I have created a text based on #1881 but I don't see how I can test that the error message is printed out correctly to a log.

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -1,3 +1,4 @@
+all
 __author__ = "Johannes Köster"
 __copyright__ = "Copyright 2022, Johannes Köster"
 __email__ = "johannes.koester@uni-due.de"
@@ -494,7 +495,7 @@ class CPUExecutor(RealExecutor):
         return ""
 
     def get_job_args(self, job, **kwargs):
-        return f"{super().get_job_args(job, **kwargs)} --quiet all"
+        return f"{super().get_job_args(job, **kwargs)} --quiet"
 
     def run(self, job, callback=None, submit_callback=None, error_callback=None):
         super()._run(job)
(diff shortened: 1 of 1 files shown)


SilasK:
Change suggested by @cclienti

johanneskoester (approved):

PR #1886, opened 2022-10-04
docs: small tweaks to flux documentation

vsoch:
After seeing it live - here are some small tweaks to make it better!

- We do not have more than one step, so it does not make sense to include Step 1.
- The spacing on the first bulleted list is off because I think I used `*` instead of `-`.
- I added an actual link to the Dockerfile in examples.
- The flag in the list was missing the correct quotes
- The first sentence will read/ look better just as "Flux Framework" as part of the sentence instead of in parentheses.

### QC


vsoch:
That was so speedy! Thank you!

vsoch:
Looks great! The one thing I'm not sure how to fix is adding more space below `docker` here:

image

I thought using - would fix it (the list below looks ok) but it did not. It's a tiny thing, not a big issue.

PR #1889, opened 2022-10-05
feat: added support for micromamba frontend

fgvieira:
### QC


johanneskoester (changes requested):
Awesome! Can you add a test case?

f0t1h:
Bump

charmoniumQ:
@f0t1h See #2013

Issue #1892, opened 2022-10-07
Snakemake crashes with AttributeError: 'str' object has no attribute 'name


Redmar-van-den-Berg:
**Snakemake version**
snakemake --version
7.15.1

**Describe the bug**
The snakemake process crashes on a minimal workflow.

**Logs**
```bash
$ snakemake --snakefile test.smk  -c 1
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Traceback (most recent call last):
  File "/home/user/miniconda3/envs/HAMLET/lib/python3.10/site-packages/snakemake/__init__.py", line 733, in snakemake
    success = workflow.execute(
  File "/home/user/miniconda3/envs/HAMLET/lib/python3.10/site-packages/snakemake/workflow.py", line 1031, in execute
    logger.run_info("\n".join(dag.stats()))
  File "/home/user/miniconda3/envs/HAMLET/lib/python3.10/site-packages/snakemake/dag.py", line 2415, in stats
    yield tabulate(rows, headers="keys")
  File "/home/user/miniconda3/envs/HAMLET/lib/python3.10/site-packages/tabulate/__init__.py", line 2048, in tabulate
    list_of_lists, headers = _normalize_tabular_data(
  File "/home/user/miniconda3/envs/HAMLET/lib/python3.10/site-packages/tabulate/__init__.py", line 1471, in _normalize_tabular_data
    rows = list(map(lambda r: r if _is_separating_line(r) else list(r), rows))
  File "/home/user/miniconda3/envs/HAMLET/lib/python3.10/site-packages/tabulate/__init__.py", line 1471, in <lambda>
    rows = list(map(lambda r: r if _is_separating_line(r) else list(r), rows))
  File "/home/user/miniconda3/envs/HAMLET/lib/python3.10/site-packages/tabulate/__init__.py", line 107, in _is_separating_line
    (len(row) >= 1 and row[0] == SEPARATING_LINE)
  File "/home/user/miniconda3/envs/HAMLET/lib/python3.10/site-packages/snakemake/rules.py", line 1222, in __eq__
    return self.name == other.name and self.output == other.output
AttributeError: 'str' object has no attribute 'name'
```

**Minimal example**
```Snakemake
rule all:
    input: "test.out"

rule make_out:
    output: "test.out"
    shell: "touch {output}"
```

**Additional context**

karl-az:
[...]

Issue #1891, opened 2022-10-07
tabulate-0.9 do not support snakemake


Hocnonsense:
**Snakemake version**
7.15.1 (latest)

**Describe the bug**
in tabulate 0.9, a check "_is_separating_line" compare whether a `<class 'snakemake.rules.Rule'>` object is equal to a string `SEPARATING_LINE = "\001"`. This leads to an AttributeError.

Supported solution:
1. limit supported tabulate lower than 0.9, or
2. consider string when comparing Rule object:
    ```python
    def __eq__(self, other):
        if isinstance(other, str): return self.name == other
        return self.name == other.name and self.output == other.output
    ```

**Logs**
```log
~/Work/2022_10-clsxx$ file_path="data/file_path.py" python -m snakemake --profile remote/trench-0.sy/ my_collect_drep -c1 -krnp
Building DAG of jobs...
other: 
Traceback (most recent call last):
  File "/home/hwrn/software/miniconda3/envs/smk/lib/python3.10/site-packages/snakemake/__init__.py", line 733, in snakemake
    success = workflow.execute(
  File "/home/hwrn/software/miniconda3/envs/smk/lib/python3.10/site-packages/snakemake/workflow.py", line 1037, in execute
    logger.run_info("\n".join(dag.stats()))
  File "/home/hwrn/software/miniconda3/envs/smk/lib/python3.10/site-packages/snakemake/dag.py", line 2415, in stats
    yield tabulate(rows, headers="keys")
  File "/home/hwrn/software/miniconda3/envs/smk/lib/python3.10/site-packages/tabulate/__init__.py", line 2048, in tabulate
    list_of_lists, headers = _normalize_tabular_data(
  File "/home/hwrn/software/miniconda3/envs/smk/lib/python3.10/site-packages/tabulate/__init__.py", line 1471, in _normalize_tabular_data
    rows = list(map(lambda r: r if _is_separating_line(r) else list(r), rows))
  File "/home/hwrn/software/miniconda3/envs/smk/lib/python3.10/site-packages/tabulate/__init__.py", line 1471, in <lambda>
    rows = list(map(lambda r: r if _is_separating_line(r) else list(r), rows))
  File "/home/hwrn/software/miniconda3/envs/smk/lib/python3.10/site-packages/tabulate/__init__.py", line 107, in _is_separating_line
    (len(row) >= 1 and row[0] == SEPARATING_LINE)
[...]

PR #1893, opened 2022-10-07
fix: tabulate dependency at < 0.9.0

ftabaro:
### Description

Address issues #1891 and #1892 by limiting tabulate version lower than 0.9.0.


johanneskoester:
Thanks! I have solved this by ensuring that a proper string is passed to tabulate now in PR #1898.

PR #1904, opened 2022-10-12
fix: allow report generation to handle pathlib objects

mbhall88:
### Description

In #1869 I added the ability to use a `pathlib.Path` object in the `script` directive. When generating a report where a rule has a Path in the script directory, this function was causing snakemake to error as it assumes the path is a string. This PR makes it explicitly a string.

### QC


johanneskoester (approved):

johanneskoester:
Thank you, good catch!

PR #1920, opened 2022-10-18
perf: cached os.pathconf() call in _record_path()

roshnipatel:
### Description

The value of `os.pathconf(self._incomplete_path, "PC_NAME_MAX")` is cached in an attribute `Persistence._max_len` -- this avoids repeatedly querying the operating system for the same configuration information every time `marked_incomplete()` is called on a file/job output. (Note: this will dramatically speed up Snakemake DAG building on compute clusters where `os.pathconf()` is slow to return and the number of files checked for incompleteness is very large.)

### QC

### DIFF
--- a/snakemake/persistence.py
+++ b/snakemake/persistence.py
@@ -33,6 +33,7 @@ class Persistence:
         shadow_prefix=None,
         warn_only=False,
     ):
+        self._max_len = None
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):
Good catch!

PR #1921, opened 2022-10-18
fix: remove redundant dot in expand call in multiext documentation

### LINKED ISSUE #1012: multiext documentation mistake
In the [multiext documentation](https://snakemake.readthedocs.io/en/v6.4.0/snakefiles/rules.html#the-multiext-function), it states

> The effect is the same as if you would write expand("some/plot.{ext}", ext=[".pdf", ".svg", ".png"])

which is not quite correct, as this would lead to two dots in front of the extensions.

ressy:
### Description

This just removes a single period in the documentation section about the multiext function, so that this, with a redundant period:

```
expand("some/plot.{ext}", ext=[".pdf", ".svg", ".png"])
```

becomes this, matching the expected output:

```
expand("some/plot{ext}", ext=[".pdf", ".svg", ".png"])
```

Fixes #1012.

### QC

### DIFF
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -233,7 +233,7 @@ The multiext function
         shell:
             ...
 
-The effect is the same as if you would write ``expand("some/plot.{ext}", ext=[".pdf", ".svg", ".png"])``, however, using a simpler syntax.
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):

johanneskoester:
thanks!

PR #1924, opened 2022-10-20
Add dhall to allowable configuration filetypes

ndwarshuis:
### Description

This adds configuration support for Dhall files in addition to JSON and YAML. Just like the other two file types, configurations will be loaded as an OrderedDict.

For those unfamiliar with Dhall, think of it as a typed version of JSON/YAML. It's a (non-Turing complete) functional language as opposed to a data-format language, which should be more concise/robust for many use cases. A few nice features:
- Repeated configuration can be collapsed into reusable functions (no YAML anchors/references)
- Everything is type-safe, which takes care of most of the bugs a schema would normally catch (and the rest can be caught with assert statements)
- Defaults can be assigned to dictionary keys, which takes care of another function a schema would normally do
- no pesky indentation :)

See more information on Dhall [here](https://dhall-lang.org/).

### QC


johanneskoester (approved):

johanneskoester:
Why not, choice in this area doesn't hurt. Thanks!

johanneskoester (changes requested):
dhall first needs to be added to conda-forge though.

ndwarshuis:
@johanneskoester FYI [here's](https://github.com/conda-forge/staged-recipes/pull/20976) the PR to get dhall into conda-forge. It's been open for a while but I'm guessing they are dealing with a backlog from the holidays

ndwarshuis:
@johanneskoester dhall-python is now part of conda forge, not sure if there's still a place to specify optional dependencies (since setup.py changed)?

johanneskoester:
@ndwarshuis sorry for missing this for so long! Since this is optional, it is fine to just import dhall-python and fail with a readable error suggesting the respective python package to be installed. In case I miss this PR again, please send me a friend request on and contact me there.

PR #1925, opened 2022-10-21
snakemake.vim: add envmodules keyword

mnsmar:
### QC


johanneskoester (approved):
Thanks!

PR #1939, opened 2022-10-28
feat: first try to match output files against input files while persisting wildcard values from the consuming job. This can dramatically reduce ambiuity problems. Thanks to @descostesn!

johanneskoester:
Only fall back to generic regex based matching if the wildcard based approach does not work. This is inspired by a chat with @descostesn.

### QC


ShouWenWang:
Any update on this ?

PR #1951, opened 2022-11-03
fix: forwarding --keep-incomplete to cluster executor

prs513rosewood:
### Description

The option `--keep-incomplete` was not correctly forwarded in the case of cluster execution, causing jobs to clean up output files in case of failure, despite the main snakemake process being called with `--keep-incomplete`. Below is a MWE:

```snakefile
rule test:
    output:
        "file.out"
    shell:
        "touch {output[0]}"
```
Run with `snakemake -j1 --cluster "bash -x" --keep-incomplete file.out` to see that `--keep-incomplete` is absent from the `snakemake` call in the generated jobscript.

I've looked at adding a regression test, but I'm not sure how to do that in the snakemake tests, which is why the box below is unchecked. Any advice is appreciated.

### QC

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -314,6 +314,7 @@ class RealExecutor(AbstractExecutor):
                 "--no-hooks",
                 "--nolock",
                 "--ignore-incomplete",
+                format_cli_arg("--keep-incomplete", self.keepincomplete),
                 w2a("rerun_triggers"),
                 w2a("cleanup_scripts", flag="--skip-script-cleanup"),
                 w2a("shadow_prefix"),


johanneskoester:
Thanks a lot!

dariober:
I have nothing to contribute other than saying that `--cluster "bash -x"` is a really cool way to mock cluster execution (do I understand it right?). It would be nice to make it an FAQ or a Q&A on StackOverflow for future reference. Thanks!

nsiccha:
Not sure whether I should open another issue, but when using slurm I'm having the problem which was fixed here.

I.e. when running e.g. `snakemake --keep-going --keep-incomplete --slurm -j1024 --default-resources runtime=10 mem_mb=1000 cpus_per_task=1` with snakemake version `7.32.4` snakemake keeps deleting all output files in a group if a single job failed in that group.

PR #1955, opened 2022-11-04
Add Filelist cache for azure blob storage

bioinformaticsbejo:
Add Filelist cache for azure blob storage when AZ_USE_CACHE environment variable is enabled.

Then listing thousands of files, snakemake makes thousands of HTTP requests to Azure storage. This change list all files once and query this dictionary instead of making HTTP requests.

Although this is not advised when actively creating files, this speeds up dry-runs and graph creation.

### QC

### DIFF
--- a/snakemake/remote/AzBlob.py
+++ b/snakemake/remote/AzBlob.py
@@ -162,8 +162,6 @@ class RemoteObject(AbstractRemoteRetryObject):
 
 
 # Actual Azure specific functions, adapted from S3.py
-
-
 class AzureStorageHelper(object):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


bioinformaticsbejo:
@johanneskoester used *black* to format code

johanneskoester (commented):
Did you consider using Snakemake's inventory functionality for a speedup instead? This would avoid the risk of using something outdated, and also mean less custom code. For example, see here the inventory implementation for google storage:
https://github.com/snakemake/snakemake/blob/a5c3523992656ec8230d59ee12f9652313c6dd96/snakemake/remote/GS.py#L167

PR #1958, opened 2022-11-04
fix: Change ratelimiter dependency to throttler

### LINKED ISSUE #1952: Snakemake incompatible with Python 3.11 due to `ratelimiter` dependency
**Snakemake version**

Any version

**Describe the bug**
Snakemake uses the `ratelimiter` library to manage rate-limited api calls (mainly for checking if remote jobs are finished). This library uses a deprecated method on `asyncio`: `asyncio.coroutine`, that has been fully removed in Python 3.11. As a result, Snakemake will not run at all on Python 3.11

**Logs**
```
  File "/home/runner/work/snakebids/snakebids/.venv/lib/python3.11/site-packages/snakemake/__init__.py", line 747, in snakemake
    success = workflow.execute(
              ^^^^^^^^^^^^^^^^^
  File "/home/runner/work/snakebids/snakebids/.venv/lib/python3.11/site-packages/snakemake/workflow.py", line 949, in execute
    self.scheduler = JobScheduler(
                     ^^^^^^^^^^^^^
  File "/home/runner/work/snakebids/snakebids/.venv/lib/python3.11/site-packages/snakemake/scheduler.py", line 109, in __init__
    from ratelimiter import RateLimiter
  File "/home/runner/work/snakebids/snakebids/.venv/lib/python3.11/site-packages/ratelimiter.py", line 36, in <module>
    class RateLimiter(object):
  File "/home/runner/work/snakebids/snakebids/.venv/lib/python3.11/site-packages/ratelimiter.py", line 127, in RateLimiter
    __aexit__ = asyncio.coroutine(__exit__)
```

**Minimal example**
Any snakemake call will trigger this

**Additional context**
There's another library, [throttle](https://github.com/uburuntu/throttler), that should be a drop-in replacement for ratelimiter and doesn't have the same problems.

pvandyken:
ratelimiter used outdated asyncio code that no longer works in Python 3.11. Throttler is more up to date

As part of the change, _wait_for_jobs in all ClusterExecutors have been made async to be compatible with throttler. The Documentation on cluster design has been updated accordingly.

Resolves #1952

### QC


johanneskoester (approved):

snakemake/common/__init__.py
```diff
@@ -268,3 +271,17 @@ def get_input_function_aux_params(func, candidate_params):
...
+
+@contextlib.asynccontextmanager
+async def async_lock(_lock: threading.Lock):
+    """Use a threaded lock form threading.Lock in an async context"""
+    loop = asyncio.get_event_loop()
+    await loop.run_in_executor(_pool, _lock.acquire)
+    try:
+        yield  # the lock is held
+    finally:
+        _lock.release()
```
johanneskoester, line 287: Instead of this, I am pretty sure we should just asyncio.Lock: https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock

johanneskoester: Can you modify the PR accordingly?

johanneskoester (changes requested):

johanneskoester:
PR for adding throttler to conda-forge is here: https://github.com/conda-forge/staged-recipes/pull/21058

snakemake/common/__init__.py
pvandyken: The issue here is that `asyncio.Lock` is not thread safe. So for portions of the program coming at the resources from different threads, the asyncio locks will seize up and freeze the program (at least based on my tests). So long as there's mixed threading and async, it will need to use this function

pvandyken: (I meant to leave an attribution to the stackoverflow source for this... I'll do that now)

johanneskoester (approved):

PR #1963, opened 2022-11-07
DOC: Explain how to get scattergather.split value

### LINKED ISSUE #1961: DOC: Documentation doesn't explain how I can get `scattergather.split` value
**Describe the bug**
The documentation of scattergather lacks information on how to retrieve the value for the number of splits that are chosen in the run. `scattergather` doesn't seem to be part of config. The example in the docs is too simplistic and simply creates the split output by `touch` which doesn't help.

Here's the relevant line of code:
https://github.com/snakemake/snakemake/blame/main/docs/snakefiles/rules.rst#L1663

I also asked on Stackoverflow but I feel like this is a docs problem rather than a usage problem as I couldn't find any info on this, and looking at source code didn't help either.

https://stackoverflow.com/questions/74340191/how-to-get-the-number-of-total-scattergather-items-in-snakemake-scatter-gather

SultanOrazbayev:
Resolves #1961

### QC

### DIFF
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -1703,6 +1703,18 @@ For more complex workflows it's possible to define multiple processes, for examp
         
 The calls to ``scatter`` and ``gather`` would need to reference the appropriate process name, e.g. ``scatter.split_a`` and ``gather.split_a`` to use the ``split_a`` settings.
 
+For custom use cases, the value assigned to each `scattergather` process can be acessed via an internal property, `workflow._scatter`:
+
+.. code-block:: python
+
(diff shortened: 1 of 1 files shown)


johanneskoester (commented):
Such private fields should not be documented, as they are not guaranteed to stay the same. Can you elaborate on the use case of this? If it is really needed, I would rather first make the field public.

dlaehnemann:
This PR helped me figure out something with re-setting scattergather in a module context, so I figured I'd at least bring it into shape for further discussion. Thanks for providing it! And let's see if we can find a good solution to make this or the needed functionality public.

PR #1966, opened 2022-11-11
feat: add tes token

svedziok:
### Description

This PR adds a parameter for forwarding a JWT to a TES instance for authentication.

### QC


johanneskoester (changes requested):
Thanks! One thing: when configuring like this, the token ends up in the shell history. I think it is better to use an environment variable for this instead of a CLI arg. Is there a default env var used in TES implementations already? If not we could name it just TES_TOKEN?

svedziok:
> Thanks! One thing: when configuring like this, the token ends up in the shell history. I think it is better to use an environment variable for this instead of a CLI arg. Is there a default env var used in TES implementations already? If not we could name it just TES_TOKEN?

Hi Johannses, thanks for the review. Do you think, using env vars is required? Using an arg was in accordance to cwl-tes (see here: https://github.com/ohsu-comp-bio/cwl-tes). We also do not see a big security risk as the tokens' validity is generally short. E.g. in WESkit we document arguments as well as env vars that are set by the users anyhow.

johanneskoester:
Mhm, I think an env var is indeed better here. Also, it can be more convenient because it can be set without needing to tell it to Snakemake upon each invocation.

svedziok:
> Thanks! One thing: when configuring like this, the token ends up in the shell history. I think it is better to use an environment variable for this instead of a CLI arg. Is there a default env var used in TES implementations already? If not we could name it just TES_TOKEN?

solved - added env var and documentation

svedziok:
@johanneskoester all is resolved - can we merge now?

svedziok:
> Mhm, I think an env var is indeed better here. Also, it can be more convenient because it can be set without needing to tell it to Snakemake upon each invocation.

done

svedziok:
> Thanks! One thing: when configuring like this, the token ends up in the shell history. I think it is better to use an environment variable for this instead of a CLI arg. Is there a default env var used in TES implementations already? If not we could name it just TES_TOKEN?

done

johanneskoester (approved):

johanneskoester:
Thanks a lot!

Issue #1882, opened 2022-09-29
WorkflowError: Metadata can't be created as it already exists (Windows)


DaminK:
[Others ](https://github.com/snakemake/snakemake/issues/1788#issuecomment-1221312000) seem to have the same problem on windows, but it was commented on an unrelated Issue so I'll open a seperate issue. 

**Snakemake version:** 7.14.2 (latest stable release)

**Describe the bug**
If a previous job is run again (due to changes in code, config or manually removing output files that are required by a subsequent rule), it sometimes fails to create the corresponding meta data as it is already existing. 

This can be circumvented by either:
- manually removing the metadata folder
- restarting the rule/file again (but likely fails with the same error on a subsequent job that was also already executed, leading to several restarts of a rule/file to get the final results)

**Logs**
```
WorkflowError:
Error recording metadata for finished job ([WinError 183] Cannot create a file when that file already exists: 'D:\\WIPAR\\repo\\.snakemake\\metadata\\tmp6cvr9il8.cmVzdWx0' -> 'D:\\WIPAR\\repo\\.snakemake\\metadata\\cmVzdWx0cy9HTjA2XzIwMjEtMDEtMjBfMTAtMTUtMTYvYW5hdG9taWNhbC9BbGwvRmVhdHVyZXMvdmlzdWFsL2F1dG9jb3ZhcmlhbmNlX3RpbWVsYWdzflsxXS9mZWF0dXJlcy5oNQ=='). Please ensure write permissions for the directory D:\WIPAR\repo\.snakemake
Removing output files of failed job feature_calculation since they might be corrupted:
results/GN06_2021-01-20_10-15-16/anatomical/All/Features/visual/autocovariance_timelags~[1]/features.h5, results/GN06_2021-01-20_10-15-16/anatomical/All/Features/visual/autocovariance_timelags~[1]/features.npy, results/GN06_2021-01-20_10-15-16/anatomical/All/Features/visual/autocovariance_timelags~[1]/conf.yaml
```
The WinError was in the language of my OS (German), so I substituted the English translation.

**Minimal example**
[...]

PR #1971, opened 2022-11-17
fix: issue #1882 WorkflowError: Metadata can't be created as it already exists (Windows)

stanmart:
### Description

Fixes issue #1882: WorkflowError: Metadata can't be created as it already exists (Windows)

The underlying cause of the problem is that `os.rename` was used to overwrite the target file. Unfortunately, it has a different behavior under Unix/Linux (overwrites target) and Windows (fails). `os.replace` is an alternative to `os.rename` that simulates the  Unix behavior.

The required change is extremely small (commit 4ad1ab1) and I have also added a test case covering it (commit e12955f).

### QC

### DIFF
--- a/snakemake/persistence.py
+++ b/snakemake/persistence.py
@@ -459,7 +459,7 @@ class Persistence:
             suffix=f".{os.path.basename(recpath)[:8]}",
         ) as tmpfile:
             json.dump(json_value, tmpfile)
-        os.rename(tmpfile.name, recpath)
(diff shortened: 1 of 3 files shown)


johanneskoester (approved):

johanneskoester:
May I ask you to update to the latest upstream version of the main branch?

stanmart:
Sure! Though it seems to me that this branch is up to date with main unless I'm missing something.

Issue #1032, opened 2021-06-07
Notebook validation failed. 


melund:
I was working on fixing conda environment support for windows. 

In the process I stumbled on a bug in "jupyter notebook" tests.  

I get a validation error from the temporary notebooks where snakemake has added a preamble using nbformat. 

The error is triggered when snakemake runs the notebooks, but also if I just open the notebook in jupyter. 

Has anyone see something similar?

```
Notebook validation failed: Additional properties are not allowed ('id' was unexpected):
{
 "metadata": {
  "tags": [
   "snakemake-job-properties"
  ],
  "trusted": true
 },
 "id": "5d9f3b5a",
 "cell_type": "code",
 "source": "\n######## snakemake preamble start 

```

johanneskoester:
Yes, I have seen that as well. This somehow seems to be an incompatibility between nbformat and jupyter. Luckily it does not seem to be a critical error, but of course any hint on how to solve this would be nice.

Christian-Heyer:
Running into the same issue. A fix for now is to pin `nbformat<5.1` in the snakemake enviroment. 

It seems nbformat added cell id tags in 5.1.0 which isn't playing nice with jupyter and nbconvert.

khughitt:
> As an aside: is it possible to run notebooks through snakemake without also creating a separate output?

I was also wondering about this.. Since the "output" for notebooks is usually the notebook itself, it would be nice to be able to drop the ~`output: test.txt` line.

Alternatively, it could be useful to support specifying an arbitrary ".html" output path, and having snakemake automatically call `nbconvert` and attempt to generate an HTML output for a notebook. Just a thought..

johanneskoester:
The id issue is finally fixed now. It was suprisingly difficult to find the right solution, as the behavior seems a bit undocumented. And chatgpt hallucinates heavily about this :-/.
For documentation:
nbformat adds the id field, but it it not part of the notebook v4 schema. During execution, depending on the jupyter or papermill version and how strictly the notebook is validated this can fail. I now manually remove the id field after cell creation with nbformat: #3767.

PR #1986, opened 2022-12-01
fix: json validation error with markdown cells

### LINKED ISSUE #718: Markdown cells in edited notebook have invalid "outputs"
**Snakemake version**
`5.26.1`

**Describe the bug**
For notebooks that contain markdown cells, when loading a previously edited notebook `nbformat.validate` raises the error `Notebook JSON is invalid: Additional properties are not allowed ('outputs' was unexpected)`

**Logs**
```
[E 16:26:01.489 NotebookApp] Notebook JSON is invalid: Additional properties are not allowed ('outputs' was unexpected)

    Failed validating 'additionalProperties' in markdown_cell:

    On instance['cells'][5]:
    {'cell_type': 'markdown',
     'metadata': {},
     'outputs': ['...0 outputs...'],
     'source': '# Test markdown\n'
               '\n'}
```

**Minimal example**
Any markdown cell should do

**Additional context**
All cell outputs are set to an empty list when saving an edited notebook, which might be the source of the error:
https://github.com/snakemake/snakemake/blob/a16ed719fbee96f0119c663eaa13a46c24025d03/snakemake/notebook.py#L95

prs513rosewood:
### Description

This should fix #718. The cleanup routine after editing a notebook would add the `"outputs"` attribute to all cells in the notebook, which would cause the JSON validation of nbconvert to fail for cells where no output is expected.

I've also added a cleanup of the `"execution_count"` property, which should make SCM management of the notebook easier (execution count of cells is prompt to frequent changes in interactive sessions). I can remove it if that change is superfluous.

Finally, I think #1032 can be safely closed, I have not encountered it with recent versions of snakemake.

### QC

### DIFF
--- a/snakemake/notebook.py
+++ b/snakemake/notebook.py
@@ -117,6 +117,8 @@ class JupyterNotebook(ScriptBase):
                 for cell in nb["cells"]:
                     if "outputs" in cell:
                         cell["outputs"] = []
+                    if "execution_count" in cell:
+                        cell["execution_count"] = None
 
                 nbformat.write(nb, self.local_path)
 
--- a/snakemake/notebook.py
+++ b/snakemake/notebook.py
@@ -115,7 +115,8 @@ class JupyterNotebook(ScriptBase):
 
                 # clean up all outputs
                 for cell in nb["cells"]:
-                    cell["outputs"] = []
+                    if "outputs" in cell:
+                        cell["outputs"] = []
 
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):

prs513rosewood:
It looks like tests are failing due to a failed deployment of slurm. I don't really know what to do about that.

johanneskoester:
Should be fixed with the merge I just did

Issue #1992, opened 2022-12-06
Failed GLS job log files are not uploaded to bucket AND --show-failed-logs does not work on instance


cademirch:
**Snakemake version**

`7.18.2`
**Describe the bug**

When a job fails on GLS the stdout/stderr from the instance is captured and uploaded to the specified bucket. However, if the stderr/stdout is captured in a log file, that file is not uploaded to the bucket. Looking at the executor code, it doesn't seem like this is a feature - so this isn't a bug - but it would be a nice feature.

I attempted a quick work around by passing the option `--show-failed-logs`, expecting that would cat the log file to stdout (on the instance), which _would be_ captured. However, it seems that `--show-failed-logs` is not passed to the snakemake command on the instance, rather it tries to display the log file locally, which of course does not exist.
**Logs**

- Execute the ME with `--show-failed-logs`:
```snakemake --google-lifesciences --default-remote-prefix gls-log-test -j1 --show-failed-logs```
 - It fails as expected and tries to display the log file (which does not exist locally):
 ```
Error in rule hi:
    jobid: 1
    output: gls-log-test/hi.txt
    log: gls-log-test/logs/hi/log.txt (check log file(s) for error message)
    shell:
        hi 2> gls-log-test/logs/hi/log.txt
        (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)
    jobid: 6527137249714480570
Logfile gls-log-test/logs/hi/log.txt not found.
```
 - Check the command executed on the cloud:
 ```
> gcloud beta lifesciences operations describe projects/snaketest/locations/us-central1/operations/4249048113650117362

/tmp/workdir.tar.gz && tar -xzvf /tmp/workdir.tar.gz && python -m snakemake
        --snakefile 'Snakefile' --target-jobs 'hi:' --allowed-rules 'hi' --cores 'all'
        --attempt 1 --force-use-threads  --resources 'mem_mb=1000' 'disk_mb=1000'  --force
        --keep-target-files --keep-remote --max-inventory-time 0 --nocolor --notemp
        --no-hooks --nolock --ignore-incomplete --rerun-triggers 'params' 'mtime'
        'software-env' 'code' 'input' --skip-script-cleanup  --conda-frontend 'mamba'
        --wrapper-prefix 'https://github.com/snakemake/snakemake-wrappers/raw/' --latency-wait
        5 --scheduler 'ilp' --default-remote-prefix 'gls-log-test' --default-remote-provider
[...]

PR #1993, opened 2022-12-06
add '--show-failed-logs` to executor args

cademirch:
### Description

See #1992 

Adds `--show-failed-logs` to Snakemake call in cloud containers. This is a workaround to capture the stderr/stdout that is redirected to a rule log file which is not saved when a job fails on GLS. 

@johanneskoester would appreciate your feedback/discussion.

### QC

### DIFF
--- a/snakemake/executors/__init__.py
+++ b/snakemake/executors/__init__.py
@@ -318,6 +318,7 @@ class RealExecutor(AbstractExecutor):
                 "--no-hooks",
                 "--nolock",
                 "--ignore-incomplete",
+                "--show-failed-logs",
                 format_cli_arg("--keep-incomplete", self.keepincomplete),
                 w2a("rerun_triggers"),
                 w2a("cleanup_scripts", flag="--skip-script-cleanup"),


johanneskoester:
Mhm, why is the logfile not saved though?

cademirch:
> Mhm, why is the logfile not saved though?

I'm not entirely sure to be honest. I'm not super familiar with the other remote executors, but for GLS, only the stderr and stdout from the container is captured and uploaded to the bucket. This is not helpful if you've defined a log for a rule and redirect stderr/stdout to the log as is best practices. 

Ideally we would upload the defined log on rule failure, and I could probably figure that out for GLS, but for other remote executors I'm not sure.

cademirch:
> Mhm, why is the logfile not saved though?

I looked at this some more and think I may understand why this is. For GLS, 2 'actions' always occur on the cloud instance. 1) The actual Snakemake command creating some file for the pipeline. 2)`google_lifesciences_helper.py` is run which uploads the captured stderr and stdout from the instance to the bucket. This script only has knowledge of the job name and jobid.

If the snakemake action finishes without error, then all files (outputs and logs) are uploaded to the bucket by Snakemake itself. However, if the job fails - then Snakemake cannot/does not upload anything. I think the real problem lies here - if a job fails when using a remote provider (i.e. `--default-remote-prefix`) then on job failure, Snakemake should at least attempt to upload the log file to the remote prefix. What do you think @johanneskoester? I'm not super familiar with the remote codebase, but I can start to dig into this.

PR #2016, opened 2022-12-15
fix: Correctly parse UserDicts in executors

### LINKED ISSUE #1968: Cluster jobs won't run when --set-resource-scopes is used
**Snakemake version**
7.18.2+0.ga5c35239.dirty

**Describe the bug**
Bug in `get_resource_scopes_args()` in `snakemake/executors/__init__.py` line 117 - and as far as I can see this method does nothing useful and should be removed entirely.

**Logs**
After running:
```
$ snakemake --drmaa --set-resource-scopes foo=local
```

The main workflow starts but all cluster jobs fail on startup with an error like this in the log:
```
$ cat ~/tmp/slurm_job.GROUP.3572731.err
Invalid resource scopes: entries must be defined as RESOURCE=SCOPE pairs, where SCOPE is either 'local', 'global', or 'excluded' (Unparseable value: 'foo')
```
Adding a `set -x` line to the jobscript confirms that the local invocation of snakemake does indeed include `--set-resource-scopes foo` when the original command correctly said `--set-resource-scopes foo=local`.

**Diagnosis**

The function `get_resource_scopes_args(self)` in `snakemake/executors/__init__.py` simply says:

```
return format_cli_arg(
            "--set-resource-scopes", self.workflow.overwrite_resource_scopes
        )
```
`format_cli_arg()` in turn calls `format_cli_pos_arg(value)` which performs the test `isinstance(value, dict)` (at `snakemake/executors/common.py` line 15). It looks like `workflow.overwrite_resource_scopes` should pass this test and thus be formatted as `key=value` pairs but in fact it's an instance of `<class 'snakemake.resources.ResourceScopes'>` which inherits from `collections.UserDict` and does not pass the `isinstance(value, dict)` test. Therefore it gets formatted as regular iterable (ie. only the keys are added) and we get the error described above.

So the most obvious fix is to alter the test in `snakemake/executors/common.py` to say:

```
if isinstance(value, dict) or isinstance(value, collections.UserDict):
```

But I don't think this is right; just passing the original resource scopes down to the job is incorrect. From the perspective of the individual cluster job (be it a job to run a single rule or a group job) all resources are global resources, so passing `--set-resource-scopes foo=local` into the job makes no sense. In the case of my workflow, removing the call to the `get_resource_scopes_args()` method entirely yields the expected behaviour.

So you should consider simply removing this method and the single line that calls it, but this may break other use cases I don't understand.

TIM

pvandyken:
ResourceScopes, a subclass of UserDict (not dict), was getting parsed as a regular iterator, not a dict, so only the keys were being passed on.

Resolves #1968

### QC

### DIFF
--- a/snakemake/executors/common.py
+++ b/snakemake/executors/common.py
@@ -1,3 +1,4 @@
+from collections import UserDict
 from snakemake.io import not_iterable
 
 
@@ -12,13 +13,12 @@ def format_cli_arg(flag, value, quote=True, skip=False):
 
 
 def format_cli_pos_arg(value, quote=True):
-    try:
+    if isinstance(value, (dict, UserDict)):
         return join_cli_args(repr(f"{key}={val}") for key, val in value.items())
-    except AttributeError:
-        if not_iterable(value):
-            return repr(value)
-        else:
-            return join_cli_args(repr(v) for v in value)
+    elif not_iterable(value):
+        return repr(value)
+    else:
+        return join_cli_args(repr(v) for v in value)
 
 
 def join_cli_args(args):
--- a/snakemake/executors/common.py
+++ b/snakemake/executors/common.py
@@ -12,12 +12,13 @@ def format_cli_arg(flag, value, quote=True, skip=False):
 
 
 def format_cli_pos_arg(value, quote=True):
-    if isinstance(value, dict):
+    try:
         return join_cli_args(repr(f"{key}={val}") for key, val in value.items())
-    elif not_iterable(value):
-        return repr(value)
-    else:
-        return join_cli_args(repr(v) for v in value)
+    except AttributeError:
+        if not_iterable(value):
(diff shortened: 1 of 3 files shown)


snakemake/executors/common.py
```diff
@@ -12,12 +12,13 @@ def format_cli_arg(flag, value, quote=True, skip=False):
 
 
 def format_cli_pos_arg(value, quote=True):
-    if isinstance(value, dict):
+    try:
```
johanneskoester, line 15: I'd rather explicitly check for isinstace of dict or UserDict instead.

johanneskoester (changes requested):

johanneskoester (approved):

pvandyken:
Something went wrong in the tests... I'll figure that out and merge

PR #2025, opened 2022-12-20
Add CITATION.cff

mwtoews:
### Description

This adds a [CITATION.cff](https://citation-file-format.github.io/) file to aid citation, as supported by [GitHub](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files). GitHub will default to show the preferred citation. The software citation can be used by tools like Zenodo. And if Zenodo is enabled, use the stable DOI and be sure to update the release version and date before pushing the release. There are probably also other software authors to include [from here](https://github.com/snakemake/snakemake/graphs/contributors), but I don't have a clue who.

Tools like [cffconvert](https://github.com/citation-file-format/cff-converter-python) can be used to validate and convert the citation.

AUTHORS.rst is removed, as it was empty.


CITATION.cff
```diff
@@ -0,0 +1,35 @@
...
+    orcid: https://orcid.org/0000-0002-8536-6065
+  title: Snakemake—a scalable bioinformatics workflow engine
+  doi: 10.1093/bioinformatics/bts480
+  journal: Bioinformatics
+  volume: 28
+  issue: 19
+  start: 2520
+  end: 2522
+  year: 2012
+  month: 8
```
johanneskoester, line 35: We should use the F1000 rolling paper here: https://doi.org/10.12688/f1000research.29032.2

CITATION.cff
```diff
@@ -0,0 +1,35 @@
...
+abstract: The Snakemake workflow management system is a tool to create
+  reproducible and scalable data analyses.
+repository-artifact: https://pypi.org/project/snakemake/
+repository-code: https://github.com/snakemake/snakemake
+license: MIT
+authors:
+- family-names: Köster
+  given-names: Johannes
+  orcid: https://orcid.org/0000-0001-9818-9320
+  alias: johanneskoester
```
johanneskoester, line 17: It does not feel right to have me as only author here. Can we scrape this info from git/github and update the citation.cff automatically like every week or so in a CI job? For simplicity, as this is yaml, the ci job you use https://yte-template-engine.github.io/ for templating, basically get contributors with pygithub and insert them here sorted by number of commits or so.
Also the release and the date could be inserted automatically like that.

johanneskoester (changes requested):
Nice thing! I am worried about the fact that this will quickly outdate though. See below.

CITATION.cff
johanneskoester: Thought further about this. I think the update should be done directly in .github/workflows/release-please.yml.
Whenever a new release is created.

johanneskoester: I.e. no weekly job needed, just upon every release. If you want, this could even be a generic new github action, so that it can be applied in all other snakemake related repos and by any project outside. Suprising that there is no such action yet, seems like a common problem to me. But there are seemingly only citation.cff validators so far.

cmeesters:
@johanneskoester what needs to be done here? How is the paper update coming along? edit: I asked because it was on the hackathon project list.

fgvieira:
This PR adds a `CITATION.cff` file, but @johanneskoester  suggested that, instead of having a static file, this should be automatically generated by scrapping the info from git/github.

Maybe scrapping the DOI from:

https://github.com/snakemake/snakemake/blob/9030f352b6a0e4b2def508cc62260b9cba613693/docs/project_info/citations.rst?plain=1#L18

and generating the CFF file with [doi2cff](https://github.com/citation-file-format/doi2cff).

cmeesters:
@fgvieira cool, thanks! Wasn't ware of that tool. Will do for the teaching material repo. Here, I feel not equipped to draw conclusions.

Also, I wonder when the paper update will appear ;-).

CITATION.cff
cmeesters: or, meanwhile, v3

btraven00:
I took the suggestions in this issue and created https://github.com/btraven00/cffbump, that is available as a github action. I'm gonna be testing the workflow in the coming days and then will open a PR reusing the patch in this one.

Feel free to ping me if you're interested in discussing features for the action or want to co-maintain it.

cmeesters:
@btraven00 cool! Actually, the preprint could also be there in the cff. But that should be discussed.

PR #2026, opened 2022-12-21
refactor: move static metadata from setup.py to pyproject.toml

mwtoews:
### Description

This PR moves the project metadata from setup.py to pyproject.toml, as described by [PEP 621](https://peps.python.org/pep-0621/) and supported by [setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). It is a close 1:1 conversion, with a few changes:

- The description is shortened, since this should be a one-line summary of what the distribution does
- The previous "description" content is set as the "readme" content that you'd see rendered on PyPI [here](https://pypi.org/project/snakemake/)
- The "Home" URL is replaced with a "Documentation" URL
- Dependencies and extra dependencies are alphasorted
- Packages are discovered automatically via "find", rather than explicitly listed
- Before publishing a release, run `twine check --strict dist/*`


mwtoews:
Hi @johanneskoester the only conflicting edits with setup.py is from #2020, but this PR already expands the Python version classifiers. The only other trivial change from master in this PR is:
```diff
-     "Programming Language :: Python :: Only",
+     "Programming Language :: Python :: 3 :: Only",
```

johanneskoester (approved):

PR #2040, opened 2022-12-28
fix: use short argument name for `--chdir` for compatibility with Slurm <=v17

### LINKED ISSUE #2014: Using a slurm cluster that doesn't support --chdir=<folder>

I am running `snakemake` version `7.19.1` on a Slurm cluster that doesn't support `--chdir=<folder>` option so the command below:
```
snakemake --slurm --default-resources slurm_account=<account> slurm_partition=<partition> --cores 1 --use-conda --conda-frontend mamba --jobs 1 plot_data
```

Throws:
```
WorkflowError:
SLURM job submission failed. The error message was sbatch: unrecognized option '--chdir=<folder>'
Try "sbatch --help" for more information
```

I checked with the Slurm sysadmin and confirmed that we should be using `--workdir=<folder>` instead of `--chdir=<folder>`.

How can I tell snakemake to use `--workdir=<folder>` instead of `--chdir=<folder>`?

I tried adding `slurm_extra="--workdir=<folder>"` to:
```
snakemake --slurm --default-resources slurm_account=<account> slurm_partition=<partition> slurm_extra="--workdir=<folder>" --cores 1 --use-conda --conda-frontend mamba --jobs 1 plot_data
```

But I got:
```
too many values to unpack (expected 2)
```

tdido:
### QC

This would fix #2014 since the short argument is the same (`-D`) even if the long argument changed from `--workdir` to `--chdir` after Slurm 17:

https://slurm.schedmd.com/archive/slurm-17.02.0/sbatch.html
https://slurm.schedmd.com/sbatch.html

### DIFF
--- a/snakemake/executors/slurm/slurm_submit.py
+++ b/snakemake/executors/slurm/slurm_submit.py
@@ -253,6 +253,8 @@ class SlurmExecutor(ClusterExecutor):
 
         exec_job = self.format_job_exec(job)
         # ensure that workdir is set correctly
+        # use short argument as this is the same in all slurm versions
+        # (see https://github.com/snakemake/snakemake/issues/2014)
         call += f" -D {self.workflow.workdir_init}"
         # and finally the job to execute with all the snakemake parameters
         call += f" --wrap={repr(exec_job)}"
--- a/snakemake/executors/slurm/slurm_submit.py
+++ b/snakemake/executors/slurm/slurm_submit.py
@@ -253,7 +253,7 @@ class SlurmExecutor(ClusterExecutor):
 
         exec_job = self.format_job_exec(job)
         # ensure that workdir is set correctly
-        call += f" --chdir={self.workflow.workdir_init}"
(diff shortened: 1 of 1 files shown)


snakemake/executors/slurm/slurm_submit.py
```diff
@@ -253,7 +253,7 @@ def run(self, job, callback=None, submit_callback=None, error_callback=None):
 
         exec_job = self.format_job_exec(job)
         # ensure that workdir is set correctly
-        call += f" --chdir={self.workflow.workdir_init}"
+        call += f" -D {self.workflow.workdir_init}"
```
johanneskoester, line 256:
```suggestion
        # use short argument as this is the same in all slurm versions
        # (see https://github.com/snakemake/snakemake/issues/2014)
        call += f" -D {self.workflow.workdir_init}"
```

johanneskoester (changes requested):

johanneskoester (approved):

PR #2042, opened 2023-01-01
fix: ga4gh executor resources

xhejtman:
### Description

Setting resources via ga4gh executor actually never(?) worked. This PR fixes setting resources properly.


snakemake/executors/ga4gh_tes.py
```diff
@@ -329,12 +330,16 @@ def _get_task(self, job, jobscript):
...
-        if "mem_mb" in job.resources:
-            task["resources"]["ram_gb"] = job.resources["mem_mb"] / 1000
-        if "disk_mb" in job.resources:
-            task["resources"]["disk_gb"] = job.resources["disk_mb"] / 1000
+        if job.resources.get("_cores"):
+            setattr(task["resources"], "cpu_cores", job.resources["_cores"])
+        if job.resources.get("mem_mb"):
+            setattr(
+                task["resources"], "ram_gb", math.ceil(job.resources["mem_mb"] / 1000)
+            )
```
johanneskoester, line 338: Can't you just write the following (and the same below)?

```suggestion
            task["resources"].ram_gb = math.ceil(job.resources["mem_mb"] / 1000
```

johanneskoester (changes requested):

snakemake/executors/ga4gh_tes.py
xhejtman: no, it does not compile then, it has to be via setattr. at least after some google search.

uniqueg: Shouldn't this rather be:

```python
task.resources.cpu_cores = job.resources["_cores"]
```

...and so on?

snakemake/executors/ga4gh_tes.py
```diff
@@ -329,12 +330,12 @@ def _get_task(self, job, jobscript):
...
-        if "mem_mb" in job.resources:
-            task["resources"]["ram_gb"] = job.resources["mem_mb"] / 1000
-        if "disk_mb" in job.resources:
-            task["resources"]["disk_gb"] = job.resources["disk_mb"] / 1000
+        if job.resources.get("_cores"):
+            task.resource.cpu_cores = job.resources["_cores"]
+        if job.resources.get("mem_mb"):
+            task.resources.ram_gb = math.ceil(job.resources["mem_mb"] / 1000)
+        if job.resources.get("disk_mb"):
+            task.resources.disk_gb = math.ceil(job.resources["disk_mb"] / 1000)
```
uniqueg, line 338: I think for the first case, you can really just use `.get()` in the assignment, like so:

```python
task.resources.cpu_cores = job.resource.get("_cores")
```

And for testing membership in a dictionary, the more conventional/Pythonic/readable way would be to simply write:

```python
if "mem_mb" in job.resources:
    task.resources.ram_gb = math.ceil(job.resources["mem_mb"] / 1000)
if "disk_mb" in job.resources:
    task.resources.disk_gb = math.ceil(job.resources["disk_mb"] / 1000)
```

In principle you could write these as one-liners, too, and then do the whole assignment in the model construction:

```python
task["resources"] = tes.models.Resources(
    cpu_cores = job.resource.get("_cores")
    ram_gb = math.ceil(job.resources["mem_mb"] / 1000) if "mem_mb" in job.resources else None
    disk_gb = math.ceil(job.resources["disk_mb"] / 1000) if "disk_mb" in job.resources else None
)
```

However, that is probably too unwieldy and makes it a lot harder to read. But the one-liner with `.get()` for `_cores` and the `in` operator for the RAM/disk space I think would be the cleanest.

I am wondering about another things though: Is there any particular reason you don't allow floating-point numbers for `ram_gb` and `disk_gb`? The TES specs explicitly allow them and TES clients will generally produce them. Do HPCs like Slurm, PBS etc. not support fractions of Gbs of RAM/disk resources?

xhejtman: fixed

snakemake/executors/ga4gh_tes.py
```diff
@@ -7,6 +7,7 @@
 import os
 import stat
 import time
+import math
```
uniqueg, line 10: Seems like the imports are sorted alphabetically, with whole library imports preceding partial imports. So perhaps move this up between `asyncio` and `io`.

uniqueg (changes requested):
Apart from the import order, LGTM

snakemake/executors/ga4gh_tes.py
xhejtman: done

uniqueg (approved):
LGTM. @johanneskoester, could you please run the CI pipeline? Would be important for us to get this merged for a Snakemake-TES demonstrator in an upcoming paper on hybrid cloud solutions :pray:

johanneskoester:
I don't have the permissions to push to your fork. Please enable that, then I can retrigger the CI. The other option is that you update your branch with latest upstream yourself, this should also trigger the CI.

xhejtman:
> I don't have the permissions to push to your fork. Please enable that, then I can retrigger the CI. The other option is that you update your branch with latest upstream yourself, this should also trigger the CI.

I added you to the repo

johanneskoester (approved):

PR #2043, opened 2023-01-02
fix: slurm default_resources quoting

dlaehnemann:
* to start with, I'll try to generate a failing test by including quoted runtime specification in test_slurm_complex()
* the fix is to use [`shlex.quote()`](https://docs.python.org/3/library/shlex.html#shlex.quote) instead of [`repr()`](https://docs.python.org/3/library/functions.html#repr) for quoting shell command strings (which should also be safer)

Some background on proper shell quoting style can be found in this [stackoverflow answer on `[How to escape single quotes within single quoted strings`](https://stackoverflow.com/a/1250279)

### Description

This fixes errors in slurm execution that are generated due to incorrect quote escaping with `repr()`. These can for example be:

```bash
WorkflowError:
SLURM job submission failed. The error message was /bin/sh: 1: Syntax error: "(" unexpected
```

or:

```bash
WorkflowError:
SLURM job submission failed. The error message was /bin/sh: -c: line 0: unexpected EOF while looking for matching `''
/bin/sh: -c: line 1: syntax error: unexpected end of file
```

### QC


dlaehnemann:
It might make sense to check other uses of `repr()` to see if they should also be replaced by `shlex.quote()`. But I'm just glad to have fixed my current problem and currently don't have the bandwidth to go looking further.

johanneskoester (changes requested):
Using shlex seems like a good idea here. There is one additional issue though: runtime is supposed to be given as an integer, in minutes (see [here](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#standard-resources)). Actually, there is a check missing which ensures that it is an integer and provides a reasonable error message if it is not.

johanneskoester:
> Using shlex seems like a good idea here. There is one additional issue though: runtime is supposed to be given as an integer, in minutes (see [here](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#standard-resources)). Actually, there is a check missing which ensures that it is an integer and provides a reasonable error message if it is not.

I know, sbatch accepts anything from minutes over to more complex formats, but as always, Snakemake should ensure that workflow definitions are as portable as possible, and not slurm specific. For more convenient specifications, we will have PR #1861 soon.

dlaehnemann:
So, is it enough to just change the test to some fake resource (non-`runtime=`) and wait for the other PR to also address the respective `int` (or whatever more complex) testing?

johanneskoester:
Yes, but it must be one of the resources that the slurm executor translates into an sbatch argument. For example `constraint`.

dlaehnemann:
Just to quickly document the long history of commits here: I tried to also include proper quoting and testing for it for `constraint=` annotations, but I could not get the slurm testing harness for snakemake to properly configure [`Features` annotations for nodes](https://slurm.schedmd.com/slurm.conf.html#OPT_Features).

johanneskoester (approved):

PR #2044, opened 2023-01-03
fix: bug with using flux with conda

vsoch:
Flux can in fact have a shared filesystem, and I did some testing today with a workflow (that I wanted to use conda) and given the flag that there was not a shared filesystem, I could not use conda (it would start the workflow and not attempt to create any environments until I removed the `--flux` flag) and then doing this tweak seemed to fix it. So I think there are two possible approaches:

- assume flux has a shared filesystem (possibly not if the flux operator is being run without volumes)
- assume flux has a shared filesystem if conda is wanted

Both could be possible, and I've chosen this approach so we can assume the shared filesystem in cases where it is absolutely needed, or at least this is a good approach to start.I will do more testing in the next few weeks, as I'd like to get a workflow working in our Flux Operator (on Kubernetes).

### QC


johanneskoester (changes requested):
I think we should better make this explicit. Snakemake already has an option --no-shared-fs. Hence, we should by default not assume that there is no shared fs with flux. And then, if the flux instance indeed has no shared fs, the user just has to set --no-shared-fs in addition to --flux. Can you change code and docs accordingly?

vsoch:
And the important detail is that conda _should_ work given a shared fs, and it won't work otherwise. Those two things need to be paired, to some respect. I'm worried that people will do:

```bash
$ snakemake --flux --use-conda
```
which is a common pattern, and it sort of starts to run but doesn't actually work because the environments aren't created. I'm good to switch up the logic a bit, but I want to make sure we handle the use conda case in a way that isn't confusing for the user. What do you think?

johanneskoester:
So, if there is no shared FS, conda should still work. Snakemake should just create the required environments within each job. This happens for example in the kubernetes executor.

vsoch:
@johanneskoester I tested this scenario with the current setup (without the changes for the PR) and it does not work. The flag to add conda doesn’t really do anything.

johanneskoester:
@vsoch, sorry for the delay. THat sounds like a bug/missing feature in the executor then. Can you create a separate PR with your testcase? Then, I'll try to help with debugging, or maybe I can spot the problem immediately.

vsoch:
@johanneskoester all set! I use the use_conda test case, and you can see https://github.com/snakemake/snakemake/pull/2075 that it doesn't attempt to make any conda envs (and then the test fails).

johanneskoester:
I think this can be closed as we fixed the bug in the other PR. Feel free to reopen if I am wrong.

vsoch:
Yep that sounds good, and if we are moving this to a separate executor plugin it shouldn't be an issue to fix here! I'll test this again when we do that.

PR #2063, opened 2023-01-13
fix: Catch missing error stream in Slurm executor

rabdill:
### Description

I ran into a bug in the `job_status` method in which the variable `sacct_error` is set to `None`, but then later is printed using `sacct_error.strip()`, which throws this error (with gently redacted filepaths):

```
Job 146 has been submitted with SLURM jobid 154129239 (log: .snakemake/slurm_logs/rule_sra_to_fastq/154129239.log)
  File "/path_to_project/venv/lib/python3.8/site-packages/snakemake/executors/slurm/slurm_submit.py", line 310, in job_status
    out = subprocess.check_output(
  File "/shared/anaconda/python3-2020.07-mamba/lib/python3.8/subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/shared/anaconda/python3-2020.07-mamba/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'scontrol show jobid -dd 154129189' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

  File "/path_to_project/venv/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 838, in _wait_thread
    asyncio.run(self._wait_for_jobs())
  File "/shared/anaconda/python3-2020.07-mamba/lib/python3.8/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/shared/anaconda/python3-2020.07-mamba/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/path_to_project/venv/lib/python3.8/site-packages/snakemake/executors/slurm/slurm_submit.py", line 354, in _wait_for_jobs
    status = await self.job_status(j.jobid)
  File "/path_to_project/venv/lib/python3.8/site-packages/snakemake/executors/slurm/slurm_submit.py", line 322, in job_status
    f"Error getting status of slurm job {jobid}:\n    sacct error: {sacct_error.strip()}\n    scontrol error: {e.stderr.strip()}"
AttributeError: 'NoneType' object has no attribute 'strip'
Complete log: .snakemake/log/2023-01-13T024214.712457.snakemake.log
'NoneType' object has no attribute 'strip'
``` 
If `sacct_error` is still `None` on line 322, I think the only explanations would be:

1. The call to `subprocess.check_output` on line 292 returned an `IndexError`, which was caught on line 304 and ignored, or
2. The call to `subprocess.check_output` on line 292 returned `None` for the standard error stream, which it shouldn't have done: [That only happens](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.subprocess.Process.stderr) if called with the `stderr=None` parameter, which it wasn't.

There's nothing I can see that's checking whether `sacct_error` is `None`, so my first thought was just to change the declaration of the variable on line 289 to something like `sacct_error = '(none)'` so it prints nicely, but that looks funny *and* doesn't avoid scenario number 2 above, which I suppose could happen if the parameters are changed one day, or something freaky happens with the `subprocess.PIPE` that I honestly don't understand very well.

In any case, this fix isn't particularly sophisticated, but it fixes the `AttributeError` I ran into without modifying any of the logic and accommodates a path (`except IndexError as e`) that's at least common enough to be included here.

### QC


johanneskoester:
Thanks a lot! I've made some slight changes.

Issue #2055, opened 2023-01-10
Wrong batch behavior with non-round batch division


jsgounot:
**Snakemake version**

7.8.2

**Describe the bug**

When `(rule outputs) / (batch number)` is not a round number, batches does not behave correctly.

Here is a short procedural translation of [the current batch process](https://github.com/snakemake/snakemake/blob/main/snakemake/dag.py#L62) with a theoretical job having 476 outputs and 100 batches. Each loop output items indexes for each batch.

```python3
def get_batch(items, idx, batches):
    
    batch_len = math.floor(len(items) / batches)
    # self.batch is one-based, hence we have to subtract 1
    idx = idx - 1
    i = idx * batch_len
    
    if idx == (batches-1):
        # extend the last batch to cover rest of list
        return idx, i, len(items[i:])
    else:
        return idx, i, len(items[i : i + batch_len])
    
items = list(range(476))
for i in range(1, 101):
    print(i, get_batch(items, i, 100))
```

Output tail:

```
90 (89, 356, 4)
91 (90, 360, 4)
92 (91, 364, 4)
93 (92, 368, 4)
94 (93, 372, 4)
95 (94, 376, 4)
96 (95, 380, 4)
97 (96, 384, 4)
98 (97, 388, 4)
99 (98, 392, 4)
100 (99, 396, 80)
```

You can see that the first index is not following a correct portioning of the items list. Batch 100 having `i == 396`, leaving 80 outputs for the last batch while others have 4. This is due to the fact that `i` is based on a rounded value.

**Potential fix**

Inspired by this [stackoverflow post](https://stackoverflow.com/questions/2130016/splitting-a-list-into-n-parts-of-approximately-equal-length):

Procedural code:

```python3
def get_batch(items, idx, batches):
    
    k, m = divmod(len(items), batches)
    
    # self.batch is one-based, hence we have to subtract 1
    idx = idx - 1
    i = idx * k + min(idx, m)
    batch_len = (idx + 1) * k + min(idx + 1, m) 
    
    if idx == batches-1:
        # extend the last batch to cover rest of list
        return idx, i, len(items[i:])
    else:
        return idx, i, len(items[i : batch_len])
    
items = list(range(476))
for i in range(1, 101):
    print(i, get_batch(items, i, 100))
```

```
1 (0, 0, 5)
2 (1, 5, 5)
3 (2, 10, 5)
4 (3, 15, 5)
5 (4, 20, 5)
6 (5, 25, 5)
[...]

PR #2064, opened 2023-01-16
fix: rounding for batch calculation

jsgounot:
Fix batch issue #2055

### Description

This is a PR related to issue #2055. 

### QC

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -69,15 +69,18 @@ class Batch:
                 "Please choose a smaller number of batches.".format(self.rulename)
             )
         items = sorted(items)
-        batch_len = math.floor(len(items) / self.batches)
+        k, m = divmod(len(items), self.batches)
+
         # self.batch is one-based, hence we have to subtract 1
         idx = self.idx - 1
-        i = idx * batch_len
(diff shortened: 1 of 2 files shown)


johanneskoester:
Thanks! May I ask you to format the code with black?

snakemake/dag.py
```diff
@@ -69,15 +69,18 @@ def get_batch(self, items: list):
...
             )
         items = sorted(items)
-        batch_len = math.floor(len(items) / self.batches)
+        k, m = divmod(len(items), self.batches)
+
         # self.batch is one-based, hence we have to subtract 1
         idx = self.idx - 1
-        i = idx * batch_len
+        i = idx * k + min(idx, m)
+        batch_len = (idx + 1) * k + min(idx + 1, m)
```
johanneskoester, line 77: May I ask you to add some comments explaining each line for the future?

johanneskoester (changes requested):

snakemake/dag.py
jsgounot: Good point, I just pushed a new version with comments.

PR #2065, opened 2023-01-16
refactor: update versioneer from 0.18->0.28, clean-up setup.py/.cfg

mwtoews:
### Description

This PR does two main things:

- Update the vendored versioneer, which uses a configuration in `pyproject.toml`, so clean-up `setup.cfg`
- A follow-up from [#2026](https://github.com/snakemake/snakemake/pull/2026) is that most of `setup.py` was to be cleaned-up, but this somehow missed that PR! Keep the repeated `name="snakemake"` metadata in setup.py for now to avoid a [GitHub bug](https://github.com/orgs/community/discussions/6456)

### QC


johanneskoester (approved):

johanneskoester:
Thanks a lot, highly appreciated!

johanneskoester:
@mwtoews release-please action now fails with the following error:
https://github.com/snakemake/snakemake/actions/runs/3937545745/jobs/6735196445

Seemingly the pyproject.toml is invalid. Any idea?

mwtoews:
Hmm, no ideas yet. I've checked with two validators, and the TOML format is not invalid, so it must be something else.

johanneskoester:
The toml does not contain a version string I guess, and this is what it might complain about here: https://github.com/googleapis/release-please/blob/adb805357d6344f1737bb3541ec7354e76f45cdc/src/updaters/python/pyproject-toml.ts#L55

johanneskoester:
I think this hybrid of having classical versioneer and modern style pyproject is not supported by release-please. It is probably best to migrate to poetry based package building and remove versioneer. I can try to do that during the week, but in case you are interested to do that, I would of course greatly appreciate that.

mwtoews:
> The toml does not contain a version string I guess, and this is what it might complain about here: https://github.com/googleapis/release-please/blob/adb805357d6344f1737bb3541ec7354e76f45cdc/src/updaters/python/pyproject-toml.ts#L55

I've independently arrived at this conclusion too! This is handled via `dynamic = ["version"]`, which release-please does not understand. I'll try to submit this issue upstream.

mwtoews:
Another option is to roll-back the versioneer changes and migrate the static metadata to [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html).

johanneskoester:
> Another option is to roll-back the versioneer changes and migrate the static metadata to [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html).

Yes, maybe that is the best option for now. Do you want to do that?

mwtoews:
Sure, I can prepare this.

Issue #2060, opened 2023-01-12
Parallel execution of job groups on SLURM


koen-vg:
**Snakemake version**
7.19.1

**Describe the bug**
I'm sure this is on the agenda, but since e6cdb32fd20a9899f441b3ec8aca3f36710f7a4a, job groups are run entirely sequentially on SLURM even if multiple jobs in the group could be run in parallel. Even so, the runtime for the whole group is calculated with the assumption that jobs that can be run in parallel are run in parallel. This means that job groups can easily time out even though the runtime specification for each individual rule is entirely correct.

**Minimal example**
Something like
```
rule all: [...]

rule a:
    ...
    resources: runtime=1
    group: "foo"
    shell: "sleep 0.8"

rule b:
    ...
    resources: runtime=1
    group: "foo"
    shell: "sleep 0.8"
```
called with `snakemake --slurm -j2 -c2 all` will trigger the problem.

**Additional context**
Please let me know if I can get started in solving this in any way! I'm more than happy to contribute, and I'm starting to know my way around the snakemake codebase a little bit.

pvandyken:
I'm not so familiar with the slurm executor code (@cmeesters is who we want), but I did write the runtime resource calculation for group resources.

My question is why are slurm jobs not being run in parallel? Is there a decisive reason, or is it an oversight or mistake?

cmeesters:
I'm afraid, but I cannot reproduce this behaviour. To confirm, I changed the `test_group_job` Snakefile to contain `touch {output} && sleep 300` to give me enough time. I submitted with `snakemake -j unlimited -c 4 --slurm --default-resources ...`, logged in the job node and:

```
z0133:~ meesters$ ps aux | grep meesters | grep sleep
meesters  142081  0.0  0.0 222836  3476 ?        S    18:57   0:00 /usr/bin/bash -c set -euo pipefail;  touch a/2.out && sleep 300
meesters  142083  0.0  0.0 217144   976 ?        S    18:57   0:00 sleep 300
meesters  142486  0.0  0.0 222836  3376 ?        S    18:57   0:00 /usr/bin/bash -c set -euo pipefail;  touch a/5.out && sleep 300
meesters  142488  0.0  0.0 217144   968 ?        S    18:57   0:00 sleep 300
meesters  142591  0.0  0.0 222836  3480 ?        S    18:57   0:00 /usr/bin/bash -c set -euo pipefail;  touch a/1.out && sleep 300
[...]

Issue #2059, opened 2023-01-12
Clash in job group names across modules leads to scheduling error


koen-vg:
**Snakemake version**
Latest: 7.19.1

**Describe the bug**
In cluster execution, when two different modules specify job groups with the same name, the scheduler doesn't find any ready jobs.

**Minimal example**
I haven't gotten time to produce a complete MWE, but this problem seems to occur when you have to modules which both use the same job group name, i.e. in module A:
```
rule moduleArule:
    ...
    group: "foo"
```
In module B:
```
rule moduleBrule:
    ...
    group: "foo"
```

**Additional context**
The problem seems to be in the determination of whether any jobs are ready, in the `dag._ready(job)` function. This is how it checks whether a job is ready:
```python
n_internal_deps = lambda job: sum(
    self._group.get(dep) == group for dep in self.dependencies[job]
)
return all(
    (self._n_until_ready[job] - n_internal_deps(job)) == 0 for job in group
)
```
I think the arithmetic is somehow wrong when the same group might span several modules.

PR #2066, opened 2023-01-17
fix: slurm group job steps

cmeesters:
### Description

Re-introduces job step execution for group jobs in a SLURM environment. Addresses issues #2060, possibly #2059 

### QC


cmeesters:
Had to trigger the CI twice, reason: unknown.

snakemake/executors/slurm/slurm_jobstep.py
```diff
@@ -64,42 +64,45 @@ async def _wait_for_jobs(self):
...
+                # SLURM's limits: it is not able to limit the memory if we divide the job
+                # per CPU by itself.
+
+                # check whether level_job.resources.mem_mb is TBD
+                # '100' shall be the (tiny) default, if no value has been specified
+                level_mem = (
+                    100
+                    if level_job.resources.mem_mb != type(int)
+                    else level_job.resources.mem_mb
+                )
```
johanneskoester, line 80: I think now I remember the problem again. This 100 can be completely wrong. mem_mb can be TBD if the memory depends on some input file that is created during execution of the group. If we just execute the entire group via Snakemake as it happened when I commented out the job.is_group() case, the job will be started with the correct mem_mb because Snakemake re-evaluates that.

johanneskoester: One way to fix this is to wait for the previous levels to finish whenever a TBD in mem_mb.

johanneskoester:
The test case for parallel group jobs passes: https://github.com/snakemake/snakemake/pull/2092

Hence, this PR is not about fixing the issue (which I would for now consider to be not present (anymore)), but in general a potential improvement. Feel free to contact me (ideally via mail) about how to pursue this further.

johanneskoester:
This does not apply here anymore, as slurm support has been moved into the slurm plugin.

Issue #789, opened 2020-12-07
Group jobs in pipeline messes up file modification time


ManavalanG:
**Snakemake version**
Tested in v5.9.1 and v5.30.0

**Describe the bug**

When snakemake’s `group:` feature is used when pipeline is run in cluster, it seemingly modifies the file modification time of all output files produced in that group. Unfortunately and not surprisingly, sometimes this results in output files being newer than its input files and therefore when rerunning the pipeline, snakemake thinks those output files need to be created again even though they shouldn't be. And this issue happens even if such output files were `protected`.

If the same pipeline is run in local mode or in cluster without the grouping feature, this problem doesn't exist.

**Minimal example**

```py
rule all:
    input:
        f"data/b.txt"

rule first:
    output:
        protected(f"data/a.txt")
    group: 'xyz'
    shell:
        r"""
        touch {output}
        """

rule second:
    input:
        f"data/a.txt"
    output:
        f"data/b.txt"
    group: 'xyz'
    shell:
        r"""
        sleep 5
        touch {output}
        """
```

**Logs**

1. When run in local mode, notice output files are at least 5 secs apart as expected.

```
# Command used: snakemake -s Snakefile
$ stat data/* | grep 'File\|Modify'
  File: ‘data/a.txt’
Modify: 2020-12-07 09:18:51.203767000 -0600
  File: ‘data/b.txt’
Modify: 2020-12-07 09:18:56.218426000 -0600
```

2. When run in cluster mode, notice output files are NOT at least 5 secs apart.

```
# Command used: snakemake --snakefile Snakefile --cluster 'sbatch' -j 4
$ stat data/* | grep 'File\|Modify'
  File: ‘data/a.txt’
Modify: 2020-12-07 09:21:03.519490744 -0600
  File: ‘data/b.txt’
Modify: 2020-12-07 09:21:03.514490722 -0600
```

3. When run in cluster mode with grouping part commented out, notice output files are again at least 5 secs apart as expected.

```
# Command used: snakemake --snakefile Snakefile --cluster 'sbatch' -j 4
$ stat data/* | grep 'File\|Modify'
  File: ‘data/a.txt’
Modify: 2020-12-07 09:36:23.160759361 -0600
  File: ‘data/b.txt’
Modify: 2020-12-07 09:37:03.208906144 -0600
```

**Additional context**
My Environment:
* SLURM 
* Red Hat

twillis209:
[...]

PR #2073, opened 2023-01-23
fix: Postprocess job groups in toposorted order for correct touch times

koen-vg:
### Description

If I understand the mechanics correctly, jobs in job groups executed on clusters get "postprocessed", and as part of this their output files are touched. However, the order in which this happened was not necessarily a topological order, meaning that output files of dependent rules could be touched in the "wrong" order. See https://github.com/snakemake/snakemake/issues/789

This simple change just ensures that jobs are post-processed in topological order, which should solve the above issue.

If desirable, I could also writing a test for the above behaviour to catch potential future issues.

### QC

### DIFF
--- a/snakemake/jobs.py
+++ b/snakemake/jobs.py
@@ -1461,7 +1461,9 @@ class GroupJob(AbstractJob):
             job.cleanup()
 
     def postprocess(self, error=False, **kwargs):
-        for job in self.jobs:
+        # Iterate over jobs in toposorted order (see self.__iter__) to
+        # ensure that outputs are touched in correct order.
+        for job in self:
             job.postprocess(error=error, **kwargs)
         # remove all pipe and service outputs since all jobs of this group are done and the
         # outputs are no longer needed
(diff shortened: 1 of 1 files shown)


snakemake/jobs.py
```diff
@@ -1461,7 +1461,9 @@ def cleanup(self):
             job.cleanup()
 
     def postprocess(self, error=False, **kwargs):
-        for job in self.jobs:
+        # Iterate over jobs in toposorted order (see self.__iter__) to
+        # ensure that outputs are touched in correct order.
+        for job in self:
             job.postprocess(error=error, **kwargs)
```
johanneskoester, line 1467:
```suggestion
        for level in self.toposorted:
            for job in level:
                # postprocessing involves touching output files (to ensure that
                # modification times are always correct. This has to happen in
                # topological order, such that they are not mixed up.
                job.postprocess(error=error, **kwargs)
```

johanneskoester (changes requested):
Good catch! I have a slight change below, as I think it is better to make this explicit to avoid future breakage.

snakemake/jobs.py
koen-vg: Sure, this looks good and it's easier to understand the meaning.

koen-vg:
Feel free to merge then :)

johanneskoester (approved):

johanneskoester:
Thank you so much!

PR #2075, opened 2023-01-23
fix: assume shared filesystem by default when running with --flux

vsoch:

### Description

The `--use-conda` flag does not work with Flux! This is my attempt to reproduce what I see locally. I'll mark this as ready when I see it reproduced here.

### QC


johanneskoester:
@vsoch, the --use-conda flag is correctly passed to the snakemake invocation inside the flux job. I've improved the logging a bit so that we can now see what happens inside the flux job. THe problem is that there is no conda (and probably also no mamba) in the container you use there, or alternatively that the shell is not configured such that it is found.

vsoch:
Are you sure? When I tested, making the change here resulted in the job successfully running. I was running in a container with snakemake and conda, so it should work akin to if there wasn't flux - the conda environment should be created by snakemake first (and then flux uses it). I'm happy to test again to try and understand what you are saying.

vsoch:
Oh I see - let's just install conda there for this test then!

vsoch:
@johanneskoester should we not see the conda environments being created? https://github.com/snakemake/snakemake/actions/runs/4087695318/jobs/7048509767 :thinking:

johanneskoester:
> @johanneskoester should we not see the conda environments being created? https://github.com/snakemake/snakemake/actions/runs/4087695318/jobs/7048509767 thinking

Good catch. Indeed, and this was because there was still the code that enforced to not assume a shared FS when using flux. I've changed this now such that you always have to specify whether your flux uses a shared FS or not (via `--no-shared-fs true|false`). Would you mind adapting the docs in this PR?

vsoch:
Yes of course! Will do right now!

So with the changes - the user is always required to specify `--no-shared-fs` as true or false for Flux - is there a reason we cannot have it default to False unless True is specified? It seems like (most cases) will have a shared filesystem, and it would be rare not to (and thus should be a default). I'm surprised the `--no-shared-fs` is a flag that takes true/false and not just a boolean flag that defaults to False and when set turns to True.

vsoch:
I'll push these changes, but I think we should have the old way of no_shared_fs defaulting to False, and not need this extra argument.

johanneskoester:
Ok, I understand your argument. I agree then, let's have the old style of --no-shared-fs being just a flag and let's assume that by default it is not specified.

vsoch:
Let me know if you'd like any other changes @johanneskoester !

johanneskoester (approved):

johanneskoester:
Thanks a lot @vsoch! Nice work as always!

vsoch:
Thank you for your help here @johanneskoester! I’m actually just this weekend getting the first workflows going that require object storage for the flux operator and I’m using Snakemake! I’m hoping to figure out a sort of plug and play way to do it.

PR #2088, opened 2023-02-02
feat: add cleanup containers option

### LINKED ISSUE #2078: Cleanup containers
I have a pipeline where I update a container quite regularly, so I end up pulling lots of different versions of it. As a result, the singularity directory in the snakemake hidden directory is quite large.

An option `--cleanup-containers` that acts similar to `--conda-cleanup-envs` and removes any containers which are not required by the workflow in it's current state.

mbhall88:
### Description

Closes #2078. I've created an isolated test case, and have also tested it in the wild. Couldn't figure out how to make a test for a proper pipeline run though. Hope the test case I've added is sufficient though.

### QC

### DIFF
--- a/snakemake/persistence.py
+++ b/snakemake/persistence.py
@@ -38,7 +38,6 @@ class Persistence:
         shadow_prefix=None,
         warn_only=False,
     ):
-
         try:
             import pandas as pd
 
--- a/snakemake/persistence.py
+++ b/snakemake/persistence.py
@@ -16,6 +16,8 @@ from functools import lru_cache, partial
 from itertools import filterfalse, count
 from pathlib import Path
 
+from humanfriendly import format_size
+
 import snakemake.exceptions
 from snakemake.logging import logger
 from snakemake.jobs import jobfiles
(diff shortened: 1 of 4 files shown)


mbhall88:
I think the formatting has failed because of the recent major version bump of black (now on v23.1). When I checked formatting with v22.6.0 it just wanted to format `tests/test_slurm.py` and `versioneer.py`, but with v23.1.0 it wants to reformat 28 files.

I'm assuming you don't want me to reformat the whole codebase within this PR? If you do I can easily do that too... I guess the downside is that without passing the formatting check I can't run the CI tests

mbhall88:
@johanneskoester the linux test `tests/tests.py::test_singularity` is failing because

```
ModuleNotFoundError: No module named 'humanfriendly'
```

but that module is in the test environment...

And the windows test `tests/tests.py::test_jupyter_notebook` is failing for the same reason...

Any ideas? Is there somewhere else `humanfriendly` needs to be specified?

snakemake/persistence.py
```diff
@@ -16,6 +16,8 @@
 from itertools import filterfalse, count
 from pathlib import Path
 
+from humanfriendly import format_size
```
johanneskoester, line 19: You have to move this import into the method where you use it. The reason is that Snakemake pickles a part of its code (the class for the snakemake object in scripts) into scripts when executing them. And this fails in the container case whenever there is an import to an external dependency. I am not happy about this and for long already wanted to change that behavior by replacing the pickling with a code generator. But I haven't had the time yet.

johanneskoester (changes requested):

snakemake/persistence.py
mbhall88: Ohhhh I see! Okay, I'll do that now

mbhall88: That did the trick

johanneskoester (approved):

Issue #2084, opened 2023-02-01
Unintuitive and limitting behavior in naming and prefixing nested modules


jotelha:
**Snakemake version**

```
$ snakemake --version
7.19.1
```

and I checked against the changelog, apparently nothing related has been worked on since 7.19.1.

**Describe the bug**
When modularizing Snakemake workflows, one would naturally expect (and need) the workflows to apply rule renaming and path prefixing strictly recursively. 

In short: when `top.Snakefile` modularizes `nested-1st-level.Snakefile` and applies prefix `prefix_a/`,  and `nested-1st-level.Snakefile` modularizes `nested-2nd-level.Snakefile`  and applies `prefix_b/`, then outputs of the latter `nested-2nd-level.Snakefile` should be prefixed `prefix_a/prefix_b/` when executed via the former `top.Snakefile`, and the same rationale should apply to rule renaming.

This is not the case, as the following example demonstrates,

**Minimal example**
We have three Snakefiles,
```console
$ ls -1
nested-1st-level.Snakefile
nested-2nd-level.Snakefile
top.Snakefile
```

with their content

```console
$ cat nested-2nd-level.Snakefile 
from snakemake.utils import min_version
min_version("6.0")

rule leaf:
    output:
        ".done"
    shell:
        "touch {output}"

rule default:
    input:
        rules.leaf.output
    default_target: True

$ cat nested-1st-level.Snakefile 
from snakemake.utils import min_version
min_version("6.0")

module module_2:
    snakefile: "nested-2nd-level.Snakefile"
    prefix: "nested-2nd-level"

use rule * from module_2 as nested_2nd_level_*

rule default:
    input:
       rules.nested_2nd_level_default.input
    default_target: True

$ cat top.Snakefile 
from snakemake.utils import min_version
min_version("6.0")

module module_1:
    snakefile: "nested-1st-level.Snakefile"
    prefix: "nested-1st-level"

use rule * from module_1 as nested_1st_level_*

rule default:
    input:
       rules.nested_1st_level_default.input
    default_target: True
```

so `top.Snakefile` imports, renames, and prefixes `nested-1st-level.Snakefile`, and `nested-1st-level.Snakefile` in turn imports, renames, and prefixes `nested-1st-level.Snakefile`.

All Snakefiles are syntactically correct:
```console
$ snakemake --lint --snakefile nested-2nd-level.Snakefile 
[...]

Issue #1626, opened 2022-05-02
unexpected behavior when using snakemake-modules that import other snakemake-modules


Smeds:
**Snakemake version**
7.50

**Describe the bug**

I have a few different pipelines that I try to combine into one pipeline without success. I have setup a few simple examples that should emulate my problem, see pull-request #1625. The problem is that rules aren't renamed and prefix isn't added to input and output when I have a pipeline with the following format:

pipeline1: with regular rules
pipeline2: contain regular rules and imports pipeline1 as a module and adds a prefix ( this works)
pipeline3: imports pipeline2 as a module and adds a prefix (this doesn't behave as I expect):
- only rules originating from pipeline2 are renamed, not the rules imported as module into pipeline2
- prefix is only added to input/output belonging to pipeline2, and not to pipeline1

**Example**
If one run a dry-run on the different modules the result would be as follows:

***Module level2***
``` 
snakemake -s tests/test_module_multilevel/level1/level2/Snakefile  -d tests/test_module_multilevel/level1/level2/
``` 
``` 
Building DAG of jobs...
Job stats:
job          count    min threads    max threads
---------  -------  -------------  -------------
a                1              1              1
b                1              1              1
joint_all        1              1              1
total            3              1              1


[Mon May  2 14:17:14 2022]
rule a:
    input: test1.txt
    output: test_intermediate.txt
    jobid: 2
    resources: tmpdir=/tmp


[Mon May  2 14:17:14 2022]
rule b:
    input: test_intermediate.txt
    output: test_final_b.txt
    jobid: 1
    resources: tmpdir=/tmp


[Mon May  2 14:17:14 2022]
localrule joint_all:
    input: test_final_b.txt
    jobid: 0
    resources: tmpdir=/tmp

Job stats:
job          count    min threads    max threads
---------  -------  -------------  -------------
a                1              1              1
b                1              1              1
joint_all        1              1              1
total            3              1              1

This was a dry-run (flag -n). The order of jobs does not reflect the order of execution.

``` 
which  have expected result

***Module level1***
``` 
snakemake -s tests/test_module_multilevel/level1/Snakefile  -d tests/test_module_multilevel/level1/ -n
``` 
``` 
Building DAG of jobs...
Job stats:
job          count    min threads    max threads
---------  -------  -------------  -------------
c                1              1              1
joint_all        1              1              1
l2_1_a           1              1              1
l2_1_b           1              1              1
l2_2_a           1              1              1
[...]

PR #2094, opened 2023-02-05
feat: Proper recursive naming of rules and construction of prefixes in nested modules

jotelha:
### Description

This PR suggests a quick and dirty modification to achieve recursive renaming of rules and prefixing of paths in nested modules, addressing https://github.com/snakemake/snakemake/issues/2084, or the earlier https://github.com/snakemake/snakemake/issues/1626 .

I expect this to break stuff. But it does what I need it to do.

After applying the modification, I realized that another PR with the same aim exists already, https://github.com/snakemake/snakemake/pull/1625 . I am suggesting this PR anyway, since the other has not been discussed yet neither, and the implementations differ.

Will fill in samples and tests when able to find some time.

### QC

### DIFF
--- a/snakemake/modules.py
+++ b/snakemake/modules.py
@@ -185,7 +185,11 @@ class WorkflowModifier:
         self.rule_exclude_list = rule_exclude_list
         self.ruleinfo_overwrite = ruleinfo_overwrite
         self.allow_rule_overwrite = allow_rule_overwrite
+
+        self.replace_prefix = replace_prefix
+        self.prefix = prefix
         self.path_modifier = PathModifier(replace_prefix, prefix, workflow)
+
         self.replace_wrapper_tag = replace_wrapper_tag
         self.namespace = namespace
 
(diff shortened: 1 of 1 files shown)


jotelha:
Have to understand how test System works. At the current point, the inserted test case succeeds even If I intentionally corrupt the expected results.

jotelha:
Test now works as expected, issue were . (dot)-prefixed filenames not being checked against.

jotelha:
@johanneskoester would you please consider rerunning the CI workflows and discussing this

pastewka:
@johanneskoester Is there an update on this PR? My research group is making extensive use of SnakeMake (a fantastic tool, thank you) and it would be important for us to get this or at the functionality of this PR merged.

fgvieira:
@jotelha has this been fixed on #1817?

jonasfreimuth:
Heyo,

I recently ran into the issue with recursive name prefixing not working with the Snakemake version I had been using (9.1.3). I had not seen an obvious fix for it in the version history up until the most recent version (currently 9.14.6). So I tried to adapt this PR for the current repo state and discovered the introduced tests passing without the actual Snakemake changes of this PR. It was only then I discovered that [9.11.4](https://github.com/snakemake/snakemake/releases/tag/v9.11.4) (or more precisely https://github.com/snakemake/snakemake/pull/3714) fixed the issue with recursive prefixing not working as expected.

If that is desired I have the adaptation of the tests available at https://github.com/TRON-Bioinformatics/snakemake/tree/prefix-test-on-main (little more than squashing the test commits by @jotelha and adpting the paths slightly), I could merge them.

It seems to me this PR is no longer needed...

Hocnonsense:
Thanks for looking into PR #3714! It has already introduced tests for this behavior, which can be found here:
https://github.com/snakemake/snakemake/blob/main/tests/test_module_nested/

If you notice that some relevant cases are still missing from those tests, please add them, additional coverage would certainly be welcome.

jonasfreimuth:
Nope, the tests from https://github.com/snakemake/snakemake/pull/3714 should be sufficient. This can be closed from my view. 👍

PR #2100, opened 2023-02-07
Adding option to redirect shell stdout and stderr to log files

dfm:
### Description

This PR picks up where @tylergannon left off in #289. Since I don't have access to that branch, I've opened a new PR to try to get that contribution over the finish line, but if it would be better to do this some other way, I'm happy to do that!

Besides the changes that @tylergannon already implemented, I've started addressing @johanneskoester's comments from that PR:

1. I've added a test to cover the case where a user explicitly pipes to a log file called `stdout`. This works as expected: the log file contains both the process' stdout and the piped output.
2. The log files are now explicitly closed after the process exits.

My main open question is that in #289, @johanneskoester [commented](https://github.com/snakemake/snakemake/pull/289#discussion_r399346400):

> The log files should be opened in append mode, such that multiple shell commands below each other don't overwrite their output.

I switched the log files to be opened in append mode, but I'm not sure exactly what kind of workflow @johanneskoester was referring to here, or even really what the expected behavior is for multiple uses of the same log file. I added a test (which fails) for what I thought this was supposed to mean, but if anyone has tips for how to test this, I'd love to hear them!

### QC

### DIFF
--- a/snakemake/io.py
+++ b/snakemake/io.py
@@ -1682,7 +1682,7 @@ class Log(Namedlist):
             dirname = os.path.dirname(path)
             if any(dirname) and not os.path.exists(dirname):
                 os.makedirs(dirname)
-        return {name: open(value, "w") for name, value in _streams.items()}
+        return {name: open(value, "a") for name, value in _streams.items()}
 
 
 def _load_configfile(configpath_or_obj, filetype="Config"):
--- a/snakemake/io.py
+++ b/snakemake/io.py
@@ -3,9 +3,11 @@ __copyright__ = "Copyright 2022, Johannes Köster"
 __email__ = "johannes.koester@uni-due.de"
 __license__ = "MIT"
 
+import _io
 import collections
 from hashlib import sha256
 import os
+import sys
 import shutil
 from pathlib import Path
 import re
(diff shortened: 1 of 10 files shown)


cademirch:
+1 for this. Would be super helpful - sometimes I miss redirecting to log in big(ger shell commands, which then makes troubleshooting more difficult. Would also be a DX improvement. 

As for your question:

> The log files should be opened in append mode, such that multiple shell commands below each other don't overwrite their output.

I think @johanneskoester is referring to the case of multiple commands in one shell block, something like:
```
rule a:
    input: "hi.txt"
    output: "bye.txt"
    log: "log.txt"
    shell:
        """
        cmd1 {input} > {output} 2> {log}
        cmd2 {output} &>> {log}
        """ 
```

It seems like your implementation should cover this, though?

kjohnsen:
Another +1 for this feature. In addition to the use case people have previously outlined, not having to repeat `> log.stdout 2> log.stderr` for `shell:` directives, it would be even nicer for external scripts. i.e., in my Python script I would like to simply capture `print()` statements rather than explicitly write to the log file.

@johanneskoester is there anything I can do to help here?

PR #2105, opened 2023-02-10
feat: Publish docker images for amd64 & arm64

qw-in:
### Description

Hello! New to snakemake, but excited to be here. Had to make a couple of tweaks to the Dockerfile as I'm running on a newer mac and figured I'd try to contribute back.

1. Updated Dockerfile to download correct Mambaforge based on container architecture
    - Used `uname -m` rather than [docker's `$TARGETARCH`](https://docs.docker.com/build/building/multi-platform/) as it outputs what Mambaforge is expecting (`aarch64` & `x86_64`)
    - Can test building for different platforms using ex. `docker build . --platform linux/amd64`
2. Update github action to publish both architectures via Github actions 
    - Was not able to test this directly, but [copied right from the action documentation](https://github.com/elgohr/Publish-Docker-Github-Action#platforms)
3. Noticed VSCode complaining about the `MAINTAINER` key so swapped that out
    - [See the deprecation notice in the docs here](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated)

### QC

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -10,9 +10,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v2
     - name: Publish to Registry
       uses: elgohr/Publish-Docker-Github-Action@v5
       with:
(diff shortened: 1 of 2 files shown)


johanneskoester (approved):
Thanks a lot!

PR #2114, opened 2023-02-15
fix: batch collect jobs for scancel

### LINKED ISSUE #2113: Slurm executer overuses scancel
**Snakemake version**
7.22.0

**Describe the bug**
When using --slurm, exiting snakemake by SIGINT or ctrl-c helpfully cancels spawned jobs. However, this is quite unreliable as it often hangs for a few minutes and exits without canceling the jobs in the end (without indicating that it failed either).

Slurm documentation on [scancel](https://slurm.schedmd.com/scancel.html#SECTION_PERFORMANCE) notes that a large number of scancel calls at the same time may result in denial of service.

Snakemake runs scancel on each job [individually](https://github.com/snakemake/snakemake/blob/main/snakemake/executors/slurm/slurm_submit.py#L136). Instead, job ids should be collected and cancelled all at once:
```python
def cancel(self):
    jobids = " ".join([job.jobid for job in self.active_jobs])
    try:
        subprocess.check_output(
            f"scancel {jobids}",
            text=True,
            shell=True,
            timeout=60,
            stderr=subprocess.PIPE,
        )
    except subprocess.TimeoutExpired:
        logger.warning(f"Unable to cancel jobs within a minute.")
    self.shutdown()

```

I will open a PR with these changes.

**Minimal example**

The issue is intermittent with 8 concurrent jobs, but I am able to reproduce it every time with 20 concurrent jobs. Different slurm installations may have a lower or higher tolerance.
<details>

<summary> File tree </summary>

```
.
├── in
│   ├── 10.txt
│   ├── 11.txt
│   ├── 12.txt
│   ├── 13.txt
│   ├── 14.txt
│   ├── 15.txt
│   ├── 16.txt
│   ├── 17.txt
│   ├── 18.txt
│   ├── 19.txt
│   ├── 1.txt
│   ├── 20.txt
│   ├── 2.txt
│   ├── 3.txt
│   ├── 4.txt
│   ├── 5.txt
│   ├── 6.txt
│   ├── 7.txt
│   ├── 8.txt
│   └── 9.txt
├── out
└── Snakefile
```
</details>

<details>

<summary> Snakefile </summary>

```python
SAMPLES, = glob_wildcards("in/{sample}.txt")

rule all:
    input: expand("out/{sample}.txt", sample=SAMPLES)

rule mkout:
    input: "in/{sample}.txt"
    output: "out/{sample}.txt"
    # Sleep to allow time to cancel
    shell: "sleep 10000 && cp {input} {output}"
```
</details>

1. `snakemake --slurm --default-resources slurm_account=[SLURM ACCOUNT] slurm_partition=[SLURM PARTITION] runtime=40 nodes=1 cpus_per_task=1 mem_mb=7000 qos=normal -j 20`
2. Wait for jobs to start running in Slurm
3. Press Ctrl-C
4. Output will be something like:
```
^CTerminating processes on user request, this might take some time.
(several minutes of waiting)
Complete log: .snakemake/log/2023-02-15T130415.436417.snakemake.log
(snakemake ends)
```
5. `squeue -u $USER` will show that snakemake jobs were not cancelled and are still running (not state=completing).

Trying the same steps but with my patch applied results in the jobs cancelling properly in about 10 seconds.

**Logs**
<details>
<summary>Failed log</summary>

This is an example of the end of a log where `^C` shows the point of exiting snakemake. The program does not print any output for "this might take some time" and then starts printing sacct status output. After waiting for a few minutes, snakemake exits without canceling the jobs.

```
The sacct output is: '47843535|RUNNING
47843535.batch|RUNNING
47843535.extern|RUNNING
'
^CTerminating processes on user request, this might take some time.
The sacct output is: '47843536|RUNNING
47843536.batch|RUNNING
47843536.extern|RUNNING
47843536.0|RUNNING
'
The sacct output is: '47843537|RUNNING
47843537.batch|RUNNING
47843537.extern|RUNNING
47843537.0|RUNNING
'
The sacct output is: '47843538|RUNNING
47843538.batch|RUNNING
47843538.extern|RUNNING
47843538.0|RUNNING
'
The sacct output is: '47843539|RUNNING
47843539.batch|RUNNING
47843539.extern|RUNNING
47843539.0|RUNNING
'
The sacct output is: '47843540|RUNNING
47843540.batch|RUNNING
47843540.extern|RUNNING
47843540.0|RUNNING
'
Complete log: .snakemake/log/2023-02-14T182901.208983.snakemake.log
unlocking
removing lock
removing lock
removed all locks
```
</details>

**Additional context**

slurm 21.08.8-2

yamanq:
### Description

When using --slurm, exiting snakemake by SIGINT or ctrl-c helpfully cancels spawned jobs. However, this is quite unreliable as it often hangs for a few minutes and exits without canceling the jobs in the end (without indicating that it failed either).

Slurm documentation on [scancel](https://slurm.schedmd.com/scancel.html#SECTION_PERFORMANCE) notes that a large number of scancel calls at the same time may result in denial of service.

Snakemake runs scancel on each job [individually](https://github.com/snakemake/snakemake/blob/main/snakemake/executors/slurm/slurm_submit.py#L136). Instead, job ids should be collected and cancelled all at once.

fixes #2113

### QC

### DIFF
--- a/snakemake/executors/slurm/slurm_submit.py
+++ b/snakemake/executors/slurm/slurm_submit.py
@@ -134,22 +134,23 @@ class SlurmExecutor(ClusterExecutor):
         return [" --slurm-jobstep", "--jobs 1"]
 
     def cancel(self):
-        for job in self.active_jobs:
-            jobid = job.jobid
+        # Jobs are collected to reduce load on slurmctld
+        jobids = " ".join([job.jobid for job in self.active_jobs])
+        if len(jobids) > 0:
             try:
                 # timeout set to 60, because a scheduler cycle usually is
                 # about 30 sec, but can be longer in extreme cases.
                 # Under 'normal' circumstances, 'scancel' is executed in
                 # virtually no time.
                 subprocess.check_output(
-                    f"scancel {jobid}",
+                    f"scancel {jobids}",
                     text=True,
                     shell=True,
                     timeout=60,
                     stderr=subprocess.PIPE,
                 )
             except subprocess.TimeoutExpired:
-                logger.warning(f"Unable to cancel job {jobid} within a minute.")
+                logger.warning(f"Unable to cancel jobs within a minute.")
         self.shutdown()
 
     def get_account_arg(self, job):


johanneskoester (approved):

johanneskoester:
Great! I think it will also make sense to do something similar for querying job status, as (if I am not wrong) slurm also allows to query that for multiple jobids, right? Would you be interested in doing that as well?

johanneskoester:
> I am interested in improving the SlurmExecutor (and other Executors) in future PRs:
> 
> * The jobname argument of snakemake does not apply to the executors since the value is not propagated from [here](https://github.com/snakemake/snakemake/blob/c7ea0590c396c67fa5d56042e21f678c20784d3b/snakemake/scheduler.py#L203) to [here](https://github.com/snakemake/snakemake/blob/main/snakemake/executors/slurm/slurm_submit.py#L103).
> * The asynchronous handling of the `self.active_jobs` list does not take into account that cancellation can occur at any time. If snakemake receives SIGINT after [this line](https://github.com/snakemake/snakemake/blob/main/snakemake/executors/slurm/slurm_submit.py#L363) but before the active_jobs list is updated, snakemake will not have any jobs to cancel. I believe this could be fixed by (1) putting the entire `_wait_for_jobs` function inside async lock and (2) using async lock to get the active jobs in `cancel()` but I will need to try this.
> * Like you said, `sacct` is able to query multiple jobs at once which will reduce the load on the SLURM scheduler. I occasionally get snakemake erroring from job status timing out due to the number of calls.
> 
> All of these are issues that I am thinking about, but I would like to work on them in separate PRs after this one is accepted. For now, I added a small change to check that the job list is nonempty before trying to cancel the jobs. Thanks!

Sounds great! One other thing that comes to my mind is that I think that at least some slurm commands can be called with --json (if I remember correctly), which would make the parsing of their output simpler and probably more future-proof.

PR #2128, opened 2023-02-21
perf: Gfal2 remote provider using gfal2-python instead of  gfal2-utils.

maarten-k:
### Description

Added support of gfal2-python library instead of gfal2-utils. Time(wallclock) of a dry run with 100 files on a gridftp  server was reduced by a 7 to 11 fold (from about a minute to 6 seconds!), while the system and user time was also significantly less. This speed-up is caused because no shell->python->gfal2 objects have to be generated for every file (gfal2-utils are  gfal2-python commands wrapped with a command line interface).

If gfal2-python is not installed, it will fall back to gfal2-utils. This makes the code unclear, but this will not break the current installation when upgrading. A deprecation warning is given if gfal2-python is not installed.

### QC

Test case was not written because of lack of generic available server and adding dependencies
* [X] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behaviour or functionalities of Snakemake).
Should not change behaviour


maarten-k:
It is not clear to me why CI / testing (pull_request)   failed. Can someone take a look?

johanneskoester:
Thanks a lot, sounds great! I would vote for ditching the gfal2-utils dependency though. I don't think it is a problem to have a message that gfal2-python needs to be installed. Also, as the new dependency is much faster, I don't want people to silently use the old one without noticing.

maarten-k:
I agree.  (I thought removing code was easier than introducing backward compatibility). I will run this code in production next week to verify it really works.

maarten-k:
I tested the code and I did not introduce bugs while removing the gfal-utils code.

johanneskoester (approved):

PR #2132, opened 2023-02-22
fix: ensure user and group rw permissions for metadata files and source cache

johanneskoester:
This way, multiple users can more conveniently work on the same workdir.

### QC


dlaehnemann (approved):
LGTM.

Although I don't have a clue if this is all the places where this might need fixing... :grimacing:

Issue #990, opened 2021-05-11
Provide support for Kubernetes Persistent Volumes


gh-kgangineni:
Provide support to use Kubernetes Persistent Volumes .

For applications which need higher I/O, cloud storage like s3 will be slow. 

Instead of this please provide support for using EBS , EFS (AWS) volumes.

PR #2148, opened 2023-02-28
feat: Support for Kubernetes Persistent Volumes

LarsStegemanGT:
### Description
This PR adds the option "--persistent-volume-claim [NAME]" to snakemake. It was requested in #990 
When the persistent-volume-claim option is used the spawned kubernetes pods will have an additional mount on the filesystem of the pod. Using this it is no longer needed to download/upload the files from/to a remote provider for each job.

Because the initial snakemake process has to access these files as well (when generating the DAG) another remote provider is added "LocalDirRemoteProvider".
This remote provider basically adds a way to prefix all the path in the snakefile. The prefix is also passed on to the pod creation and mounted on the same location. 

I didn't add a test case or any documentation yet. I'm curious what you think of this feature.

### QC


johanneskoester:
I love it! Thanks a lot!
One thing: I would like to use minikube on github actions to test the kubernetes backend, instead of gcloud as it is now. Would you be interested in adding that along with adding a test case for this PR?

LarsStegemanGT:
Sure, i can give it a try

colingarvey:
Hello. 

This would be an awesome feature-add, as we are trying to use a parallel filesystem for speeding up execution in our EKS environment by avoiding the staging in of data from S3.

johanneskoester:
This is part of the executor plugin for kubernetes since a while: https://github.com/snakemake/snakemake-executor-plugin-kubernetes/blob/c15750a7364c54d59fadd5ee186c9b8e20f46b54/snakemake_executor_plugin_kubernetes/__init__.py#L88

Thanks a lot for this initial work!

PR #2152, opened 2023-03-06
fix: prevents DeprecationWarning caused by using old  draft of json schema

maarten-k:
### Description
Prevents DeprecationWarning caused by using an old draft of JSON schema when running tests. updated JSON schema and use https.  

### QC

### Context

Run into this issue while sorting out documentation about testing.


johanneskoester (approved):

johanneskoester:
Thanks!

PR #2167, opened 2023-03-10
fix: slurm batch job status queries

dlaehnemann:
### Description

This is a follow-up pull request to #2136 , providing a more thorough solution to putting less strain on the slurm cluster system with job status queries.

The solution suggested here:
* assigns the same UUID4 as the job name during `sbatch` job submission for all jobs during a run of snakemake -- this goes back to [an idea by Sean Maxwell on the `slurm-users` mailing list](https://lists.schedmd.com/pipermail/slurm-users/2023-February/009713.html)
* uses `sacct` as the job status query command, as this [will put less strain on `slurmctld` and instead query the `slurmdbd`](https://github.com/snakemake/snakemake/pull/2136#issuecomment-1442618490) and we assume that any reasonable cluster setup will have at least basic [job accounting](https://slurm.schedmd.com/accounting.html) set up
* uses `sacct --name <uuid4>` to make more efficient database queries to `slurmdbd`, according to [a suggestion by @wickberg](https://github.com/snakemake/snakemake/pull/2136#issuecomment-1446757458)
* adaptively adjusts the minimum wait time between job status query calls (although this only applies to retries when previously seen jobs are somehow not present -- this usually shouldn't happen, and usually queries are spaced by at least 20 seconds, see below), which will be the maximum (longer waiting time period) from either (this also goes back to [an idea by Sean Maxwell on the `slurm-users` mailing list`](https://lists.schedmd.com/pipermail/slurm-users/2023-February/009713.html)):
  *  the specified rate limit (default now set to 2 seconds between calls, but users can set this differently), or
  * 5 times the query time by the last call to `sacct`
* increase the sleep time between checking on all the running jobs by 10 seconds every time no job had finished at the last check; reset to a minimum wait time of 20 seconds once a job finishing was detected; cap at a maximum sleep time of 180 seconds

The solution explicitly DOES NOT use:
* `squeue --noheader --format=\"%F|%T\" --name <uuid4> --jobs <job_ids_of_jobs_with_no_sacct_status>` as the fallback job status query command any more, because we assume basic accounting is always set up on a cluster where workflows would be run (it would be better than `scontrol show job`, because it can also do the `--name` based filtering)
* either the `--yaml` or `--json` flags of slurm commands, as these seem to require (optional) plugins to be installed and they weren't on the system I tested -- so we cannot assume these to usually work. (This came up as an [idea by Bas van der Vlies on the `slurm-users` malining list](https://lists.schedmd.com/pipermail/slurm-users/2023-February/009728.html)
* use some kind of mechanism to wait for jobs finishing (this came up as an [idea by Brian Andrus on the `slurm-users` mailing list](https://lists.schedmd.com/pipermail/slurm-users/2023-February/009724.html), because the only easy solution I saw, was to have [`sbatch --wait`](https://slurm.schedmd.com/sbatch.html#OPT_wait) until a job finishes and only return then. However, this would keep a thread running for every job currently active, likely overstraining the login node in cases of high parallelism in the workflow (or oversubmissions).

### QC

### DIFF
--- a/snakemake/executors/slurm/slurm_jobstep.py
+++ b/snakemake/executors/slurm/slurm_jobstep.py
@@ -28,7 +28,7 @@ class SlurmJobstepExecutor(ClusterExecutor):
         printshellcmds=False,
         cluster_config=None,
         restart_times=0,
-        max_status_checks_per_second=1,
+        max_status_checks_per_second=0.5,
         **kwargs,
     ):
         # overwrite the command to execute a single snakemake job if necessary
--- a/snakemake/executors/slurm/slurm_submit.py
+++ b/snakemake/executors/slurm/slurm_submit.py
@@ -313,7 +313,9 @@ class SlurmExecutor(ClusterExecutor):
                     pd.read_csv(
                         StringIO(
                             command_res
-                        )
+                        ),
+                        header=None,
+                        delimiter="|",
                     )
                 )
             except pd.errors.EmptyDataError:
@@ -383,7 +385,9 @@ class SlurmExecutor(ClusterExecutor):
                         # -X: only show main job, no substeps
                         f"sacct -X --parsable2 --noheader --format=JobIdRaw,State --name {self.run_uuid}"
                     )
+                    logger.debug(f"status_of_jobs after sacct is: {status_of_jobs}")
                     no_sacct_status = active_jobs_ids - set(status_of_jobs.keys())
+                    logger.debug(f"no_sacct_status after sacct is: {no_sacct_status}")
                     if not no_sacct_status:
                         break
                     else:
@@ -395,9 +399,11 @@ class SlurmExecutor(ClusterExecutor):
                             # * %T: long job status
                             f"squeue --noheader --format=\"%F|%T\" --name {self.run_uuid} --jobs {job_ids_string}"
                         )
-                        status_of_jobs.update(squeue_stati)
+                        status_of_jobs = status_of_jobs | squeue_stati
+                        logger.debug(f"status_of_jobs after squeue is: {status_of_jobs}")
 
-                        remaining_without_status = active_jobs_ids - status_of_jobs
+                        remaining_without_status = active_jobs_ids - set(status_of_jobs.keys())
+                        logger.debug(f"remaining_without_status after squeue is: {remaining_without_status}")
                         if not remaining_without_status:
                             break
                         else:
--- a/snakemake/executors/slurm/slurm_submit.py
+++ b/snakemake/executors/slurm/slurm_submit.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 from functools import partial
+from io import StringIO
 import os
 import re
 import stat
@@ -10,6 +11,8 @@ import shutil
 import subprocess
 import tarfile
 import tempfile
+import uuid
+import pandas as pd
 
 from snakemake.jobs import Job
 from snakemake.logging import logger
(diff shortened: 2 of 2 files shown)


johanneskoester (approved):

johanneskoester:
@cmeesters do you want to have a look at this as well?

dlaehnemann:
Caution, this is still a work in progress. Through testing on a slurm cluster locally, I have identified issues that still remain with the code, but that the CI setup does not catch. So I turned it into a draft for now.

But **reviews are already welcome**!

cmeesters:
@johanneskoester sure – just need some time, will not get to it today (school is closed)

@dlaehnemann at first I thought `--yaml` and `--json` for SLURM commands? What is this?? And this illustrates your point exactly well: we admins do update if there are security issues or a new system or a killer feature. In all other cases, a batch system silently ages. So yes, we may safely assume, that NOT all current clusters support this feature. However, it raises the question of how we should deal with this in the future. (NOT in this thread, but elsewhere: How backwards compatible shall snakemake be to dependencies, such as SLURM - regardless of whether we need a particular feature? IMO the minimum requirement may be a version which introduces security fixes. Anyone not supporting those may lose features of snakemake - and we know, that security fixes rarely introduce new feature or lead to API changes.)

dlaehnemann:
@wickberg, as you commented on the initial basic fix for snakemake's job status queries, would you maybe find a couple of minutes to take a quick peek at the more thorough solution provided in this pull request. Your feedback on whether the new mechanisms make sense and whether the defaults are good, would be very welcome. And should you need more info or prefer a quick call to walk through the code, feel free to ping me directly.

dlaehnemann:
And also @cmeesters, no worries, we'll wait for your input. I also asked for more feedback on the `slurm-users` mailing list. So let's see if we can further improve this suggestion or if we are good to go in the coming days. But a couple of extra pairs of eyes definitely don't hurt on this, as the CI can currently only test so much...

many-states:
This looks good. I only see one potential edge case in the changes, where the original version parsed job state as the first word, i.e. "JobState=(\w)" in the output, but in the new version it appears to take the entire value delimited by |. Job states for cancelled jobs may include a suffix, e.g. JobState="CANCELLED by 123" indicating the uid of the user who cancelled the job, so not parsing the first word out would cause them to miss the set of error states defined. If I'm reading it wrong, I apologize. Again, nice work.

dlaehnemann:
Ah, thanks for taking the time to look through the code in such detail!

I had actually missed that parsing part in the original code, so thanks for highlighting it. I went and double-checked, and the parsing was only done for the `scontrol show jobid -dd {jobid}` command, which had rather verbose output that needed this regex parsing. Luckily, we don't have to use that command any more, but instead rely on `sacct -X --parsable2 --noheader --format=JobIdRaw,State --name {self.run_uuid}`. This should always yield something like `12345|FAILED`.

However, if I read [the documentation for the `State` Job Accounting Field](https://slurm.schedmd.com/sacct.html#OPT_State) correctly, you are right that even the `State` in this `--format=JobIdRaw,State` could contain further info. So our command could return `12345|CANCELLED by 123`. I haven't seen this, but then I only tested on one cluster. So the defensive way to deal with this is probably to do the first word regex parsing, after all. Will add a commit in a sec.

cmeesters:
@many-states I agree with @dlaehnemann! Very well spotted. This suggestion will cover a very nasty corner case.

Also, excellent work by @dlaehnemann. If only the CI hiccups (connectivity issues and such) will disappear such that this PR gets into main!

johanneskoester:
CI problems fixed, this should now pass.

PR #2169, opened 2023-03-13
feat: Tes auth

vschnei:
### QC


docs/executing/cloud.rst
```diff
@@ -438,3 +441,16 @@ you can forward your token by setting the `TES_TOKEN` environmental variable.
...
     $ export TES_TOKEN=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
+
+**Funnel basic authentication:** 
+In order to execute individual tasks for Funnel based TES servers, a basic authentication is required .
+An authentication via AOuth2 access token is not supported yet.
+
+.. code-block:: console
+
+    $ snakemake \
+        --tes $TES_URL --tes_user $FUNNEL_USER --tes_password $FUNNEL_PASSWORD\
```
johanneskoester, line 452: I would not pass credentials like this (cumbersome cause it needs to be repeated, can lead to leaking passwords via bash history). Rather, the credentials should be given as env vars.
Suggestion: `TES_USER` and `TES_PASSWORD`. Or maybe funnel already has some expected env vars for this to be used?

johanneskoester (changes requested):
Thanks! One thing below

docs/executing/cloud.rst
vschnei: That make sense.

Nobody should be able to see the password!

johanneskoester (approved):

johanneskoester:
Ci issues are fixed, but the TES tests fail now.

PR #2174, opened 2023-03-16
feat: proof of concept for flux operator executor

vsoch:
### Description

This is a proof of concept to add the [Flux Operator](https://github.com/flux-framework/flux-operator) executor. The Flux Operator is a Kubernetes operator that submits jobs to a Flux "minicluster" - or a Flux cluster running in networked pods on Kubernetes.

The basic (starting) idea is that we dynamically create and submit jobs to one or more MiniClusters as we go, either re-using or spinning up new ones based on unique container images + nodes requested. This is a very basic algorithm that has huge room for improvement - for example it might make sense to calculate an "optimized" set of MiniClusters before based on the DAG, and visually show the user before launching anything. It could be we do the same, but instead only allow one MiniCluster to run at a time. We likely want to add a node limit given the size of the Kubernetes cluster. There is a short asciinema of (what is the test added here) running snakemake with the flux operator executor: https://asciinema.org/a/567278?speed=2

It's also a WIP because the model (submitting locally) assumes that the same files are found in a mounted volume on storage. We'd need to figure out a way to run from a local machine and do checks without straining the kubernetes API. It would also largely depend on the kind of storage being used (some filesystem mount vs. object storage, etc).

I tried to walk through my ideas for some of the "calculate the set of MiniClusters we need" from the DAG here: https://github.com/alpha-unito/streamflow/issues/88. It would be great to be able to think about this with others, because arguably the idea could be extended to other workflow tools. We should only ask for what we need, spin it up when we need it, and bring it down when we are done. 

### QC


johanneskoester:
Very interesting,thanks a lot! Also thanks for your mail and sorry that I did not yet find the time to answer that. I'll just answer right here on both:

I like the minicluster idea a lot, and indeed one can think about ways to optimize the scheduling and the choice based on the topology and the avaialble miniclusters.

I think one should make use of two things:
1. We can extend the interplay between executor and scheduler such that the scheduler can ask the executor about hints which jobs it would prefer for the next round of submissions. Given that your executor currently has a minicluster running, it could check the job resource requirements and mark all those as preferred that would work well on the currently active minicluster(s).
2. The executor already can access the dag (self.dag). It can use that to check which resource requirements to expect. For this, you can explore the dag structure via self.dag.dependencies (a dict of jobs and their dependencies, resolved by individual files). Or you can access self.dag.toposorted() to get jobs in topological sorting (as iterator over levels, each level is a list of jobs that could run in parallel). One thing to keep in mind: when the workflow involves checkpoints, the dag will be updated after those have been executed. Hence, the executor should repeat its decision about miniclusters whenever a checkpoint job is finished.

vsoch:
@johanneskoester I've updated the flux executor to have more of an algorithm that figures out scheduling apriori, specifically in the documentation:

```rst
How does it work?
:::::::::::::::::

MiniClusters are unique based on the container base needed, and the number of nodes for a specific level,
where the level include a set of jobs that can be run in parallel. Since we are always paying for the
Kubernetes cluster, no matter what, we will do best to optimize its usage, meaning running as many
jobs in parallel at once as we can. For this reason, we use a simple algorithm that calculates
the maximum number of nodes that might be used at a single level (up to the max size of your Kubernetes cluster)
and use that paired with the container image to define a unique ID for a cluster. This means that when you are 
using the Flux Operator executor you should:

 - Have a container image defined for every step.
 - Be aware of the different container bases you are using, and (for small or quick jobs) try to share them.
 - Don't be afraid to increase the number of jobs to a high number, as the flux scheduler on the MiniCluser is going to manage things.
 - Define your Kubernetes cluster size by exporting ``SNAKEMAKE_FLUX_OPERATOR_MAX_NODES``

This is a fairly simple workflow, meaning that it uses a single container and is able to run
all the jobs on the same MiniCluster.  
```

I was wondering if (for a next step) you could point me to a workflow (with some dummy data) that uses multiple different container images, and possibly has a wider range of resource requrements. I'd like to test this in a more complex context.

johanneskoester:
Unfortunately I don't have such a test workflow ready for you, sorry.

Apart from that, two questions:
1. What about the idea that the scheduler asks the executor for additional hints which jobs to prefer among the set of ready jobs? Wouldn't this help here, by allowing basically one minicluster to stay as busy as possible? The change would be very simple, I could do that in a separate PR so that you can directly use it.
2. I am a bit unsure about the handling of container images here. Every Snakemake job has to run in the Snakemake container, which will internally start singularity for launching container images that might be annotated in the rule. The reason is all the stuff that Snakemake does around the actual shell command (e.g. remote file handling). So, I would have said that by definition the container is always the same, but inside that container, different ones are launched.

vsoch:
> What about the idea that the scheduler asks the executor for additional hints which jobs to prefer among the set of ready jobs? Wouldn't this help here, by allowing basically one minicluster to stay as busy as possible? The change would be very simple, I could do that in a separate PR so that you can directly use it.

That would work for me! I think I assumed that jobs in the same level would be launched at once - e.g., all given to a MiniCluster scheduler to use, and then I also assumed that all jobs in the same level would share the same container base / resource needs and snakemake would already provide them to launch quickly in sequence.

> I am a bit unsure about the handling of container images here. Every Snakemake job has to run in the Snakemake container, which will internally start singularity for launching container images that might be annotated in the rule. The reason is all the stuff that Snakemake does around the actual shell command (e.g. remote file handling). So, I would have said that by definition the container is always the same, but inside that container, different ones are launched.

I did hijack the container field a bit! :laughing:  I'm not sure we are able to run containers _inside_ MiniCluster containers, hence my approach.It would either have to be like the Google Life sciences executor with only one container used (as a base container for the job) and then conda used, and then don't allow any containers, or this approach that is flexible to run the step inside any chosen container. We definitely can't launch singularity or docker, however.  If we do the first and then choose a default base container, this would also be something to maintain (it currently requires flux to be installed and would also need snakemake).

johanneskoester:
I believe that there are container runtimes around that we could add in addition to singularity which allow to run inside other containers without priviledged mode. I remember that I had reviewed a paper about that some time ago. Do you happen to know which runtimes would be suitable? If not, I will try to dig up that paper.

johanneskoester:
I think it boils down to enroot or udocker. On the one hand, Udocker is already on anaconda.org. On the other hand, enroot seems to support GPU computing more out of the box (at least for nvidia).

vsoch:
okay - I'll do some experiments soon to try these out.

vsoch:
Is there a way to tell it to make the conda envs as it goes? When I have the local bind and launch locally, it's trying to create them on my host (not desired!) I got pretty far with this refactor but I'm stuck here.

```
$ snakemake --cores 1 --jobs 1 --flux-operator --use-conda
Building DAG of jobs...
Your conda installation is not configured to use strict channel priorities. This is however crucial for having robust and correct environments (for details, see https://conda-forge.org/docs/user/tipsandtricks.html). Please consider to configure strict priorities by executing 'conda config --set channel_priority strict'.
Creating conda environment environment.yaml...
CreateCondaEnvironmentException:
Could not create conda environment from /tmp/workflow/environment.yaml:
Command:
mamba env create --quiet --file "/tmp/workflow/.snakemake/conda/4675c636da88fd40566d02f5d6d00f77_.yaml" --prefix "/tmp/workflow/.snakemake/conda/4675c636da88fd40566d02f5d6d00f77_"
Output:
  File "/home/vanessa/anaconda3/bin/mamba", line 11, in <module>
    sys.exit(main())
  File "/home/vanessa/anaconda3/lib/python3.9/site-packages/mamba/mamba.py", line 848, in main
    from conda.common.compat import ensure_text_type, init_std_stream_encoding
ImportError: cannot import name 'init_std_stream_encoding' from 'conda.common.compat' (/home/vanessa/anaconda3/lib/python3.9/site-packages/conda/common/compat.py)

```

vsoch:
@johanneskoester how long does the conda environment take to create? It's been going over 20 minutes...
```
Removing incomplete Conda environment environment.yaml...
Creating conda environment environment.yaml...

```
Is there something wonky going on possibly?

vsoch:
Ok I've narrowed this down to filesystem issues with MiniKube - when it's bound to the host it's too slow to make it feasible to do anything. When I remove the bind, I lose permissions to execute binaries (saw this bug yesterday too). I'm not sure what to do because I need this small tutorial to work for the docs - I don't want to require an actual cloud. I could likely try kind but I suspect it might have similar issues.

vsoch:
okay tested with singularity (added priviledged to cluster) but doesn't seem to work 
image

vsoch:
I suspect I'll run into similar with docker (container in container). Anyway I'll open another PR with these updates - there are quite a few - and we can pick up conversation.

vsoch:
This is replaced by the kueue plugin, which now has support for using the flux operator (and MiniClusters) https://github.com/snakemake/snakemake-executor-plugin-kueue

Issue #1591, opened 2022-04-21
Additional methods to declare local rules


troycomi:
**Is your feature request related to a problem? Please describe.**
This is related to issue #945 and a post I made on [stackoverflow](https://stackoverflow.com/questions/71912278/named-rules-and-local-rules).  Basically the current implementation of listing all `localrules` in a separate directive is challenging for modules and using "anonymous" rules that are named with the `name` directive.

**Describe the solution you'd like**
Adding another rule directive would address both issues:
```python
for key, url in config['urls'].items():
    rule:
        name: f'download_{key}'
        localrule: True
        output: config['paths'][key]
        params: url=url
        shell: # do download

# example with modules
use rule some_task from other_workflow as other_some_task with:
    output:
        "results/some-result.txt"
    localrule: True
```

In either case, placing the information that a rule is local closer to the actual rule makes the workflow more readable in my opinion. The current `localrules` directive should remain for compatibility and as a separate method to set localrules. 

**Describe alternatives you've considered**
You can list local rules that don't exist, so both issues can be worked around.  In the case of anonymous rules, you need to give them a name and manually add it to the `localrules` list.  You can modify the `workflow._localrules`, but that's bad form.

**Additional context**
The `name` directive isn't very clearly presented in the docs and should probably be expanded.

fgvieira:
`localrule` directive added on #2180. Can you test if it works for you?

troycomi:
Nice, thank you!  Can you also add it to the syntax files for vim and nano?

PR #2180, opened 2023-03-20
feat: added localrule directive

### LINKED ISSUE #945: Localrules not compatible with new module mechanism
**Snakemake version**
6.1.1

**Describe the bug**
Rules defined as ``localrules`` aren't localrules anymore when imported from a module. 

**Logs**
```bash
localrules directive specifies rules that are not present in the Snakefile:
	a
```

**Minimal example**
```Python
# module.smk

localrules: a

rule a:
    output: "something.txt"
    shell: "touch {output}"

rule b:
    input: rules.a.output
    output: "another.txt"
    shell: "touch {output}"

# Snakefile

module other_workflow:
    snakefile: "module.smk"

use rule * from other_workflow as other_*
```

fgvieira:
### Description

Added support for rule directive `localrule`, as described in #1591 and to fix #945.

### QC


johanneskoester (approved):

corneliusroemer:
Does this require changes in snakefmt @mbhall88?

mbhall88:
> Does this require changes in snakefmt @mbhall88?

Yes it will.

corneliusroemer:
@johanneskoester would be great if you pinged snakefmt if something requires changes - I just stumbled over this when reading through changelog, which I don't always do. Would be good to be proactively ping so that changes are ready when the version comes out.

mbhall88:
@corneliusroemer we would have detected it next time we update the minimum pin for the snakemake (dev) dependency. We have a test that checks that all ruleparams etc. are covered by snakefmt.

PR #2193, opened 2023-03-24
feat: add flux operator test single container

vsoch:
### Description

This is a continuation to #2174 to (as suggested) fall back to the original model to use a common base container, and allow directives on top of that (e.g., conda).

Update: singularity and conda both work! I can try udocker (although I've never used it before, and definitely not in the context of snakemake). @johanneskoester this is probably ready for another look! Let me know if you want me to try udocker.

### QC


snakemake/__init__.py
```diff
@@ -2486,6 +2493,25 @@ def get_argument_parser(profile=None):
...
+        default="flux-operator",
+        help="Namespace for the Flux Operator MiniCluster(s).",
+    )
+    group_flux_operator.add_argument(
+        "--container-image",
+        metavar="IMAGE",
+        help="Docker image to use, e.g., when submitting jobs to the Flux MiniCluster"
+        "on Kubernetes. It must contain a Flux install and have active user root. "
+        "Defaults to ghcr.io/rse-ops/mamba:app-mamba",
+    )
```
johanneskoester, line 2513: There is already the --container-image argument, which can be also used for the flux operator.

snakemake/script.py
```diff
@@ -617,6 +617,8 @@ def get_preamble(self):
         )
 
     def write_script(self, preamble, fd):
+        print(preamble)
+        print(self.source)
```
johanneskoester, line 621: I guess these and below were for debugging? Can you remove them?

snakemake/workflow.py
```diff
@@ -527,6 +527,8 @@ def execute(
         kubernetes=None,
         k8s_cpu_scalar=1.0,
         flux=None,
+        flux_operator=None,
+        flux_operator_ns=None,
```
johanneskoester, line 531: if flux_operator is combined with use_singularity, this should log a warning mentioning that it requires the cluster to run in priviledged mode, and that depending on the settings this might fail? Should also be mentioned in the docs on Snakemake's flux operator backend.

johanneskoester (changes requested):
Awesome! Just a few things below. Regarding udocker, I guess that can be done in a separate PR then.

vsoch:
@johanneskoester I think I'd like to test it in this PR - it's fairly important to get working! And I definitely don't mind having this one open a little bit longer.

I haven't used it before but I'll look at it tonight!

snakemake/__init__.py
vsoch: I tried using the same name, but it got angry at me. Are you saying to remove it from the kubernetes group and add to the flux operator group? or is it possible to define the same param for two groups?

vsoch:
Will pick up on this tomorrow / later this week - I was testing using kind over minikube but I'm not sure the difference in performance is significant.

vsoch:
@johanneskoester an option to get around this bind (and the permissions issues, which weirdly I haven't seen locally) would be to have a directive that allows the user to specify a snakemake workflow (in GitHub) to download to the operator - this would be done as a pre command. The con then is that the output wouldn't be saved to the filesystem. We probably need to think harder about the ideal execution modes because likely someone will want to run the workflow from GitHub, or some cloud storage mount.

.github/workflows/test-flux.yaml
```diff
@@ -1,9 +1,78 @@
...
 jobs:
-  build:
+  test-flux-operator-executor:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Note that we cannot test conda because the bind of the filesystem is too slow
+        commands: [["singularity", "cp -R ./tests/test_singularity_module/* /tmp/workflow/", "snakemake --cores 1 --jobs 2 --flux-operator --use-singularity"],
+                   ["vanilla", "", "snakemake --cores 1 --jobs 1 --flux-operator --minicluster-container-image ghcr.io/rse-ops/atacseq-vanilla:app-latest"]]
```
johanneskoester, line 13: I think the option should be called --flux-minicluster-container-image. And can you remind me again in what sense this differs from the ordinary --container-image option? Do we really need a separate one?

johanneskoester (commented):
If singularity works now, implementing --use-udocker  as an alternative to --use-singularity does not have a priority for me, and is definitely separate of this PR.

An additional question is below.

.github/workflows/test-flux.yaml
vsoch: I only created a separate one because the first container image was in a different argument group. Do you have a suggested way of implementing this?

vsoch:
@johanneskoester question for you - are there specific snakemake workflows that benchmark something (tools or similar?) that the larger bio-sciences community is interested in? I'm putting a plan for a paper for the flux operator and it would be interesting to run such a workflow to hit that target audience. Pinging @pditommaso in case he knows of similar ones for nextflow.

.github/workflows/test-flux.yaml
johanneskoester: Just move the --container-image argument into group_exec and adapt the help text of --container-image.

johanneskoester:
> @johanneskoester question for you - are there specific snakemake workflows that benchmark something (tools or similar?) that the larger bio-sciences community is interested in? I'm putting a plan for a paper for the flux operator and it would be interesting to run such a workflow to hit that target audience. Pinging @pditommaso in case he knows of similar ones for nextflow.

You mean example workflows that run long enough to not be just toy examples and of general enough nature to be interesting for many people? I guess there are plenty of such workflows. I might have one for you. We are in the process of publishing a paper about a benchmarking initiative where we also will publish such a workflow. It will be made publicly available in the coming days or at most weeks. Would that help?

vsoch:
That would be absolutely perfect! We are just setting up our cloud spaces this week so a week or even a few would be great timing. What should I follow to be notified of the workflow being made public?

johanneskoester:
@vsoch the workflow is now public and the paper will be out soon. You find the homepage with results and in particular links to the workflow source code here: https://ncbench.github.io.
Feel free to contact me by discord or mail in case there are issues or something is unclear.

vsoch:
Fantastic! Also - I've been doing work with elasticity in the operator (https://flux-framework.org/flux-operator/tutorials/elasticity.html and full story https://vsoch.github.io/2023/elasticity/) and I think this might be a better design for Snakemake. E.g., the implementation would be similar to as it is here, but instead of controlling creation / destruction of MiniClusters, it would request resizing. I've just started experiments with resizing Kubernetes too, although that is more vendor specific so likely would be harder to add here.

Question for you - have you thought about allowing something like an executor plugin? To me these implementations seem too experimental to add to snakemake, but it would be really cool if I could maintain them in a separate repository, and then have them be easily used by Snakemake (and the maintainer logic separated, removing the burden from you here). What do you think?

johanneskoester:
> Fantastic! Also - I've been doing work with elasticity in the operator (https://flux-framework.org/flux-operator/tutorials/elasticity.html and full story https://vsoch.github.io/2023/elasticity/) and I think this might be a better design for Snakemake. E.g., the implementation would be similar to as it is here, but instead of controlling creation / destruction of MiniClusters, it would request resizing. I've just started experiments with resizing Kubernetes too, although that is more vendor specific so likely would be harder to add here.
> 
> Question for you - have you thought about allowing something like an executor plugin? To me these implementations seem too experimental to add to snakemake, but it would be really cool if I could maintain them in a separate repository, and then have them be easily used by Snakemake (and the maintainer logic separated, removing the burden from you here). What do you think?

Yes, plugins for executors are indeed a great idea. Some students have worked on that in the past but the implementation never made it into the main branch. I would be happy to merge a PR that adds plugin functionality for executors! I think there are several good plugin frameworks out there for python. Feel free to choose the one you think is best and provide a PR! Thanks a lot!

vsoch:
This is replaced by the kueue plugin, which now has support for using the flux operator (and MiniClusters) https://github.com/snakemake/snakemake-executor-plugin-kueue

PR #2194, opened 2023-03-27
Sql remote

baffelli:
### Description

Added a simple remote for SQL databases. This enables users to specify dependencies on a relational database table. 

### QC

### DIFF
--- a/snakemake/remote/SQL.py
+++ b/snakemake/remote/SQL.py
@@ -1,0 +1,360 @@
+__author__ = "Simone Baffelli"
+__copyright__ = "Copyright 2022, Simone Baffelli"
+__email__ = "simone.baffelli@gmail.com"
+__license__ = "MIT"
+
+import os
+import re
+import time
+from pathlib import Path
+from re import Match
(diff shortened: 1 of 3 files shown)


johanneskoester (commented):
Interesting idea! I wonder whether this is better than just accessing the DB via pandas from a script. What is your opinion on that?

baffelli:
> Interesting idea! I wonder whether this is better than just accessing the DB via pandas from a script. What is your opinion on that?

Thanks. The idea is exactly that users are free to access the DB using any API they like. The idea is just to have changes in the DB tracked in snakemake as part of the DAG. This way, when new rows are appended or when a table is modified, the jobs taking that table as an input are run again.

Depending on the database, the "metadata" on a table (the modification time) can be taken from `information_schema.tables` or from an user supplied query. I prototyped both solutions, but I am happy for inputs if someone has a better API.

Let me know whether this is something interesting for you.

baffelli:
Any interest in this? It is been a while since I opened the PR

Issue #2197, opened 2023-03-28
long account names prevent jobs from executing under slurm


jgbaum:
**Snakemake version**
7.25.0

**Describe the bug**
When a slurm account name is longer than 20 characters, it is impossible to submit jobs using slurm.  Snakemake seems to query slurm for valid account names and names longer than 20 characters are truncated with a '+' at position 21.  If I don't specify my account name, snakemake will use the truncated version (dept-bioinformatics+) and get rejected by slurm.  If I do specify my full account name with ```slurm_account=dept-bioinformatics-facility```, it is rejected by snakemake.

**Minimal example**
The account name that my user is assigned to is 'dept-bioinformatics-facility'.  When I run snakemake with slurm, without specifying the account, I get:
```
No SLURM account given, trying to guess.
Guessed SLURM account: dept-bioinformatics+
WorkflowError:
SLURM job submission failed. The error message was sbatch: error: Batch job submission failed: Invalid account or account/partition combination specified
```

If I do specify my full account name in my profile, I get:
```
WorkflowError:
The given account dept-bioinformatics-facility appears to be invalid. Available accounts:
dept-bioinformatics+
```

And for good measure, if I explicitly specify the truncated version I get a snakemake error...although maybe I'm not specifying it properly:
```
Error:
  WorkflowError:
    Failed to evaluate default resources value 'dept-bioinformatics+'.
```

FYI, here is how the truncated version is specified in my config:
```yaml
default-resources:
  - slurm_walltime=3600
  - slurm_partition=compute
  - slurm_account=dept-bioinformatics+
```

For now, our sysadmin has been kind enough to add a secondary account for my department with a shorter name.  That works for the time being, but this likely should be fixed in the snakemake code base.    I'd be happy to contribute if I could be directed to the relevant section of the code that pulls this info.

jgbaum:
I submitted #2198 to fix the issue with the existing calls to sacct & acctmgr.   It works, but I wanted to throw out there that our sysadmin suggested this might be a preferred way to get the default account, rather than going through the recent jobs in sacct:

```bash
sacctmgr -nP show user user="$USER" format=DefaultAccount
```

PR #2198, opened 2023-03-28
fix: addressing #2197 by allowing 256 character account names in slurm

jgbaum:
### Description

This changes the format string on the calls to 'sacct' and 'sacctmgr' to allow for long slurm account names.  This fix seems to work on our slurm cluster.

### QC

### DIFF
--- a/snakemake/executors/slurm/slurm_submit.py
+++ b/snakemake/executors/slurm/slurm_submit.py
@@ -33,7 +33,7 @@ def get_account():
     tries to deduce the acccount from recent jobs,
     returns None, if none is found
     """
-    cmd = f'sacct -nu "{os.environ["USER"]}" -o Account%20 | head -n1'
+    cmd = f'sacct -nu "{os.environ["USER"]}" -o Account%256 | head -n1'
     try:
         sacct_out = subprocess.check_output(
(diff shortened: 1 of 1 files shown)


dlaehnemann (approved):
Seems like a trivial change to me, where I can't see any negative side effects. So if this fixes the problem locally, this should be fine to merge.

dlaehnemann:
Also, we should follow up on your suggestion for a more thorough fix:
https://github.com/snakemake/snakemake/issues/2197#issuecomment-1487687115

But this fix works for now, and doesn't need any further digging into the slurm docs... 😅

ReubenJ:
Would greatly appreciate this change. Most of TU Delft's accounts are longer than 20 characters ([examples](https://doc.dhpc.tudelft.nl/delftblue/Accounting-and-shares/#eemcs-ewi_1)), meaning we cannot make use of `--slurm` on our cluster. Sorry to bump, but any reason this can't be merged @johanneskoester?

johanneskoester:
Sorry for the delay. This slipped my attention. If something like that happens, always feel free to ping me via discord (username johanneskoester).

ReubenJ:
No worries! Thank you 🙌

PR #2204, opened 2023-04-05
fix: Limit python versions for testing

maxim-h:
### Description

Starting with Python3.10 collections.Callable moved to collections.abc.Callable, which [breaks nose](https://github.com/nose-devs/nose/issues/1122).
 Seems that for now testing can only be done with earlier versions.
 In the long run some other testing solution might be needed.

### QC

### DIFF
--- a/test-environment.yml
+++ b/test-environment.yml
@@ -2,7 +2,7 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - python
+  - python < 3.10 # 3.10 breaks nose
   - yte
   - stopit
   - datrie


johanneskoester:
Actually we use pytest, not nose. It works fine with 3.10. Do we still mention or recommend nose somewhere?

maxim-h:
Ah, ok. 
Then it's just a documentation issue.
https://github.com/snakemake/snakemake/blob/50b8f1609a597dc9f25d2fd86c9cdda531bdc041/docs/project_info/contributing.rst#testing-guidelines

Issue #2021, opened 2022-12-17
Strange behavior when using checkpoints in cloud execution


cademirch:
**Snakemake version**

`7.18.2`
**Describe the bug**

I am trying to run a workflow with a checkpoint in the cloud. The workflow runs as expected when executed locally.  However,  in the cloud (using `--google-lifesciences`) it seems that on the checkpoint aggregating step (see below), the whole DAG is rebuilt within that cloud job. 

**Minimal example**

```
import glob

localrules: first_step, second_step
samples = ["sample_a"]
rule all:
    input: expand("results/{sample}/done.txt", sample=samples)

def get_files(wc):
    """
    Globs files from second_step's output to create input for third_step
    """
    checkpoint_output = checkpoints.second_step.get(**wc).output[0]
    query = os.path.join(checkpoint_output, "*.txt") # only care about txt files
    files = [os.path.basename(f) for f in glob.glob(query)]
    numbers = [f.replace("_interval.txt", "") for f in files]
    
    out = expand("results/{{sample}}/third_step/intermediate_{i}.txt", i=numbers)
    
    return out
    
rule first_step:
    output: "results/{sample}/hi.txt"
    shell: "echo 'hi' > {output}"

checkpoint second_step:
    """
    Generates random number of files
    """
    input: "results/{sample}/hi.txt"
    output: directory("results/{sample}/second_step")
    run:
        import random
        import os
        num = random.randint(1,10)
        if not os.path.exists(output[0]):
            os.mkdir(output[0])
        for i in range(num):
            with open(os.path.join(output[0], f"{i}_interval.txt"), "w") as f:
                print(f"this is a message from file interval_{i}.txt", file=f)

rule third_step:
    input: "results/{sample}/second_step/{i}_interval.txt"
    output: "results/{sample}/third_step/intermediate_{i}.txt"
    shell: "cat {input} > {output}"


rule aggregate_step:
    """
    Aggregates files from third_step
    """
    input: get_files
    output: "results/{sample}/done.txt"
    shell: "cat {input} > {output}"
```
run using:
`snakemake --google-lifesciences --default-remote-prefix cade-test -j 10`

This is the stdout/stderr from the aggregate_step job, notice how it rebuilds the DAG
```
--2022-12-17 05:53:54--  https://raw.githubusercontent.com/snakemake/snakemake/main/snakemake/executors/google_lifesciences_helper.py
[...]

PR #2220, opened 2023-04-17
fix: Fix checkpoints rerunning on cloud execution

cademirch:
### Description

When executing in a cloud environment, nodes executing a rule that depends on a checkpoint output rerun the checkpoint  inside the node because the checkpoint output is not available to that node. See #2021 for more details on this issue.

To fix this, in `checkpoints.Checkpoint.get`, we check if the checkpoint output exists remotely and does not exist locally, if both of these are true then download the checkpoint output. In order to test this I build a container of my development version of Snakemake here: `cademirch/snakemake:fix-2021`. Using this container for cloud execution fixes this issue. 

Note: This PR does not address the issue of duplicated remote prefixes returned from input functions (see the issue above for more details). I think this is a separate issue and perhaps even not an issue, but I can open a separate issue or PR to work on that.   

### QC


cademirch:
Seems like there is an issue with setting up slurm on the testing. Also the test_ancient on windows testing failed, but it seems unrelated to these fixes? @johanneskoester

cademirch:
So, I have realized that this PR is likely not needed as this 'issue' is covered in the docs and I completely missed it 🤦. Always read the docs!! 

In the input function example [in the docs ](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#data-dependent-conditional-execution): 

```python
    # decision based on content of output file
    # Important: use the method open() of the returned file!
    # This way, Snakemake is able to automatically download the file if it is generated in
    # a cloud environment without a shared filesystem.
    with checkpoints.somestep.get(sample=wildcards.sample).output[0].open() as f:
        if f.read().strip() == "a":
            return "post/{sample}.txt"
        else:
            return "alt/{sample}.txt"
```

PR #2225, opened 2023-04-20
fix: Fix race condition when creating lock directory

sirno:
### Description

This PR makes two changes:

1. when `--nolock` is set, it will avoid creating a lock directory
2. when multiple instances of snakemake are running there is a race condition on the `os.mkdir` call. Switching to `os.makedirs` and setting `exist_ok=True` will simply avoid raising an exception in this case.

### QC


sirno:
Removing the conditions and using `os.makedirs(path, exist_ok=True)` which is perfectly able to deal with concurrency is probably best.

johanneskoester (approved):

PR #2254, opened 2023-05-09
feat: add support for Kubernetes service account name spec

hnawar:
### Description

Add support for Custom Kubernetes service account which enables the use of GKE Autopilot and other Kuberenetes clusters using Workload identity

### QC

I tested it with a GATK pipeline but will add relevant test cases


hnawar:
Hi @johanneskoester ,
It's been a while since I used snakemake, but I was trying to use it with GKE Autopilot when I ran into some issues. This PR enables the use of GKE autopilot and clusters configured with Workload identity.

Can you let me know your thoughts and I will start adding test cases and documentation.

Please note that there is a limit on ephemeral storage per pod of 10GiB which does limit the pipelines you can run.
This is not an urgent PR

hnawar:
I noticed that docs should be added automatically by the CLI parser, so I'm assuming I don't need to add any more documentation unless you want me to add a note about Kubernetes workload identity.
Are there any examples of tests that would be needed. I've already tested with multiple snakefiles on different clusters using workload identity

johanneskoester (approved):
Thanks a lot!

johanneskoester:
May I ask you to fix the formatting with black?

johanneskoester:
You have to run [black](https://github.com/psf/black) on the source code to get the formatting fixed according to the best practices. This is what the failing test means.

PR #2258, opened 2023-05-14
feat: allow forced local execution regardless of cluster flags

mrvollger:
### Description
This PR adds a `--force-local` command line argument that forces local execution regardless of other flags. This is useful when you have a default profile environment variable but you want to test a workflow locally. This is my first contribution so hopefully I am following all the conventions/directions!

### QC

**I added a test case, but there may be a cleaner way to test.**

**If I understand the code correctly the new command line argument will add its help information to the cli page in the docs.**


snakemake/__init__.py
```diff
@@ -2697,6 +2702,18 @@ def adjust_path(f):
...
+        args.cluster = None
+        args.slurm = False
+        args.slurm_jobstep = False
+        args.cluster_sync = None
+        args.tibanna = False
+        args.kubernetes = None
+        args.tes = None
+        args.google_lifesciences = False
+        args.drmaa = None
+        args.flux = False
```
johanneskoester, line 2715: This is dangerous, as we might miss a future backend here. A safer way would be to pass force_local to the scheduler and let it override all other cluster/cloud statements considered there.

Apart from that: what about just having a second profile for the local execution? Wouldn't that be sufficient?

mrvollger: Hi @johanneskoester, I agree this might be missed in the future. So I just made a commit that integrated it with the run_local variable that seems to be used in `Workflow` that should accomplish the same thing. I am not super familiar with the codebase so I apologize if this is not what you were suggesting. 

I agree a second profile is a pretty reasonable option to get around this, but on the other hand if this solution could be made acceptable I would much prefer a command line argument. 

Thanks!
Mitchell

mrvollger:
Hi @johanneskoester,

The `workflow/profiles/` you recently added removed enough of the boiler plate options I was including with my more complex pipelines that I don't feel the need for this so much anymore, and I agree with some of the risks you pointed out, so I am going to close unless someone objects. 

Thanks for taking a look!

Issue #1710, opened 2022-06-10
Use S3 timestamp metadata to avoid pulling remote data locally when jobs do not need to be run


nh13:
S3 does not preserve timestamps, which affects snakemake's ability to reason about files localized from S3 (see #752).  One solution is to use `rclone` to store last modified time as metadata on an S3 object, and set that appropriately when localizing it again.  Using `rclone` to transfer files to/from S3 allows snakemake to skip already executed jobs, versus thinking files need to be re-generated falsely.  

The question or feature request really is can/does the snakemake S3 remote provider leverage S3 object metadata to preserve and set timestamps when move data to and from S3?  And if not, can we do so? 

Ideally, I could avoid having to sync large data from S3 to a local disk (with `rclone`), just to run snakemake to see if my workflow has to update any outputs (i.e. no jobs to run).  Thnughts?

Issue #752, opened 2020-11-18
Best practice for syncing with Amazon S3 and therefore losing timestamps?


tavinathanson:
Hello! I store all my files in S3 buckets, and run a sync with S3 before and after every Snakemake job.

S3 does not support preserving the original last modified date of files, so whenever I pull files from S3 the timestamps are all based on the time of upload rather than when they were created/modified by Snakemake.

This, of course, prompts Snakemake to think that certain files need to be re-created; it re-creates some arbitrary set of files depending on the order in which the files were uploaded to S3.

To address this, I’ve tried using `ancient` in all my rules, but this is very cumbersome.

Currently, I run a job with `--touch` after pulling the files from S3 before running the actual job. This does the trick, but again feels quite hacky.

I’m curious if there is a best practice for this situation in Snakemake. Is there is a reason that there is no global `--ancient` command, or something to that effect, to tell Snakemake that my timestamps are bogus and to completely ignore them?

nh13:
@tavinathanson I have found using rclone to preserve timestamps works.  Unfortunately, I have to pull the data locally still then run snakemake.  See #1710.

PR #2264, opened 2023-05-18
Add the --ancient command line option.

nh13:
### Description

Motivated by #752 and #1710, snakemake will treat all input files as ancient, thus ignoring timestamps.

### QC

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -123,6 +123,7 @@ class DAG:
         notemp=False,
         keep_remote_local=False,
         batch=None,
+        mark_all_ancient=False,
     ):
         self.dryrun = dryrun
         self.dependencies = defaultdict(partial(defaultdict, set))
@@ -200,6 +201,8 @@ class DAG:
         self.periodic_wildcard_detector = PeriodicityDetector()
 
         self.update_output_index()
+
+        self.mark_all_ancient = mark_all_ancient
 
     def init(self, progress=False):
         """Initialise the DAG."""
@@ -1140,6 +1143,7 @@ class DAG:
                         for f in job.input
                         if f.exists
                         and f.is_newer(output_mintime_)
+                        and not self.mark_all_ancient
                         and not is_same_checksum(f, job)
                     ]
                     reason.updated_input.update(updated_input)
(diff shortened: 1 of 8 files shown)


tests/tests.py
```diff
@@ -228,6 +228,10 @@ def test_ancient():
     run(dpath("test_ancient"), targets=["D", "C", "old_file"])
 
 
+def test_ancient_cli():
```
yfarjoun, line 231: perhaps run all the tests also without the `ancient` flag to "prove" that the output is then generated?

nh13: Done

snakemake/dag.py
```diff
@@ -1140,6 +1143,7 @@ def update_needrun(job):
                         for f in job.input
                         if f.exists
                         and f.is_newer(output_mintime_)
+                        and not self.mark_all_ancient
                         and not is_same_checksum(f, job)
```
nh13, line 1147:
```suggestion
                        if not self.mark_all_ancient 
                        and f.exists
                        and f.is_newer(output_mintime_)
                        and not is_same_checksum(f, job)
```

snakemake/dag.py
```diff
@@ -1132,7 +1135,7 @@ def update_needrun(job):
             if not reason:
                 output_mintime_ = output_mintime.get(job)
                 updated_input = None
-                if output_mintime_:
+                if not self.mark_all_ancient and output_mintime_:
```
nh13, line 1138: could put in the loop for `update_input` below, but it's faster and clearer

nh13:
@johanneskoester all tests pass; I think this is ready for review

johanneskoester (changes requested):
Thanks  a lot! Just a tiny change request below.

nh13:
@johanneskoester done, waiting for tests to pass

nh13:
Closing this PR as the underlying use case is now addressed by the `--consider-ancient` flag added in release 8.25.0 (PR #3171).

While this PR implemented a simpler `--all-ancient` boolean flag, the granular `--consider-ancient RULE=INPUTITEMS` approach on main provides more flexibility for the same timestamp-ignoring use case.

Additionally, this branch is significantly out of date (based on v7.25.4 while main is at v9.14.5) and would require a complete reimplementation rather than a rebase.

Fixes #1

PR #2266, opened 2023-05-18
fix: remote GS builds too many inventories; io:collect_mtime always uses uncached mtime

cademirch:
### Description
This PR includes two small changes that drastically reduce inventory building time when using Google remote files, as well as perhaps unnecessary extra mtime checks.

The `IOCache` class uses [`ExistsDicts`](https://github.com/snakemake/snakemake/blob/6f602a3cceba81b1c2281db3156f9a696bc50b6a/snakemake/io.py#LL106C1-L122C16) to quickly lookup if a file is in the cache or not. Specifically, in https://github.com/snakemake/snakemake/blob/6f602a3cceba81b1c2281db3156f9a696bc50b6a/snakemake/io.py#L119-L122 `ExistsDict` checks if a `parent` of the query file has been inventoried. 

This is problematic, because when we build the inventory for a remote GS file, we add `bucket/subfolder` to `has_inventory` https://github.com/snakemake/snakemake/blob/6f602a3cceba81b1c2281db3156f9a696bc50b6a/snakemake/remote/GS.py#L195
But only return `self.bucket` from `get_inventory_parent` https://github.com/snakemake/snakemake/blob/6f602a3cceba81b1c2281db3156f9a696bc50b6a/snakemake/remote/GS.py#L199-L200

Which leads to many extra inventories being built. We fix this by returning `self.bucket/subfolder`.

The second commit in this PR I am less sure about. In [`IOCache`](https://github.com/snakemake/snakemake/blob/6f602a3cceba81b1c2281db3156f9a696bc50b6a/snakemake/io.py#L125) , we build an [`mtime_inventory` asynchronously ](https://github.com/snakemake/snakemake/blob/6f602a3cceba81b1c2281db3156f9a696bc50b6a/snakemake/io.py#L135-L172). However, for reasons I'm not sure about, each async worker uses https://github.com/snakemake/snakemake/blob/6f602a3cceba81b1c2281db3156f9a696bc50b6a/snakemake/io.py#L174-L175 to get `mtime` info for a file. This is confusing to me as it explicitly uses `mtime_uncached`, which directly checks the `mtime` for a file, and for remote files this means expensive API calls. I'm not sure why `mtime_uncached` is used here, but it would be good to know if this is intentional.

Benchmarking these two changes shows some pretty impressive improvement.

Here is a simple workflow:
```
rule all:
    input: expand("results/{i}.txt", i=range(100))

rule make_file:
    output: "results/{i}.txt"
    shell: "echo {wildcards.i} > {output}"
```
Timed with `GNU time` three times.

Timing (no fix):
```
0:31.33 elapsed
0:35.06 elapsed
0:31.10 elapsed
```

Timing (with fix):
```
0:06.13 elapsed
0:06.12 elapsed
0:03.15 elapsed
```

The improvements are even more significant when testing with some of my larger workflows.

Would appreciate discussion/thoughts on this @johanneskoester and @vsoch and of course anyone else!

### QC


vsoch:
I would guess the timing difference comes down to using a cached value vs. requesting that one is not used. The second fix definitely looks like a bugfix (to include the dirname of the blob in the parent path). Did you test workflows without updating the mtime but with the second fix?

cademirch:
> I would guess the timing difference comes down to using a cached value vs. requesting that one is not used. The second fix definitely looks like a bugfix (to include the dirname of the blob in the parent path). Did you test workflows without updating the mtime but with the second fix?

I did but I didn't record it, I can do that now. I'm curious why we would directly use the uncached mtime when building the mtime inventory?

vsoch:
> I did but I didn't record it, I can do that now. I'm curious why we would directly use the uncached mtime when building the mtime inventory?

Mostly because a cached value may not be the most recent time it was modified. The question for snakemake is where / why would that matter? E.g., if we knew it was modified within the last day, is that enough, vs knowing it was an hour ago vs. three?

cademirch:
> > I did but I didn't record it, I can do that now. I'm curious why we would directly use the uncached mtime when building the mtime inventory?
> 
> 
> 
> Mostly because a cached value may not be the most recent time it was modified. The question for snakemake is where / why would that matter? E.g., if we knew it was modified within the last day, is that enough, vs knowing it was an hour ago vs. three? 

Hmm I guess I'm confused about how IOCache is used then. I thought it's built every execution of snakemake so it should be the most up to date information.

vsoch:
>  Hmm I guess I'm confused about how IOCache is used then. I thought it's built every execution of snakemake so it should be the most up to date information.

Yeah - so we probably would not want a cached modified time, but I'll defer to @johanneskoester maybe that specificity doesn't matter as long as we can determine it was modified.

cademirch:
> >  Hmm I guess I'm confused about how IOCache is used then. I thought it's built every execution of snakemake so it should be the most up to date information.
> 
> 
> 
> Yeah - so we probably would not want a cached modified time, but I'll defer to @johanneskoester maybe that specificity doesn't matter as long as we can determine it was modified.
> 
> 
> 
> 

I'm not sure I agree. The iocache is created in workflow.py then passed to the DAG. Where it's used and updated by target files to check existence and mtime of needed files. So the cache is always up to date relative to the start of the workflow.

cademirch:
@johanneskoester Any thoughts?

snakemake/remote/GS.py
```diff
@@ -197,7 +197,7 @@ async def inventory(self, cache: snakemake.io.IOCache):
     # === Implementations of abstract class members ===
 
     def get_inventory_parent(self):
-        return self.bucket_name
+        return "%s/%s" % (self.bucket_name, os.path.dirname(self.blob.name))
```
johanneskoester, line 200: Let us rather use an f-string here:
```suggestion
        return f"{self.bucket_name}/{os.path.dirname(self.blob.name)}"
```

snakemake/io.py
```diff
@@ -172,7 +172,7 @@ async def worker(queue):
         await asyncio.gather(*tasks)
 
     async def collect_mtime(self, path):
-        return path.mtime_uncached
+        return path.mtime
```
johanneskoester, line 175: Let us not change this here. I have just created a separate PR with a more efficient change that in this special case involves fewer lookups: #2284.

```suggestion
        return path.mtime_uncached
```

johanneskoester (changes requested):

cademirch:
@johanneskoester Thanks, applied your suggestions. Should be ready to merge.

Issue #4, opened 2019-10-05
Remove some deprecated functionality


johanneskoester:
* [ ] remove `dynamic` as this is now supported via checkpoints, keep a dummy that fails with an explanatory error message.
* [ ] remove `version` keyword, as this is now much better handled via conda and singularity integration
* [ ] remove `R()` helper function, as this is now handled via the `script` keyword. Keep a dummy that fails with an explanatory error message.

epruesse:
`version` might be handy for development of custom code that isn't packaged? To rerun everything that used an outdated version of the code.

johanneskoester:
Removed already, so closing this. We are pretty save on that axis by checking script contents nowadays.

PR #2272, opened 2023-05-23
feat!: #2271 Replace list-version-changes with list-conda-env-changes

### LINKED ISSUE #2271: --list-version-changes should become --list-software-env-changes

Sometimes I update the YAML file of a conda environment that is used by multiple rules, but I don’t want to re-run all the rules that depends on it. It happens for instance when the n-th rule that uses this conda environment requires an extra package that was not needed by the previous rules.

There’s actually no way to easily cleanup metadata for the rules that are triggered by software-env, because contrary to code, input and mtime changes we lack the corresponding `--list-<>-changes` option.

I discovered there’s actually a deprecated `--list-version-changes` but it does not seems to work with conda env (and it will be removed as mentioned in issue [#4](https://github.com/snakemake/snakemake/issues/4)). It actually took me a long time to understand what this parameter was doing since it’s not documented anymore (I guess rules can have a "version" keyword?).

Note that the option `--rerun-triggers software-env` does the job of identifying the correct rules, but there’s no easy way to combine it with `--cleanup-metadata` like `--list-<>-changes`, so you end up having to run your snakemake pipeline with `--rerun-triggers mtime` forever...

The idea is to have a `--list-software-env-changes` (name is chosen for consistency with `--rerun-triggers`) option for this exact purpose. It will list all the outputs that are generated due to a change in their software environment. Its main purpose will be to cleanup metadata for the given files using: `snakemake --cleanup-metadata $(snakemake --list-software-env-changes)`

Eventually, a `--list-changes` that lists all the files that will be updated could serve the same purpose as it could be combined to obtain the desired result: `snakemake --rerun-triggers software-env --list-changes`

nigiord:
### Description

Replace list-version-changes with list-conda-env-changes to easily list the files impacted by a change in their rule’s conda environment. This enhancement removes the deprecated `--list-version-changes` that was flagged for removal since version 6.0.0 (see issue [#4](https://github.com/snakemake/snakemake/issues/4)) and implements enhancement request #2271 .

### QC


johanneskoester (approved):

nigiord:
Hello, seems like conflicts have been introduced since this PR was approved without it being merged?

Is there something I should do for it to be merged? I feel like the main branch is currently under heavy changes and I’m not even sure it’s functional, so I prefer asking before trying to resolve the conflicts.

fgvieira:
@nigiord sorry for the delay, but can you fix the conflicts now?

Issue #2147, opened 2023-02-28
Benchmark only counts CPU time for the last subprocess of the rules


VPestel:
**Snakemake version**
Behavior can be reproduced with v7.23.1 (and also on the main branch)

**Describe the bug**
When running multiple command in the rule `shell` instruction, only the CPU contribution of the last one is accounted when using the `benchmark` procedure.

**Minimal example**
The following `Snakefile` illustrate the problem.
```
rule all:
    input: "test.file"

rule test_benchmark:
    output: temp("test.file")
    benchmark: "test.tsv"

    shell:
        """
        # Task A
        time stress --cpu 1 --timeout 110
        
        # Task B
        time stress --cpu 1 --timeout 95

        # Task C
        time stress --cpu 1 --timeout 65

        # Task D
        time stress --cpu 1 --timeout 35

        touch {output}
        """
```

The created `test.tsv` contains:
```
s	h:m:s	max_rss	max_vms	max_uss	max_pss	io_in	io_out	mean_load	cpu_time
305.0059	0:05:05	4.64	19.42	0.49	0.72	0.00	0.00	63.83	15.35
```
because it only account for task C (the inaccuracy is due to the sampling time of 30 seconds).

The processes tree generated by this rules looks like follow:
```
bash
├── stress # <- task A
├── stress # <- task B
├── stress # <- task C
└── stress # <- task D
```

But when the last call of [`self._update_record()`](https://github.com/snakemake/snakemake/blob/1a754fd094bd13bb4a201f1c80a077656c89f995/snakemake/benchmark.py#L262) happen only `bash` and `stress` corresponding to task D are accounted, because they are the only ones still running.

**Proposed solution**

A simple way to solve the problem would be to turn `self.processed_procs` into a dictionary, and save the maximum CPU time for each pair of `(proc.pid, proc.name())` and sum the total CPU time when all the active processes have been parsed. I'll be happy to open a merge request to propose a solution.

cademirch:
I ran into this issue too recently and I think your proposed solution makes a lot of sense. If you open a PR I'd be happy to help!

fgvieira:
@VPestel can you check if this is fixed in the latest `snakemake` version?

VPestel:
> @VPestel can you check if this is fixed in the latest `snakemake` version?

Just tested it with 8.4.8, it seems like it works.

PR #2278, opened 2023-05-25
fix: only last daughter process accounted in benchmark cpu time

VPestel:
Solves #2147  by keeping track of CPU times in a dictionary. 

### Description

The CPU time was computed at each call of ``BenchmarkTimer:_update_record`` by looping over the current  ``chain((self.main,), self.main.children(recursive=True))`` and summing the CPU times.

This approach only account for the total CPU times of processes that are still alive when the last ``BenchmarkTimer:_update_record`` is done, so when 2 or more processes are started sequentially, the CPU times value only capture the running time of the last process.

To correct this bug, I propose to replace ``BenchmarkTimer:processed_procs`` by a dictionary and keep track of the total CPU times of all processes spawn by the main current process.  It know gets the proper result when using the example provided in #2147.

### QC


VPestel:
I'm not sure on how to implement a test for this case. Looking at [tests/test_benchmark](https://github.com/snakemake/snakemake/tree/main/tests/test_benchmark) I see it only checks the benchmark files are produced (makes sense, hard to guess what be in it). As it only corrects edge case bug, not sure if we need to test it though.

johanneskoester (approved):
Thanks a lot for the fix. I think we can live with not really having a direct test. The changes make a lot of sense.

johanneskoester:
May I ask you to run black on the codebase, so that your changes are formatted according to the standards?

VPestel:
> May I ask you to run black on the codebase, so that your changes are formatted according to the standards?

Done !

johanneskoester:
Thanks a lot!

Issue #1619, opened 2022-04-30
Why can't we run conda and run together in a rule?


animesh-workplace:
```python3
Conda environments are only allowed with shell, script, notebook, or wrapper directives (not with run or template_engine).
```

I'm receiving the above error when I'm trying to run a snakemake rule something like this

```python3
rule bcftools_consensus_1_15_1:

    input:
        file = rules.gatk_haplotype_caller_4_2_6_1.output.file,
        reference = "resources/reference/NC_045512.2.fna",
    output:
        file = "output/bcftools/A_consensus.fasta",
    threads: 1
    conda: "env.yaml"
    log: "output/bcftools/consensus_1_15_1.log"
    run:
        # few python statements run here
        # ...
        # ...
        shell(
            '''
                bcftools consensus -f {input.reference} {input.file} \
                -o {output.file} && sed -i "1s/.*/>A_fasta/" {output.file}
            '''
        )
```

Is there a way to achieve the use of conda and run together??

cademirch:
From the [docs](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#integrated-package-management):
> Note that conda environments are only used with `shell`, `script`, `notebook` and the `wrapper` directive, not the `run` directive. The reason is that the run directive has access to the rest of the Snakefile (e.g. globally defined variables) and therefore must be executed in the same process as Snakemake itself. If used with `notebook` directive, the associated conda environment should have package `jupyter` installed (this package contains dependencies required to execute the notebook).

animesh-workplace:
The best option would then be to use a wrapper directive that would have the conda environment as well as the pythonic scripting ability.

dariober:
(I guess this issue can be closed now?) - I'm not familiar with wrappers as suggested by @animesh-workplace so I would use one of these two options:

1. Wrap the "few python statements" in a script and use it as:

```
shell:
    r"""
    scripts/my-bcftools_consensus_1_15_1.py {input.foo} {output.bar} {params if any}
    bcftools ...
    """
```

2. Run python using the "few python statements" as command line string:

```
shell:
    r"""
    python -c "import pandas
dat = pandas.read_csv('{input.foo}')
out = dat.to_csv('{output.bar}')
"

    bcftools ...
    """
```

These solutions will "see" only variables defined inside the rule, like `input`, `output`, `params` and in my opionion this is a good thing.

davidecrs:
Hi,

I have the same problem. 
[...]

Issue #163, opened 2019-12-20
Conditional rule with conda environment


mevers:
Hi.

Following advice given by Johannes Köster in [Add complex conditional file dependency
](https://bitbucket.org/snakemake/snakemake/issues/37/add-complex-conditional-file-dependency) on how to include a conditional I've got the following conditional rule

```python
rule bowtie2_align_reads:
    input:
        idx = IDX,
        reads = lambda wildcards: expand(join(
            config["fastqdir"],
            "{file}"),
            file = config["units"][wildcards.sample])
    output:
        join(
            config["analysisdir"],
            "{reference_version}/alignment/{sample}.bam")
    log:
        "logs/bowtie2_{reference_version}_{sample}.log"
    params:
        cmd      = config["bowtie2"]["cmd"],
        ref      = re.sub(".fa", "", REF),
        in_fmt   = config["bowtie2"]["in_fmt"],
        phred    = config["bowtie2"]["phred"],
        maxins   = config["bowtie2"]["maxins"],
        mismatch = config["bowtie2"]["mismatch"],
        threads  = config["bowtie2"]["threads"]
    version: "1.0"
    run:
        if (len(input.reads) == 2):
            shell(" \
                {params.cmd} \
                {params.phred} \
                --no-mixed \
                --no-discordant \
                --maxins {params.maxins} \
                -N {params.mismatch} \
                --threads {params.threads} \
                -x {params.ref} \
                -1 {input.reads[0]} \
                -2 {input.reads[1]} \
                2> {log} \
                | samtools view -bS -F4 - > {output} \
            ")
        elif (len(input.reads) == 1):
            shell(" \
                {params.cmd} \
                {params.phred} \
                -N {params.mismatch} \
                --threads {params.threads} \
                -x {params.ref} \
                -U {input.reads} \
                2> {log} \
                | samtools view -bS -F4 - > {output} \
            ")
```

In an attempt to "robustify" the workflow I am trying to use conda environments per rule. However, when I add 

```python
conda:
    "envs/bowtie2.yaml"
``` 

I get the following error
```
RuleException in line 42 of /home/mevers/Projects/ChIPseq_UBF_reanalysis/snakemake/rules/alignment.smk:
Conda environments are only allowed with shell, script, or wrapper directives (not with run).
```

My questions are:

[...]

PR #2289, opened 2023-06-07
feat: shell function calls inside of the 'run' directive now use conda, container, or envmodules specifications

mhulsman:
### Description

Rules that combine python and shell code can be quite useful, e.g.:
```
  run:
       x = parse_input_file(input)
       shell('command --param {x}')
```

Conda integration for the shell commands in such a rule is currently only supported like this:
```           
       shell('command --param {x}', conda_env='named_preexisting_environment')
```
Named pre-existing conda environments make a workflow however more difficult to distribute and could lead to reproducibility issues.

Better would be to use conda specification files (e.g. 'envs/my_env.yaml'). This is however not supported by 'shell' function calls. Implementing it would not be straightforward, because these shell-calls are only executed during the processing of a rule, and snakemake therefore would not know at startup which conda environments would be needed. Creating the environments as-needed during rule execution would lead to issues with rules that are executed in parallel. 

This issue has bee noted before #163, #1619. One possible solution is to put all code within e.g. a separate python script which in turn executes 'command'. Then everything can run within the conda environment. This however requires quite a bit of boilerplate code to substitute what snakemake normally does (e.g. argument parsing, formatting of the shell command based on params/resources/input/output files, subprocess machinery). Another alternative is to implement this with  a bash script and use the 'shell' directive, but this is usually not as readable as a python solution, particularly if one needs to do more complex operations, e.g. parse an input file. 

This pull request therefore enables the following rule structure: 
```
  conda: 'envs/my_env.yaml'
  run:
       x = parse_input_file(input)
       shell('command --param {x}')
```
This will perform 'command --param {x}' within the specified conda environment. Note that it will not perform the whole run block within the conda environment, but only shell function calls that are executed from the run block. A note specifying this is added to the documentation. For symmetry, it makes this change also for the 'container' directive. 

No large code changes are required, as the functionality is already present, only a check has to be removed. 

One limitation of this solution is that each shell call executed from the run script will be performed in the same conda environment (it is though possible to execute without the conda environment using `shell('command', conda_env=None)`). 
A future implementation could make this more flexible by enabling something like this:
```
conda:
      env1='envs/env1.yaml',
      env2='envs/env2.yaml'
run:
    if <condition>: 
         shell('command1 {input}', conda_env=conda_env.env1)
    else:
         shell('command2 {input}', conda_env=conda_env.env2)
```

### QC

## Summary by CodeRabbit

- **Documentation**
  - Revised guidance now allows using conda environments and container setups with an additional workflow directive, enhancing clarity on environment execution.
  
- **New Features**
  - Expanded support for environment and container configurations increases workflow flexibility.
  - Introduced a new rule for executing a Python script within a conda environment.
  
- **Tests**
  - Added new test cases to validate the execution of conda environments and updated existing tests to reflect changes in script paths and outputs.


mhulsman:
Just to clarify:  the example above is simplified, and I realize it can also be accomplished by defining a separate params function. However, I frequently have rules that have some python control flow (if/else, for) mixed with shell commands, for which this is not the case.

ZYF1997CN:
How is it going? This enhancement will make if-else conditions much easier in snakemake.

YuanfengZhang:
That would be very intuitive and more convenient. Hope this PR get reviewed and approved soon.

tests/test_conda_run/Snakefile
```diff
@@ -0,0 +1,9 @@
+rule random_python_conda_script:
+    output:
+        "test.txt"
+    conda:
+        "test_python_env.yaml"
+    params: 
+        script=workflow.source_path('test_script_run.py')
+    run:
+        shell("python {params.script}")
```
johanneskoester, line 9:
```suggestion
        # Calling a script like this is usually unnecessary with Snakemake
        # we use this pattern just for testing purposes here.
        # In a real workflow, use the script directive instead which
        # is cleaner and gives you many more benefits.
        shell("python {params.script}")
```

johanneskoester (approved):

johanneskoester (approved):

PR #2300, opened 2023-06-14
fix: Detect pandas availability to select serializer

DrYak:
### Description

- previous try catch solution wasn't triggered
- use detecting availability of the module instead

### QC


DrYak:
for a potential test, see [this Discord post](https://discord.com/channels/753690260830945390/1117899234243514368/1118630685461205113) but I would need the help of someone more fulent into the CI/CD infrastructure of snakemake to guide me.

PR #2305, opened 2023-06-16
add support for external executor plugins

vsoch:
Hi @johanneskoester! :wave: As we chat about in a thread somewhere, I think it would be really powerful to allow for installing (and discovering) external plugins to Snakemake. Specifically for the Flux Operator, I have easily three designs I'm testing, and it's not really appropriate to add them proper to snakemake - but I believe the developer user should be empowered to flexibly add/remove and test them out. 

This pull request is first try demo of how snakemake could allow external executor plugins. I say "first try" because it's the first time I've experimented with plugins, and I tried to choose a design that optimizes simplicity and flexibility without requiring external packages, or specific features of setuptools or similar (that are likely to change). The basic design here uses pkgutil to discover snakemake_executor_* plugins, and then provides them to the client (to add arguments) and scheduler to select using one with `--executor`.

I've written up an entire tutorial and the basic design in this early prototype, which is basically the current Flux integration as a plugin! https://github.com/snakemake/snakemake-executor-flux. The user would basically do:

```bash
# Assuming this was released on pypi (it's not yet)
$ pip install snakemake-executor-flux

# Run the workflow using the flux custom executor
$ snakemake --jobs 1 --executor flux
```
I've designed it so that plugins are validated only when chosen, and each plugin can add or otherwise customize the parser, and then (after parsing) further tweak the args if chosen. Then in scheduler.py, we simply look if the user selected a plugin, and call the main executor (and local_executor) classes if this is the case.

The one hard piece is having a flexible way to pass forward all those custom arguments. The current snakemake design has basically a custom boolean for every executor hard coded (e.g., `--flux` or `--slurm`) and while we don't want to blow that up, I'm worried moving forward passing all these custom namespaced arguments through the init, workflow, scheduler/dag, is going to get very messy. So the approach here is a suggested way to handle the expanding space of additional executors by way of passing forward full args, and then allowing the plugins to customize the parser before or after. If we were to, for example, turn current executors into plugins (something I expect we might want to do for the Google Life Sciences API that is going to be deprecated in favor of batch) we could write out a more hardened spec - some configuration class that is passed from the argument parser through the executor and through execution (instead of all the one off arguments).

Anyway - this is just a first shot and I'm hoping to start some discussion! This is a totally separate thing from TBA work with Google Batch - this is something that I've wanted to try for a while as I've wanted to add more executors and have seen the executor space exploding. :laughing:  I haven't written tests or updated any docs yet pending our discussion!

### QC


johanneskoester (changes requested):
@vsoch this is so great, thanks a lot! And a very nice and simple approach. 

# Topic 1:
Regarding the args, I have the following thought:
Since Snakemake is not only a CLI app but also is from time to time used as a python library (people calling the `snakemake` function directly), I think it is probably better to not configure executor plugins via args. Instead I think it makes sense to use a [dataclass](https://docs.python.org/3/library/dataclasses.html) for the executor plugin settings. Each plugin could then provide its own dataclass, by convention named e.g. `ExecutorParameters`. An object of this dataclass can then be expected as argument in the `snakemake` function. Further, the dataclass itself can be used to extend the argument parser of Snakemake, e.g. using [this](https://pypi.org/project/argparse-dataclass/) package which allows to annotate data classes with help texts for the argument parser.

Maybe there is also a better way, but this would be my first idea.

# Topic 2:
In addition, I think it makes sense to decide about the namespacing of executor arguments while designing this. I think there are two types of arguments relevant for an executor.

1. Those that are executor specific (like `--k8s-cpu-scalar`).
2. Those that are relevant for multiple executors (like --preemptible).

For the executor specific ones, we should define a prefix that comes from the plugin name (e.g. just the name of the executor). For the others, the name is kind of free to choose. In order to let the executor access all those general args, it makes probably sense to put everything that is currently passed as individual parameters to the `snakemake` function into a dataclass as well. Then, the executor plugin can just get two dataclass objects (the global args and the executor specific args).

# Topic 3:
I think the plugin implementation should require a plugin to provide the minimum compatible Snakemake version. A natural value for the beginning would be the first release with plugin functionality. But later, some plugins might require more recent versions of Snakemake and the user should get a proper error message instead of just some cryptic exceptions.

# Topic 4:
And finally: yes, every executor should eventually be migrated into a plugin.

vsoch:
@johanneskoester an update!

I completely agree with your sentiment - I was not happy with the args design either, and I think using a dataclass is a great strategy. Specifically what I decided to try for this first go is:

1. Allowing the custom plugin to define a vanilla dataclasses.dataclass, and with arguments that have a prefix specific to the plugin (to namespace them in the argument space)
2. Using a function from https://github.com/mivade/argparse_dataclass that can add the dataclass to the parser (and it's a simple conversion of a datatype to an arguiment). This is why the namespacing is important, otherwise you'd get a conflicting name error)
3. The args (with user choices for the plugin too!) are then converted back to the dataclass, which is passed forward to the custom executor plugin!

Note that the above approach doesn't require an extra dependency in snakemake, but just in the plugin, which I like. I also provide the "parse your args back into the dataclass" as a `snakemake.plugins` function, which (I hope) along with documentation / examples will encourage folks to honor their plugin's namespace. We only (by default) return custom dataclass arguments that also adhere to the executor plugin namespace. Anyone that wants to customize that can proceed at their own risk! And note that outside of the dataclass, any "global" snakemake arg that is used by a plugin would be provided as it is now - passed from args into some other argument to the executor. We could eventually clean this up and have some way for multiple plugins to add the same argument, but that's probably out of scope for this early work.

> For the executor specific ones, we should define a prefix that comes from the plugin name (e.g. just the name of the executor). For the others, the name is kind of free to choose. In order to let the executor access all those general args, it makes probably sense to put everything that is currently passed as individual parameters to the snakemake function into a dataclass as well. Then, the executor plugin can just get two dataclass objects (the global args and the executor specific args).

Oh that's a fantastic idea - I could follow up with this PR, unless you think it's absolutely essential here (but would be a lot of moving pieces for one PR).

And I'm not sure where those other commits came from - I hope I didn't bork the git history. I can try a rebase, although now I'm nervous to screw the entire thing up. Perhaps when it is definitely done I will just redo the whole thing over freshly.

> I think the plugin implementation should require a plugin to provide the minimum compatible Snakemake version. A natural value for the beginning would be the first release with plugin functionality. But later, some plugins might require more recent versions of Snakemake and the user should get a proper error message instead of just some cryptic exceptions.

I didn't add this yet, but I will after posting this!

> And finally: yes, every executor should eventually be migrated into a plugin.

Agree! This init file and passing forward arguments has seriously been giving me nightmares for years... I'm glad we have a plan to clean things up a bit!

For details on the current state of the plugin setup, we have an early writeup here: https://github.com/snakemake/snakemake-executor-flux#how-does-it-work. I'll update it soon for the snakemake version (and will validate it's defined).

johanneskoester:
I think the other commits are just coming from merging in upstream. That should be fine.

In order to get the plugin interface right, it seems to me that the dataclass approach for the other arguments should be done first. But indeed that can happen in a separate PR. It would be awesome if you really would like to do that!
Then, the plugins would be also simpler because I guess the add_args and parse functionality can then be moved into Snakemake itself.

vsoch:
> Then, the plugins would be also simpler because I guess the add_args and parse functionality can then be moved into Snakemake itself.

Most of the functionality within those functions can be provided by snakemake, but I expose them on the level of the plugin to give the plugins control to dictate logic. But actually, we could remove them from validation and then do a check (and call the default provided in snakemake) if they aren't present!

What did you have in mind for porting current executors to plugins? Having the same structure but adding them to snakemake dependencies so they are always installed?

vsoch:
Just rebased - I'm not sure why the local test is reporting there is no module packaging, because it's definitely in test-environment.yaml!

vsoch:
@johanneskoester it is reporting that packaging isn't installed, however I've added it to the snakemake install requires, the test step, and the container. I'm not familiar with this setup - can you take a look and advise where it needs to be added?

```
__real_file__ = __file__; __file__ = '/tmp/snakemake-lmz9cooy/script.py';
  File "/mnt/snakemake/snakemake/__init__.py", line 23, in <module>
    from snakemake.workflow import Workflow
  File "/mnt/snakemake/snakemake/workflow.py", line 28, in <module>
    from snakemake.scheduler import JobScheduler
  File "/mnt/snakemake/snakemake/scheduler.py", line 27, in <module>
    from snakemake.plugins import executor_plugins
  File "/mnt/snakemake/snakemake/plugins.py", line 10, in <module>
    import packaging.version
ModuleNotFoundError: No module named 'packaging'
```

johanneskoester:
> What did you have in mind for porting current executors to plugins? Having the same structure but adding them to snakemake dependencies so they are always installed?

Yes, exactly that.

snakemake/plugins.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def validate_snakemake_version(minimum_version):
+    """
+    Ensure that minimum snakemake version is satisfied.
+
+    If an unknown or unparseable version is provided, we cannot
+    determine and are more leniant and allow it.
+    """
+    from snakemake import __version__
```
johanneskoester, line 85:
```suggestion
    from snakemake import __version__
    import packaging.version
```

snakemake/plugins.py
```diff
@@ -0,0 +1,128 @@
+__author__ = "Johannes Köster"
+__copyright__ = "Copyright 2023, Johannes Köster"
+__email__ = "johannes.koester@uni-due.de"
+__license__ = "MIT"
+
+
+import importlib
+import os
+import pkgutil
+import packaging.version
```
johanneskoester, line 10:
```suggestion
```

johanneskoester (changes requested):
Since Snakemake pickles parameters into scripts executed within singularity, any imports of external dependencies have to be done at the function/method level, not the module level. Otherwise, pickling will require them to be present in the container.

vsoch:
@johanneskoester how would we get custom arguments then?

johanneskoester:
The executor provides information about its required input via the dataclass that it defines. Like this:

```
from dataclasses import dataclass, field

@dataclass
class ExecutorParameters:
    machine: str = field(default="standard-e2", metadata={"help": "Machine type to use for Google Batch execution."})
```

For the two use cases:
* CLI: we add the fields to the argument parser of Snakemake, and prefix every field with the name of the executor (e.g. in the above case we generate `--google-batch-machine`).
* API: an instance of the dataclass is simply created by the developer who uses the API and passed to the snakemake function.

I was wondering whether we should have a way to specify and access non executor specific arguments, but I think this is not necessary. At least not beyond those parameters that are already passed to the current exectuors by the scheduler. This also means that we don't have to modify the entire Snakemake towards dataclass usage for now. We can simply start with the executor plugins.

vsoch:
> CLI: we add the fields to the argument parser of Snakemake, and prefix every field with the name of the executor (e.g. in the above case we generate --google-batch-machine if the executor plugin package is called snakemake-executor-google-batch).

So you are OK with the plugin dynamically adding arguments (meaning snakemake adding them based on the dataclass, the opposite direction that we have now), you just don't want to explicitly hand over args? What about cases where we need further customization of the args for the plugin to work?

That could work, although if more than one plugin is registered, we can't enforce any of the arguments to be required, and if we can't pass the args forward to validate, it will be more of a "try to run and hope it works."

Let me know if that is what you have in mind and I can make changes.

vsoch:
Taking a look again - the case for just having the arguments from the dataclass added, as is, is already easy - we can just tweak this function:

```
def add_args(parser):
    """
    Allow the custom executor to modify the parser as needed.

    Note that we do not here, but it's recommended to use your custom
    parser GROUP_IDENTIFIER as a prefix to honor the namespace.
    """
    _add_dataclass_options(ExecutorParameters, parser)
```
To add the plugin prefix to the ExecutorParameters before we provide to that function. But what about needing to tweak logic within snakemake apriori? We have a function for that too:

```
plugins.args_to_dataclass(args, ExecutorParameters)
``` 
But I'm thinking about cases in the current init where we have big if statements like "if this executor, change the filesystem setting to X." I'm guessing your thinking is that we will eventually have some kind of shared executor parameters that somehow handle that? Should we not consider that for now, and just move these existing functions to (instead of being called by the plugin) to be called by snakemake directly with the context of the plugin?

johanneskoester:
> To add the plugin prefix to the ExecutorParameters before we provide to that function. But what about needing to tweak logic within snakemake apriori? We have a function for that too:
> 
> ```
> plugins.args_to_dataclass(args, ExecutorParameters)
> ```
> 
> But I'm thinking about cases in the current init where we have big if statements like "if this executor, change the filesystem setting to X." I'm guessing your thinking is that we will eventually have some kind of shared executor parameters that somehow handle that? Should we not consider that for now, and just move these existing functions to (instead of being called by the plugin) to be called by snakemake directly with the context of the plugin?

I would handle that by requiring that the plugin has a dict called `snakemake_behavior`.
In Snakemake, this dict can then be queried for certain things (always with defaults so that things added in future versions of Snakemake are in any case backwards compatible and so that the plugin can choose to just use the default for certain things), e.g.
```
executor_plugin.snakemake_behavior.get("use_threads", False)
executor_plugin.snakemake_behavior.get("assume_shared_fs", True)
executor_plugin.snakemake_behavior.get("default_remote_provider", None)
executor_plugin.snakemake_behavior.get("default_remote_prefix", None)
```

This queries should happen inside of the `snakemake` function (of course by calling a helper), because they have to happen both in the API and in the CLI case.

vsoch:
@johanneskoester ignoring the black failure for now (I somehow have the same version 23.3.0 installed but it is saying there are no changes so we can look at that detail later) I've refactored the plugin structure so that the plugin itself does no additional parsing of either the parser, or the args. The plugin now can define whatever fields they like, eg..,

```
@dataclass
class ExecutorParameters:
    help: Optional[str] = None
    description: Optional[str] = None
```
And then snakemake iterates through each plugin's fields, and adds them with the plugin prefix. E.g., the above becomes:

```
--flux-help
--flux-description
```

And then to go the other way around, we take that dataclass, remove the prefix, and parse back into the original. An edge case I'm thinking abiout

More detail updated here:

https://github.com/snakemake/snakemake-executor-flux. I think this is probably closer to what you had in mind - let me know what we should work on next (and what might be scoped to this PR vs done in a follow up).

johanneskoester (changes requested):
I have thought further about this (all morning basically).

Instead of the minimum version requirerement, we can do something else: The execuctor plugin and Snakemake itself can depend on a third package, called snakemake-executor-plugin-interface. This package implements the logic for interacting with the plugin (so mainly what you now have in snakemake/plugins.py). This way, we achieve the following:
The versioning of Snakemake is decoupled from the plugin interface. This allows the plugin to e.g. depend on the major version of the interface package, and thereby easily avoid future breakages. Further, we don't need custom attributes and functions for checking compatibility. The package dependency management solves the problem for us. Also, a user choosing a specific plugin that has not been updated for a while will automatically get the latest compatible Snakemake version suggested by the package manager.

I have prepared a skeleton for this here: https://github.com/snakemake/snakemake-executor-plugin-interface
While doing that, I have also redesigned the interface another time. It is now more object oriented, which has the advantage to assert compliance with the interface via the types.
The basic implementation is already in place, and I was wondering if you would like to migrate your plugin code from here into the missing pieces of the new interface via a PR on that repository. Mainly, we now need

1. The code that translates the dataclass into CLI args [here](https://github.com/snakemake/snakemake-executor-plugin-interface/blob/5771bba8346efb6e62b9725f71d86c0f8daa5907/snakemake_executor_plugin_interface/__init__.py#L36)
2. The code that translates args into the dataclass [here](https://github.com/snakemake/snakemake-executor-plugin-interface/blob/5771bba8346efb6e62b9725f71d86c0f8daa5907/snakemake_executor_plugin_interface/__init__.py#L40)
3. The code that collects all plugins [here](https://github.com/snakemake/snakemake-executor-plugin-interface/blob/5771bba8346efb6e62b9725f71d86c0f8daa5907/snakemake_executor_plugin_interface/__init__.py#L64)

I have added an example how the executor would implement this to the [README](https://github.com/snakemake/snakemake-executor-plugin-interface/tree/main).

What do you think?

vsoch:
>  I was wondering if you would like to migrate your plugin code from here into the missing pieces of the new interface via a PR on that repository.

> What do you think?

If that's the direction you like, it works for me! One suggestion - should we have a different prefix name for this library so it doesn't detect itself as a plugin? And also - if you have vision for other components of snakemake to be plugins, would you want to manage another interface (or go off of this one?)

snakemake/plugins.py
```diff
@@ -0,0 +1,183 @@
...
+    for _, name, _ in pkgutil.iter_modules()
+    if name.startswith(executor_plugin_prefix)
+}
+
+# Valid Argument types (to distinguish from empty dataclasses)
+ArgTypes = (str, int, float, bool)
+
+# Required executor plugin attributes
+executor_plugin_attributes = [
+    "local_executor",
```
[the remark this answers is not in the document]
vsoch: I noticed there is logic in scheduler.py to have a different local executor, but I think we could fairly safely assume the CPU executor for now, and think about how to handle it if that isn't the case.

vsoch:
WIP updates pushed! Here are other relevant links:

- https://github.com/snakemake/snakemake-executor-plugin-interface/pull/5
- https://github.com/snakemake/snakemake-executor-flux/pull/3

None of the tests will pass here for the time being because the plugin interface package is not available via pip.

Overall I think adding this interface is a good idea, although it does seem higher in complexity in the design. In that you are happy with this direction I think I'm OK with it too. :)

johanneskoester:
I think we will benefit a lot from a controlled and stable interface between Snakemake and its plugins. Avoiding accidental breakages will also encourage devs to join and write plugins.

vsoch:
You’ll definitely have my contributions! 🎉

johanneskoester:
I am currently defining stable ABCs for everything executors currently interact with: https://github.com/snakemake/snakemake/pull/2341.
* Once that PR works and is merged, I would move those ABCs into `snakemake-executor-plugin-interface`.
* Once that is done, we can remove those interfaces from Snakemake in this PR here and instead import them from the plugin interface package.

This way, any change in those methods will automatically lead to version bumps in the Snakemake interface package, such that compatibility with executors is automatically handled. E.g., when we extend the interface, there will be a minor version bump. When we break the interface, there will be a major version bump. Executor plugins should basically just pin to a major version of the interface package and that's it.

vsoch:
We might need an updated release for this to work? I think the location of the registry (with registry and plugin classes) changed. https://github.com/snakemake/snakemake-executor-plugin-interface/pull/3

johanneskoester:
Yes, I want to move some things around before I release it. I hope to get this done tomorrow morning.

johanneskoester:
Hey @vsoch. I am currently drafting a fix for the issue we see in the tests now. I will try to finalize my thoughts during my vacation time in the next two weeks.

vsoch:
Sounds good! I'll be around to help if needed. I'm working on adding ORAS and then the kueue executor plugin, and once we get this merged in I can start transforming current "hard coded" plugins into external ones - likely starting with Google Life Sciences makes sense since it's getting deprecated.

johanneskoester (approved):

vsoch:
Woohoo!! What would you like to do next @johanneskoester ? I can start refactoring some the current executors into plugins, or since we already have one for flux, remove it here. Let me know your preferences!

PR #2306, opened 2023-06-16
fix: remove distutils from snakemake

maarten-k:
### Description

Python 3.12 does not support the distutils library any more. In most cases, the go-to solution would be using the `packaging` package. This works in general cases well, but when using singularity this fails because the external packages are not in the container binding  (pytest tests/tests.py::test_singularity would trigger this)

I wrote a poor man version check and wrote test cases for the release numbers available in conda singularity and apptainer.  This works well for the version numbers of these programs, but I would not use this **untested** for other programs' version numbers.

### QC

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -214,7 +214,7 @@ jobs:
           ZENODO_SANDBOX_PAT: "${{ secrets.ZENODO_SANDBOX_PAT }}"
         shell: bash -el {0}
         run: |
-          pytest --show-capture=stderr -v -x tests/test_expand.py tests/test_io.py tests/test_schema.py tests/test_linting.py tests/tests.py tests/test_schema.py tests/test_linting.py tests/tests.py
+          pytest --show-capture=stderr -v -x tests/test_expand.py tests/test_io.py tests/test_schema.py tests/test_linting.py tests/tests.py tests/test_schema.py tests/test_linting.py tests/test_version.py
       - name: Build and publish docker image
         if: >-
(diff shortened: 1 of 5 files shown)


maarten-k:
Unit tests did not run in CI. There is a faulty import in the unit test file, but does not trigger an error. I will try to trigger an error before fixing this faulty import.

johanneskoester:
Thanks! I think it would be fine to add packaging as a dependency to Snakemake. The packaging package is lightweight and has no further dependencies. Do you probably just want to add it here in this PR?

maarten-k:
I tried using packaging at https://github.com/maarten-k/snakemake/tree/packaging_fail_singularity  and this would also be my go-to solution. However,  pytest tests/tests.py::test_singularity  fails, because a pickled dump of the snakemake object is made https://github.com/snakemake/snakemake/blob/ab58c652847c03a9f1529d2d7632f2788a5fadc4/snakemake/script.py#L541. Inside this pickled dump, there is also an indirect reference to the packaging package, which is not available inside the container, since only the snakemake directory itself is bound to the container and not the other packages. I am not sure how to sort this out in a good/clean way, so that is why I wrote this ugly workaround.

Please let me know If there is a better solution.

johanneskoester:
You can import the packaging module within the function that uses it instead of adding an import at the module level. This way, it won't be considered during the pickle.

maarten-k:
That 

> You can import the packaging module within the function that uses it instead of adding an import at the module level. This way, it won't be considered during the pickle.

That would do it(always wondered why the imports were inside the function). I create pull request  #2312

PR #2318, opened 2023-06-23
feat: Add support for Google Service Accounts and GCE VM network configuration

johntbates:
### Description

feat: Adds support for Google Service Accounts and GCE VM network configuration
* Google Life Sciences executors can now be run with user-specified service accounts
* As a result, application default credentials can be used on GCE without requiring the `GOOGLE_APPLICATION_CREDENTIALS` environment variable.
* Added flags to support user-specified service accounts, VM network and subnetwork
* Updated tests and documentation

### QC


johanneskoester (approved):

johanneskoester:
Thanks a lot!

PR #2340, opened 2023-07-03
fix: Enable values with an = sign in default_resources

### LINKED ISSUE #2290: Specification of default_resources in a profile cannot contain = signs
**Snakemake version**
7.26

**Describe the bug**
In the default_resources block in a profile, variables cannot contain = signs; this is, however, needed to specify slurm_extra (e.g., --mail-type=none). The reason is in https://github.com/snakemake/snakemake/blob/main/snakemake/resources.py#L23: the argument is split by the = sign; this leads to an error if the result contains more than 2 elements (which occurs when there is more than one = sign).

I could make it work by replacing
```
            return arg.split("=")
```
by
```
            return arg.split("=")[0], "=".join(arg.split("=")[1:])
```

which basically concatenates the rest of the value again.

**Minimal example**
Add the following to a profile file:
```
default_resources:
  - slurm_extra="--mail-type=none"
```
and then run snakemake using this profile.

ohlmann:
### Description

In the default_resources block in a profile, variables cannot contain = signs; this is, however, needed to specify slurm_extra (e.g., --mail-type=none). The reason is in https://github.com/snakemake/snakemake/blob/main/snakemake/resources.py#L23: the argument is split by the = sign; this leads to an error if the result contains more than 2 elements (which occurs when there is more than one = sign).

Closes #2290 

### QC


snakemake/resources.py
```diff
@@ -20,7 +20,7 @@ class DefaultResources:
     @classmethod
     def decode_arg(cls, arg):
         try:
-            return arg.split("=")
+            return arg.split("=")[0], "=".join(arg.split("=")[1:])
```
johanneskoester, line 23: Wouldn't this be the more idiomatic way?
```suggestion
            return arg.split("=", maxsplit=1)
```

johanneskoester (changes requested):
Thanks! Please find a change request below. Further, may I ask to add a test case?

snakemake/resources.py
ohlmann: Yes, thanks for the suggestion! I wasn't aware of the maxsplit argument.

ohlmann:
Thanks for reviewing the change! I have added a test as well.

johanneskoester (approved):

ohlmann:
Thanks for reviewing and approving!

PR #2341, opened 2023-07-04
refactor: define consistent abstract base classes for jobs

johanneskoester:
### QC


tsibley:
This PR bumped the Python requirement from >=3.7 to >=3.9 and was then released as part of a minor version bump 7.30.2.

PR #2364, opened 2023-07-19
fix: updating of non-dict config values gives error

padix-key:
### Description

This PR addresses an issue in config merging: If a config value that is not a `dict` (e.g. it is `None`) is replaced by a `dict` an exception is raised, as `update_config()` tries to update the keys of the config, which is not possible (e.g. `None` has no keys). This PR fixes this problem by just overwriting the original value in this case (e.g. `None` is replaced by the new dictionary).

### QC


johanneskoester:
Seems like a testcase fails because the result is not as expected?

padix-key:
Hi, The problem was the removal of the inner function: When replacing `None` with a new `dict` this is no in-place modification anymore. Hence, without a return value the config does not get updated in this case. I reenabled the inner function with return value to solve this issue and edited the PR description accordingly.

johanneskoester (approved):

padix-key:
Conceptually, the function was correct, only formatting changes of the module and the test output file was required in the failed jobs. My changes should hopefully fix this now.

PR #2368, opened 2023-07-24
docs: basics.rst: suggest VS Code instead of deprecated Atom as IDE

dlaehnemann:
### Description

Atom has been deprecated for quite a while, so we should suggest another editor / IDE. We have been using VS Code quite a bit and are very satisfied with its capabilities for developing Snakemake workflows, so that I think this is a reasonable suggestion. But other suggestions are welcome.

### QC

### DIFF
--- a/docs/tutorial/basics.rst
+++ b/docs/tutorial/basics.rst
@@ -85,7 +85,7 @@ Step 1: Mapping reads
 Our first Snakemake rule maps reads of a given sample to a given reference genome (see :ref:`tutorial-background`).
 For this, we will use the tool bwa_, specifically the subcommand ``bwa mem``.
 In the working directory, **create a new file** called ``Snakefile`` with an editor of your choice.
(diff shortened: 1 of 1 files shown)


docs/tutorial/basics.rst
```diff
@@ -83,7 +85,7 @@ Step 1: Mapping reads
 Our first Snakemake rule maps reads of a given sample to a given reference genome (see :ref:`tutorial-background`).
 For this, we will use the tool bwa_, specifically the subcommand ``bwa mem``.
 In the working directory, **create a new file** called ``Snakefile`` with an editor of your choice.
-We propose to use the Atom_ editor, since it provides out-of-the-box syntax highlighting for Snakemake.
+We propose to use the integrated development environment (IDE) tool _`Visual Studio Code`, since it provides a good syntax highlighting _`Snakemake extension` and a _`remote extension` for directly using the IDE on a remote server.
```
johanneskoester, line 88:
```suggestion
We propose to use the integrated development environment (IDE) tool `Visual Studio Code`_, since it provides a good syntax highlighting `Snakemake extension`_ and a `remote extension`_ for directly using the IDE on a remote server.
```

johanneskoester (changes requested):

PR #2370, opened 2023-07-24
docs: update CHANGELOG.md: add minimum Python version bump

corneliusroemer:
Resolves #2369 

Add changelog entry for Python version bump

See https://github.com/snakemake/snakemake/pull/2341#issuecomment-1648487223

Issue #2369, opened 2023-07-24
Add minimum Python version requirement bump to changelog for 


corneliusroemer:
>This PR bumped the Python requirement from >=3.7 to >=3.9 and was then released as part of a minor version bump 7.30.2.
>_Originally posted by @tsibley in https://github.com/snakemake/snakemake/issues/2341#issuecomment-1648487223_

Changelog should mention the Python version requirement bump (ideally ahead of time next time)

johanneskoester:
That version bump was my mistake. I forgot about the implications for scripts. I changed the implementation so that it is not necessary anymore. The next release will be python 3.7 again, with a corresponding note in the changelog.

PR #2372, opened 2023-07-28
fix: require python >=3.7 again (the python 3.9 dependency was unnecessary)

dlaehnemann:
### Description

The respective bug was introduced with PR #2341, so it exists starting with version `7.30.2`. Also see issue #2369 and PR #2370 for further context.

This code from PR #2341 seems to get run every time a conda environment is used:

https://github.com/snakemake/snakemake/blob/13b5d4d11d7475f942531c51b6013962a88a72b7/snakemake/__init__.py#L8-L9

Thus, any conda environment in a rule with an older python version included, will trigger a:
```
  File "/opt/conda/envs/snakemake/lib/python3.11/site-packages/snakemake/__init__.py", line 9, in <module>
    raise ValueError("Snakemake requires at least Python 3.9.")
ValueError: Snakemake requires at least Python 3.9.
```

This would preclude snakemake from running any software that needs an older python version, so this is clearly an unwanted side-effect. But I am not sure how to fix this.

Currently, this PR is only a minimal example test to trigger this behaviour in the CI tests. Hopefully, this will make debugging it quicker for @johanneskoester (or someone else?).

### QC


dlaehnemann:
A quick workaround for anybody getting this bug, until it is fixed: Downgrade your `snakemake` to `<=7.30.1`.

dlaehnemann:
So, this bug gets triggered in the Windows CI already, right here is the actual error message:
https://github.com/snakemake/snakemake/actions/runs/5690154966/job/15423010580?pr=2372#step:6:526

Currently rerunning the Linux CI, as this failed on slurm cluster creation earlier in the tests.

dlaehnemann:
The Linux CI failure seem to be an unrelated and obscure error, where the general GitHub Actions runner is `Ubuntu 22.04.2`, but the slurm setup tries to download `Ubuntu 22.04.1` packages, with the respective version not available on the Azure Ubuntu package mirror. I have no clue, why these versions would be out of sync...

dlaehnemann:
Here's a related PR for the Linux CI failure with the slurm setup #2377 .

johanneskoester:
I expect this to be fixed with the changes in https://github.com/snakemake/snakemake/pull/2305. However, that will become Snakemake 8.0. I'll think about a fix for the 7.30 series, in order to have a final release that works.

PR #2375, opened 2023-07-30
style: add type linting

Hocnonsense:
### Description

- Some code formatting
    - sort the imports
    - add type hints for some classes in order to fully utilize the code prompts of modern editors. 
        Is it necessary to add type hints gradually in a series of PR (I'd like in the future), or add all hints for all files in a single PR?


johanneskoester (changes requested):
Thanks a lot for the cleanup and the annotations. I always wanted to do that but never found the time so far. May I ask to use just the typing of Python 3.7, because I just lowered the min python version to 3.7 again (i.e. use `Dict[str, str]` instead of `dict[str, str]`).

Hocnonsense:
@johanneskoester Thanks for your approval! I've move typing to python37 style.

Meanwhile, I've move two functions, i.e., refer_to_remote and iocache, from class _IOFile to package snakemake.io. Is it necessary? Thanks!

johanneskoester:
Mhm, fails on windows though...

Hocnonsense:
`CI / testing-windows (pull_request)` successful after rerun, it seems to be an accidental error

johanneskoester (approved):

PR #2377, opened 2023-07-31
chore: use latest release of ansible roles for slurm setup

johanneskoester:
### QC

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,8 @@ jobs:
 
   testing:
     runs-on: ubuntu-latest
-    needs: formatting
+    # TODO comment in once slurm works again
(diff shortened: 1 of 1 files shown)


dlaehnemann:
I think it is the incompatible version of Ubuntu `deb` files being pulled in here:
https://github.com/snakemake/snakemake/actions/runs/5711993695/job/15474626077?pr=2377#step:8:38

It looks for `22.04.1` files, while only `22.04.2` files are available at the Azure mirror. However, I don't understand why it looks for `22.04.1` in the first place, as the [GitHub Actions runner `ubuntu-latest` (on which this workflow runs) has been on `22.04.2` for more than five months, now](https://github.com/actions/runner-images/pull/7149).

PR #2386, opened 2023-08-02
docs: fix a copy&paste (?) mistake

knarrff:
### Description

The current documentation mentions, within the tutorial:

```
set-threads:
    b: 16

to set the memory requirement of rule b to 1000 MB.
```
which is apparently a copy&paste left-over from a similar section within the same document:
```
set-resources:
    b:
        mem_mb: 1000

to set the memory requirement of rule b to 1000 MB.
```
This request fixes this.

### QC

    - **does not apply**
    - **This is only a documentation change.**


johanneskoester (approved):
Thanks!

PR #2394, opened 2023-08-04
fix: default resources setter

jakevc:
### Description

Compatibility with the new DefaultResources setter. The batch executor was failing with the following error:

```
    self.workflow.default_resources = DefaultResources(mode="bare")
AttributeError: can't set attribute 'default_resources'
```

This change fixes that.
```
 self.workflow._default_resources = DefaultResources(mode="bare")
```

### QC

* [x ] The PR contains a test case for the changes or the changes are already covered by an existing test case.

### DIFF
--- a/snakemake/executors/azure_batch.py
+++ b/snakemake/executors/azure_batch.py
@@ -25,8 +25,13 @@ from snakemake.interfaces import (
     WorkflowExecutorInterface,
 )
 from snakemake.logging import logger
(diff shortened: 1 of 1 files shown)


johanneskoester:
Thanks for pointing to the problem! Actually setting default resources in the executor does not make any sense, since that happens before in the cli handling. I have removed the statement entirely here: https://github.com/snakemake/snakemake/pull/2395

jakevc:
Awesome thanks!

PR #2396, opened 2023-08-05
style: ExecutorJobInterface->JobExecutorInterface

Hocnonsense:
### Description

This is a refactor of `JobExecutorInterface` affected snakemake/interfaces.py, snakemake/jobs.py, and snakemake/executors/*

Before, ExecutorJobInterface is nammed different from SingleJobExecutorInterface and GroupJobExecutorInterface.

### QC


johanneskoester:
Thanks, indeed a good idea. Done in #2452 instead.

PR #2403, opened 2023-08-08
feat!: redesigned Snakemake API. It now uses a modern, dataclass based approach

johanneskoester:
### QC


snakemake/cli.py
```diff
@@ -2480,3 +2481,120 @@ def print_candidates(candidates):
...
+                "--force",
+                "--keep-target-files",
+                "--keep-remote",
+                "--max-inventory-time 0",
+                "--nocolor",
+                "--notemp",
+                "--no-hooks",
+                "--nolock",
+                "--ignore-incomplete",
+                w2a("execution_settings.keep_incomplete")
```
Hocnonsense, line 2561: It seems that a comma is lost here

kyleabeauchamp:
I *think* the recommendation for migration involves the `local` flag but it might be useful to add to the 7->8 migration guide.

kyleabeauchamp:
Actually it's a find and replace of `srcdir` with `workflow.source_path`.  Probably good adding to the migration guide.

claczny:
> Actually it's a find and replace of `srcdir` with `workflow.source_path`. Probably good adding to the migration guide.

Either I am doing something wrong or this is not a drop-in replacement.
I worked around it via `workflow.basedir`, e.g., `ENVS = Path(workflow.basedir) / "envs".

Issue #2411, opened 2023-08-16
SACCT Error.


g-pacheco:
Dear Snakemake Community, 

I have been running a large pipeline using **Snakemake v7.31.0** and I have repeatedly seen this error related to sacct:

```
Job 371 has been submitted with SLURM jobid 9051536 (log: .snakemake/slurm_logs/rule_snp_call_gatk_genomicsdbimport/9051536.log).
[Wed Aug 16 06:51:38 2023]
Finished job 881.
1 of 15 steps (7%) done
[Wed Aug 16 07:08:59 2023]
Finished job 808.
2 of 15 steps (13%) done
The job status query failed with command: sacct -X --parsable2 --noheader --format=JobIdRaw,State --name c8aa4481-55ca-4e61-8d9f-21757f3b8a0f
Error message: sacct: error: slurm_persist_conn_open_without_init: failed to open persistent connection to host:los:6819: No route to host
sacct: error: Sending PersistInit msg: No route to host
sacct: error: Problem talking to the database: No route to host
```

Interestingly, I am able to run the command normally:

```
sacct -X --parsable2 --noheader --format=JobIdRaw,State --name c8aa4481-55ca-4e61-8d9f-21757f3b8a0f
9051531|RUNNING
9051532|RUNNING
9051533|COMPLETED
9051534|COMPLETED
9051535|RUNNING
9051536|RUNNING
```

Thus, it seems that this error only occurs occasionally but I do not know what is causing it. May I ask if this error has been noticed before and if there would be a solution to it?

Please let me know should you need any additional information from my end.

Many thanks in advance, George.

fgvieira:
From the error, it looks like that `sacct` could not contact the database. In this case, I think the problem is that when this command fails:
https://github.com/snakemake/snakemake/blob/27b224ed12448df8aebc7d1ff8f25e3bf7622232/snakemake/executors/slurm/slurm_submit.py#L293-L295

The exception is caught, but variables `res` and `query_duration` are not set:
https://github.com/snakemake/snakemake/blob/27b224ed12448df8aebc7d1ff8f25e3bf7622232/snakemake/executors/slurm/slurm_submit.py#L309-L323

PR #2413, opened 2023-08-17
fix: bug when sacct fails

fgvieira:
### Description

Fix issue #2411 

### QC


cmeesters:
the commit looks just fine - can we merge this PR?

PR #2416, opened 2023-08-18
Print to standard error instead of printing sys.stderr itself

terrycojones:
### Description

When running `snakemake` with no `--cores` option, the output looks like this:

```
Error: you need to specify the maximum number of CPU cores to be used at the same time. If you want to use N cores, say --cores N or -cN. For all cores on your system (be sure that this is appropriate) use --cores all. For no parallelization use --cores 1 or -c1. <_io.TextIOWrapper name='<stderr>' mode='w' encoding='utf-8'>
```

This pr prints _to_ standard error instead of printing `sys.stderr` itself. And I added `file=sys.stderr` to a couple other `print` statements in `snakemake/cli.py`.

### DIFF
--- a/snakemake/cli.py
+++ b/snakemake/cli.py
@@ -2032,7 +2032,7 @@ def main(argv=None):
         args.cores = cores
         args.jobs = jobs
     except CliException as err:
-        print(err.msg, sys.stderr)
+        print(err.msg, file=sys.stderr)
         sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


jlumpe:
Was going to do this myself but looks like you beat me to it. I see you forgot to add a comma at the end of line 2450, which is causing the formatting check to fail. Other than that I think you just need to add the `fix: ` prefix to the PR title and it should pass.

fgvieira:
Already fixed.

PR #2419, opened 2023-08-22
fix: #1938 DAG is now properly displayed in report layout of the workflow

### LINKED ISSUE #1938: Report.html linear workflow graph
**Snakemake version**

7.17.0

**Describe the bug**

After I upgraded to version 7.17.0 I started to get an unexpected workflow graph.

The workflow graph in the report.html is linear as the following:

image

**Logs**

When I look at the error log I realized the following console log error:

```
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'GroupItem'
    |     property 'mark' -> object with constructor 'Object'
    |     property 'items' -> object with constructor 'Array'
    --- index 0 closes the circle
    at JSON.stringify (<anonymous>)
    at _stringify (report.html:101010:23)
    at stringify (report.html:101069:8)
    at HTMLAnchorElement.<anonymous> (report.html:105072:21)
```

**Minimal example**

I was not able to reproduce the error with a minimal example.

**Additional context**

Thanks

nigiord:
### Description

This PR fixes the way the DAG was displayed in Snakemake reports, but it adds back the pygraphviz dependency. I did not find a way to properly format the DAG layout without this dependency though. Since it is only used for reports, I kept the previous option of just displaying a warning to install it if one wants to use the report functionnality.

There is also some redundancy with the "links" and "links_direct" options in the VEGA template. I kept it for compatibility but I think it would be safe to remove the links_direct altogether (unless it has some purposes I did not understand?).

### QC


sroener:
For me, this pull request looks like a good solutione for fixing the graph view in the reports. If ultimately, another solution would be preferred (e.g. without the pygraphviz dependency), these changes look like a good temporary solution. 

@johanneskoester do you have any input, if anything else needs to be done?

johanneskoester:
Thanks a lot! However, I think I would rather like to offer more options dynamically directly within the report, without needing to make a decision about the layout when generating the report. Options are:
* https://github.com/erikbrinkman/d3-dag
* https://www.npmjs.com/package/graphology
* https://www.npmjs.com/package/sigma

Issue #2451, opened 2023-09-18
Google life sciences executor down


erikenbody:
Snakemake v7.32.3

Using the google life sciences executer seems to currently be broken. As of 9/16, my submissions to GLS result in the following error:
[gls_snakemake_log.txt](https://github.com/snakemake/snakemake/files/12653967/gls_snakemake_log.txt)

```
stderr: |
        --2023-09-18 22:09:53--  https://raw.githubusercontent.com/snakemake/snakemake/main/snakemake/executors/google_lifesciences_helper.py
        Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.111.133, 185.199.108.133, 185.199.109.133, ...
        Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.111.133|:443... connected.
        HTTP request sent, awaiting response... 404 Not Found
        2023-09-18 22:09:53 ERROR 404: Not Found.

        tar (child): /tmp/workdir.tar.gz: Cannot open: No such file or directory
        tar (child): Error is not recoverable: exiting now
        tar: Child returned status 2
        tar: Error is not recoverable: exiting now
```

The URL is indeed missing, so I don't believe this is an issue with 7.32.3. 

The full output of ```gcloud beta lifesciences operations describe``` is attached

I realize that GLS is going to unfortunately stop being supported by Google soon, but it would be useful to know if this is true for snakemake in the short term. 

Thanks!

CowanCS1:
I think the recent changes to the API are interacting with a `wget -O /gls.py https://raw.githubusercontent.com/snakemake/snakemake/master/snakemake/executors/google_lifesciences_helper.py` that was previously hardcoded inside `snakemake/executors/google_lifesciences.py`. 

This file now lives over in [this plugin repo](https://github.com/snakemake/snakemake-executor-plugin-google-lifesciences/tree/main) - this is the same with other executors, and to my knowledge doesn't have anything to do with GLS being marked for end of life. 

The download link in the plugin repo appears to not be updated: `wget -O /gls.py https://raw.githubusercontent.com/snakemake/snakemake/main/snakemake/executors/google_lifesciences_helper.py` and it looked to me like test_google_lifesciences is not being run. 

[...]

PR #2461, opened 2023-09-28
fix: Restore backward compatibility for Google Life Sciences executor

CowanCS1:
### Description

fix: Previous versions of snakemake had a hardcoded link to the file `snakemake/executors/google_lifesciences_helper.py` which was moved in recent API changes. All previous versions of snakemake are now unable to run the Google Life Sciences executor. This pull request would temporarily add this file back in the "expected" location to provide backward compatibility for existing projects using previous versions of snakemake. This is a temporary fix which can be reverted once GLS is deprecated. #2451 

### QC

### DIFF
--- a/snakemake/executors/google_lifesciences_helper.py
+++ b/snakemake/executors/google_lifesciences_helper.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python
+
+# This is a helper script for the Google Life Sciences instance to be able to:
+# 1. download a blob from storage, which is required at the onset of the Snakemake
+#     gls.py download <bucket> <source> <destination>
+# workflow step to obtain the working directory.
+# 2. Upload logs back to storage (or some specified directory of files)
+#    gls.py save <bucket> <source-dir> <destination-dir>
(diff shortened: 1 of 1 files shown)


johanneskoester:
All the executors except local, touch and dryrun have been moved into plugins. There is also one for gls (https://GitHub.com/Snakemake/Snakemake-executor-Plugin-Google-lifesciences). All these changes will stabilize and work well again soon. Sorry for the hassle, but be assured it is worth it.

CowanCS1:
Hi Johannes,

I'm aware of the API changes and GLS plugin. I have an open pull request over there on the GLS plugin as well. 

The change proposed in this pull request isn't to address issues with the plugins, but rather to fix a (presumably unanticipated) issue that breaks all previous versions of the GLS executor. 

Very briefly, the helper file added back in this pull request was directly linked to by past versions of the GLS executor using a hyperlink that assumes it is on the main branch. By moving it, all previous GLS executors stopped working. See the discussion in #2451

johanneskoester:
Ah sorry, I missed your explanation! Indeed this is an uncritical merge.

Issue #1123, opened 2021-08-04
--print-compilation does not consider include statements


LustigePerson:
**Snakemake Version**
6.6.1

**Bug Description**
When using `include`to include another file with rules, `--print-compilation` is not listing the respective rules.
With the two files `Snakefie` and `rules/copy.smk`, e.g.:

```python
# Snakefile

SAMPLES = ["A", "B"]


rule all:
    input:
        expand("data/{sample}.out", sample=SAMPLES),


include: "rules/copy.smk"
```

```python
# copy.smk

rule copy_input:
    input:
        "data/{sample}.in",
    output:
        "data/{sample}.out",
    shell:
        "cp {input} {output}"

```

the result of `--print-compilation` is:

```bash
SAMPLES = [ "A" , "B" ]


@workflow.rule(name='all', lineno=17, snakefile='/home/reizbe/snakemake_test/workflow/Snakefile')

@workflow.input(
                                expand ( "data/{sample}.out" , sample = SAMPLES ) ,


)
@workflow.norun()
@workflow.run
def __rule_all(input, output, params, wildcards, threads, resources, log, version, rule, conda_env, container_img, singularity_args, use_singularity, env_modules, bench_record, jobid, is_shell, bench_iteration, cleanup_scripts, shadow_dir, edit_notebook, conda_base_path, basedir):
        pass
workflow.include( "rules/copy.smk"
)
```

**whereas in one file**
```python
SAMPLES = ["A", "B"]


rule all:
    input:
        expand("data/{sample}.out", sample=SAMPLES),


rule bwa_map:
    input:
        "data/{sample}.in",
    output:
        "data/{sample}.out",
    shell:
        "cp {input} {output}"
```

the Results of `--print-compilation` is:

```bash
SAMPLES = [ "A" , "B" ]


@workflow.rule(name='all', lineno=17, snakefile='/home/reizbe/snakemake_test/workflow/Snakefile')

@workflow.input(
                                expand ( "data/{sample}.out" , sample = SAMPLES ) ,


)
@workflow.norun()
@workflow.run
def __rule_all(input, output, params, wildcards, threads, resources, log, version, rule, conda_env, container_img, singularity_args, use_singularity, env_modules, bench_record, jobid, is_shell, bench_iteration, cleanup_scripts, shadow_dir, edit_notebook, conda_base_path, basedir):
        pass
[...]

PR #2469, opened 2023-10-04
Compiled  `workflow.include` should carry print_compilation keyword

kennyworkman:
### Description

As described here #1123, `--print-compilation` will display `workflow.include(<Snakefile>)` statements instead of the compiled code within these included Snakefiles as desired. We simply introduce `print_compilation` keywords to these compiled statements to fix this.

@johanneskoester can you please guide me as to the best place to introduce a test

### QC

### DIFF
--- a/snakemake/parser.py
+++ b/snakemake/parser.py
@@ -71,6 +71,7 @@ class StopAutomaton(Exception):
 class TokenAutomaton:
     subautomata: Dict[str, Any] = {}
     deprecated: Dict[str, str] = {}
+    print_compilation = False
 
     def __init__(self, snakefile: "Snakefile", base_indent=0, dedent=0, root=True):
         self.root = root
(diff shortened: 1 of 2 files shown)


kennyworkman:
Hi @johanneskoester, wanted to check in on this

johanneskoester:
TBH, I am hesitating a bit. If one is interested in the compilation code of an included Snakefile, one can always just run snakemake --print-compilation on that file. Seeing how include is compiled into Python tokens is also important. Hence, I would rather not like this to happen, not by default at least I think. However, happy to consider additional arguments I might be unaware of right now.

matthewfeickert:
@kennyworkman Just rebasing this to resolve conflicts from a PR of mine that was merged earlier.

To get this in your local repository, assuming that your fork is named `origin`, you can run the following commands:

```
git fetch --all --prune
git checkout kenny/fix-include-print-compilation
git reset --hard origin/kenny/fix-include-print-compilation
```

Hocnonsense:
Thanks for the contribution. However, I don't think this is the right approach: `--print-compilation` is intended to be a dry-run inspection tool and should not trigger execution (see #3041), and `include:` only takes effect after the included file is actually executed. So propagating `print_compilation` into `workflow.include` would conflate the two concerns.

As noted above, if you want to inspect the compiled output of an included Snakefile, running `snakemake --print-compilation` directly on that file is the right way to do it.

Closing for now — feel free to reopen if the branch is updated to address this.

PR #2474, opened 2023-10-10
Let wrapper log to tow files

SilasK:
### QC


SilasK:
This changes the behavior of test_get_log_complex, where the log from the wrapper is truncated.

I think, error s from the wrapper should always be visible.

SilasK:
This would be ready

Issue #2475, opened 2023-10-10
'inf' value cannot pass to R script


ghbore:
**Snakemake version**

7.32.4

**Describe the bug**

The infinite value from either the __config__ file:
```yaml
threshold:
  - -.inf
  - .inf
```
or from the rule parameter:
```python
params:
  threshold = [- float("inf"), float("inf")]
```
will raise an error when evoking __external R__ script:

> Error in initialize(value, ...) : object 'inf' not found
> Calls: Snakemake -> new -> initialize -> initialize
> Execution halted

**Logs**

> Error in initialize(value, ...) : object 'inf' not found
> Calls: Snakemake -> new -> initialize -> initialize
> Execution halted

**Minimal example**

```python
# Snakefile
rule all:
  output: "final.rst"
  params:
    threshold = [-float("inf"), float("inf")]
  script: "test.R"
```
```R
#! /usr/bin/env Rscript
# test.R

library(tidyverse)

run <- function (output, threshold){
  log <- str_c(
    str_glue("The type: {class(threshold[1])}.\n"),
    str_glue("The value: {threshold}.\n"),
  )
  conn <- file(output)
  writeLines(log, conn)
  close(conn)
}

run(
  snakemake@output[0],
  snakemake@params["threshold"]
)
```

**Additional context**

I checked the wrapper R script used for execution, and noticed grammar errors in the preamble code section:
  - `inf` should be `Inf` in R

```R
######## snakemake preamble start (automatically inserted, do not edit) ########
library(methods)
Snakemake <- setClass(
    "Snakemake",
    slots = c(
        input = "list",
        output = "list",
        params = "list",
        wildcards = "list",
        threads = "numeric",
        log = "list",
        resources = "list",
        config = "list",
        rule = "character",
        bench_iteration = "numeric",
        scriptdir = "character",
        source = "function"
    )
)
snakemake <- Snakemake(
    input = list(),
    output = list('final.rst'),
    params = list(c(-inf, inf), "threshold" = c(-inf, inf)),    <<<<<<-FIXME
    wildcards = list(),
    threads = 1,
    log = list(),
    resources = list('tmpdir', "tmpdir" = '/var/folders/kv/6fn35g1d7q7dg3_hh7bs8h300000gn/T'),
    config = list("threshold" = c(-inf, inf)),
    rule = 'all',
    bench_iteration = as.numeric(NA),
   ...
```

PR #2476, opened 2023-10-11
fix: 'infinite' value cannot pass to R script

ghbore:
### Description

An **Infinite** value will raise an error when evoking an external R script (#2475), because the current encoding schema ([snakemake/script.py REncoder::encode_value](https://github.com/snakemake/snakemake/blob/5e3a46476d78d5d52340a9ffa327d18a5e7e9828/snakemake/script.py#L181)) will translate Python `float("inf")` into `inf` for R, not the expected `Inf`.

To fix this, two specific encoding schemes are introduced to support the correct translation of Python `float("inf")` to R `Inf`, and Python `float("-inf")` to R `-Inf`.

### QC

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - series-7
   pull_request:
     branches_ignore: []
 jobs:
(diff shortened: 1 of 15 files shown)


padix-key:
There are also additional special values, like NaN that do not work properly in the R preamble: #2425

johanneskoester:
Thanks as lot! There are unfortunately many unrelated changes here. I have therefore cleaned it up in #3238, but while giving you credit there.

PR #2485, opened 2023-10-18
fix: handle different f-string tokens in py3.12

### LINKED ISSUE #2480: Python 3.12 break f string inside Snakemake
**Snakemake version**
Tested with snakemake 7.34.2
(older version are affected, too. Tested down to 7.18.2)

Affected Python version: 3.12.0
(Last known good: Python 3.11.6)

**Describe the bug**

When snakemake is installed with python version 3.12.0, f-strings get unexpected extra space.

For example:
```python
output: f'{PREFIX}.txt'
```
will produce...

**Logs**

...this output in the log (note the unexpected extra spaces):
```text
Building DAG of jobs...
File path ' SID23454678 .txt ' starts with whitespace. This is likely unintended. It can also lead to inconsistent results of the file-matching approach used by Snakemake.
File path ' SID23454678 .txt ' ends with whitespace. This is likely unintended. It can also lead to inconsistent results of the file-matching approach used by Snakemake.
```
When run on python 3.12.0

**Minimal example**

UnitTest.smk:
```python
#!/usr/bin/env python3
PREFIX = 'SID23454678'
rule unit1:
 input:
 output:
  f'{PREFIX}.txt'
 run:     
   "sleep 1"
```

Running with:
```shell
snakemake --dry-run -s UnitTest.smk --cores 1
```

In this environment... :
```shell
mamba create -n snake-test -c conda-forge -c bioconda snakemake-minimal=7.32.4 python=3.11
```
...everything work as expected:
```text
[Thu Oct 12 15:55:02 2023]
rule unit1:
    output: SID23454678.txt
    jobid: 0
    reason: Missing output files: SID23454678.txt
    resources: tmpdir=/tmp
```

But in that environment...:
```shell
mamba create -n snake-test -c conda-forge -c bioconda snakemake-minimal=7.32.4 python=3.12
```
...the bug is triggered:
```text
Building DAG of jobs...
File path ' SID23454678 .txt ' starts with whitespace. This is likely unintended. It can also lead to inconsistent results of the file-matching approach used by Snakemake.
File path ' SID23454678 .txt ' ends with whitespace. This is likely unintended. It can also lead to inconsistent results of the file-matching approach used by Snakemake.
```

**Additional context**

Apparently, the f-string have been [completely overhauled as part of python 3.12](https://realpython.com/python312-f-strings/). I suspect the new f-string parser somehow clashes with the parse of snakemake files.

Hocnonsense:
### Description

fix #2480 

### QC

In this fix, f-string such as `f"{any}"` will be formated as `f"{ any }"`.

However, f-string such as `f"{1+1=}"` will be formated as `f"{ 1 + 1 = }"`.
in a pure python environment, `f"{1+1=}"` == `"1+1=2"`, while `f"{ 1 + 1 = }"` == `" 1 + 1 = 2"`.
As f-string such as `f"{1+1=}"` cannot be recognized by snakefmt, I didn't try to handle this problem


johanneskoester (changes requested):

Hocnonsense:
Ah, It seems that conda cannot install mamba with python=3.12

Now. I'm trying to install conda only via micromamba. However, snakemake will break when using conda in PATH.

johanneskoester:
Mhm, mamba and python 3.12 work fine together. I just tested installing them into the same env on my local machine.

johanneskoester:
I don't get all those changes here. Can you give me a summary? Why are they needed? If it is just for CWL, you can also rather remove cwl stuff from the test env and we install it dynamically in the python 3.11 case.

Hocnonsense:
The new commit (54d893b6456644957d64a5916458b9c3e0c08339) is just for trying to use conda in python3.12 environment. However, it failed. I think it is necessary to revert this commit.

johanneskoester:
No need to worry about the history here, we anyway do a squash merge in the end. So, just edit as you need.

johanneskoester (approved):

Issue #2473, opened 2023-10-08
On windows, the new storage plugin functionality seems to fail to properly match s3 objects between input and output of rules


johanneskoester:
**Snakemake version**
Snakemake 8.0 (in development)

**Describe the bug**
On windows, the new storage plugin functionality seems to fail to properly match s3 objects between input and output of rules

**Logs**
```
# snakemake.exceptions.MissingInputException: Missing input files for rule all:
# affected files:
#   s3://snakemake-ec6c5e4c267d4a1e890976b57c99ee55/test2.out (storage)
#
```

**Minimal example**
This can for example be observed when activating the test case test_storage on windows.

@melund, your help would be greatly appreciated! (at the moment, the test fails in PR #5440, but I will merge that PR ASAP even before this is fixed in order to proceed with the finalization of Snakemake 8.0).

melund:
Yes. Just merge it. I will be happy to help fix the issue on Windows if I can. However, I will not have time to really look into it until next weekend.

johanneskoester:
Thank you so much! PR is merged, the problematic test is https://github.com/snakemake/snakemake/blob/fb65c3300c0c509a702e3dc01666fd86bab57e7c/tests/tests.py#L655

melund:
@johanneskoester. What do you suspect. Is it snakemake itself or the new storage-plugin that is causing the bug?

melund:
@johanneskoester. I gave a look while the kids were sleeping. But it took a me a while to get into the code base again. Just getting the test running locally proved difficult.

Turns out that "Git for Windows" have reorganized its internal unix tools so they are no longer on path (Things like "cp", "touch", "cat" etc.) I guess it works in the cloud because the GitHub actions container VM comes prepackaged with all the tools MSYS2 tools on path. 

I [fixed that](https://github.com/snakemake/snakemake/compare/main...melund:snakemake:master) and got the tests suite running, but for some reason a few tests still fail for me:

```
FAILED tests/tests.py::test_glpk_solver - AssertionError: expected successful execution
FAILED tests/tests.py::test_output_file_cache - AssertionError: expected successful execution
FAILED tests/tests.py::test_module_complex - AssertionError: expected successful execution
FAILED tests/tests.py::test_module_complex2 - AssertionError: expected successful execution
FAILED tests/tests.py::test_module_with_script - AssertionError: expected successful execution
```

I will disregard that for now and begin looking at the `test_storage` test. 

[...]

PR #2487, opened 2023-10-20
fix: Storage plugin bug on Windows 

melund:
### Description

This PR attempts to fix the storage bug plugin test failures on Windows. 
Fixes: #2473

@johanneskoester, I am not sure this is how you want to fix the bug, but it should give you an idea what the problem is. 

Currently, on Windows all files in rules etc. are converted to forward slashes to avoid the 'backslash' mesh. However, the new storage plugin uses 'pathlib' for some paths internally. Converting pathlib.Path objects to strings gives paths with 'backslash'es on Windows. So I have added code to handle those cases. 

I have also improved the CI test setup on windows.  The setup time for conda environment creation is down from 12 min to 2.5 min on Windows. Much more is possible, but I have split that work into a separate PR (#2489)

The PR also does a few other things for local testing on Windows:

> Turns ou that "Git for Windows" have reorganized its internal Unix tools so they are no longer on path (Things like "cp", "touch", "cat" etc.) I guess it works in the cloud because the GitHub actions VM comes prepackaged with all tools on path from MSYS package. 
>
> A few tests still fail for me locally but work in the cloud, but I haven't dug into that.
### QC


.gitattributes
```diff
@@ -1,3 +1,5 @@
+* text=auto eol=lf
```
melund, line 1: This ensures files are checked-out with linux file endings on Windows.That is usually not the recommendation for windows. But this is the best option given how many input-output files in the tests system that are manipulated by posix tools.

pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+python_files = tests/tests.py tests/test_expand.py tests/test_io.py tests/test_schema.py tests/test_linting.py tests/test_executor_test_suite.py
```
melund, line 2: The test suite contains plenty of files which doesn't follow the standard pytest conventions. This tripped the vs-code test plugin which tried to inspect non-test files which it thought caused it to crash.

melund:
@johanneskoester. Test pass and it is ready for review.

johanneskoester (changes requested):
Thanks a lot! Sorry, the review was done since last week, but I forgot to hit the submit button. One thing below.

snakemake/io.py
```diff
@@ -733,10 +733,10 @@ def apply_wildcards(self, wildcards):
...
                     rule=self.rule,
                 )
-
-            file_with_wildcards_applied = IOFile(
-                str(storage_object.local_path()), rule=self.rule
-            )
+            local_path_str = str(storage_object.local_path())
+            if ON_WINDOWS:
+                local_path_str = local_path_str.replace("\\", "/")
+            file_with_wildcards_applied = IOFile(local_path_str, rule=self.rule)
```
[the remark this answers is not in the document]
melund: @johanneskoester, and have done that and pushed a new version.

johanneskoester:
Seems like touching directories on windows does not work anymore with the changes here.

melund:
> Seems like touching directories on windows does not work anymore with the changes here.

I will try to roll back so the check doesn't happen inside IOFile. The test were able to pass before I did that.

snakemake/io.py
johanneskoester, line 739: The Path object has a method `as_posix()` which does exactly what you do here. I don't understand though why we need this at all. What is the problem with having a windows path returned here?

johanneskoester (changes requested):

johanneskoester:
I have now implemented a simpler version of this in #2519. Thanks a lot for this work here, which pointer me to the right places for a simplified fix!

PR #2489, opened 2023-10-21
feat: Faster ci test setup 

melund:
### Description

This PR speeds up the CI tests.

| Before: | After |
| --- | ----| 
| image |  image |

This was done by using [setup-micromamba](https://github.com/mamba-org/setup-micromamba) to install mamba and the Python environment. This is especially fast because the environment is cached. 

Here are the speed improvements for micromamba install + environment creation:

| Platform |  Before: | After: |
| --- | ----| ---- | 
| Linux | 3 min | 11 sec |
| Win | 12 min | 2.5 min | 

(maybe snakemake should use micromamba for its conda-jobs)...

The black formatting job is also much faster: Total time: 60 sec -> 11 sec.

A small speed up also comes from replacing the `cancel-previous` job with a github-action built-feature (`concurrency`) which does the same. 

Finally, a lot of speedup comes from adding a `.test_duration` file so `pytest-split` can minimize test time by spreading the tests across jobs in an optimal way. This cuts 30% of the actual test time.

### QC


melund:
@johanneskoester. This is ready for review.

johanneskoester (approved):
Awesome, thanks a lot!

.test_durations
```diff
@@ -0,0 +1,308 @@
+{
```
johanneskoester, line 1: I guess new tests that do not occur in this file are just put in one of the buckets and not ignored? Is there a way to update the durations file from the main branch, via a separate job that records and updates the timings?

melund: They are not ignored. New tests are just assumed to have an average duration. So if the test suite changes significantly it can be good to update the file. It may also be smart to keep separate duration files for linux and windows.

 > Is there a way to update the durations file from the main branch, via a separate job that records and updates the timings?
 
 Yes. That would be easy enough. I could make separate workflow which would grab the duration files as artifacts, combine them to a single file and then make automated pull request. I know conda does something similar. Maybe it could run once a week/month. 

I will make a follow-up PR if you think it is worth it?

johanneskoester: Yes, sounds great and would be much appreciated!

PR #2509, opened 2023-11-10
feat: add function get unnamed arguments from name

SilasK:
### Description

Follow up from https://github.com/snakemake/snakemake/pull/2474

### QC

### DIFF
--- a/snakemake/io.py
+++ b/snakemake/io.py
@@ -1543,6 +1543,21 @@ class Namedlist(list):
             elif i == index:
                 self._set_name(name, i, end=i + len(items))
 
+    def _get_unnamed_arguments(self):
+        """
+        Get unnamed arguments and return them as a list.
+        """
+
+        keys_with_positions = self._names
+        if len(keys_with_positions) == 0:
+            # all arguments are unnamed
+            return list(self)
+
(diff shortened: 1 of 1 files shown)


snakemake/io.py
```diff
@@ -1517,6 +1517,21 @@ def _insert_items(self, index, items):
             elif i == index:
                 self._set_name(name, i, end=i + len(items))
 
+    def _get_unnamed_arguments(self):
```
johanneskoester, line 1520: This should not be private as we want to use it in snakemake wrappers.

johanneskoester: I would suggest `def get_positional_items(self) -> Sequence[str]:`.

johanneskoester (changes requested):

SilasK:
Name changed nas requested. Do I need to write a test case?

snakemake/io.py
SilasK: Name changed as requested.

johanneskoester:
> Name changed nas requested. Do I need to write a test case?

That would be awesome!

fgvieira:
Can you fix the conflicts?

PR #2521, opened 2023-11-22
feat: allow python expressions in --set-resources

johanneskoester:
### QC

### DIFF
--- a/docs/executing/cli.rst
+++ b/docs/executing/cli.rst
@@ -115,8 +115,8 @@ Therefore, since Snakemake 4.1, it is possible to specify configuration profiles
 to be used to obtain default options.
 Since Snakemake 7.29, two kinds of profiles are supported:
 
-* A global profile that is defined in a system-wide or user-specific configuration directory (on Linux, this will be ``$HOME/.config/snakemake`` and ``/etc/xdg/snakemake``, you can find the answer for your system via ``snakemake --help``).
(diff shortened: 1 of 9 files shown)


dlaehnemann:
I now added and cleaned up some of the documentation. Also, I amended the profile test case to cover the different profile resource keywords and actually test them. This exposed two things for `--set-threads`:

1. Using a callable as suggested in the docs for `threads:`, does not work for `set-threads:` in a profile. I am using `input.size_mb`, as the docs suggest that only `wildcards, [input]` are valid usable arguments for the callable here.
2. When only specifying a number for a `rule x` via `set-threads:`, this does get parsed correctly. But interestingly, it only gets used, if the `rule x` already has a value specified with the `threads:` directive. Otherwise, whatever you specify via `set-threads:` in the profile is just ignored.

I think these two things make sense to already fix here, right?

johanneskoester:
> I now added and cleaned up some of the documentation. Also, I amended the profile test case to cover the different profile resource keywords and actually test them. This exposed two things for `--set-threads`:
> 
>     1. Using a callable as suggested in the docs for `threads:`, does not work for `set-threads:` in a profile. I am using `input.size_mb`, as the docs suggest that only `wildcards, [input]` are valid usable arguments for the callable here.
> 
>     2. When only specifying a number for a `rule x` via `set-threads:`, this does get parsed correctly. But interestingly, it only gets used, if the `rule x` already has a value specified with the `threads:` directive. Otherwise, whatever you specify via `set-threads:` in the profile is just ignored.
> 
> 
> I think these two things make sense to already fix here, right?

Yes, fixing it already makes sense.

PR #2537, opened 2023-12-07
fix: Slack log service

vincentczhou:
### Description

This updates the Slack ``--log-service``  feature added in #170.
The ``Slacker`` package is no longer maintained, and legacy slack tokens are no longer generatable.
I've replace it with the officially supported [Slack SDK](https://github.com/slackapi/python-slack-sdk) (see ``setup.cfg``), preserving the original functionality with minimal adjustments.

This has been minimally tested for functionality. However, the original implementation has a couple issues (which this PR **does not** address) - 

1. Lack of `docs/` on how to set up the ``--log-service slack`` feature via Slack
2. Logic is not robust - e.g. when `msg["done"] == msg["total"]`, this does not mean the workflow is complete (checkpoints)
3. The implementation of  ``--log-service slack`` is synchronous, meaning that Snakemake workflows will wait for the implemented `log_handler` function to finish. For normal print statements, this isn't an issue, but repeated calls of Slack WebAPI's `postMessage` can significantly slow things down. Depending on the workflow (Snakefile), this can become noticeable.
4. The implementation does not comply with Slack WebAPI's rate limiting.

The current implementation will look like this - messages on slack are sent by the user, using a user OAuth token.
image

I've already addressed points 2, 3, and 4 in a private repository using ``--log-handler`` that should be integrable into this current implementation, but it has not been robustly tested (integration can happen at a later PR).
image

Thanks!

### QC

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,7 +63,7 @@ install_requires =
     conda-inject >=1.3.1,<2.0
 
 [options.extras_require]
-messaging = slacker
+messaging = slack_sdk
 pep =
     eido
     peppy
--- a/snakemake/logging.py
+++ b/snakemake/logging.py
@@ -118,9 +118,7 @@ class SlackLogger:
 
         if msg["level"] == "progress" and msg["done"] == msg["total"]:
             # workflow finished
-            self.slack.chat_postMessage(
-                channel=self.own_id, text="Workflow complete."
-            )
+            self.slack.chat_postMessage(channel=self.own_id, text="Workflow complete.")
 
 
 class WMSLogger:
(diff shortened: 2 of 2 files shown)


johanneskoester:
Cool! Log handlers like this are also candidates for turning them into plugins. But I first have to design the interface for that. Of course, if you are interested, you could also give this a try. One would basically follow the same approach as with snakemake-interface-storage-plugins. We can however merge this first. May I ask you to format the code with black?

christopher-schroeder (approved):

johanneskoester (approved):

PR #2547, opened 2023-12-19
fix: use conda as the default frontend

victorlin:
### Description

With conda's October 2023 release¹, the default solver uses the same backend as mamba, making it equally fast.

¹ https://github.com/conda/conda/blob/f9fe520956490f7d218abbe6f6fdc0c361535847/CHANGELOG.md#23100-2023-10-30

### QC

_This PR has been opened for discussion. Further changes can be made if this seems like a good direction._

### DIFF
--- a/snakemake/cli.py
+++ b/snakemake/cli.py
@@ -1570,10 +1570,9 @@ def get_argument_parser(profiles=None):
     )
     group_conda.add_argument(
         "--conda-frontend",
-        default="mamba",
+        default="conda",
         choices=["conda", "mamba"],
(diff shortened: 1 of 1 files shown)


BEFH:
Are you sure they are equally fast? The solver isn't the only requirement for building environments

victorlin:
@BEFH No, I haven't done any benchmarking. I imagine it's comparable though. The benefit would be one less runtime dependency (mamba).

johanneskoester:
Thanks, this is implemented since a while in the main branch, sorry for missing this PR.

PR #2551, opened 2023-12-21
feat: add --sdm short opt for --deployment

mbhall88:
### Description

The new `--software-deployment-method`/`--deployment-method`/`--deployment` is probably a heavily used option for most users, so I think it would benefit from a shorthand option as it is quite verbose. Seems `-m` is available and is kind of related.

I've also updated the docs with some references to this shorthand version and replaced some longform usage with the short option, ensuring it is made clear what the shorthand option is an alias for. Happy to change this though.

### QC


johanneskoester:
Good idea, but I think I would prefer `--sdm` instead (instead of a single letter code). THis way, it should be easier to remember and decipher.

mbhall88:
> Good idea, but I think I would prefer `--sdm` instead (instead of a single letter code). THis way, it should be easier to remember and decipher.

Done

mbhall88:
@johanneskoester i didn’t really touch any code so I don’t know why test are failing…

mbhall88:
Ah it was because when I went from `-m` to `--sdm` it renames the option to `args.sdm` rather than `args.software_deployment_method`

johanneskoester:
Thanks!

Issue #2574, opened 2024-01-03
Wrong error line numbers reported in version 8.0.1


yardimcilab:
**Snakemake version** 8.0.1

**Describe the bug**
As described in [this issue](https://github.com/snakemake/snakemake/issues/1256), Snakemake reports incorrect line numbers for errors.

**Minimal example**
1. Create the following "Snakefile":
```
rule all:
    input:
        "complete"

rule start:
    output:
        "complete"
    params:
        a = 1    # Edit - this syntax error (the missing comma) is deliberate, generating the erroneous line number in the error message.
        b = 2
    shell:
        "touch complete"
```
2. run snakemake --cores 1

This generates the error:
`SyntaxError in file [path]/Snakefile, line 20:
invalid syntax. Perhaps you forgot a comma? (Snakefile, line 20)`

It is the fact that the error message states "line 20" rather than "line 9" or "line 10" that is the issue I am pointing out.

The file is only 12 lines long and the SyntaxError actually occurs on line 9/10 with the deliberately introduced "forgotten" comma, which is meant to generate the SyntaxError for illustrative purposes.

Hocnonsense:
sure, a comma is required after `a＝1`

In fact, these several lines will be wrapped as something like `workflow.params(a＝1,b＝2)`

yardimcilab:
The syntax error is deliberately introduced to illustrate that snakemake is giving an erroneous line number. The erroneous line number is the bug, not the illustrative syntax error.

dariober:
I confirm the issue has resurfaced and I reopened #1256. It would be good to have a test checking for this bug. If I get a bit of time I'll try to submit a PR with it.

@yardimcilab your link to issue #1256 seems to be broken, can you fix it?

dariober:
I suppose this issue can be closed now with version 8.1.0?

y9c:
This bug still exists in the latest version (8.2.0)

dariober:
@y9c : Version 8.2.0 works fine with the simple test file below from #1256. Can you provide an example reproducing the bug?

---

```
rule all:
    input:
        'done.txt',
        
rule trim_transcripts:
    output:
        done.txt,
    run:
        if 1 = 2:  # Syntax error: It should be 1 == 2
            print('foo')
```

```
snakemake -v
8.2.0

snakemake -p -n -j 1
SyntaxError in file /home/dario/Downloads/Snakefile, line 9:
[...]

Issue #1256, opened 2021-11-11
Line number of error is incorrect


dariober:
**Snakemake version**: 6.10.0 installed via bioconda/mamba

**Describe the bug**

When a snakefile contains a syntax error or some other bug,  snakemake reports the wrong line number of the error. More specifically, it seems to me that:

* The reported line number is consistently after the actual line number
* As the Snakefile becomes longer, the difference between reported and actual line number grows
* Earlier versions do not show this bug (e.g. v5.4.3 works)

This is quite a problem since it makes it difficult to locate bugs!

**Minimal example**

Here the syntax error is on **line 9**:

```
rule all:
    input:
        'done.txt',
        
rule trim_transcripts:
    output:
        done.txt,
    run:
        if 1 = 2:  # Syntax error: It should be 1 == 2
            print('foo')
```

Snakemake 6.10.0 reports the error on **line 18**:

```
snakemake -j 1 -n -s Snakefile2
SyntaxError in line 18 of /export/III-data/wcmp_bioinformatics/db291g/git_repos/wcip/20211109_kasia_asaj_snp/Snakefile2:
invalid syntax (Snakefile2, line 18)
```

Snakemake 5.4.3 is correct:

```
/export/projects/III-data/wcmp_bioinformatics/db291g/miniconda3/envs/20190312_andy_alba3_motifs/bin/snakemake -j 1 -n -s Snakefile2
SyntaxError in line 9 of /export/III-data/wcmp_bioinformatics/db291g/git_repos/wcip/20211109_kasia_asaj_snp/Snakefile2:
invalid syntax
```

Linkrw12:
This also exists in Snakemake 6.12.1

dariober:
I think it's been fixed in v6.12.3 (thanks!) - please correct me if I'm wrong...

yardimcilab:
This error persists in version 8.0.1.

PR #2583, opened 2024-01-05
test: Add test for issue #1256 and #2574

dariober:
### Description

Add test for issues #1256 and #2574.

The test Snakefile has a syntax error on line 9. Some versions of Snakemake, including 8.0.1, incorrectly report the error on line 20. I'm only submitting the test case, not the fix.

### QC


johanneskoester:
Thank you! I merge this into a temporary branch now, in order to include it in PR #2584.

PR #2597, opened 2024-01-09
feat: allow definition of access pattern for input files

johanneskoester:
### QC


cmeesters:
This PR is feature incomplete. Shall we find a time to complete it? I don't feel confident enough to proceed.

johanneskoester:
Already implemented in a different PR.

PR #2613, opened 2024-01-12
fix: Handle pulp future backward compatibility.

henrykironde:
### QC

### DIFF
--- a/snakemake/cli.py
+++ b/snakemake/cli.py
@@ -767,7 +767,12 @@ def get_argument_parser(profiles=None):
     try:
         import pulp
 
-        lp_solvers = pulp.listSolvers(onlyAvailable=True)
+        # Check if pulp version is 2.7.0 or above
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -46,7 +46,7 @@ install_requires =
     nbformat
     packaging
     psutil
-    pulp >=2.3.1,<2.9 # pulp introduced a breaking change from 2.7 to 2.8. Hence we should always pin to the minor version.
+    pulp >=2.0
```
johanneskoester, line 49: Mhm, I get the idea, but who knows what pulp will break next on a minor release. Check poetry, it is nowadays default behavior to be conservative with dependency pinning. I think we should gradually do that for all dependencies and rather rely on dependabot for updating them.

johanneskoester:
Closing in favor of the mentioned strategy then.

PR #2627, opened 2024-01-17
fix: fix string resource definition in CLI and profile

cmeesters:
### Description

The PR is related to the[ issue described for the slurm submit executor](https://github.com/snakemake/snakemake-executor-plugin-slurm/issues/18) plugin to not correctly hand over arguments. 

The test folder is named `test_slurm_resource_propagation meesters` and may not be complete, as it covers multiple related scenarios to provoke the error. 

Note, that the tests were carried out when using a cluster:

- `$ snakemake --executor slurm -j2 --workflow-profile ./profiles/ --default-resources slurm_account=<account>`
  yields
  ``Error:
  WorkflowError:
    Failed to evaluate default resources value '<function eval_resource_expression.<locals>.callable at 0x7f5c19ed0e00>'.``` 
    
    when the patition is defined at the rule level in a profile. (see the directory of the test case)
    
- likewise `$ snakemake --executor slurm -j2 --workflow-profile ./profiles/ --default-resources slurm_account=<account>  --set-resources slurm_extra="--nice=150"`
  yields
  ```ValueError: Invalid resource definition: entries have to be defined as RULE:RESOURCE=VALUE, with VALUE being a positive integer a quoted string, or a Python expression (e.g. min(max(2*input.size_mb, 1000), 8000)).``` 
  
These errors can both be triggered on the command line and in the config file, of course.

### QC


cmeesters:
@johanneskoester does this make sense? Any progress on this side?

johanneskoester (approved):

cmeesters:
Thank you for the fix! Alas, case no. 1 is not covered, yet - as it seems.

Issue #2648, opened 2024-01-25
Python 3.12 f-strings still introduce unwanted whitespace in Snakemake 8


sdruskat:
**Snakemake version**
- Snakemake 8.2.3 from bioconda
- Python 3.12.1

**Describe the bug**
- It was [reported in #2480](https://github.com/snakemake/snakemake/issues/2480#issuecomment-1882770815) that Snakemake >8 fixes bugs in handling Python 3.12 f-strings, and pull requests addressing this were merged (#2485, with the release chore PR #2401 still open).
- However, when using f-strings with the above-mentioned setup, whitespace is still being introduced.

**Logs**
```bash
Building DAG of jobs...
Retrieving input from storage.
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job      count
-----  -------
all          1
total        1

Select jobs to execute...
Execute 1 jobs...

[Thu Jan 25 13:24:27 2024]
localrule all:
    output: out/ test.out
    jobid: 0
    reason: Missing output files: out/ test.out     <---------------- Added whitespace
    resources: tmpdir=/tmp

/usr/bin/bash: line 1: out/: Is a directory         <---------------- Actual error
[Thu Jan 25 13:24:27 2024]
Error in rule all:
    jobid: 0
    output: out/ test.out                           <---------------- Added whitespace
    shell:
        echo "SUCCESS!" > out/ test.out             <---------------- Added whitespace
        (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)

Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: .snakemake/log/2024-01-25T132427.614936.snakemake.log
WorkflowError:
At least one job did not complete successfully.
```

**Minimal example**
```bash
conda create -n snakemake8 -c bioconda -c conda-forge python=3.12 snakemake=8.2.3
```

```python
NAME = 'test'
OUTDIR = "out/"

rule all:
    output:
        f"{OUTDIR}{NAME}.out"
    shell:
        'echo "SUCCESS!" > {output}'
```

```bash
snakemake -c 1
```

**Additional context**
```bash
Python 3.12.1 | packaged by conda-forge | (main, Dec 23 2023, 08:03:24) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> s1 = "out/"
>>> s2 = "test"
[...]

PR #2649, opened 2024-01-25
fix: f-string in a more robust style?

Hocnonsense:
### Description

fix: #2648

now, the f-string will never be parsed by `self.state(token)`

### QC

### DIFF
--- a/snakemake/parser.py
+++ b/snakemake/parser.py
@@ -62,6 +62,10 @@ def is_string(token):
     return token.type == tokenize.STRING
 
 
+def is_fstring(token):
+    if sys.version_info >= (3, 12):
+        return token.type == tokenize.FSTRING_START
+
 def is_eof(token):
     return token.type == tokenize.ENDMARKER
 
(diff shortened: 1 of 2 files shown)


snakemake/parser.py
```diff
@@ -103,11 +107,33 @@ def indentation(self, token):
...
+                isin_fstring -= 1
+                t += t1.string
+            elif t1.type == tokenize.FSTRING_MIDDLE:
+                t += t1.string.replace("{", "{{").replace("}", "}}")
+            else:
+                t += t1.string
+            if isin_fstring == 0:
+                break
+        if hasattr(self, "cmd") and self.cmd[-1][1] == token:
+            self.cmd[-1] = t, token
```
johanneskoester, line 127: Can you please add a comment why this works for both >=3.12 and <3.12?

johanneskoester (changes requested):

snakemake/parser.py
Hocnonsense: Line 113-127 will never be touched in python<3.12, because in L135, it will test `is_fstring(token)`, which `return sys.version_info >= (3, 12) and token.type == tokenize.FSTRING_START`

johanneskoester: Thanks!

johanneskoester (approved):

sdruskat:
Beautiful, thanks @Hocnonsense for the quick fix and @johanneskoester for the quick merge.
Is there an ETA for a v8.3.3 including this fix in bioconda?

PR #2659, opened 2024-01-30
added support for comma in wildcards

Feelx234:
### Description

This is one part of two PR necessary to allow for commas in wildcards. This PR introduces the necessary decoding changes. 

The test case will currently fail as the other PR is necessary to properly handle the encoding.

### QC

### DIFF
--- a/snakemake/target_jobs.py
+++ b/snakemake/target_jobs.py
@@ -1,6 +1,6 @@
 from collections import namedtuple
 import typing
-
+import re
 from snakemake_interface_executor_plugins.utils import TargetSpec
 
 from snakemake.common import parse_key_value_arg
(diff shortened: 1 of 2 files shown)


johanneskoester:
I don't really get it. Commas in wildcards are already supported now:
```
rule a:
    output:
	"test.{foo}.txt"
    shell:
	"touch {output}"
```
with
```
snakemake -n test.abc,d.txt
```
works just fine.

Feelx234:
Hey johannes,

Thanks for your reply. Apparently my test case was ill chosen. It took me a moment to figure out what went wrong. Apparently for the "bug" to show up we need to have a rule calling another rule and the other rule needs to have a `run` field rather than a `shell` field.
```python
import subprocess

rule a:
    output:
	    "test2/{bar}_{i}.txt"
    run:
        subprocess.run(["touch", output[0]])

rule b:
    input:
        "test2/{bar}_1.txt"
    output:
        "test_{bar}.txt",
    shell:
        'touch "{output[0]}"'

```

Now if we do a dry run it works
`snakemake "test_abcd,.txt" --cores 1 -n`
but the actual execution errors
`snakemake "test_abcd,.txt" --cores 1`
with the traceback
```
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "(...)/snakemake/__main__.py", line 4, in <module>
    main()
  File "(...)/snakemake/__init__.py", line 3167, in main
    target_jobs=parse_target_jobs_cli_args(args),
  File "(...)/snakemake/target_jobs.py", line 21, in parse_target_jobs_cli_args
    wildcards = dict(
  File "(...)/snakemake/target_jobs.py", line 22, in <genexpr>
    parse_wildcard(entry) for entry in wildcards.split(",")
  File "(...)/snakemake/target_jobs.py", line 19, in parse_wildcard
    return parse_key_value_arg(entry, errmsg)
  File "(...)/snakemake/common/__init__.py", line 49, in parse_key_value_arg
    raise ValueError(errmsg + f" (Unparseable value: {repr(arg)})")
ValueError: Invalid target wildcards definition: entries have to be defined as WILDCARD=VALUE pairs (Unparseable value: '')
```


Also if we have no comma it works: ;)

`snakemake "test_abcd.txt" --cores 1`

Issue #2701, opened 2024-02-15
Snakemake 8 not usable with Slurm clusters offering GPUs


w8jcik:
In order to use GPUs one needs to be able to ask Slurm for a GPU.

The usual way to do this is `--gres gpu:1` or `--gpus 1` switches, which both go to `slurm_extra` in the Snakemake profile.

`slurm_extra` does not work in Snakemake 8, it is rejected.

You might be aware of the issue, but maybe you think it is already fixed.

I am creating this issue as a reminder.

Prior discussion https://github.com/snakemake/snakemake-executor-plugin-slurm/issues/18

Some workflows really need GPUs, otherwise they do not finish in a feasible time.

cmeesters:
As far as I can see, a profile with

```
slurm_extra: "'--gres gpu:1'" 
```
gets accepted on submission, but fails in a job context, right?

cmeesters:
I will prepare a simple PR.

w8jcik:
> As far as I can see, a profile with
> 
> ```
> slurm_extra: "'--gres gpu:1'" 
> ```
> 
> gets accepted on submission, but fails in a job context, right?

That is correct.

w8jcik:
Was this fixed in https://github.com/snakemake/snakemake/pull/2711?

I have tried the fix by updating Snakemake to 8.4.12

1. With following profile:

```
default-resources:
  ...
  slurm_extra: "'--gres=gpu:1'"
```

it fails when it gets a slot in the Slurm queue

```
Error:
  WorkflowError:
    Failed to evaluate resources value '--gres=gpu:1'.
        String arguments may need additional quoting. E.g.: --default-resources "tmpdir='/home/user/tmp'" or --set-resources "compute1:slurm_extra='--nice=100'".
    SyntaxError: invalid syntax (<string>, line 1)
```

2. With following profile

```
default-resources:
  ...
  slurm_extra: "--gres=gpu:1"
```

it fails right away

```
Error:
  WorkflowError:
    Failed to evaluate resources value '--gres=gpu:1'.
        String arguments may need additional quoting. E.g.: --default-resources "tmpdir='/home/user/tmp'" or --set-resources "compute1:slurm_extra='--nice=100'".
```

w8jcik:
> Please try: `slurm_extra: "'--gres=gpu:1'"`.

I am not sure if I understand. This is exactly what I used, am I missing some detail?

cmeesters:
[...]

PR #2702, opened 2024-02-16
fix: issue #2701

cmeesters:
## Description

The issue is caused by the attempt to parse CLI strings like

'--set-resources "<rule>:slurm_extra='--flag=arg'"'

The solution is simply to ignore all values starting with a hyphen.

Will also fix issues #18 and #19 of the SLURM executor plugin.

### QC

The PR should be covered by the tests - unsure why it has not been spotted, so far.

### DIFF
--- a/snakemake/resources.py
+++ b/snakemake/resources.py
@@ -523,7 +523,7 @@ def eval_resource_expression(val, threads_arg=True):
         # the form of '--flag=arg' non-Snakemake parameters
         # require parsing, which is not possible. They also do
         # not play any role after submission. Hence, we do:
-        if val.startswith("-"):
+        if "-" in val:
             return None
         if threads_arg:
             args["threads"] = kwargs["threads"]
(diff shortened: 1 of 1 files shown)


snakemake/resources.py
```diff
@@ -518,6 +518,13 @@ def generic_callable(val, threads_arg, **kwargs):
             "attempt": kwargs["attempt"],
             "system_tmpdir": tempfile.gettempdir(),
         }
+        # As the 'slurm_extra' parameter from the SLURM
+        # executor plugin permits full strings flags in
+        # the form of '--flag=arg' non-Snakemake parameters
+        # require parsing, which is not possible. They also do
+        # not play any role after submission. Hence, we do:
+        if val.lstrip().startswith("-"):
+            return None
```
johanneskoester, line 527: Mhm, sorry, but I don't understand the explanation yet.

johanneskoester:
Thanks again for the exploration of the issue! Closing in favor of fixing the primary cause in #2711.

PR #2716, opened 2024-02-22
fix: proper interpretation of standard resources given as strings (e.g. runtime as '5m'). Avoid the need to set additional quotes around size or timespan resources. Improved error messages for resource handling. 

cmeesters:
### Description

Eventually, human-readable resource parsing for runtimes from workflow-profiles ought to be enabled.

### QC

* [ x] The PR contains a test case for the changes, or the changes are already covered by an existing test case.


cmeesters:
Breaks as expected.

Issue #1697, opened 2022-06-01
Report generation fails when script path contains wildcard


gibsramen:
**Snakemake version**

7.8.1

**Describe the bug**

When a rule has a wildcard in the path to an external script, `--report` is unable to parse the script path for inclusion.

**Logs**

<details><summary>Snakemake execution</summary>

```
$ snakemake --cores 1
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job           count    min threads    max threads
----------  -------  -------------  -------------
all               1              1              1
use_script        2              1              1
total             3              1              1

Select jobs to execute...

[Wed Jun  1 15:01:37 2022]
rule use_script:
    input: data/file1.txt
    output: results/file.bar.txt
    jobid: 2
    reason: Missing output files: results/file.bar.txt
    wildcards: filetype=bar
    resources: tmpdir=/tmp

[Wed Jun  1 15:01:38 2022]
Finished job 2.
1 of 3 steps (33%) done
Select jobs to execute...

[Wed Jun  1 15:01:38 2022]
rule use_script:
    input: data/file1.txt
    output: results/file.foo.txt
    jobid: 1
    reason: Missing output files: results/file.foo.txt
    wildcards: filetype=foo
    resources: tmpdir=/tmp

[Wed Jun  1 15:01:38 2022]
Finished job 1.
2 of 3 steps (67%) done
Select jobs to execute...

[Wed Jun  1 15:01:38 2022]
localrule all:
    input: results/file.foo.txt, results/file.bar.txt
    jobid: 0
    reason: Input files updated by another job: results/file.bar.txt, results/file.foo.txt
    resources: tmpdir=/tmp

[Wed Jun  1 15:01:38 2022]
Finished job 0.
3 of 3 steps (100%) done
Complete log: .snakemake/log/2022-06-01T150136.792964.snakemake.log
```
</details>

<details><summary>Report generation</summary>

```
$ snakemake --report report.html
Building DAG of jobs...
Creating report...
Loading script code for rule use_script
WorkflowError:
Failed to open source file /home/grahman/projects/snakemake-bug/workflow/scripts/process_{wildcards.filetype}.py
FileNotFoundError: [Errno 2] No such file or directory: '/home/grahman/projects/snakemake-bug/workflow/scripts/process_{wildcards.filetype}.py'
```
</details>

<details><summary>File tree</summary>

[...]

PR #2752, opened 2024-03-12
chore: add testcases for issue #1697

mawenzy:
### Description

Add test cases such as "scipts_{wildcards.name}.py", which contains a wildcard but is not correctly detected.

These tests will fail for now, because they depend on a change in an underlying package, see: https://github.com/snakemake/snakemake-interface-storage-plugins/pull/40


mawenzy:
I think the overall better solution would be to move the test `test_wildcard_regex` to the other package, where `WILDCARD_REGEX` is actually defined.

fgvieira:
Can you format your code with `black`?

fgvieira (approved):

fgvieira:
Ping @johanneskoester

PR #2761, opened 2024-03-15
fix: ensure that dynamic `resources:` determination can depend on dynamically determined `threads:`

dlaehnemann:
The first step is to include a testcase where `resources: mem_mb=` depend on dynamically determined `threads:`. If the test fails as expected, a fix should be easier to track.

### QC


dlaehnemann:
It seems like the problem runs deeper. It's probably only triggered in the context of remote executors. I'll try to come up with a failing test case...

PR #2800, opened 2024-04-09
fix: failure message in cleanup_metadata

sjpfenninger:
### Description

`cleanup_metadata()` in `workflow.py` is passed a list of `pathlib.Path` objects. This leads to an error when it ought to raise a WorkflowError on cleanup failing for some paths. Given how the function is written, the `failed` list should consist of strings, rather than `Path` objects, otherwise we get:

```
TypeError: sequence item 0: expected str instance, PosixPath found
```
This seems the be a regression in 8.x. vis-a-vis 7.x. I haven't added any tests since the function seems not to be covered by tests so far.

### QC

### DIFF
--- a/snakemake/workflow.py
+++ b/snakemake/workflow.py
@@ -815,7 +815,7 @@ class Workflow(WorkflowExecutorInterface):
         for path in paths:
             success = self.persistence.cleanup_metadata(path)
             if not success:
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):

johanneskoester:
Thank you!

Issue #1614, opened 2022-04-28
Workflow error in module when using a prefix and 'expand' in a rule input


RomainFeron:
**Snakemake version**
7.3.8

**Describe the bug**

I have an issue when using a workflow as module in another workflow with a prefix. Rules from the module that are using 'expand' as input generate a WorkflowError (see below). This does not occur when removing the prefix in the module definition.

**Logs**

```
WorkflowError in line 15 of /home/romain/work/test/wfA/Snakefile:
Flags in file patterns given to expand() are invalid. Flags (e.g. temp(), directory()) have to be applied outside of expand (e.g. 'temp(expand("plots/{sample}.pdf", sample=SAMPLES))').
  File "/home/romain/work/test/wfB/Snakefile", line 5, in <module>
  File "/home/romain/work/test/wfA/Snakefile", line 15, in <module>
```

**Minimal example**

Two simple workflows with the following structure:
```
.
├──  wfA
│   └── Snakefile
└── wfB
    ├── a.txt
    ├── b.txt
    └── Snakefile
```

*wfa/Snakefile:*
```

WILDCARDS = ['a', 'b']

rule copy_data:
    input:
        '{x}.txt'
    output:
        'results/data/{x}.txt'
    shell:
        'cp {input} {output}'


rule test:
    input:
        expand(rules.copy_data.output, x=WILDCARDS)
    output:
        'results/test_wfA.txt'
    shell:
        'cat {input} > {output}'
```

*wfB/Snakefile:*
```
module wfA:
    snakefile: '../wfA/Snakefile'
    prefix: 'results/wfA'

use rule * from wfA as wfA_*

rule test:
    input:
        rules.wfA_test.output
    output:
        'results/test_wfB.txt'
    shell:
        'cp {input} {output}'
```

*Snakemake command ran from wfB:*
`snakemake -j1 -n results/test_wfB.txt`

**Additional context**

The example above generates the logs I posted but it works when commenting the "prefix" line in "module wfA". I'm not sure if it's a bug or if I missed something in modules.

fgvieira:
I am having the same issue on version `v8.10.6`, but when the input file is a `temp` (in your case, if `copy_data.output`  was `temp(copy_data.output)` ).

I think the issue might be around:
https://github.com/snakemake/snakemake/blob/293bc056b6c44f17340543b2678f8a59e808a5a6/snakemake/path_modifier.py#L55

[...]

PR #2803, opened 2024-04-11
fix: expand error on modules with prefix (#1614)

fgvieira:
### Description

Fix for #1614 

### QC

### DIFF
--- a/snakemake/io.py
+++ b/snakemake/io.py
@@ -1247,10 +1247,14 @@ def expand(*args, **wildcard_values):
 
     filepatterns = list(map(path_to_str, filepatterns))
 
-    for fp in filepatterns:
-        if fp_flags := getattr(fp, "flags", False):
+    for filepattern in filepatterns:
+        if filepattern_flags := {
+            key: value
+            for key, value in getattr(filepattern, "flags", {}).items()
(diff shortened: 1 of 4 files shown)


snakemake/io.py
```diff
@@ -1247,12 +1247,18 @@ def path_to_str(f):
...
-        raise WorkflowError(
-            "Flags in file patterns given to expand() are invalid. "
-            "Flags (e.g. temp(), directory()) have to be applied outside "
-            "of expand (e.g. 'temp(expand(\"plots/{sample}.pdf\", sample=SAMPLES))')."
-        )
+    for filepattern in filepatterns:
+        filepattern_flags = {
+            key: value
+            for key, value in getattr(filepattern, "flags", {}).items()
+            if key not in ["path_modified"]
```
johanneskoester, line 1254: Maybe rather this?
```suggestion
            if key != PATH_MODIFIER_FLAG
```

johanneskoester (changes requested):

snakemake/io.py
fgvieira: Was not sure if there might be other flags that should be ignored, but it can be changed later if needed.

johanneskoester (approved):

Issue #2797, opened 2024-04-06
Duplicate Calls to log_handler in log-handler-script


Ash1One:
**Snakemake version**

v8.10.6

**Describe the bug**

The `log_handler` function of the `log-handler-script` parameter is repeatedly added to `snakemake.logging.Logger.log_handler`, causing the `log_handler` function to be called twice for every logging message.

**Possible Reason**

The `log_handler` function seems to be added to `snakemake.logging.Logger.log_handler` twice; once here:

https://github.com/snakemake/snakemake/blob/b3c1dd4b297f01b5d08d248b8512d91b3bd5a7da/snakemake/api.py#L141

and once here:

https://github.com/snakemake/snakemake/blob/b3c1dd4b297f01b5d08d248b8512d91b3bd5a7da/snakemake/api.py#L526-L530

**Minimal example**

The problem presents itself when I run the following command with the Snakefile and log_job.py files given below. Note that every message, except the first one, is printed twice.

* Snakefile:

  ```
  rule all:
      input:
          "b.txt"


  rule test:
      output:
          touch("b.txt")
  ```

* log_job.py:

  ```
  def log_handler(msg):
      print(msg)
  ```

* command: 

  `snakemake --snakefile Snakefile -c 1 --log-handler-script log_job.py`

**Output**:
  ```
  Assuming unrestricted shared filesystem usage.
  {'level': 'info', 'msg': 'Assuming unrestricted shared filesystem usage.', 'indent': False, 'timestamp': 1712385654.750852}
  Building DAG of jobs...
  {'level': 'info', 'msg': 'Building DAG of jobs...', 'indent': False, 'timestamp': 1712385654.7939415}
  {'level': 'info', 'msg': 'Building DAG of jobs...', 'indent': False, 'timestamp': 1712385654.7939415}
  {'level': 'dag_debug', 'status': 'candidate', 'job': all, 'timestamp': 1712385654.795178}
  {'level': 'dag_debug', 'status': 'candidate', 'job': all, 'timestamp': 1712385654.795178}
  {'level': 'dag_debug', 'status': 'candidate', 'job': test, 'timestamp': 1712385654.7959971}
  {'level': 'dag_debug', 'status': 'candidate', 'job': test, 'timestamp': 1712385654.7959971}
  {'level': 'dag_debug', 'status': 'selected', 'job': test, 'timestamp': 1712385654.796126}
[...]

PR #2809, opened 2024-04-13
remove extra call to setup_logger()

cademirch:
### Description

Removes seemingly extra call to setup_logger() in api, see #2797 for details. 

Still not super familiar with API changes, so @johanneskoester please advise.

### QC

### DIFF
--- a/snakemake/api.py
+++ b/snakemake/api.py
@@ -523,12 +523,6 @@ class DAGApi(ApiBase):
                 "For local execution, --shared-fs-usage has to be unrestricted."
             )
 
-        self.snakemake_api.setup_logger(
-            stdout=executor_plugin.common_settings.dryrun_exec,
-            mode=self.workflow_api.workflow_settings.exec_mode,
-            dryrun=executor_plugin.common_settings.dryrun_exec,
-        )
-
         if executor_plugin.common_settings.local_exec:
             if (
                 not executor_plugin.common_settings.dryrun_exec


cademirch:
@johanneskoester could you take a look at this when you get a chance? Would be good to get this fixed before moving forward with monitoring plugin.

snakemake/api.py
```diff
@@ -523,12 +523,6 @@ def execute_workflow(
                 "For local execution, --shared-fs-usage has to be unrestricted."
             )
 
-        self.snakemake_api.setup_logger(
-            stdout=executor_plugin.common_settings.dryrun_exec,
-            mode=self.workflow_api.workflow_settings.exec_mode,
-            dryrun=executor_plugin.common_settings.dryrun_exec,
-        )
```
johanneskoester, line 530: The reason for this extra call is that with calling execute, the dryrun info might be new, as well as the exec mode. One could of course offer a method to just update these in the logger api.

cademirch: Hmm, I'm not sure I follow. It seems setup logger is called here https://github.com/snakemake/snakemake/blob/8a80bdaed0d8524dffbdce71dd07eb123f71726d/snakemake/api.py#L141
How can the the info be new by time execute is called? Regardless, I think you're right, we can provide setters for these in the logger

cademirch: Could we just do this instead of calling setup_logger()?

```suggestion
         logger.stdout = executor_plugin.common_settings.dryrun_exec
         logger.mode = self.workflow_api.workflow_settings.exec_mode
         logger.dryrun = executor_plugin.common_settings.dryrun_exec
```

johanneskoester: Yes, I think so.

cademirch:
Closing because logging refactor makes this obsolete.

PR #2814, opened 2024-04-16
fix: module prefix added twice on expand

fgvieira:
### Description

When using modules with prefixes, sometimes the prefix is added twice (e.g. if there are two rules with expands). For example, using the `prefix_replace`:
```
    replace_prefix:
        {"results/": "results/testmodule/"}
```

Snakemake throws the error:
```
WorkflowError:
MissingInputException: Missing input files for rule test_b:
    output: results/testmodule/b/test.out
    wildcards: name=test
    affected files:
        results/testmodule/testmodule/a/test.out
MissingInputException: Missing input files for rule test_c:
    output: results/testmodule/b/test.out
    wildcards: name=test
    affected files:
        results/testmodule/testmodule/b/test.out
```

The same happens if using the prefix:
```
    prefix:
	"test"
```
with Snakemake throwing the error:
```
WorkflowError:
MissingInputException: Missing input files for rule test_c:
    output: test/results/b/test.out
    wildcards: name=test
    affected files:
        test/test/results/b/test.out
MissingInputException: Missing input files for rule test_b:
    output: test/results/b/test.out
    wildcards: name=test
    affected files:
        test/test/results/a/test.out
```

With more verbose logs, it seems that the prefix is being added before and after `expand`:
```
[...]
Path results/a/{name}.out, converted to test/results/a/{name}.out
Path test/results/a/test.out, converted to test/test/results/a/test.out
Path results/b/{name}.out, converted to test/results/b/{name}.out
Path test/results/b/test.out, converted to test/test/results/b/test.out
Path results/b/{name}.out, converted to test/results/b/{name}.out
Path test/results/b/test.out, converted to test/test/results/b/test.out
Path test/results/b/test.out, converted to test/results/b/test.out
Path did not change after prefix replace
[...]
```
I think `prefix` should not be changed after `expand`, since `prefix_replace` might replace the wildcards, so I added a function to copy the file attributes after `expand`.

### QC


snakemake/io.py
```diff
@@ -1305,10 +1306,20 @@ def flatten(wildcard_values):
...
 
+        def copy_flags(from_path, dest_path):
+            if hasattr(from_path, "flags"):
+                dest_path = AnnotatedString(dest_path)
+                dest_path.flags.update(from_path.flags)
+                if is_flagged(dest_path, "multiext"):
+                    dest_path.flags["multiext"] = self.apply_default_storage(
+                        self.replace_prefix(dest_path.flags["multiext"], property)
+                    )
+            return dest_path
```
johanneskoester, line 1317: I fear that we forget the reason for this later. Can you add some comments?

johanneskoester: And why the special case for multiext?

johanneskoester: And maybe multiext should then also occur in the test case?

johanneskoester (changes requested):
Thanks a lot! I think the code needs some documentation though

snakemake/io.py
fgvieira: Sure, I can add some comments.
Actually I am not sure if we need that special case for `multiext`.. I'll add it to the test case and check.

johanneskoester (approved):

PR #2817, opened 2024-04-16
docs: add snakemake and reportplugins to req.txt to show all cli options

### LINKED ISSUE #2813: Missing All Flags section in Snakemake docs in at least several 8.x versions
The [command line interface](https://snakemake.readthedocs.io/en/stable/executing/cli.html) section of the 8.x Snakemake docs has an All Flags section (also present in various forms in older versions) that is currently blank in the latest 8.10.7 version of the docs and also in some older 8.x versions I've looked at. It appears to be present at least up to [7.0.0](https://snakemake.readthedocs.io/en/v7.0.0/executing/cli.html). I'm not sure of the exact set of documentation versions in which it's missing.

cademirch:
### Description

Fixes #2813. `snakemake` and `snakemake-interface-report-plugins` were missing from the req.txt, which are needed to get the argument parser from `snakemake.cli`.

Note: it seems the default report plugin's arguments leak into the docs (see below)
Screenshot 2024-04-16 at 2 48 53 PM

### QC

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -7,6 +7,8 @@ configargparse
 appdirs
 immutables
 sphinxawesome-theme
+snakemake
 snakemake-interface-common
 snakemake-interface-executor-plugins
 snakemake-interface-storage-plugins
+snakemake-interface-report-plugins
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -7,7 +7,6 @@ configargparse
 appdirs
 immutables
 sphinxawesome-theme
-snakemake
 snakemake-interface-common
 snakemake-interface-executor-plugins
 snakemake-interface-storage-plugins


docs/requirements.txt
```diff
@@ -7,6 +7,8 @@ configargparse
 appdirs
 immutables
 sphinxawesome-theme
+snakemake
```
johanneskoester, line 10:
```suggestion
```
Snakemake itself is imported from the repo checkout. It should not be needed and I would think it is even harmful here.

johanneskoester (changes requested):

cademirch:
Ah, I figured. I’ll remove it.

Thanks

johanneskoester (approved):

johanneskoester:
Thanks a lot!

Issue #2826, opened 2024-04-19
`--show-failed-logs` if binary file used as `log` file


votti:
**Snakemake version**
`main` banch

**Describe the bug**
If a binary file is set as `log` output and `--show-failed-logs` (or other options triggering logs to be shown) is activated, if a job crashes the whole workflow execution crashes as snakemake tries to print the binary `log` file as text.

**Logs**
```

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/site-packages/snakemake/executors/local.py", line 259, in _callback
    raise ex
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/site-packages/snakemake/executors/local.py", line 245, in cached_or_run
    run_func(*args)
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/site-packages/snakemake/executors/local.py", line 232, in spawn_job
    raise SpawnedJobError()
snakemake.exceptions.SpawnedJobError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/concurrent/futures/_base.py", line 340, in _invoke_callbacks
    callback(self)
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/site-packages/snakemake/executors/local.py", line 266, in _callback
    self.report_job_error(job_info)
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/site-packages/snakemake_interface_executor_plugins/executors/base.py", line 92, in report_job_error
    self.print_job_error(job_info, msg, **kwargs)
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/site-packages/snakemake_interface_executor_plugins/executors/base.py", line 113, in print_job_error
    job_info.job.log_error(msg, **kwargs)
  File "/project/home/vizano/mambaforge/envs/snakemake8/lib/python3.12/site-packages/snakemake/jobs.py", line 1023, in log_error
[...]

PR #2827, opened 2024-04-19
fix: Prevent binary log files to crash snakemake execution with `show-failed-logs`

votti:
### Description

As described in #2826 currently the whole snakemake execution crashes if a log file is binary and `show-failed-logs` is used.

This fix prevents the crash by capturing the `UnicodeDecodeError` and returning an error message that the log file is not a text instead.

### QC

### DIFF
--- a/snakemake/logging.py
+++ b/snakemake/logging.py
@@ -506,6 +506,9 @@ class Logger:
                 except FileNotFoundError:
                     yield f"Logfile {f} not found."
                     return
+                except UnicodeDecodeError:
+                    yield f"Logfile {f} is not a text file."
+                    return
                 lines = content.splitlines()
                 logfile_header = f"Logfile {f}:"
                 if not lines:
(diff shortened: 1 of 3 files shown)


votti:
@Test: I am not extremely happy about the test: currently the test is more on the integration test level, so I think it may be better to also add a unittest for `show_logs`.
Currently this function is defined nested within the `text_handler` function, thus making it hard to test. To increase the testability , factoring it out as either a separate function in the` logging` module (eg `_show_logs`) or as a static method (`Logger._show_logs`).
Maybe it should be considered to factor out `text_handler` as a separate class containing all the methods defined withing the current `text_handler` function. This overall could be cleaner.

johanneskoester:
Sure, a follow-up PR in that direction would be appreciated. Maybe it would need synchronization with #2830

johanneskoester (approved):

PR #2830, opened 2024-04-20
feat: add monitoring plugin interface

cademirch:
### Description

@johanneskoester, just wanted to start the discussion on adding the monitoring plugin interface. 

I've implemented a monitoring plugin interface [in this repo](https://github.com/cademirch/snakemake-interface-monitoring-plugins). In this PR I've replaced the `--log-service` and `--wms` CLI options with `--monitoring-provider`, as well as wired up the use of the monitoring provider. 

I designed the `MonitoringProvider` base class around using the `log_handler`, as this is what `logging.WMSLogger` and `logging.SlackLogger` used. New monitoring plugins will just need to define a `log_handler` and will be good to go.

In order to test this, I reimplemented the `WMSLogger` into a monitoring plugin [in this repo](https://github.com/cademirch/snakemake-monitoring-plugin-wms).

Off the top of my head, things left to do are:
- Write tests and docs
- Publish the interface to PyPi
- Implement SlackLogger as a plugin and publish
- Publish WMSLogger plugin
- Remove WMSLogger and SlackLogger from logging.py

Shouldn't take long to do all of the above, just want to make sure this is the way to go first.. I'll keep this as a draft until we decide on the way forward.

**Extra:**
It would be nice to pass the rulegraph to the monitor, so that it can visualize the workflow's DAG. Perhaps something like this:
```Diff
# workflow.py
def _build_dag(self):
        logger.info("Building DAG of jobs...")
        async_run(self.dag.init())
        async_run(self.dag.update_checkpoint_dependencies())
+      logger.debug_dag(self.dag.rule_dot()) # Could also define new func for this instead of debug_dag

```
Doing this in `_build_dag` would allow the monitor to update its DAG visualization if the workflow has checkpoints. Of course this assumes we use the log_handler for the monitoring plugin interface. I can't think of any issues logging the rulegraph in `_build_dag` could cause, but feedback here would be helpful.


cademirch:
Hi @johanneskoester making this a full PR so you can review.

johanneskoester:
> It would be nice to pass the rulegraph to the monitor, so that it can visualize the workflow's DAG. Perhaps something like this:

Yes, good idea! Should be a separate method then, instead of debug_dag I think.

snakemake/cli.py
```diff
@@ -1482,12 +1463,8 @@ def get_argument_parser(profiles=None):
...
     )
     group_behavior.add_argument(
-        "--log-service",
-        default=None,
-        choices=["none", "slack", "wms"],
-        help="Set a specific messaging service for logging output."
-        "Snakemake will notify the service on errors and completed execution."
-        "Currently slack and workflow management system (wms) are supported.",
+        "--monitoring-provider",
+        help="Specify a custom monitoring provider, available via an monitoring plugin: snakemake_monitoring_<name>",
```
johanneskoester, line 1467:
```suggestion
        help="Specify a custom monitoring provider, available via an monitoring plugin: snakemake-monitoring-plugin-<name>",
```

johanneskoester: Thinking about it, why not just `snakemake-monitor-plugin-...` (without the ing)?

cademirch:
> > It would be nice to pass the rulegraph to the monitor, so that it can visualize the workflow's DAG. Perhaps something like this:
> 
> Yes, good idea! Should be a separate method then, instead of debug_dag I think.

Great, I agree. What do you think about the monitoring interface? If this is the approach we take, I'm happy to start publishing the interface and plugins to pypi and write some tests for snakemake.

johanneskoester (commented):
Awesome, only a few things.

snakemake/cli.py
cademirch: I'm not sure tbh - I think b/c I had "monitoring-provider" stuck in my head. I like `snakemake-monitor-plugin` better, I'll make that change.

jjjermiah:
nice work @cademirch glad to see this going. If theres a new logging call it might be useful to consider specific logs that might provide more/structured info regarding mapping to be ingested by the plugin

cademirch:
> nice work @cademirch glad to see this going. If theres a new logging call it might be useful to consider specific logs that might provide more/structured info regarding mapping to be ingested by the plugin

Thanks. Yes, I agree. I think working out the suggestion in #2827 to break out the `show_failed_logs` into its own method would be a good place to start, as it would be nice to offer failed logs to the monitor plugin

johanneskoester:
One further thought: couldn't we instead call the interface snakemake-interface-logger-plugins, and unify monitoring and logging? Isn't it in the end just the question where the log messages are directed to? For the cli, I would then imagine that people can specify any number of loggers based on what plugins are installed, e.g. `snakemake --logger json slack` would run snakemake with snakemake-logger-plugin-json and snakemake-logger-plugin-slack, whereas `snakemake --logger text wms-monitor` would use the term logger (which will stay built into snakemake analogously to the local executor) and snakemake-logger-plugin-wms-monitor. Importantly, loggers that write to the terminal should be mutually exclusive, i.e. Snakemake should give an error if e.g. `snakemake --logger text json` is executed.

jjjermiah:
that sounds about right, this is directly related to the logging itself. It might help to think about how the logs might be used. 

I for one, was first interested in structuring logs to be posted directly to the Google Cloud Logging API and use that as a "database" where I can build a monitor that queries logs using filters on the fields that the login plugin defines.
- I've tested this out using the `log-handler-script` argument to circumvent the issue regarding the google-batch plugin's logging issues. 

Taking a step back it becomes clear that this is resembles more of a pub/sub approach which opens the door for using an actual pub/sub architecture. 

This might be complicating things more than it needs to be, but it also allows for the remote jobs that are being executed to post their logs (instead of having to i.e do some `kubectl logs snakejob-......` as the only other way to get a job's logs

cademirch:
@johanneskoester I think unifying monitoring and logging is a great idea. Just to confirm though, the interface design would stay the same? Just the name would change? Now would be a good time to change the design. 


If we go this direction, perhaps we should compile any changes to logging.py we would want to see and implement them here. 

I'm currently traveling so I won't be able to make any code progress, but am happy to keep discussing!

cademirch:
@johanneskoester would you be open to changes to the logger class in this PR? I think it could be beneficial to simplify the class by reducing the number of logging methods to just the basic info, warn, debug, error, etc. And then the handler can handle the emission of that log record. I can flesh this out further, but just want to start a discussion on this, as it would be useful to think about in the context of logging plugins. And make sure the logger is where we want it before finalizing the logging interface. 

Furthermore, we could consider using a different logging package like [structlog](https://www.structlog.org/en/stable/). While this would take a bit of work and add a dependency, I think it is worth thinking about.

johanneskoester:
@cademirch yes absolutely. The logging code of Snakemake has never seen a proper design and has only been extended ad hoc when needed. Structlog looks great and I am completely open for a rewrite of all these parts. Looking forward to your changes!

cademirch:
Sounds good. I'll leave this PR where its at for now to revisit down the line and open a new one for logging refactoring once. And I agree re presenting info on the terminal... I have some ideas already :)

johanneskoester:
Not sure, I think the current logging is better than json because it is more readable. If anything it should rather be yaml as a default. However, with plugins and options of those plugins the possibilities are endless and people could use profiles to customize to whatever defaults they like most.

cademirch:
That makes sense. I had made some progress in the logging refactoring last month, but had to put it down due to some deadlines. Those are wrapping up this week so I plan to start working on this again more next week. For now I'm planning on reimplementing the current stdout/file log formatting as closely as possible, then adding different output options (progress bars, JSON, yaml, etc) as plugins and or built in options.

cademirch:
closed in favor of #3107

PR #2843, opened 2024-04-24
fix: checkpoints inside modules are overwritten

fgvieira:
### Description

When including two modules with checkpoints (or the same module twice with different prefixes), only the checkpoint of the last loaded module is used.

In the example here, the checkpoint `somestep` is run twice the module `test_b` (under prefix `b`) while it should be run one for each prefix (as exemplified under `expected-results`).

```
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Job stats:
job                 count
----------------  -------
all                     1
test_a_aggregate        1
test_b_aggregate        1
test_b_somestep         2
total                   5

Select jobs to execute...
Execute 1 jobs...

[Wed Apr 24 13:36:30 2024]
localcheckpoint test_b_somestep:
    output: b/somestep/a.txt
    jobid: 2
    reason: Missing output files: <TBD>
    wildcards: sample=a
    resources: tmpdir=/tmp
DAG of jobs will be updated after completion.

[Wed Apr 24 13:36:31 2024]
Finished job 2.
1 of 5 steps (20%) done
Select jobs to execute...
Execute 1 jobs...

[Wed Apr 24 13:36:31 2024]
localrule test_a_alt_intermediate:
    output: a/alt/a.txt
    jobid: 8
    reason: Missing output files: a/alt/a.txt
    wildcards: sample=a
    resources: tmpdir=/tmp

[Wed Apr 24 13:36:31 2024]
Finished job 8.
2 of 6 steps (33%) done
Select jobs to execute...
Execute 1 jobs...

[Wed Apr 24 13:36:31 2024]
localcheckpoint test_b_somestep:
    output: b/somestep/b.txt
    jobid: 4
    reason: Missing output files: <TBD>
    wildcards: sample=b
    resources: tmpdir=/tmp
DAG of jobs will be updated after completion.
[...]
```

### QC

## Summary by CodeRabbit

- **Tests**
	- Enhanced test validations with additional assertions ensuring functions produce the expected results.
	- Introduced a new helper utility that returns a consistent preset value to support reliable testing.
	- Added an expected output file for improved test verification.


johanneskoester:
ping @fgvieira

fgvieira:
Sorry for the late reply. The test does not fail because both modules have the same rules (since it is the same `Snakefile` being added twice) and rule `test_b_somestep` ends up being called twice. As I see it, each module should call its own rules, so the dryrun should look like:
```
job                 count
----------------  -------
all                     1
test_a_aggregate        1
test_b_aggregate        1
test_a_somestep         1
test_b_somestep         1
total                   5
```

Hocnonsense:
@fgvieira Hello, may I push some commits based on your code?

fgvieira:
Sure! 😄

Hocnonsense:
@fgvieira @matthewfeickert Hello, do you think this solution #3359 accepable? I just don't know how to merge that to this PR

fgvieira:
@Hocnonsense seems to be working. Thanks!
I guess there is no need for this PR then.

PR #2852, opened 2024-04-26
Feature/jjjermiah/doc update

jjjermiah:
This is just a small typo fix. 
Working on a fix for https://github.com/snakemake/snakemake-executor-plugin-kubernetes/issues/19 and eventually ended up at the following error when deploying to kubernetes:
```
Error: Error when applying default storage provider gcs to upload workflow sources. {query_validity}
```
Fixing this here so the error is more clear since it looks like kubernetes uses the docker container from main.


jjjermiah:
this was the wrong pr. sorry, ignore

PR #2856, opened 2024-04-30
feat: allow to set shell_exec (e.g. bash, sh, zsh) per rule

johanneskoester:
Use case: https://stackoverflow.com/questions/78381413/how-to-set-apptainer-shell-when-using-software-deployment-method-apptainer-wit

### QC


johanneskoester:
I will probably close this in favor of #2862.

johanneskoester:
Yes the new one is better, it makes more sense to define shell_exec within resources instead of introducing a new directive.

Issue #2657, opened 2024-01-30
parsing fstring with if else on params


sravel:
**Snakemake version 8.4.0**
**Python 3.12.1**

**Describe the bug**
SyntaxError in file Snakefile, line 20:
expected 'else' after 'if' expression (Snakefile, line 20)

**Minimal example**
```
from collections import defaultdict
from pprint import pp
config = defaultdict(defaultdict)

output_dir="."
config['params']['SMARTDENOVO'] = {}
config['params']['SMARTDENOVO']['KMER_SIZE'] = 10
config['params']['SMARTDENOVO']['OPTIONS'] = "zmo"
config['CIRCULAR'] = False

pp(config)
################################ SMARTDENOVO ####################################
rule test:
    threads: 1
    input:
        fastq = "/shared/home/sravel/glp701/sandbox/test_snake8/Data-Xoo-sub/refBAI3.fasta"
    params:
        prefix = "SMART",
        kmersize = f"16 " if f"{config['params']['SMARTDENOVO']['KMER_SIZE']}" == '' else f"{config['params']['SMARTDENOVO']['KMER_SIZE']}",
        out_tmp = f"{output_dir}{{fastq}}/ASSEMBLERS/SMARTDENOVO/ASSEMBLER/SMART.{'zmo' if 'zmo' in config['params']['SMARTDENOVO']['OPTIONS'] else 'dmo'}.cns"
    output:
        fasta = f"{output_dir}{{fastq}}/ASSEMBLERS/SMARTDENOVO/ASSEMBLER/assembly2Circ.fasta" if config['CIRCULAR'] else f"{output_dir}{{fastq}}/ASSEMBLERS/SMARTDENOVO/ASSEMBLER/assembly.fasta"
    shell:
        "smartdenovo.pl -t {threads} -k {params.kmersize} -p  {params.prefix} -c 1 {input.fastq} 1>{params.mak}"
```

**Additional context**
`snakemake --show-failed-logs -p -s Snakefile  --lint`

dariober:
```
f"{output_dir}{{fastq}}/ASSEMBLERS/SMARTDENOVO/ASSEMBLER/SMART.{'zmo' if 'zmo' in config['params']['SMARTDENOVO']['OPTIONS'] else 'dmo'}.cns"
^^^                                                                                                                                        ^^^
```

I suspect this line is not right since everything is contained in double quotes.

sravel:
the code works in a python console but snakemake raises an error:
[...]

PR #2864, opened 2024-05-03
fix: parsing fstring with if else on params #2657

sravel:
fix:  parsing fstring with if else on params #2657

### QC


johanneskoester (changes requested):
May I ask you to add a testcase?

sravel:
Hello I don't now how can do that, I open issue since 4 months and just to need fix to switch my workflow on snakemake 8....

tedil:
Hi @sravel I've modified the code to use tokenize.LBRACE/RBRACE and added a testcase which makes use of f-strings (and should simply error when parsing goes awry). Is that okay with you?

johanneskoester:
Fixed already and test included by PR https://github.com/snakemake/snakemake/pull/2930

johanneskoester:
Nevertheless, thanks a lot for helping with this!!

PR #2888, opened 2024-05-21
feat: add flag to mark files whose path should not be modified

fgvieira:
### QC


tests/test_module_no_path_modified/module1/Snakefile
```diff
@@ -0,0 +1,7 @@
+rule a:
+    input:
+        path_modified("input.txt"),
```
johanneskoester, line 3: I don't really see why one would want this to be annotated manually. Can you explain?

fgvieira: Described the problem a bit more in the description, so it is more visible.

johanneskoester: Ah, understood. I think this is covered semantically already by the `local` flag, see https://snakemake.readthedocs.io/en/stable/snakefiles/storage.html#local-input-output-files. 
However, it is not properly documented to also apply to module handling with prefixes. Can you check whether that works for you?

fgvieira: No, the `local` flag does not work as the path is still modified.

fgvieira: Ok, so reverted most changes and add a check for the `local` flag when modifying the path.

johanneskoester (approved):

docs/snakefiles/storage.rst
```diff
@@ -60,8 +60,9 @@ Local input/output files
 """"""""""""""""""""""""
 
 Despite using a default storage provider, you might have certain files in your workflow
-that still come from the local filesystem. In this case, you can use the ``local``
-flag::
+that still come from the local filesystem. Likewise, when importing a module while
+specifying a prefix (see :ref:`snakefiles-modules`), you might have some input files
+that come from outside the workflow. In either cases, you can use the ``local`` flag::
```
johanneskoester, line 65: I think one should in addition refer to this from the modules sections where it talks about the prefixes.

johanneskoester (changes requested):

docs/snakefiles/modularization.rst
```diff
@@ -165,8 +165,9 @@ It is possible to overwrite the global config dictionary for the module, which i
 
 In this case, any ``configfile`` statements inside the module are ignored.
 In addition, it is possible to skip any :ref:`validation <snakefiles_config_validation>` statements in the module, by specifying ``skip_validation: True`` in the module statement.
-Moreover, one can automatically move all relative input and output files of a module into a dedicated folder: by specifying ``prefix: "foo"`` in the module definition, e.g. any output file ``path/to/output.txt`` in the module would be stored under ``foo/path/to/output.txt`` instead.
+Moreover, one can automatically move all relative input and output files of a module into a dedicated folder by specifying ``prefix: "foo"`` in the module definition. In this case any output file ``path/to/output.txt`` in the module would be stored under ``foo/path/to/output.txt`` instead.
 This becomes particularly useful when combining multiple modules, see :ref:`use_with_modules`.
+However, if you have some input files that come from outside the workflow, you can use the ``local`` flag so that their path is not modified.
```
johanneskoester, line 170: Maybe also refer to the section about the local flag from here (you have to add a label to the section so that you can use :ref: ... here)?

johanneskoester (changes requested):

johanneskoester (approved):

PR #2904, opened 2024-05-31
Add assets and use local file links

### LINKED ISSUE #282: offline snakemake --report
I am often working in an offline environment and I would still like to generate reports for my snakemake runs. Currently it seems that the reports need to pull css files from a cdn server. 

it would be great if I could simply download the css files and pass the path to the report command. For example like this.

```
snakemake --report docs/report.html -css docs/css
```

fs82:
This PR solves the issue with the offline reports. Several files are downloaded on the fly while generating the report(js, css and licenses). They are now added to the package as assets.

I used the `test_report` test case to verify that the generated report is the same and works as expected. One minor change in the report appears, as I thought it is more logical to use major.minor version numbering for the `react` package as well.

To my opinion a update of the documentation is not necessary.

Closes #282 

### QC


fs82:
@johanneskoester The checks are failing on the new files added to the repository as assets for the reports. To my opinion they should not be checked by the ci of snakemake, as they are obtained from other projects. However, my knowledge about Github and its CI is limited. Would be good if you can comment on how to work around this.

fs82:
The only other way I see is to compress the files and extract the on the fly?

fs82:
> I think the files can simply be excluded via adding a sonar-project.properties file to the repo: https://docs.sonarsource.com/sonarcloud/advanced-setup/analysis-scope/#file-exclusion-and-inclusion

Ok, sounds good. I will have a look asap.

fs82:
> I think the files can simply be excluded via adding a sonar-project.properties file to the repo: https://docs.sonarsource.com/sonarcloud/advanced-setup/analysis-scope/#file-exclusion-and-inclusion

I am give up, is there a way to see somehow the log output of the sonar cloud? I tried several ways to ignore the assets folder, but it does not seem to be that the sonar cloud respects the settings in `sonar-project.properties`. It is honestly not my field of knowledge :smile:

daaaaande:
would be happy to see that feature. using snakemake on a cluster that is offline, and the reports could help. is there maybe a workaround for this issue?

fs82:
> > I think the files can simply be excluded via adding a sonar-project.properties file to the repo: https://docs.sonarsource.com/sonarcloud/advanced-setup/analysis-scope/#file-exclusion-and-inclusion
> 
> I give up, is there a way to see somehow the log output of the sonar cloud? I tried several ways to ignore the assets folder, but it does not seem to be that the sonar cloud respects the settings in `sonar-project.properties`. It is honestly not my field of knowledge 😄

Ok from cross reading the sonar cloud manual, it looks like there are automatic analysis enabled for snakemake repository, due to the presence of the .sonarcloud.properties file. This seems to conflict with the sonar-project.properties file, which is only for ci-based analysis. I honestly dont get the difference, but the manual states, this is conflicting. I move the exclude settings now to the .sonarcloud.properties, lets see what happens

johanneskoester:
Actually, I think we can solve the issue in a way that fixes two problems at a time.
It is in general not so desirable to have copies of foreign code within the repo.
So why don't we make their download part of the package building process, e.g. via [setuptools-download](https://github.com/asottile/setuptools-download)? This way, they won't be checked by any linter, and the source tree and git history remains clean of foreign code. Further, an update is just doable by changing the URLs in the setup.cfg. Alternatively, if more complex logic is needed, one can also put some download code using requests into the setup.py.

johanneskoester:
Thanks for trying! So, setuptools-download has to be extended to provide a user agent to the cdn server. I've implemented this here: https://github.com/asottile/setuptools-download/pull/25
If the change is accepted and released, we can proceed with the updated version of setuptools-download.

fs82:
> Thanks for trying! So, setuptools-download has to be extended to provide a user agent to the cdn server. I've implemented this here: [asottile/setuptools-download#25](https://github.com/asottile/setuptools-download/pull/25) If the change is accepted and released, we can proceed with the updated version of setuptools-download.

Yes, indeed, this is the fix required for setuotools-download. Ok, I will wait until the https://github.com/asottile/setuptools-download/pull/25 gets merged and try with tailwindcss again afterwards.

fs82:
@johanneskoester New setuptools-download version works as expected. I modified the PR accordingly, download seems to work and also report is generated.

johanneskoester (approved):

johanneskoester:
Oh, that is clearly a bug with automerge. It merged before all the tests were done, and some even failed!!

johanneskoester:
I will check the main branch and see if I can fix it in a quick follow-up.

PR #2913, opened 2024-06-11
docs: output files within output directories is an error (#2848)

### LINKED ISSUE #2848: document ChildIOException
I have just been bitten by the undocumented constraint that no rules can output files within a `directory` marked output file. had to dig into the source to figure out what ChildIOException. consider adding a few sentences to [the relevant section](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#directories-as-outputs)...

NelisDrost:
Closes #2848, by providing the documentation suggested, and fixes a bug found while investigating:

The `check_directory_outputs()` function relies on the sorting of file paths to detect nested outputs. This breaks on windows as file paths are separated with `\\` rather than `/`, which is sorted after alphanumeric characters, rather than before.

I've fixed this, and provided a test case, but as there's not currently a mechanism for distinguishing failures before/during execution (before being the intended behavior, as this provides better feedback to the user), it would pass either way. A better test would be the code below, but I don't want to stray from the use of `run()` in the other test cases. It may also be possible to extend run with an `expected_exception` parameter for non-shell execution.

```
def test_github_issue2848():
    # Should fail on parsing with ChildIOException, on Windows previously failed during execution
    with api.SnakemakeApi() as snakemake_api:

        workflow_api = snakemake_api.workflow(
            resource_settings=settings.ResourceSettings(cores=1),
            snakefile=Path(dpath("test_github_issue2848/Snakefile")),
            workdir=Path(dpath("test_github_issue2848")),
        )

        dag_api = workflow_api.dag(dag_settings=settings.DAGSettings(targets={"output/c.txt", "output/dir2/d.txt"}))

        with pytest.raises(ChildIOException):
            dag_api.execute_workflow()
```

### QC

## Summary by CodeRabbit

- Documentation
  - Added cautionary sidebar notes warning that directory-designated outputs are cleaned before execution; avoid placing other outputs inside those directories.
- Bug Fixes
  - Improved internal path handling for directory-output checks to reduce cross-platform errors (notably on Windows).
- Tests
  - Added a test covering workflows mixing file and directory outputs to validate error reporting in edge cases (expected failure).


johanneskoester:
There is `shouldfail=True` as a parameter in the run function. Would that help?

NelisDrost:
`shouldfail` doesn't specify when it should fail - the old behavior (passing parsing, failing during execution) and the desired behavior (failing during parsing, and displaying a useful error message) can't be distinguished this way, i.e.: the test included in the PR (below) passes either way:

```
def test_github_issue2848():
    # Should fail on parsing with ChildIOException, on Windows previously failed during execution
    run(
        dpath("test_github_issue2848"),
        targets=["output/c.txt", "output/dir2/d.txt"],
        shouldfail=True,
    )
```

johanneskoester:
I see. So let's be happy with the kind of test we have for now, and happy to take a separate PR that refines the testing in the future.

snakemake/dag.py
```diff
@@ -232,7 +232,7 @@ def get_unneeded_temp_files(self, job: AbstractJob) -> Iterable[str]:
     def check_directory_outputs(self):
         """Check that no output file is contained in a directory output of the same or another rule."""
         outputs = sorted(
-            {(os.path.abspath(f), job) for job in self.jobs for f in job.output}
+            {(Path(os.path.abspath(f)), job) for job in self.jobs for f in job.output}
```
johanneskoester, line 235:
```suggestion
            {(Path(f).absolute(), job) for job in self.jobs for f in job.output}
```

johanneskoester (approved):

johanneskoester:
Mhm, sorry, did I break something with my little change?

fgvieira:
@NelisDrost can you fix the conflicts? Just so I can merge and close this PR.

johanneskoester:
Done that now! Thanks a lot everybody!

PR #2917, opened 2024-06-14
feat: added snakemake.script.snakemake for type hinting

Hugovdberg:
The current magic injection of the `snakemake` object in python scripts prevents linters and type checkers from recognizing that object. To allow this the `snakemake.script` module is updated to include a `snakemake` variable with only a type annotation of `snakemake.script.Snakemake`, which has added type hints as well. 
The preamble of a python script now also includes the following code (here expanded over separate lines for readability):
```python
from snakemake import script
script.snakemake = snakemake # At this point the snakemake object is already unpickled
del script # This is not really necessary, but I think users should not expect this reference to be present
```

Existing scripts don't have to change, but should you want type checking you can add
```python
from snakemake.script import snakemake
```
in your script, which then makes the object explicitly available in your code.

There is no cross-talking between scripts because each script patches its own copy of the `snakemake.script` module in memory as they all run in separate interpreters.
Load time might be slightly increased, but when unpickling the snakemake object the `snakemake.script` module is already imported. Memory usage is also hardly changed as it is only an extra reference to the `snakemake` object that needs to be stored.

### QC


Hugovdberg:
I did not add a type hint to the snakemake.bench_iteration object as I wasn't sure about the type. It would be nice to have that hint as well. 

Also, I feel like this explicit import is more pythonic, even though the monkey patching isn't less "magic" than the current injection method. For now it should definitely stay backwards compatible, but food for thought for a future major release where such breaking changes might be more expected.

johanneskoester:
Thanks! Two things:
1. Why not just `from snakemake.script import snakemake`? I guess I miss some special effect on the type hinting if you do it the way you do now. Maybe you could formulate the answer also as a comment in the code where you do the import.
2. Could you please format your PR with black?

Hugovdberg:
@johanneskoester I reformatted the code using black.

However, I'm not quite sure what you mean with your first question. Do you mean why I do `from snakemake import script` instead of `from snakemake.script import snakemake`? If I would do that the preamble would look something like this:
```python
import sys
sys.path.extend({repr(searchpaths)})
import pickle
from snakemake.script import snakemake 
# now snakemake is only a local reference the not-yet-set object in the snakemake.script module
snakemake = pickle.loads({snakemake})
# now the local reference is changed to the unpickled object.
# However, the snakemake.script.snakemake reference is untouched
{shell_exec_stmt} logger.printshellcmds = {logger.printshellcmds}
{preamble_addendum}
```
As the reference inside the module is untouched an import inside the user code of `snakemake.script.snakemake` would actually overwrite the local reference to the unpickled object with the unset value inside the module.

The way I implemented it both the local snakemake variable, and the snakemake.script.snakemake variable reference the same object. Therefore, if you manually import the object the local variable just stays the same during runtime. However, the added value is that your IDE actually knowns the variable.
```python
import sys
sys.path.extend({repr(searchpaths)})
import pickle
snakemake = pickle.loads({snakemake})
# now snakemake is a local reference to the unpickled object.
{shell_exec_stmt} logger.printshellcmds = {logger.printshellcmds}
from snakemake import script
# script is a reference to the module
script.snakemake=snakemake
# now the snakemake attribute inside the script module is changed to a reference to the unpickled object
# if you now do from snakemake.script import snakemake you actually get the unpickled object
del script
{preamble_addendum}
```

I think we could change it as follows as well to make it perhaps a little cleaner:
```python
import sys
sys.path.extend({repr(searchpaths)})
import pickle
from snakemake import script
# script is a reference to the module
script.snakemake = pickle.loads({snakemake})
# now script.snakemake references to the unpickled object, but it is not in the local scope
del script
from snakemake.script import snakemake 
# now the snakemake object is also in the local scope
{shell_exec_stmt} logger.printshellcmds = {logger.printshellcmds}
{preamble_addendum}
```

Does this answer your question? And do you prefer either of the current implementation or the version where the module level reference is set directly to the unpickled object?

johanneskoester:
The last suggestion seems indeed clearer to me. Would be great if you could change it accordingly!

Hugovdberg:
I reorganised the preamble, and for readability expanded the preamble in the code, with an explicit minification in order to get the original formatting back.
For some reason my version of black generates slightly different formatted code in case of the abstract stubs. I couldn't find a way to get this to work without changing the functions to explicitly `return NotImplemented`.

snakemake/io.py
```diff
@@ -68,11 +65,11 @@ class AnnotatedStringInterface(ABC):
     @property
     @abstractmethod
     def flags(self) -> Dict[str, Any]:
-        ...
+        return NotImplemented
```
johanneskoester, line 68: Why this change? Using the ellipsis literal for abstract methods is common practice. Is it because of the type hint?

johanneskoester (commented):
Final question below. In line with this: would you mind adding the type checking to the CI as well? So that we are protected against future violations?

snakemake/io.py
Hugovdberg: I commented on the PR to explain this, but for some reason the CI version of black insists the ellipsis belongs on a separate line, and my version of black insists it belongs at the end of the line, and I couldn't find why this keeps happening.

Hugovdberg:
@johanneskoester testing raises this error, but I thought snakemake was Python 3.11+, and this seems to run on Python 3.9: https://github.com/snakemake/snakemake/actions/runs/9641698937/job/26588094767#step:7:414
Do you have any idea why this error occurs?

snakemake/io.py
Hugovdberg: Ahh, I see, this is related to changes in black version 24+, they changed their opinions on several fronts. CI uses black 23.12, I use 24.4. I will pin it to 23.12 in my dev environment and revert this if that fixes it.

Hugovdberg:
> Final question below. In line with this: would you mind adding the type checking to the CI as well? So that we are protected against future violations?

I think this could be useful, but I think that adding that would best be done in a separate PR, and initially only optional, as large amounts of the codebase are not type hinted. It would be useful to see if we can configure is such that only the additions require type hints, without forcing contributors to first add type hints to the entire code base.

Hugovdberg:
> @johanneskoester testing raises this error, but I thought snakemake was Python 3.11+, and this seems to run on Python 3.9: https://github.com/snakemake/snakemake/actions/runs/9641698937/job/26588094767#step:7:414 Do you have any idea why this error occurs?

Ah, I found the issue: https://github.com/snakemake/snakemake/blob/1b75087fa10cfcfb06e41a6286cb51d10df3ee3f/tests/test_singularity/Snakefile#L14-L20

For now I replaced the Self type hint with the verbose option it's meant to replace.

Hugovdberg:
@johanneskoester could you please allow the CI to run so I can see if all errors are fixed now?

Hocnonsense:
Is it possible to use a var from libaray "typing"?

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from snakemake.srcipt import Snakemake

    snakemake: Snakemake = None 

# snakemake.input
```

Hugovdberg:
> Is it possible to use a var from libaray "typing"?
> 
> ```python
> from typing import TYPE_CHECKING
> 
> if TYPE_CHECKING:  # never run into this block
>     from snakemake.srcipt import Snakemake
> 
>     snakemake: Snakemake = None 
> 
> # snakemake.input
> ```

of course this is possible, see snakemake/snakemake-lang-vscode-plugin/#32, but it is a lot more cumbersome. I also put it in a common file and say `from common import *`, such that the snakemake object is only imported during type checking. But the proposed solution in this PR is much cleaner, doesn't require the `TYPE_CHECKING` barrier at all, and also includes more type hints on the Snakemake object itself.

Hugovdberg:
@johanneskoester I merged all changes from `main` and reformatted with the new `black` version. Could you please allow the CI to run once again?

johanneskoester:
@Hugovdberg do you want an invite for the snakemake github org? I highly appreciate your contributions here, and it would simplify your dev process.

Hugovdberg:
@johanneskoester sure, it would be an honour

johanneskoester:
done!

johanneskoester (approved):

johanneskoester:
Thansk a lot!

PR #2921, opened 2024-06-17
feat: add `can_transfer_local_files` to executor plugin interface

jhiemstrawisc:
Some plugins that don't assume a shared filesystem are capable of managing their own file transfers from the workflow host to the remote execution points. An example of this is HTCondor, which rarely assumes a shared filesystem between the access point and the execution point, and where HTCondor is responsible for transferring the files.

The `can_transfer_local_files` flag indicates the plugin -- and not snakemake -- is responsible for managing file movement between these two locations.

### QC


jhiemstrawisc:
@johanneskoester, we briefly discussed this parameter when I mentioned that I'm working to make the HTCondor executor work with non-shared filesystems. I've gotten things to work locally, and I'm starting to clean things up to merge things back upstream, and after working on the executor, these were the places I found that interfered with allowing the executor to manage its own file transfers when passing `--shared-fs-usage none`. However, this is my first real traipse around the Snakemake source code, so let me know if I should change approach or if I missed something else.

The value is added to the executor interface in https://github.com/snakemake/snakemake-interface-executor-plugins/pull/67

PR #2930, opened 2024-06-25
fix: parse f-string as-is from source

### LINKED ISSUE #2657: parsing fstring with if else on params

### LINKED ISSUE #2881: nested for loop in f-string
**Snakemake version**
Bug in latest versions. Tested versions 8.7.0 and 8.11.3.
Works in version 7.18.2

**Describe the bug**
I often use nested 'for' loops to create a list (outside of a rule). This works fine, but if the loop is placed WITHIN an f-string it does not work with the latest versions of snakemake. Before version 8 it was working.

**Logs**
$ snakemake -j1
SyntaxError in file Snakefile, line 6:
invalid syntax. Perhaps you forgot a comma? (Snakefile, line 6)

**Minimal example**
# works for both versions
a = [(x, y) for x in [1,2,3] for y in [3,1,4]]
print(f"works for both versions: {a}")

## works for older versions only
print(f"works only in old version: {[(x, y) for x in [1,2,3] for y in [3,1,4]]}")

Hocnonsense:
This is another (and hopefully the last) fix for f-string change in python3.12.

Now everthing inner f-string are ignored, and only care about where f-string start and end, and read it from snakefile as-is.

will fix #2881 and fix #2657
related PR: #2864, #2649

### QC

### DIFF
--- a/snakemake/parser.py
+++ b/snakemake/parser.py
@@ -108,24 +108,31 @@ class TokenAutomaton:
             self.was_indented |= self.indent > 0
 
     def parse_fstring(self, token: tokenize.TokenInfo):
-        # only for python >= 3.12, since then python changed the
-        # parsing manner of f-string, see
-        # [pep-0701](https://peps.python.org/pep-0701)
-        related_lines = {token.start[0]}
+        """
+        only for python >= 3.12, since then python changed the
+        parsing manner of f-string, see
+        [pep-0701](https://peps.python.org/pep-0701)
+
+        Here, we just read where the f-string start and end from tokens.
+        Luckily, each token records the content of the line,
+        and we can just take what we want there.
+        """
+        related_lines = token.start[0]
         s = token.line
         isin_fstring = 1
         for t1 in self.snakefile:
-            if t1.start[0] not in related_lines:
+            if related_lines < t1.start[0]:
+                # go to the next line
+                related_lines = t1.start[0]
                 s += t1.line
-                related_lines.add(t1.start[0])
             if t1.type == tokenize.FSTRING_START:
                 isin_fstring += 1
             elif t1.type == tokenize.FSTRING_END:
                 isin_fstring -= 1
             if isin_fstring == 0:
                 break
+        # trim those around the f-string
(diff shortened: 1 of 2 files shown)


cmeesters (approved):
Awesome! It would merit a few comments in the code, though.

johanneskoester (approved):

PR #2937, opened 2024-07-01
fix: duplicate wildcards

Hocnonsense:
will fix 1849

### QC


Hocnonsense:
@johanneskoester 

I've found that the black defined in [.github/workflows/main.yml](https://github.com/snakemake/snakemake/blob/ab8d2ddc38b610b0bf5b2e5368cfaf349921470a/.github/workflows/main.yml) seems to be with too low version (23.12.1) compared to the current version, and if the code is formatted by black=24.*, it will be rejected by black=23.*

tests/test_globwildcards/Snakefile
```diff
@@ -1,11 +1,25 @@
+(IDS,) = glob_wildcards("test.{id}.txt")
```
johanneskoester, line 1: I think the files for this first glob are missing. And further, please add a corresponding test function to tests/tests.py so that the testcase is actually executed.

johanneskoester (changes requested):
Thank you so much! One thing below.

tests/test_globwildcards/Snakefile
Hocnonsense: Thanks for your approvel!

test_globwildcards is an existed test, and I just modified a little bit. It is announced in
https://github.com/snakemake/snakemake/blob/bdd732d1f98d0af0362115f74baa728bfc976851/tests/tests.py#L289-L290

The [tests](https://github.com/snakemake/snakemake/tree/main/tests/test_globwildcards/expected-results) are added  11 years ago

johanneskoester: Ah, sorry. Still too early in the morning for me :-)

johanneskoester (approved):

Issue #2949, opened 2024-07-08
Slow DAG building and job completion due to large file checksumming


laf070810:
I have several TB data in my DAG, and file checksumming happens in both DAG building and jobs completion stage, causing the DAG building and job completion during running be extremely slow (more than 30mins or even 1h). See the diff in https://github.com/laf070810/snakemake for the affected parts. 

I suggest adding some file size checking to avoid large file checksumming, or adding a config option to completely disable file checksumming.

Related to #2920

johanneskoester:
There is a filesize limit for checksumming, but maybe that is too permissive? Should probably also be configurable. The limit is currently hardcoded in the method is_checksum_eligible of snakemake.io._IOFile.

johanneskoester:
The setting should go into snakemake.settings.types.DAGSettings.

Issue #2920, opened 2024-06-16
Slow DAG postprocessing on high-latency filesystems


benjeffery:
I'm running a workflow with ~100k jobs in on am HPC where filesystem latency is relatively high. (Compute and filesystem are not co-located for reasons outside my control).

DAG building for this is taking ~45 minutes, almost all of which is in `DAG.update_needrun` while it performs file existence and `mtime` checks. For the file existence checks I've hacked in a cache (only active during `update_needrun`) which uses `os.listdir` in `_ioFile.exists_local`:

```python
    async def exists_local(self):
        if not cache_on:
            return os.path.exists(self.file)
        _dir = os.path.dirname(self.file)
        if _dir not in _local_cache:
            if not os.path.exists(_dir):
                _local_cache[_dir] = []
            else:
                _local_cache[_dir] = os.listdir(_dir)
        return os.path.basename(self.file) in _local_cache[_dir]
```

Using `os.listdir` greatly reduces the hits to the filesystem. This halves the runtime of `update_needrun`, the remaining time being `mtime` checks. 

Does a mechanism like this make sense? I would be happy to work up a proper PR if so.

Many thanks!

benjeffery:
Digging further it seems that `rule.workflow.iocache.exists_local` is populated, but not used by `_IOFile.exists_local`. I'll try and use that instead.

cmeesters:
> I would be happy to work up a proper PR if so.

That would be great. Thank you! Feel free to request a review from me.

One question though: Usually clusters differentiate between HOME (a file system optimized for many small files) and a project file system (various types) optimized for fast linear reads of big files. Can you sketch your layout, please? I would like to compare.

I am facing a similar issue now: We have a gpfs HOME (native on login nodes and NFS on compute nodes) and a lustre project file system (native mounts). Alas, I have to store the conda environment for joint development on the lustre FS, also because we have a narrow file limit on the HOME FS.

tbooth:
I have a fork of Snakemake 7 where I tried to eliminate as many `stat()` calls as possible: https://github.com/edinburghGenomics/snakemake

It's not suitable for general consumption because it breaks some Snakemake features, but it's good enough for my purposes and is used for some production jobs here. It speeds up DAG building drastically (on Lustre). Among other things it skips time checks entirely on files declared as "ancient".

[...]

PR #2965, opened 2024-07-15
perf: cache mtime of scripts and notebooks

laf070810:
`mtime` of `script` or `notebook` is checked repeatedly when the same rule is wildcard-expanded, causing a lot of `os.lstat` calls repeated on the same file, which is unnecessary and can be slow on some filesystems. 

This PR adds script mtime caching using `functools.lru_cache`. It can speedup DAG building time. 

Related issues: https://github.com/snakemake/snakemake/issues/2920, https://github.com/snakemake/snakemake/issues/2949

### QC

## Summary by CodeRabbit

- **New Features**
  - Introduced a utility function to efficiently retrieve and cache the modification time of script files, improving performance.
  
- **Improvements**
  - Enhanced the existing functionality to streamline file modification time checks, leading to cleaner and more maintainable logic.


johanneskoester (approved):

johanneskoester:
Very good catch!

PR #2966, opened 2024-07-15
fix: add missing output check

laf070810:
This PR proposes adding missing output check for job running. 

For example, we have two rules A and B, and B depends on A. 

```
rule A:
    input:
        "test.txt"
    output:
        "test2.txt"
    params:
        a=1
    shell:
        "cp {input} {output}"

rule B:
    input:
        "test2.txt"
    output:
        "test3.txt"
    params:
        a=2
    shell:
        "cp {input} {output}"
```

We can `touch test.txt` and run the full workflow with `snakemake -c 1 test3.txt`. Then if we change the shell command or params or something, e.g.

```
rule A:
    input:
        "test.txt"
    output:
        "test2.txt"
    params:
        a=1
    shell:
        "cp {input} nonexistent_dir/{output}"

rule B:
    input:
        "test2.txt"
    output:
        "test3.txt"
    params:
        a=2
    shell:
        "cp {input} {output}"
```

where `nonexistent_dir` is a directory which does not exist. Then rerun the workflow, it will fail on rule A. But then if we want to run the workflow again, it will say that `Nothing to be done (all requested files are present and up to date).`, because the intermediate output `test2.txt` was deleted in the first run. 

Of course, in the above example, the failure is certain. But if rule A is correct but just fails for random reasons such as IO, then the above issue will cause trouble when we want to rerun the workflow. 

Related: https://stackoverflow.com/questions/45989802/can-snakemake-be-forced-to-rerun-rules-when-files-are-missing

### QC

## Summary by CodeRabbit

- **New Features**
	- Enhanced tracking of job execution reasons with improved handling of missing output files.
	- Refined logic for determining job execution conditions based on input and output states.

- **Documentation**
	- Updated comments and documentation for better readability and maintainability.


johanneskoester:
Snakemake considers missing output if it is one of the target files. Otherwise, it intentionally and lazily ignores it.
However, I see the issue you have here. I think a more accurate fix is to also store job metadata for failed jobs, so that an evential change is properly detected when the workflow is rerun and the job has been fixed.

johanneskoester:
Fix is here: https://github.com/snakemake/snakemake/pull/3237, closing in favor of that.
Nevertheless, huge thanks for bringing this up, and sorry for the delay!

PR #2987, opened 2024-07-27
docs: Change sha256 checksum in docs to more realistic example

wimglenn:
It is strange/confusing that the example checksum in the docs does not look like a sha256.

I've changed it to a real hexdigest (the sha256 of the content b"test", as was also used in the tests: https://github.com/snakemake/snakemake/blob/e8735c1477a2a82110757ba86bbd1ccbcaf327ba/tests/test_ensure/Snakefile#L17-L24

ref: https://stackoverflow.com/questions/78801757/what-is-the-encoding-of-this-sha256-hash-in-the-snakemake-documentation


Hugovdberg:
Thank you for spotting that and suggesting a fix. However, the current testsuite fails on an unrelated issue, which I think is caused by numpy 2. The last successful run of the test suite still ran numpy 1.26. I will try to find in which spots we need to fix the numpy version to <2 in order to get the original tests running again. You can then pull in those changes to get this PR up and running again.

Hugovdberg:
Work in progress, but there are more issues with the tests, follow #2988 for more info

johanneskoester (approved):

johanneskoester:
tests fixed.

johanneskoester:
thanks!

PR #2988, opened 2024-07-28
tests: fix broken tests due to breaking changes in dependencies

Hugovdberg:
String representation of scalars changed in numpy 2 (https://numpy.org/doc/stable/release/2.0.0-notes.html#representation-of-numpy-scalars-changed), which causes tests to fail. This PR fixes the numpy version in the tests to the last pre-2.0 version for now in order for other PRs to be able to continue.
At a later stage the tests should probably be updated to work on more recent verions of numpy or contain explicit formatting.

### QC

### DIFF
--- a/test-environment.yml
+++ b/test-environment.yml
@@ -21,7 +21,7 @@ dependencies:
   - pysftp
   - requests
   - responses
-  - numpy
+  - numpy < 2
   - appdirs
   - pytools
   - docutils
--- a/test-environment.yml
+++ b/test-environment.yml
@@ -59,6 +59,7 @@ dependencies:
   - nodejs # for cwltool
   - immutables
   - conda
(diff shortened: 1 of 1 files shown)


Hugovdberg:
Also `paramiko` uses `TripleDES` from a module that is deprecated in cryptography 43, causing the literal output of some tests to include unexpected deprecation warnings. Until that is fixed limit cryptography to < 43.

Hugovdberg:
This requires quite some work, apparently all kinds of dependencies have changed and I don't think just hacking the test environment to some magic combination of packages is the right way to go. @johanneskoester what do you think we should do about this?

johanneskoester:
I have fixed the test instead. But thanks for pointing me to the issue!

Issue #2969, opened 2024-07-16
Database instead of file-based logging and book keeping? Performance issues on HPC with many Snakemake users and millions of meta files


tamasgal:
We started to use Snakemake heavily in our experiment and teaching students to do even small scientific analyses in a nice, reproducible way with the help of this tool. It works great but we have big issues with our HPC clusters...

We process large amounts of data (O(100TB)) and a large amount of files (tens of millions) and Snakemake is creating an extremely large amount of meta files in the `.snakemake` folders. The biggest problem is the impact on the I/O performance of the HPC storage systems, often optimised for large files and the bottleneck is often the IOPS due to many users acting at the same time. We basically constantly fight with non-responsive file systems (not even `ls` is working) which are caused by Snakemake files and the way Snakemake accesses those.
It's not related to a specific HPC cluster, we face the same issues everywhere around the world.

**Describe the solution you'd like**
We already use options like `shallow shadow` and try to force Snakemake to put files on local storages but the possibilities are limited since worker nodes need access to logs etc. I am wondering if how much work it would be to allow the usage of e.g. a database backend for meta data and log files. SQLite or something other serverless solution which ofheavily reduces the number of files.

**Describe alternatives you've considered**
I would happy to hear alternatives from Snakemake experts, maybe even existing solutions.

cademirch:
I think this is a great idea. Though would likely be a major undertaking to implement and test. A newer workflow manager, [redun](https://github.com/insitro/redun) uses a [DB backend](https://insitro.github.io/redun/db.html) like you are suggesting.

wintermutant:
Coming here late - any updates?

I work as a computational scientist at an HPC at Purdue University and we see similar issues. I have a working prototype that provides a solution to these, but we still need to do quite a bit of work. Uses a sqlite3 database and it basically serves as tarballing the whole directory up except we add structure so you can query (for example) all results from 1 rule or another. Looking to make it helpful for research scientists running a lot of input through similar pipelines then running meta-analyses.

cademirch:
Hi @wintermutant - thanks for reviving this. Eventually this is something I would like to tackle. I've also experimented with sqlite3 as well as KV stores for storing metadata. The KV store is easier to implement currently given Snakemake's metadata schema, and I'll put this on my list to revisit at the Snakemake hackathon next month. 

For posterity, the main blocker I encountered was that I could not guarantee only 1 Snakemake process was writing to the database. However that might have been alleviated by #3679.

[...]

Issue #2967, opened 2024-07-15
Performance degradation after running snakemake for some time


laf070810:
**Snakemake version**

`Snakemake 8.16.0`

**Describe the bug**

When running snakemake for a certain period of time, the scheduler gradually becomes lazy. Job finalization, job selection and job starting will all become slower and slower. 

Correspondingly, the `NLWP` in `htop` (which is the number of threads) of snakemake main process gradually increases, from ~100 to ~1000. CPU usage also increases. The number of opened `/proc/****/stat` in `lsof` output also increases. 

For possible reason of this issue, my current guess is some kind of performance degradation with `concurrent.futures.ThreadPoolExecutor` in the executor. 

**Logs**

**Minimal example**

**Additional context**

epruesse:
I am observing the same symptoms - after about 20k jobs, Snakemake takes a minute to submit the next (small) set of jobs.

I am _not_ seeing an increase in NLWP, however. What I'm seeing is lots of `stat()` and `openat()` calls on `.snakemake/metadata'. Always the same pattern: stat, openat, fstat, read, close, repeating 5 times for the same metadata file, then moving on. 

I'm on 7.32.2 with this .

laf070810:
> I am observing the same symptoms - after about 20k jobs, Snakemake takes a minute to submit the next (small) set of jobs.
> 
> I am _not_ seeing an increase in NLWP, however. What I'm seeing is lots of `stat()` and `openat()` calls on `.snakemake/metadata'. Always the same pattern: stat, openat, fstat, read, close, repeating 5 times for the same metadata file, then moving on.
> 
> I'm on 7.32.2 with this .

Maybe you can try removing `.snakemake/metadata` and use `--drop-metadata` when running snakemake. This improves performance a lot in my cases, though it will still degrade, just degrades slower.

epruesse:
To clarify:
This **_is not_** about general performance issues with metadata or DAG build (tag = enhancement).
This **_is_** about the **severe degradation** during run time. It can be fixed by restarting Snakemake (tag = bug).

Something is accumulating while Snakemake operates and eventually leads to <1 job/minute scheduled. There is a _bug_ somewhere.

epruesse:
It appears that my use of `--notemp` for development purposes also had a large negative impact, possibly even causative. Without this, it seems to work acceptably again.

epruesse:
[...]

PR #2991, opened 2024-07-31
Bulk, cached DAG update for multiple finished cluster jobs

epruesse:
I've been having performance issues with 7.32.4 where the checkpoint update in the DAG becomes a bottleneck and I cannot even supply a single cluster node with jobs. This patch series ameliorates the problem for me by

a) Enabling the persistence cache and the iocache for the duration of one `dag.postprocess()` pass 
b) Allowing the scheduler to process multiple finished jobs at once

I'm seeing a performance improvement >60 fold (i.e. I'm once again submitting faster than the cluster can work the jobs). 

The rationale for a) is that changing state of `mtime`, `file_exists` or metadata during the course of a `update_needrun()` DAG update is probably a recipe for disaster anyway. By enabling the cache exactly for one pass, then clearing it, we get a consistent state and avoid the multiple reads of metadata files (up to 5 times if all triggers are enabled) and repeated calls to `stat()`. I've conservatively limited the caching to the scope of one `update_needrun()` call from `postprocess()` to avoid side effects, which seems to work, but I can't claim I understand all the code that might be affected.

The implementation for b) was surprisingly simple as most of the capability is already there. This merely makes it so that the scan of active jobs in the `GenericClusterExecutor` doesn't release the `_open_jobs` semaphore for each job immediately, but waits to do this at the end of a full scan of the active jobs. Since the DAG can already handle multiple jobs at once as required by the group-job feature, only minor changes where needed there to make it so a large number of completed jobs can be addressed with one DAG update and thus one scan of the time stamps and metadata.

Caveats: 
- This only implements the change for `GenericClusterExecutor`
- It's a patch against the series-7. I checked the 8 only briefly, but it looks like this may at least conceptually be applicable there as well. 
- I've not written any unit tests because I'm not even sure whether you accept patches for series-7/
- I'm also not sure how much of my degradation issue is really caused by a different problem. I'm not sure why it's scanning all those files with every job finish (>100,000 stat calls for me). This works for me though.

Related issues:
#2967  #2969

### QC

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -1621,24 +1621,26 @@ class DAG(DAGExecutorInterface):
         """Finish a given job (e.g. remove from ready jobs, mark depending jobs
         as ready)."""
 
-        self._running.remove(job)
+        jobs = []
+        for job_ in job:
+            self._running.remove(job_)
 
-        # turn off this job's Reason
-        if job.is_group():
-            for j in job:
-                self.reason(j).mark_finished()
-        else:
-            self.reason(job).mark_finished()
+            # turn off this job's Reason
+            if job_.is_group():
+                for j in job_:
+                     self.reason(j).mark_finished()
+            else:
+                self.reason(job_).mark_finished()
 
-        try:
-            self._ready_jobs.remove(job)
-        except KeyError:
-            pass
+            try:
+                self._ready_jobs.remove(job_)
+            except KeyError:
+                pass
 
-        if job.is_group():
-            jobs = job
-        else:
-            jobs = [job]
+            if job_.is_group():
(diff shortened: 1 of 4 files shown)


epruesse:
Black formatting and SonarCloud are unfixable - they are triggered by series-7 code. Fixing that would be out of scope of this PR.

epruesse:
CI fails due to pulp version issue.

johanneskoester:
Hi! Thanks a lot. Unfortunately, we can't accept changes on series 7 anymore. May I ask you to migrate this to the current main branch with Snakemake 8 if it also applies there? Thanks a lot.

PR #2993, opened 2024-08-01
fix: parse config yaml using safe loader with datetime handler

### LINKED ISSUE #2374: 🐛  Nested config parsed as string when using --config
If nested config is passed to Snakemake via the `--config` flag all values are interpreted as strings. 

```bash
$ snakemake --config test={'test_bool': True, 'int_test': 1}
 'test': {'test_bool': 'True', 'int_test': '1'}  # Note strings
```

This is because the [yaml parser is using the `BaseLoader`](https://github.com/snakemake/snakemake/blob/836fd2a8cf856b32f10d8bb9cdf61e4999775053/snakemake/__init__.py#L1036) which does not convert YAML scalars to their corresponding Python types (instead returning them as strings). 

If I update the yaml parser to use SafeLoader then the expected behaviour is achieved. 
```python
- yaml_base_load = lambda s: yaml.load(s, Loader=yaml.loader.B̶a̶s̶e̶L̶o̶a̶d̶e̶r̶)
+ yaml_base_load = lambda s: yaml.load(s, Loader=yaml.loader.SafeLoader)
```
```bash
$ snakemake --config test={'test_bool': True, 'int_test': 1}
 'test': {'test_bool': True, 'int_test': 1}  # Note correct types
```
Should I submit a PR to fix this issue?

ollie-bell:
Fixes #2374, relates to #709 

### QC

### DIFF
--- a/snakemake/cli.py
+++ b/snakemake/cli.py
@@ -241,7 +241,16 @@ def parse_config(entries):
     """Parse config from args."""
     import yaml
 
-    yaml_base_load = lambda s: yaml.load(s, Loader=yaml.loader.BaseLoader)
+    class SafeLoaderTimestamp(yaml.SafeLoader):
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.add_constructor("tag:yaml.org,2002:timestamp", self.construct_yaml_str)
+        
+        @staticmethod
+        def construct_yaml_str(loader, node):
+            return loader.construct_scalar(node)
+
+    yaml_base_load = lambda s: yaml.load(s, Loader=SafeLoaderTimestamp)
     parsers = [int, float, _bool_parser, yaml_base_load, str]
     config = dict()
     if entries:


ollie-bell:
This needs a test using the CLI, not sure how to implement that as a test in the repo.

```shell
snakmake [...] --config 'test_config={test_bool: true, test_int: 1, test_float: 0.5, test_date: 2024-01-08, test_str: hello}'
```

this should give the config:

```python
{"test_config": {"test_bool": True, "test_int": 1, "test_float": 0.5, "test_date": "2024-01-08", "test_str": "hello"}}
```

johanneskoester:
You can pass `shellcmd="snakemake --config ..."` to the run method in the test suite. This way, also the CLI will be part of the test.

ollie-bell:
> You can pass `shellcmd="snakemake --config ..."` to the run method in the test suite. This way, also the CLI will be part of the test.

having trouble running snakemake tests locally, followed instructions https://snakemake.readthedocs.io/en/stable/project_info/contributing.html#testing-guidelines

```
___________________________________________________________________________ ERROR collecting tests/test_conda_python_script/test_script.py ___________________________________________________________________________
import file mismatch:
imported module 'test_script' has this __file__ attribute:
  /Users/ollieb/Documents/snakemake/tests/test_conda_python_3_7_script/test_script.py
which is not the same as the test file we want to collect:
  /Users/ollieb/Documents/snakemake/tests/test_conda_python_script/test_script.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
_______________________________________________________________________ ERROR collecting tests/test_script_py/scripts/test_explicit_import.py ________________________________________________________________________
ImportError while importing test module '/Users/ollieb/Documents/snakemake/tests/test_script_py/scripts/test_explicit_import.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../miniforge3/envs/snakemake-testing/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_script_py/scripts/test_explicit_import.py:1: in <module>
    from snakemake.script import snakemake
E   ImportError: cannot import name 'snakemake' from 'snakemake.script' (/Users/ollieb/Documents/snakemake/snakemake/script/__init__.py)
```

Wytamma:
I'd love to see this merged! Any way I can help? @ollie-bell @johanneskoester

snakemake/cli.py
```diff
@@ -242,8 +242,17 @@ def parse_config(entries):
     """Parse config from args."""
     import yaml
 
-    yaml_base_load = lambda s: yaml.load(s, Loader=yaml.loader.BaseLoader)
-    parsers = [int, float, _bool_parser, yaml_base_load, str]
+    class SafeLoaderTimestamp(yaml.SafeLoader):
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.add_constructor("tag:yaml.org,2002:timestamp", self.construct_yaml_str)
```
johanneskoester, line 248: I must say I am a bit unsure about this. Seems too implicit to me to automatically parse dates/timestamps. I think it is better if users do that themselves if they need it. Using the safe loader is good though.

johanneskoester (changes requested):

snakemake/cli.py
johanneskoester: Mhm, this is indeed an interesting question. I think yes, this is the default behavior of yaml in python, hence it makes sense to me.

Wytamma: Just to add my 2 cents I’d say the default safeloader behaviour is inconsistent because if you then write the config with datetime object back to YAML it’ll be different to the starting file.

@ollie-bell’s implementation overrides the default implicit behaviour (which is to parse date time strings as datetime objects) by replacing the datetime constructor with the str constructor. 

I think you could rename SafeLoaderTimestamp to something like SafeLoaderNoTimestamp to make this clear.

Issue #3001, opened 2024-08-07
snakemake --lint mistakes regex for path


cmeesters:
Hi,

**Snakemake version**
8.16.0

**Describe the bug**

with this submodule in a workflow

```Python
import os
from pathlib import Path
import sys
from itertools import product

import pandas as pd
from snakemake.remote import FTP
from snakemake.utils import validate
from snakemake.exceptions import WorkflowError

validate(config, schema="../schemas/config.schema.yaml")

samples = (
    pd.read_csv(
        os.path.join(
            os.path.realpath(os.path.dirname(workflow.configfiles[0])),
            config["samples"],
        ),
        sep=r"\s+",
        dtype={"sample": str, "condition": str, "condition2": str, "batch_effect": str},
        header=0,
        comment="#",
    )
    .set_index("sample", drop=False)
    .sort_index()
)
...
```

**Logs**

I get this in the CI logs:

```
Lints for snakefile /github/workspace/workflow/rules/commons.smk:
    * Path composition with '+' in line 19:
      This becomes quickly unreadable. Usually, it is better to endure some
      redundancy against having a more readable workflow. Hence, just repeat
      common prefixes. If path composition is unavoidable, use pathlib or
      (python >= 3.6) string formatting with f"...".
      Also see:
```

**Additional context**

On the command line with snakefmt version 0.10.2 and black (internally used, version 24.4.2), I get no error.

Apparently,  `sep=r"\s+"` is considered path concatenation, which it isn't.

We could use `\t` instead for a while, but this is semantically not identical.

Cheers,
Christian

cmeesters:
@johanneskoester I think, the cause is an overly picky regex in `snakemake/linting/snakefiles.py`, but my tinkering with a negative lookahead did not yield a fix, so far.

cmeesters:
closed as fixed in PR #3002

PR #3002, opened 2024-08-08
fix: allowing trailing '+' in name patterns

cmeesters:
This PR should fix issue #3001: 

A workflow might contain simple regexes within Python code, e.g. `r"\s+"`. Prior to this PR the `NAME_PATTERN` of the linter considered this first part, the `s` as part of a string to be concatenated with a path. 

### QC

There is no additional test (yet. Not sure it needs one). Docs do not need updating as the intended behaviour does not change.

## Summary by CodeRabbit

- **New Features**
	- Improved name validation logic by updating the regex pattern, ensuring that file names do not end with a '+' character, reducing the risk of unintended exclusions.

- **Documentation**
	- Added comments to clarify the implications of the regex changes and to suggest alternatives for different separator needs.


cmeesters:
brrrr, I consider this failure unrelated to the PR:

```
=========================== short test summary info ============================
FAILED tests/tests.py::test_paramspace - AssertionError: wrong result produced for file 'results/sep/simulations/alpha_is_2/beta_is_0.0/gamma_is_3.9.tsv':
------found------
{'alpha': np.int64(2), 'beta': np.float64(0.0), 'gamma': np.float64(3.9)}
-----expected-----
{'alpha': 2, 'beta': 0.0, 'gamma': 3.9}
-----------------
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
========= 1 failed, 22 passed, 306 deselected, 111 warnings in 51.20s ==========
```
But, of course, the rest will not run.

johanneskoester:
Seems like the floats are not correctly converted into plain string representations. A fix would be greatly appreciated, I don't have any time in the next days.

cmeesters:
Already thought, that a fix might be trivial, but whether I will find the time ... Let's see.

cmeesters:
Definitively not trivial. We could:

- change the expected results to meet the code output. 
- change the test.
- change the actual code. I found the spot, it's a tiny lambda function, which could be extended.

All three options sound like tinkering to me.

johanneskoester (approved):

johanneskoester:
Thank you!

cmeesters:
Thank you for the swift review!

zhangxiaoxing:
I'm a bit confused here, as a long-running snakefile now produces a new warning:
`"/path/to/snakefile:666: SyntaxWarning: invalid escape sequence '\d'"`
The line number could not be easily matched back in the snakefile, but somewhere I do have such a line:
```
output:
    "output/outputfile_{index,\d+}.out"
```
This is to limit the wildcard matching to only digits. Should I ignore the warning? Should I update the rules? Should I provide more detail?

Thanks!

cmeesters:
Thanks for reporting this!!!

You may ignore the warning. I will not. Can you please indicate the entire rule? Is there, perhaps, more detail to the warning message?

Actually, this PR _only_ affected the linter, not the Snakemake core. So, this might be totally unrelated. But we should sort this out, before perhaps opening a full issue.

cmeesters:
Ah, yes, this alone is a faulty syntax expression. Try

```Python
output:
     r"test{num,\d+}.txt"
```

zhangxiaoxing:
> Ah, yes, this alone is a faulty syntax expression. Try
> 
> ```python
> output:
>      r"test{num,\d+}.txt"
> ```

That indeed fixed the warning. Thanks.

zhangxiaoxing:
> Ah, yes, this alone is a faulty syntax expression. Try
> 
> ```python
> output:
>      r"test{num,\d+}.txt"
> ```

Still I suggest the documentation to be refined. Currently it's still showing this faulty syntax...
image

cmeesters:
ah, thanks. It's a Python thing. Will fix the docs.

PR #3040, opened 2024-08-22
fix: `mem` and `disk` inference fixes

### LINKED ISSUE #2774: `mem_mb` and `mem_mib` do not match `mem` when `mem` is specified with unit.
**Snakemake version**

Version 8.9.0

**Describe the bug**

When `mem` is specified with a unit, the `mem_mb` and `mem_mib` resources do not make sense. For example, when the request is `mem="2G"`, `mem_mb` is 1907 and `mem_mib` is 954.

If `2G` is supposed to be gigabyes, then `mem_mb` should be 2000 and `mem_mib` should be 1907. If `2G` is supposed to be gibibytes, then `mem_mb` should be 2147 and `mem_mib` should be 2048.

**Logs**

```
[Fri Mar 22 14:02:58 2024]
rule test:
    output: test.txt
    jobid: 1
    reason: Missing output files: test.txt
    threads: 6
    resources: mem_mb=1907, mem_mib=954, disk_mb=1000, disk_mib=954, tmpdir=<TBD>, lsf_project=acc_LOAD, lsf_queue=premium, runtime=2880, mem=2G
```

**Minimal example**

Here is a test rule:
```
rule test:
  output: "test.txt"
  threads: 6
  resources:
    runtime="2d",
    mem="2G"
  shell: ".echo test > {output}"
```

**Additional context**

This may be partially related to #2695
I am using the lsf executor plugin, and I am trying to troubleshoot if this is partially due to something I did in the plugin (`mem_mib` is not shown when not using the cluster and `mem_mb` is what `mem_mib` should be if `2G` is gigabytes), but I don't remember doing anything that could change the printed resources.

BEFH:
This fixes #2774.

1. When inferring memory or disk from a string, *_mb now divides the bytes by 1e6 (SI) instead of 2^20 (binary) and takes the ceiling instead of rounding.
2. The inference of *_mib now properly multiplies by 1.048576 instead of 0.95367431640625 (this is also equivalent to dividing by the original value, which used the wrong operation)
3. After inferring resources from string-based resources, Snakemake now updates `resources` so that the inference of mem_mib and disk_mib will use the newly inferred value.
4. mem_mib and disk_mib are now inferred regardless of whether resources are inferred from human readable values.

All space requests are now correct, but we may want to consider using `humanfriendly.parse_size(value, binary=True)` instead of `humanfriendly.parse_size(value)` for `mem`, since SI memory units are rarely used, but this would be an enhancement rather than a pure bugfix.

Many thanks to @hermidalc for their help.

### QC

## Summary by CodeRabbit

- **New Features**
  - Enhanced precision for converting megabytes to mebibytes, improving accuracy in data handling.
  - Updated resource inference to ensure memory and disk values are rounded up, increasing the minimum allocations.

- **Bug Fixes**
  - Streamlined resource inference logic for clarity and maintainability, reducing potential errors.

- **Tests**
  - Added new Snakemake rules for testing resource allocation across different capacities, ensuring accurate output in multiple units.


johanneskoester (approved):
Thanks a lot!

PR #3057, opened 2024-08-29
CLI args for `script` 

johanez:
Allow to pass command-line arguments to `script`, e.g. `scripts/script.py argument1 {params.argument2}`

##### Why?
`shell` rules can be used for passing CLI arguments to scripts, however when importing a module from a remote code repository, paths to the shell scripts are not resolved. Using `script`, the path is resolved. For some use cases, CLI arguments are preferable to the snakemake object (keep option to run script standalone) or reading in a config file from the script (e.g. foreach rules).

See related discussion [on stackoverflow](https://stackoverflow.com/questions/66818700/acess-scripts-in-external-snakemake-directory-without-defining-absolute-path)

##### How?
The `script` path is split at the first space, the second part is added as-is to the script execution command.

##### Possible issues
- spaces in script filenames won't work, even if quoted.  
- ?

### QC

Docs:  _The correct version for this update should be edited_ 

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to pass command-line arguments to scripts within the Snakemake framework, enhancing script flexibility and usability.
	- Added new rules to handle command-line arguments in both R and Python scripts, expanding workflow capabilities.
	- Implemented a command-line interface in a new R script for parsing and handling user inputs.

- **Bug Fixes**
	- Improved handling of script paths with spaces to ensure proper execution of command-line arguments.

- **Documentation**
	- Updated documentation to reflect new command-line argument features and usage examples.


johanez:
Closing this, as the same functionality can be reached using `shell` in combination with a parameter for the script with `wokspace.path(path/to/script)` see [docs](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#accessing-auxiliary-source-files)

PR #3060, opened 2024-08-30
fix: return empty set if rate limiter at max

fgvieira:
When running locally with a rate limiter, if rate limiter reaches the max number of jobs `job_selector` function returns `None` and snakemake fails on line:

https://github.com/snakemake/snakemake/blob/f9554786e9d7f81c0195e0d9ab3c22b7fe611799/snakemake/scheduler.py#L278-L281

with error:
```
TypeError: object of type 'NoneType' has no len()
```

### QC

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved job selection logic to ensure an empty set is returned when no free jobs are available, enhancing system reliability and preventing unintended behavior.


johanneskoester (approved):
Thanks, good catch!

PR #3072, opened 2024-09-05
fix: avoid downloading if assets are provided

veprbl:
The `Assets.deploy()` is invoked unconditionally by the setup.py, which forces download, even if files already present (e.g. in the source tarball). Those downloads are unnecessary and fail in sandboxed builds.

### QC

## Summary by CodeRabbit

- **New Features**
	- Enhanced the deployment process to prevent accidental overwrites of existing files by introducing a conditional check before writing asset content.

- **Bug Fixes**
	- Improved safety and integrity of the deployment process through updated logic in the asset writing method.


johanneskoester:
Yes, I had the same thought in #3076. Sorry for not seeing this before!

PR #3098, opened 2024-09-21
chore: For assets bundled at a particular version, reference the version’s license

musicinmybrain:
Referencing the license file from the release tag corresponding to the bundled release of each asset makes it less likely that the checksum will eventually start to match due to later upstream license changes – perhaps as simple as a new copyright date – and makes errors in the included licenses less likely.

Since the copyright statement in the react license changed from "Copyright (c) Facebook, Inc. and its affiliates." to "Copyright (c) Meta Platforms, Inc. and affiliates." since the bundled version of react was released – and this commit uses the older version that corresponds to the bundled library – the checksum of the react license file changes.

Similarly, in the vega license, the copyright date range changed from 2015-2021 to 2015-2023 after the bundled version was released, so using the correct license file for the bundled version changes the checksum.

### QC

* [ ] The PR contains a test case for the changes or the changes are already covered by an existing test case. **N/A – this change is not really testable – except that building the package confirms the checksums match.**
* [x] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake). **N/A – no documentation update is believed to be necessary.**

## Summary by CodeRabbit

- **New Features**
	- Updated asset URLs to point to specific tagged versions for improved stability and reliability.

- **Bug Fixes**
	- Corrected SHA256 checksums for various assets to ensure integrity and security.

### DIFF
--- a/snakemake/assets/__init__.py
+++ b/snakemake/assets/__init__.py
@@ -54,7 +54,7 @@ class Assets:
             sha256="a9d66f1d526df02e29dce73436d34e56e8632f46c275bbdffc70569e882f9f17",
         ),
         "tailwindcss/LICENSE": Asset(
-            url="https://raw.githubusercontent.com/tailwindlabs/tailwindcss/master/LICENSE",
+            url="https://raw.githubusercontent.com/tailwindlabs/tailwindcss/refs/tags/v3.0.23/LICENSE",
             sha256="60e0b68c0f35c078eef3a5d29419d0b03ff84ec1df9c3f9d6e39a519a5ae7985",
         ),
         "tailwindcss/tailwind.css": Asset(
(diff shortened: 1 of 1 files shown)


musicinmybrain:
I’ve been studying this new asset bundling scheme because I want to update Fedora’s `snakemake` package past 8.18.2, but I have to make sure I handle all the licensing details for the assets correctly.

I plan to open a follow-up PR that adds a large number of license files for things that I can find bundled in the various JavaScript assets – either because they are NPM dependencies that get included in the final bundle, as happens in `vega-embed`, or because bits of code are copied or “derived from” with attribution, as in `vega.js`. These are all under various licenses like `MIT`, `BSD-2-Clause`, `Apache-2.0`, and `ISC`, almost all of which *do* require retaining the copyright and permission notices in copies and derived works, yet the full license text is generally not present in the JavaScript files in question. These messing texts are something I need to correct – downstream if necessary, upstream if possible – before shipping the asset bundle in Fedora.

This is tedious work, and it will greatly increase the number of entries in the `Assets.spec` dictionary and the number of directories in `snakemake/assets/data`, but making a reasonable effort to ensure all required license texts are present is necessary due diligence before shipping these assets in Fedora. (In fact, it’s only very recently that shipping pre-compiled/pre-minified JavaScript in Fedora when it’s too hard to rebuild it from original sources, and a similar relaxation of the rules for CSS is still pending a vote but seems to be headed for approval.) I’ve already started this follow-up work, but please let me know if you have any general feedback or suggestions on the plan.

musicinmybrain:
The two licenses that are still referenced at `main` or `master` are:

- `snakemake` itself, because accessing the current version number here without importing `snakemake` would be painful, and you will know if you are planning to change your own license text
- and `pygments`, because Pygments CSS included in HTML reports is provided by the `pygments` Python dependency at runtime, and it’s not possible to know *a priori* exactly which version that will be.

Issue #3044, opened 2024-08-24
snakemake >8.0.0 does not respect custom logging config


samimia-swks:
With Snakemake 7.32.4, I could have a SnakeFile like this where I reconfigured the log format of snakemake to be consistent with the log format used by the rest of my python code :

```python

import logging.config

log_dict_config = {
        "version": 1,
        "disable_existing_loggers": False,
        'formatters': {
            'standard': {
                'format': '%(asctime)s [%(levelname)s] %(name)s: %(message)s'
            },
        },
        "handlers": {
            "console": {
                'level': 'INFO',
                'formatter': 'standard',
                'class': 'logging.StreamHandler',
                'stream': 'ext://sys.stdout',  # Default is stderr
            },
        },
        "loggers": {
            "snakemake": {
                "level": "DEBUG",
                "handlers": ["console"],
                "propagate": False,
            },
        },
    }

logging.config.dictConfig(log_dict_config)

```

This worked as expected: 
```
$ snakemake --cores 1
2024-08-23 19:43:38,047 [WARNING] snakemake.logging: Building DAG of jobs...
2024-08-23 19:43:38,049 [WARNING] snakemake.logging: Nothing to be done (all requested files are present and up to date).
2024-08-23 19:43:38,049 [WARNING] snakemake.logging: Complete log: .snakemake/log/2024-08-23T194337.961575.snakemake.log
```

With Snakemake 8.0.0 and higher, the custom formatted logs still print, but snakemake itself seems to be printing its own lines in yellow as well, so every log message is repeated: 
```
$ snakemake --cores 1
Building DAG of jobs...
2024-08-23 19:47:35,891 [WARNING] snakemake.logging: Building DAG of jobs...
Retrieving input from storage.
2024-08-23 19:47:35,893 [WARNING] snakemake.logging: Retrieving input from storage.
Nothing to be done (all requested files are present and up to date).
2024-08-23 19:47:35,893 [WARNING] snakemake.logging: Nothing to be done (all requested files are present and up to date).
```

How do I get the previous behavior back?

fgvieira:
Seems to be related to #2809

Issue #2974, opened 2024-07-17
Custom logging messages no longer appear in log file


lczech:
**Snakemake version**
8.15.2

**Describe the bug**
In previous versions of Snakemake (I was using 6.0.5.), I could log custom messages using `logger.info()`, and they would show up in both the terminal output, and in the log file in `.snakemake/log`. Now, testing with Snakemake 8.15.2, they only get printed on the terminal, but no longer in the log file.

**Minimal example**
```
logger.info("Running my workflow")

rule all:
    output:
        touch("result.txt")
```

Expected behavior: the line "Running my workflow" is printed to terminal, and to the log file.
Snakemake 6.0.5: works as expected
Snakemake 8.15.2: only output on the terminal, not in the log file in `.snakemake/log`.

**Additional context**
I tried to see if there were any changes in the initialization of the Logger class, but could not figure out which part in `snakemake/logging.py` was changed that resulted in this behaviour.

I also tried a simple `print()` statement, but of course that does not end of in the log file either.

This is rather unfortunate, because I was using this to log details about the runtime environment of each call of the workflow, which _greatly_ helped debugging my user's problems. Now I cannot ask them to simply send me the log file any more, but will have to ask them to copy that information manually. Please fix this, or let me know how I can change the logger to work as expected again.

lczech:
Okay, I stared at the code for a bit longer. Adding

```
logger.setup_logfile()
```

as the first line to the above minimal example creates (unsurprisingly) a second log file, because the logging is being set up twice. However, that additional file _does_ contain the extra info. So I guess the order of initialization of the logger compared to the parsing of the snakefile has changed in between the Snakemake versions. 

Please revert back to the old order, where logging worked as expected!

lczech:
For anyone who is also looking for a temporary workaround: I'm now just creating my own custom log file on top of the normal snakemake log file. Not ideal, but better than not having proper logs.

By adding it to to the internal Python logger of the snakemake logger, it also receives all messages.

```python
import logging as pylogging

extra_logfile_handler = pylogging.FileHandler("my.log")
logger.logger.addHandler(extra_logfile_handler)
```

Works for now. The first `logger` there is the Snakemake logger instance, and the second `logger` is its internal Python logger instance. Luckily, no private members in Python ;-)

ElizabethWallace97:
I am currently getting the error 
`module 'snakemake' has no attribute 'logger'`

It seems like there is some serious error with snakemake.logger at present?

[...]

Issue #2882, opened 2024-05-17
--quiet settings not intuitive


pdimens:
**Snakemake version**
8.11.3

**Describe the bug**

There is inconsistent and non-intuitive behavior using the (or omitting) the `--quiet` option. Here are examples of the 4 behaviors using a simple snakefile:
```
rule all:
    input: collect("file.{val}.txt", val = [1,2])
    message: "Checking file outputs"

rule createfile:
    output: touch("file.{val}.txt")
    message: "creating {wildcards.val}"
```

When called without `--quiet`, it prints everything, which is expected:
```
$ snakemake --cores 1
Assuming unrestricted shared filesystem usage.
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job           count
----------  -------
all               1
createfile        2
total             3

Select jobs to execute...
Execute 1 jobs...

[Fri May 17 14:21:38 2024]
Job 2: creating 2
Reason: Missing output files: file.2.txt

Touching output file file.2.txt.
[Fri May 17 14:21:38 2024]
Finished job 2.
1 of 3 steps (33%) done
Select jobs to execute...
Execute 1 jobs...

[Fri May 17 14:21:38 2024]
Job 1: creating 1
Reason: Missing output files: file.1.txt

Touching output file file.1.txt.
[Fri May 17 14:21:38 2024]
Finished job 1.
2 of 3 steps (67%) done
Select jobs to execute...
Execute 1 jobs...

[Fri May 17 14:21:38 2024]
Job 0: Checking file outputs
Reason: Rules with a run or shell declaration but no output are always executed.

[Fri May 17 14:21:38 2024]
Finished job 0.
3 of 3 steps (100%) done
Complete log: .snakemake/log/2024-05-17T142138.066438.snakemake.log
```

When using `--quiet` or `--quiet all`, it prints nothing (expected):
```
$ snakemake --cores 1 --quiet

```

When using `--quiet progress`, it only prints rule information, **but does not print the final line with the path to the Completed log**, which seems like something that should be present, as it's neither `progress` or `rule` related.
```
$ snakemake --cores 1 --quiet progress

[Fri May 17 14:23:53 2024]
Job 2: creating 2
Reason: Missing output files: file.2.txt


[Fri May 17 14:23:53 2024]
Job 1: creating 1
Reason: Missing output files: file.1.txt


[...]

PR #3107, opened 2024-09-29
feat!: Logging refactor & add LoggerPluginInterface

cademirch:
Followup to #2830.

This PR attempts to make Snakemake's logging more modular to make room for a potential logging plugin interface, or other logging/monitoring additions, as well as make logging in Snakemake more standardized.

Currently, most of logging logic is deep within `logging.Logger`, mainly in `text_handler()` with direct calls to `logger.info/warn/etc`. In this PR, this logic has been refactored to more comply with the standard `logging` practices. To do so, three new classes were added:
1. `ColorizingTextHandler` manages adding color to log records and emitting to stream.
2. `DefaultFormatter` formats log records to be printed. Much of the logic from `text_handler()` is here now.
3. `DefaultFilter` filters log records from being emitted based on level and quietness settings.

Together, these classes should output identical logs to the current logging implementation. I've tested this briefly, but further testing would be good.

With these changes, it is much easier to add new `logHandlers` via `logger.addHandler`. New handlers can decide how to format and filter logs themselves, an important characteristic missing in the current implementation that motivated this refactoring (https://github.com/snakemake/snakemake/pull/2830#issuecomment-2092731491). 

This would be a breaking change for `--log-handler-script`, as user provided log handlers will need to subclass `logging.logHandler`. Furthermore, logic for adding the user handler (and formatter/filter) will need to be added.

As for the logging plug interface (https://github.com/snakemake/snakemake/pull/2830#issuecomment-2091510808), I'm not sure that is the route to take, but would appreciate thoughts on this. First, this functionality is already largely provided by `--log-handler-script` and adding a logging plugin would be redundant. Second, I think that development of such plugins would likely be minimal, and coordinated effort in this repo to maintain a good set of logging options would be preferable. For example:
- A standard HTTP log handler send emit log records to a user supplied address on which a 3rd party monitoring app is listening. 
- A handler/formatter that uses [rich](https://github.com/Textualize/rich) to make a prettier/interactive console on which logs are written to. 

I think both of these additions would benefit being standardized to Snakemake, as well as default options to users. 

Additionally I think it would be worthwhile to unify logging in Snakemake. Currently, the use of the wrapper `logging.Logger` class and `setup_logger()` function being called in multiple places makes it difficult to work with the logger and is likely the source of some issues (#3044, #2974, #2882, #2797, #2727). Unifying the logger and its configuration as recommended by the [logging docs](https://docs.python.org/3/howto/logging-cookbook.html#a-cli-application-starter-template) would make it easier to work the the logger. Another option would be to use something like [loguru](https://loguru.readthedocs.io/en/stable/index.html). Whichever option, I think simplifying and unifying the logger should be the next step of this PR. 

One main question I have is how much we would like to preserve the current stderr output formatting. Most of logging logic is in formatting rule info and such. Moving forward we can now try different formatting. Personally, I find the current detailed rule output helpful for sanity checks when developing and doing a dry run. However, when actually running the workflow (especially a large one) the terminal is flooded and errors can quickly be lost which requires searching the log file. Of course this can be alleviated with --quiet. 

Another question is regarding the two calls of setup_logger() brought up in #2797. I still don't understand why these are necessary and understanding this would be helpful in unifying this.

I'm happy to keep pushing this forward but would appreciate your thoughts before continuing @johanneskoester! I

## Summary by CodeRabbit

- **New Features**
	- Introduced a new CLI option (--logger) for configuring custom loggers.
	- Enhanced structured logging delivers clearer runtime and resource insights, including rule graph logging.
	- Added a new method for logging rule graph information.
	- Added a new enumeration value `REASON` to the `Quietness` class.

- **Refactor**
	- Overhauled the logging architecture across execution, scheduling, and workflow components for improved traceability and error handling.

- **Chores**
	- Updated dependencies and setup configurations to include new logger plugins.
	- Adjusted test configurations to capture all output, ensuring more comprehensive test diagnostics.

- **Bug Fixes**
	- Resolved minor command and formatting inconsistencies for smoother operations.


snakemake/logging.py
```diff
@@ -346,42 +781,66 @@
...
+            "dag_debug": (
+                _logging.WARNING if self.debug_dag else _logging.DEBUG
+            ),  # bump debug_dag to warning if we want it
+            "resources_info": _logging.WARNING,
+            "host": _logging.INFO,
+            "job_stats": _logging.WARNING,
+        }.get(
+            custom_level, _logging.INFO
+        )  # Default to INFO if not recognized
+
```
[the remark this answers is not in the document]
cademirch: This was intentional to preserve the existing log output coloring:

https://github.com/snakemake/snakemake/blob/19a2ab76b475c5d7728624618bec75a7f1411e8b/snakemake/logging.py#L624-L625

johanneskoester:
> As for the logging plug interface (https://github.com/snakemake/snakemake/pull/2830#issuecomment-2091510808), I'm not sure that is the route to take, but would appreciate thoughts on this. First, this functionality is already largely provided by --log-handler-script and adding a logging plugin would be redundant. Second, I think that development of such plugins would likely be minimal, and coordinated effort in this repo to maintain a good set of logging options would be preferable. For example:
>    A standard HTTP log handler send emit log records to a user supplied address on which a 3rd party monitoring app is listening.
>    A handler/formatter that uses [rich](https://github.com/Textualize/rich) to make a prettier/interactive console on which logs are written to.
> I think both of these additions would benefit being standardized to Snakemake, as well as default options to users.

I think it is still fine to develop them as plugins. They can be default dependencies or internal plugins (like with the dryrun, local, and touch executor plugins).

johanneskoester:
> One main question I have is how much we would like to preserve the current stderr output formatting. Most of logging logic is in formatting rule info and such. Moving forward we can now try different formatting. Personally, I find the current detailed rule output helpful for sanity checks when developing and doing a dry run. However, when actually running the workflow (especially a large one) the terminal is flooded and errors can quickly be lost which requires searching the log file. Of course this can be alleviated with --quiet.

Sure, the output can be improved in any way. I have no strong feelings about how things are currently logged.

johanneskoester:
> Another question is regarding the two calls of setup_logger() brought up in https://github.com/snakemake/snakemake/issues/2797. I still don't understand why these are necessary and understanding this would be helpful in unifying this.

Answered here: https://github.com/snakemake/snakemake/issues/2797#issuecomment-2397668227
Further, one general problem I have with the Snakemake logger is that it uses a global object which kind of has a state. We can keep it like that but one could keep in mind that maybe at some point this should be somehow different and the logger instance should rather be stored in the API call or so. Would be the start of making Snakemake thread safe.

snakemake/logging.py
```diff
@@ -305,12 +243,504 @@ def log_handler(self, msg):
...
+        elif level == "rule_info":
+            return self.handle_rule_info(msg)
+        elif level == "d3dag":
+            return self.handle_d3dag(msg)
+        elif level == "dag_debug":
+            return self.handle_dag_debug(msg)
+        elif level == "run_info":
+            return self.handle_run_info(msg)
+        else:
+            return msg["msg"]
```
johanneskoester, line 289: I guess this is something for Python's new match statement.

snakemake/logging.py
```diff
@@ -305,12 +243,504 @@ def log_handler(self, msg):
...
+        while fmt(fraction) == "100%" or fmt(fraction) == "0%":
+            precision += 1
+        return fmt(fraction)
+
+
+def show_logs(logs):
+    """Helper method to show logs."""
+    for f in logs:
+        try:
+            content = open(f, "r").read()
```
johanneskoester, line 523: I always thought the idiomatic version of this uses a with statement, even if one reads the entire file into memory.

johanneskoester (changes requested):
See my answers on the PR message content above. Here comes the code review. I like the structured approach to splitting up the code a lot!

snakemake/logging.py
cademirch: Good idea, agreed.

cademirch:
Thanks for the review and answering questions @johanneskoester. This helps with the direction to take this. Currently working on consolidating the custom log levels. Will try to have more updates by end of the week.

Re logger state: I agree this is a bit tricky, I will try to keep it in mind.

cademirch:
> > Another question is regarding the two calls of setup_logger() brought up in #2797. I still don't understand why these are necessary and understanding this would be helpful in unifying this.
> 
> Answered here: [#2797 (comment)](https://github.com/snakemake/snakemake/issues/2797#issuecomment-2397668227) Further, one general problem I have with the Snakemake logger is that it uses a global object which kind of has a state. We can keep it like that but one could keep in mind that maybe at some point this should be somehow different and the logger instance should rather be stored in the API call or so. Would be the start of making Snakemake thread safe.

I've been thinking about this lately as I've been digging into the execmodes and its effect on logging. I'm still a bit confused with the API and have been looking into the LocalExecutor to try and understand, but I can't seem to understand if/where the API is constructed by the executor? Do you have any resources/diagrams of the how this works?

If I understand what you're saying here is that a logger instance could be passed to the constructor of the API instance?  

I'm also not sure how much this matters for local execution to be honest. Currently things are setup such that loggers in subprocess don't log anything (which makes sense to me):

https://github.com/snakemake/snakemake/blob/1874fa86b1ebf52ece6646f32d4a3ed025499183/snakemake/executors/local.py#L84-L85


I was thinking of making this more explicit by just attaching [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to the logger  instead of the StreamHandler when in subproccess mode.


Overall, my understanding is that we only care about logs that come from the main Snakemake process (and by extension any exectuor plugins - which should be in the same process?), so having a global logger makes sense in this way. However its possible I'm missing what you're saying here.

In a remote context, it makes sense that the logger writes to stderr/file, but that doesn't contend with the main Snakemake process so this is okay. In this scenario, we only "care" about the logs from the main Snakemake process, because remote process is totally independent.

cademirch:
@johanneskoester some more work here ready for your review.

from the commit message:

> The logger wrapper class (seemingly) largely existed to enable the custom log methods (job_info, progress, etc). However this was quite confusing, especially the usage of the logRecord msg as a dictionary to pass along conextual information.
> 
>  In this commit, the logger class is no longer used. Direct calls to log.info/warn/debug use the `extra` argument to pass a dict of contextual information (jobid, etc). Then the default log formatter/filter can use this info to format and filter logs respectively. This approach provides a much more explicit route of log records and allows potential log plugins to handle logs however they see fit.
> 
> I've put some small functions to setup the logger and such, however this is largely to get things running and I will put more effort into shoring this up if we think this is a good approach.

cademirch:
@johanneskoester, this is ready for review! 
## Removal of Logger wrapper class
(From above) In this commit, the logger class is no longer used. Direct calls to log.info/warn/debug use the extra argument to pass a dict of contextual information (jobid, etc). Then the default log formatter/filter can use this info to format and filter logs respectively. This approach provides a much more explicit route of log records and allows potential log plugins to handle logs however they see fit.
## Logger Plugins
I've added support for logger plugins. Logger plugins are relatively simple, all they must do instantiate a class that subclasses `logging.handler`. 
 - [Interface repo](https://github.com/cademirch/snakemake-interface-logger-plugins)
 - [Example plugin repo (uses rich)](https://github.com/cademirch/snakemake-logger-plugin-rich)
### Rich Logger Demo
Here is a gif of the rich logger in action. (Still very much a WIP)
t-rec_1



## Logger Manager
I created a class to hold most of the logger configuration/setup/destruction logic, `snakemake.logging.LoggerManager`.


## Todo/To discuss (loosely in order of importance)
### Deprecating `--log-handler-script`
This currently wont work with the given changes. I could probably make this work if the `log-handler-script` points to a file that defines a logging.Handler (or subclass), but at that point you almost have a plugin.
### More testing
Will likely need help here. I don't think I've broken anything (except for `--log-handler-script`, `wmslogger`, and `slacklogger`). There may be issues with executor plugins, if they were calling logger.info without a message. 
### Unintuitive quietness and verbosity
Pointed out in #2882 , and I agree. There are 3 ways to control logging verbosity: `quiet`, `debug-dag`, and `verbose`. I think we should try to unify these and make it more clear/more intuitive for a user and developer to control log verbosity. Will start thinking about this now
### Empty log messages
With switching to `logging.Logger` all log calls need a message. Some old log calls didn't have a message, so I put empty strings but we should probably do better.
 
There is probably more, but this is all I can think of now! Let me know if you have any questions.

snakemake/logging.py
```diff
@@ -453,324 +612,316 @@ def is_quiet_about(self, msg_type: str):
...
-    logger.debug_dag = debug_dag
-    logger.mode = mode
-    logger.dryrun = dryrun
-    logger.show_failed_logs = show_failed_logs
+        # Update the logger settings based on the current mode
+        if self.mode == ExecMode.SUBPROCESS:
+            self.add_stream_handler(logging.NullHandler())
+        else:
+            if handler:
+                for h in handler:
```
cademirch, line 897: need to add method for adding file/generic handler plugins

johanneskoester:
Fantastic! My thoughts:
## Deprecating --log-handler-script
Deprecating log-handler-script should be fine. Can be a breaking change then (we will get a version bump to Snakemake 9, but the change is small and should affect only very few people).
## Empty log messages
let's get rid of those empty messages entirely while doing the transition. I don't think there are any in executors, but should be easy to find them.

johanneskoester:
## Unintuitive quietness and verbosity
Yes, any improvements are  welcome.

cademirch:
@johanneskoester Sounds good. One more thing, regarding built in logger plugins, how should that be handled? I was thinking of adding a HTTP and JSON logger as builtin and would prefer to not have those as seperate repos if possible. Could just put them in a loggers subdir like executors? ie: `snakemake/loggers/http.py` and `snakemake/loggers/json.py`

snakemake/logging.py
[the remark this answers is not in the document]
st--: I agree with the coderabbit bot - Snakemake should not warn on infos, as already pointed out in https://github.com/snakemake/snakemake/issues/2727 - upgrading all infos to warnings makes it impossible to then later filter out the infos without losing all actual warnings. If info level messages should be coloured differently, that ought to be fixed elsewhere, not by conflating it with warnings.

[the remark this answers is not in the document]
st--: (Belatedly realized this is on outdated changes, so probably the whole conversation can just get resolved. Sorry for the noise!)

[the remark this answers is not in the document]
cademirch: I agree, however this mapping is now gone and we're just using direct calls to info, warning etc. Happy to accept changes to specific logging calls that you think should go to different levels

cademirch:
Hi @johanneskoester, this is ready for review again! I think this is ready to merge. I've cleaned up a couple things and addressed some of the issues linked above. Namely, I've added a queue handler in front of all logging handlers so that no handlers will block the main snakemake process. I've also added a check to see if the the logger has already been setup to prevent double logging. Finally, added some more logging info, such as printing job errors after the workflow ends (#3214) and more info when a job finishes (#3178).

I still need to work on the rich logging plugin, but that can happen separate from this PR.  I also have plans for a HTTP logger, as well as a browser monitoring webapp (like panoptes) - but those will likely come sometime early next year. I will prioritize the rich logger first. 

Looking forward to your review!

cademirch:
bump @johanneskoester

snakemake/cli.py
```diff
@@ -1851,41 +1846,16 @@ def parse_quietness(quietness) -> Set[Quietness]:
...
-        log_handler.append(wms_logger.log_handler)
+    for logger_arg in args.logger:
+        plugin = LoggerPluginRegistry().get_plugin(plugin_name=logger_arg)
+        plugin_settings = plugin.get_settings(args)
+        # logger plugin here is the subclass of LoggerPluginBase, we get handler from it in configure_logger.
+        logger_plugin = plugin.logger_plugin(plugin_settings)
+        log_handlers.append(logger_plugin)
 
-    return log_handler
+    return log_handlers
```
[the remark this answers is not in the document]
johanneskoester: No, such error messages are already created inside the plugin framework

snakemake/logging.py
```diff
@@ -22,428 +26,491 @@ def get_default_exec_mode():
...
-                # ignore any exceptions in these cases as any relevant messages have been printed before
-                pass
-            except Exception as e:
-                self.handleError(record)
+def show_logs(logs):
+    """Helper method to show logs."""
+    for f in logs:
+        try:
+            content = open(f, "r").read()
+        except FileNotFoundError:
```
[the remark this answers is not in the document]
johanneskoester: This is a good suggestion.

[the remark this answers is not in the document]
johanneskoester: Please provide a commitable comment.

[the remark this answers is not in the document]
johanneskoester: @coderabbitai please provide a commitable suggestion.

cademirch:
> > @johanneskoester Sounds good. One more thing, regarding built in logger plugins, how should that be handled? I was thinking of adding a HTTP and JSON logger as builtin and would prefer to not have those as seperate repos if possible. Could just put them in a loggers subdir like executors? ie: `snakemake/loggers/http.py` and `snakemake/loggers/json.py`
> 
> 
> 
> Mhm, I think that having them builtin creates more confusion than having them as real plugins. Will also make it easier to document them in the same way as other loggers via the plugin catalog. The reason for having dryrun, local and touch as builtin executors is rather that they are fundamentally different and more tightly integrated into Snakemake.

Agreed

cademirch:
@johanneskoester let me know what you would like to see before merging this. I had put this on the backburner last month, but will have time to work on this now.

snakemake/cli.py
```diff
@@ -1513,21 +1515,13 @@ def get_argument_parser(profiles=None):
...
-        default=None,
-        choices=["none", "slack", "wms"],
-        help="Set a specific messaging service for logging output. "
-        "Snakemake will notify the service on errors and completed execution. "
-        "Currently slack and workflow management system (wms) are supported.",
+        "--logger",
+        nargs="+",
+        default=[],
+        choices=LoggerPluginRegistry().plugins.keys(),
+        help="Specify a custom logger, available via a logger plugin: snakemake_executor_<name>. Multiple can be specified.",
```
johanneskoester, line 1524:
```suggestion
        help="Specify one or more custom loggers, available via logger plugins.",
```

snakemake/cli.py
```diff
@@ -1849,41 +1844,16 @@ def parse_quietness(quietness) -> Set[Quietness]:
...
-        metadata = generate_parser_metadata(parser, args)
-        wms_logger = logging.WMSLogger(
-            args.wms_monitor, args.wms_monitor_arg, metadata=metadata
-        )
-        log_handler.append(wms_logger.log_handler)
+    for logger_arg in args.logger:
+        plugin = LoggerPluginRegistry().get_plugin(plugin_name=logger_arg)
+        plugin_settings = plugin.get_settings(args)
+        # logger plugin here is the subclass of LoggerPluginBase, we get handler from it in configure_logger.
+        logger_plugin = plugin.logger_plugin(plugin_settings)
```
johanneskoester, line 1853: It seems confusing that the logger plugin returns a LoggerPluginBase subclass. Should that be called LogHandlerBase, and the method `plugin.log_handler(plugin_settings)`?

snakemake/cli.py
```diff
@@ -1849,41 +1844,16 @@ def parse_quietness(quietness) -> Set[Quietness]:
 
 
 def setup_log_handlers(args, parser):
```
johanneskoester, line 1846: Since we anyway produce a "hidden" breaking change in the API because the objects passed to snakemake.settings.types.OutputSettings.log_handlers are now coming from a different base class, we should probably also change the type there overall and expect a list of strings with logger plugin names in that settings class, while moving the collection of the corresponding plugins into the API classes, analogously to storage and exectuor plugins.

snakemake/logging.py
```diff
@@ -22,428 +26,491 @@ def get_default_exec_mode():
...
-            except (KeyboardInterrupt, SystemExit):
-                # ignore any exceptions in these cases as any relevant messages have been printed before
-                pass
-            except Exception as e:
-                self.handleError(record)
+def show_logs(logs):
+    """Helper method to show logs."""
+    for f in logs:
+        try:
+            content = open(f, "r").read()
```
johanneskoester, line 41:
```suggestion
            with open(f, "r") as log_file:
                content = log_file.read()
```

snakemake/logging.py
```diff
@@ -22,428 +26,491 @@ def get_default_exec_mode():
...
+            yield f"Logfile {f} is not a text file."
+            return
+        lines = content.splitlines()
+        logfile_header = f"Logfile {f}:"
+        if not lines:
+            logfile_header += " empty file"
+            yield logfile_header
+            return
+        yield logfile_header
+        max_len = min(max(max(len(l) for l in lines), len(logfile_header)), 80)
```
johanneskoester, line 55:
```suggestion
        max_len = min(max(max(len(line) for line in lines), len(logfile_header)), 80)
```

johanneskoester (changes requested):
Fantastic and very clean work! A few things below.

snakemake/cli.py
cademirch: I was confused looking at this again, so agreed. Just so we're on the same page you're suggesting changing this in: `src/snakemake_interface_logger_plugins/registry/plugin.py`


```diff
 @dataclass
 class Plugin(PluginBase):
-    logger_plugin: object
+    log_handler: object
     _logger_settings_cls: Optional[Type[LoggerPluginSettingsBase]]
     _name: str
```

And then in `src/snakemake_interface_logger_plugins/base.py`

```diff
-class LoggerPluginBase(ABC):
+class LogHandlerBase(ABC):
     def __init__(
         self,
         settings: Optional[LoggerPluginSettingsBase],
     ) -> None:
         self.settings = settings
         self.__post__init()
```

snakemake/logging.py
```diff
@@ -3,17 +3,21 @@
...
 import os
 import json
 import threading
+from queue import Queue
 from functools import partial
-import inspect
+
 import textwrap
+from typing import Optional
+from snakemake_interface_logger_plugins.base import LoggerPluginBase
```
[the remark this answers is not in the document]
cademirch: This depedency is in the test-env yaml

snakemake/cli.py
cademirch: Not sure I follow here. Is this what you're thinking:
```diff
@dataclass
class OutputSettings(SettingsBase):
    printshellcmds: bool = False
    nocolor: bool = False
    quiet: Optional[AnySet[Quietness]] = None
    debug_dag: bool = False
    verbose: bool = False
    show_failed_logs: bool = False
-   log_handlers: Sequence[object] = tuple()
+   log_handlers: Sequence[str] = tuple()
    keep_logger: bool = False
    stdout: bool = False
    benchmark_extended: bool = False
```

snakemake/cli.py
johanneskoester: Yes. In addition, I would use `log_handler: LogHandlerBase` as type annotation in `src/snakemake_interface_logger_plugins/registry/plugin.py`.

snakemake/cli.py
johanneskoester: Yes, and the strings are just the names. Then, the instantiation can happen inside of the API classes and the users of the API don't not need to instantiate those things themselves.

snakemake/logging.py
```diff
@@ -22,428 +26,493 @@ def get_default_exec_mode():
...
+                # Apply color to the formatted message
+                self.stream.write(self.decorate(record, formatted_message))
+                self.stream.write(getattr(self, "terminator", "\n"))
+                self.flush()
+            except BrokenPipeError as e:
+                raise e
+            except (KeyboardInterrupt, SystemExit):
+                pass  # Ignore exceptions for these cases
+            except Exception as e:
+                self.handleError(record)
```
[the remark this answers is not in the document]
johanneskoester:
```suggestion
            except BrokenPipeError:
                raise
            except (KeyboardInterrupt, SystemExit):
                pass  # Ignore exceptions for these cases
            except Exception as e:
                self.handleError(record)
```

snakemake/logging.py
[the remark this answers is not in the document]
johanneskoester: But it should also be in setup.cfg shouldn't it?

snakemake/cli.py
cademirch: Resolved in https://github.com/cademirch/snakemake-interface-logger-plugins/commit/2b23736a53a569d2cf9fc702d4eed39d283c39bd

snakemake/logging.py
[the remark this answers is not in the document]
cademirch: You're right, thanks!

cademirch:
> **Actionable comments posted: 0**
> 
> 🧹 Nitpick comments (2)
> > snakemake/logging.py (2)> `424-433`: **Simplify filter logic for better readability.**
> > > The filter logic can be simplified without changing functionality.
> > > Apply these simplifications:
> > > ```diff
> > > -        if level in quietness_map:
> > > -            if self.is_quiet_about(quietness_map[level]):
> > > -                return False
> > > +        if level in quietness_map and self.is_quiet_about(quietness_map[level]):
> > > +            return False
> > > 
> > > -        if level == "dag_debug" and not self.debug_dag:
> > > -            return False
> > > -        return True
> > > +        return not (level == "dag_debug" and not self.debug_dag)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 🧰 Tools
> > > 
> > > 🪛 Ruff (0.8.2)
> > > 425-426: Use a single `if` statement instead of nested `if` statements
> > > Combine `if` statements using `and`
> > > (SIM102)
> > > 
> > > 430-433: Return the condition `not (level == "dag_debug" and not self.debug_dag)` directly
> > > Replace with `return not (level == "dag_debug" and not self.debug_dag)`
> > > (SIM103)
> > > 
> > > 
> > > 
> > > `529-534`: **Remove unused exception variable.**
> > > The exception variable `e` in the final except block is not used.
> > > Apply this diff:
> > > ```diff
> > >             except BrokenPipeError:
> > >                 raise
> > >             except (KeyboardInterrupt, SystemExit):
> > >                 pass  # Ignore exceptions for these cases, all errors have been handled before.
> > > -            except Exception as e:
> > > +            except Exception:
> > >                 self.handleError(record)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 🧰 Tools
> > > 
> > > 🪛 Ruff (0.8.2)
> > > 533-533: Local variable `e` is assigned to but never used
> > > Remove assignment to unused variable `e`
> > > (F841)
> 
> 📜 Review details

The first suggestion is less readable imo. I prefer explicit return statements ie `return True` not `return condition`

johanneskoester:
> > **Actionable comments posted: 0**
> > 🧹 Nitpick comments (2)
> > > snakemake/logging.py (2)> `424-433`: **Simplify filter logic for better readability.**
> > > > The filter logic can be simplified without changing functionality.
> > > > Apply these simplifications:
> > > > ```diff
> > > > -        if level in quietness_map:
> > > > -            if self.is_quiet_about(quietness_map[level]):
> > > > -                return False
> > > > +        if level in quietness_map and self.is_quiet_about(quietness_map[level]):
> > > > +            return False
> > > > 
> > > > -        if level == "dag_debug" and not self.debug_dag:
> > > > -            return False
> > > > -        return True
> > > > +        return not (level == "dag_debug" and not self.debug_dag)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > 🧰 Tools
> > > > 🪛 Ruff (0.8.2)
> > > > 425-426: Use a single `if` statement instead of nested `if` statements
> > > > Combine `if` statements using `and`
> > > > (SIM102)
> > > > 430-433: Return the condition `not (level == "dag_debug" and not self.debug_dag)` directly
> > > > Replace with `return not (level == "dag_debug" and not self.debug_dag)`
> > > > (SIM103)
> > > > `529-534`: **Remove unused exception variable.**
> > > > The exception variable `e` in the final except block is not used.
> > > > Apply this diff:
> > > > ```diff
> > > >             except BrokenPipeError:
> > > >                 raise
> > > >             except (KeyboardInterrupt, SystemExit):
> > > >                 pass  # Ignore exceptions for these cases, all errors have been handled before.
> > > > -            except Exception as e:
> > > > +            except Exception:
> > > >                 self.handleError(record)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > 🧰 Tools
> > > > 🪛 Ruff (0.8.2)
> > > > 533-533: Local variable `e` is assigned to but never used
> > > > Remove assignment to unused variable `e`
> > > > (F841)
> > 
> > 
> > 📜 Review details
> 
> The first suggestion is less readable imo. I prefer explicit return statements ie `return True` not `return condition`

ok, fine with that, matter of style.

cademirch:
> > > **Actionable comments posted: 0**
> > > 🧹 Nitpick comments (2)
> > > > snakemake/logging.py (2)> `424-433`: **Simplify filter logic for better readability.**
> > > > > The filter logic can be simplified without changing functionality.
> > > > > Apply these simplifications:
> > > > > ```diff
> > > > > -        if level in quietness_map:
> > > > > -            if self.is_quiet_about(quietness_map[level]):
> > > > > -                return False
> > > > > +        if level in quietness_map and self.is_quiet_about(quietness_map[level]):
> > > > > +            return False
> > > > > 
> > > > > -        if level == "dag_debug" and not self.debug_dag:
> > > > > -            return False
> > > > > -        return True
> > > > > +        return not (level == "dag_debug" and not self.debug_dag)
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > 🧰 Tools
> > > > > 🪛 Ruff (0.8.2)
> > > > > 425-426: Use a single `if` statement instead of nested `if` statements
> > > > > Combine `if` statements using `and`
> > > > > (SIM102)
> > > > > 430-433: Return the condition `not (level == "dag_debug" and not self.debug_dag)` directly
> > > > > Replace with `return not (level == "dag_debug" and not self.debug_dag)`
> > > > > (SIM103)
> > > > > `529-534`: **Remove unused exception variable.**
> > > > > The exception variable `e` in the final except block is not used.
> > > > > Apply this diff:
> > > > > ```diff
> > > > >             except BrokenPipeError:
> > > > >                 raise
> > > > >             except (KeyboardInterrupt, SystemExit):
> > > > >                 pass  # Ignore exceptions for these cases, all errors have been handled before.
> > > > > -            except Exception as e:
> > > > > +            except Exception:
> > > > >                 self.handleError(record)
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > 🧰 Tools
> > > > > 🪛 Ruff (0.8.2)
> > > > > 533-533: Local variable `e` is assigned to but never used
> > > > > Remove assignment to unused variable `e`
> > > > > (F841)
> > > 
> > > 
> > > 📜 Review details
> > 
> > 
> > The first suggestion is less readable imo. I prefer explicit return statements ie `return True` not `return condition`
> 
> ok, fine with that, matter of style.

Great. Lmk what else we should do here. Happy to look into #1902 before or after merging this

cademirch:
I've added some more to this PR. 

1. `--quiet` now accepts `reason` which will suppress reason info in the logs.
2. The `LoggerManager` class has been cleaned up significantly and simplified
3. The logger is only setup exactly once, either in `snakemake.cli` or `snakemake.api`
4. `snakemake.settings.types.OutputSettings` now inherits from `snakemake_interface_logger_plugins.settings.LoggerSettingsInterface` such that LogPlugins have easy access to user defined log variables such as quietness, verbosity, etc.

@johanneskoester I think we are getting closer on this! Please let me know what you think.

cademirch:
Hi @johanneskoester - Unfortunately I can't seem to exactly replicate the environment that the actions run in, which has made troubleshooting a bit of pain, however, I've got some more of the tests passing.I caught some overlooked old logger custom method calls, causing issues with shell() in run block. I also removed the setting of `printshellcmd` from the script preamble, not sure what its purpose was there, happy to come up with a way to refactor instead of removing.


Currently the first test to fail on the actions is https://github.com/cademirch/snakemake/blob/logging-refactor/tests/tests.py#L1096, and based on the error, the issue seems more related to the test itself? I'm not very familiar with the test suite, and this one confuses me! Any thoughts? 


Would appreciate any help and advice for running the test suite.

cademirch:
@johanneskoester all tests pass, would appreciate your review. There is still room for improvement, particularly in where and what we log (handful of empty log messages, could use logging on workflow start, etc.) But I think those can come later.

snakemake/script/__init__.py
```diff
@@ -30,7 +30,7 @@
 from snakemake.common import MIN_PY_VERSION, ON_WINDOWS, get_snakemake_searchpaths
 from snakemake.deployment import singularity
 from snakemake.exceptions import WorkflowError
-from snakemake.logging import logger
+from snakemake.logging import logger_manager, logger
```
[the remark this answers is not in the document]
johanneskoester: @cademirch is the import needed?

snakemake/workflow.py
```diff
@@ -1296,7 +1317,11 @@ def execute(
             if success:
                 if self.dryrun:
                     if len(self.dag):
-                        logger.run_info("\n".join(self.dag.stats()))
+                        stats_msg, stats_dict = self.dag.stats()
+                        logger.warning(
```
johanneskoester, line 1321: logger.info here, correct?

cademirch: Warning was correct to coerce it yellow, but coloring based on the log type will resolve this. I will push some changes soon to address this.

cademirch:
Pushed some more changes here. 

I added an enum `LogEvent` to replace the `level` field  in the extra dict and renamed it to `event`. I chose "event" because it better represents what we're tracking and avoids confusion with the standard log severity levels (INFO, DEBUG, ERROR, etc.) 

LogEvent could (perhaps should?) live in logger-interface repo so that plugins can have access to this without making an interface class, but I will defer to you on this @johanneskoester. 

Using the LogEvent in the DefaultFormatter and Handler I've tried to make the log coloring match how it currently is in main. It's a bit difficult since in main, `logger.info` actually calls `logger.warn`:

https://github.com/snakemake/snakemake/blob/04f89d330dd94baa51f41bc796392f85bccbd231/snakemake/logging.py#L385-L386

https://github.com/snakemake/snakemake/blob/04f89d330dd94baa51f41bc796392f85bccbd231/snakemake/logging.py#L624-L625 

Currently all tests pass, and I've visually inspected log output on a handful of test cases but can't promise the formatting will be 1:1 to main in all scenarios!

johanneskoester:
> LogEvent could (perhaps should?) live in logger-interface repo so that plugins can have access to this without making an interface class, but I will defer to you on this @johanneskoester.
Yes, that makes sense. 

> Currently all tests pass, and I've visually inspected log output on a handful of test cases but can't promise the formatting will be 1:1 to main in all scenarios!
Fine, thanks!

So, I would be fine with merging this once the new enum is in the logger interface.

snakemake/workflow.py
cademirch: resolved w/ LogEvent enum

johanneskoester (approved):

johanneskoester (approved):

PR #3111, opened 2024-10-03
fix:  snakemake deployment update for mamba > 2.0.0

aabiddanda:
### Description 

This PR removes the `--no-default-packages` flag if the dependency engine detected is determined to be `mamba` since this flag is deprecated in `mamba >= 2.0.0`. 

### QC

This change should be covered by the pre-existing test cases that exhibit `conda` dependencies. It also should not require any documentation changes 

## Summary by CodeRabbit

- **New Features**
	- Improved handling of the `--no-default-packages` option when creating conda environments based on the selected frontend and mamba version.
	- Added a new property to retrieve the version of mamba for enhanced environment management.

- **Bug Fixes**
	- Enhanced clarity in logging messages related to environment creation.

### DIFF
--- a/snakemake/deployment/conda.py
+++ b/snakemake/deployment/conda.py
@@ -502,7 +502,7 @@ class Env:
                             "--quiet",
                             "--no-shortcuts" if ON_WINDOWS else "",
                             "--yes",
-                            "--no-default-packages",
+                            "--no-default-packages" if (self.frontend == "mamba") else "",
                             f"--prefix '{env_path}'",
                         ]
                         + packages
@@ -545,7 +545,7 @@ class Env:
                             + [
                                 "create",
                                 "--quiet",
-                                "--no-default-packages",
+                                "--no-default-packages" if (self.frontend == "mamba") else "",
                                 f'--file "{target_env_file}"',
                                 f'--prefix "{env_path}"',
                             ]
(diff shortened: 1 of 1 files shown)


maarten-k:
mamba 1.x does not ignore .condarc  "create_default_packages" option. If specified, in .condarc, this might lead to unwanted results. Does mamba 2.0x ignore this?

jjjermiah:
ran into this error in my tests as well, working solution?:

```
"--no-default-packages" if (
  self.frontend != "mamba" or not self.mamba_version.startswith("2")
) else ""

```

and 

```
    @property
    def mamba_version(self) -> str:
        if not is_mamba_available():
            return ""
        try:
            version = subprocess.check_output(["mamba", "--version"]).decode().strip()
            return version.split()[-1]
        except subprocess.CalledProcessError:
            logger.debug("Could not determine mamba version.")
            return ""
```

aabiddanda:
@jjjermiah thats a good call to version-check `mamba` to preserve the behavior of `<2.0.0`. I wonder if its more forward-facing to actually do the complement logic though (in anticipation of 3.0, etc)? 

```
"--no-default-packages" if (
  self.frontend != "mamba" or self.mamba_version.startswith("1")
```

jjjermiah:
@aabiddanda looks like #3121 and the new release deprecate mamba anyways

aabiddanda:
wow useful to know this @jjjermiah - thanks for the heads up. This seems like a more effective solution than mine at the moment. I'll close this issue/PR since its superceded by #3121

PR #3112, opened 2024-10-04
feat: subsample jobs to speed-up scheduler

fgvieira:
I am running a workflow with ~700k jobs and, at each given time, there are around 230k jobs ready to be run. The initial building of the DAG is quite slow (~2h, but I'll leave that for another PR :smile:), but the main issue is that the scheduler takes a lot of time deciding the next jobs to be submitted.

In my case, all jobs are quite fast and similar in terms of resources, so the cluster is idle most of the time. The greedy scheduler is considerably faster, but still too slow.

The ILP should switch to the greedy after 10s, but it sometimes ignores the timeout (coin-or/cbc#487) and it has been reported being quite slow instantiating large problems (coin-or/pulp#749). In my case, the ILP runs for 60s (the pulp file is 100Mb) before switching to greedy. Apart from that, and specially on slow file systems, the scheduler can still be quite slow checking all temp and input files. 

Here, I propose sampling ready jobs, so that only a subset of jobs (instead of all ready jobs) are evaluated by the scheduler. In my tests, this greatly reduces the scheduler time:

|  | ILP | greedy |
|---|---|---|
| Native |15 - 20 mins |30s - 1 min  |
| Sampling 1000 jobs |  | 1 - 2 s |

### QC

## Summary by CodeRabbit

- **New Features**
	- Introduced a new argument `--scheduler-subsample` to optimize job scheduling by limiting the number of jobs considered for execution.
	- Added a method for inferring resource requirements, enhancing user experience with better error handling.
	- Updated settings to include a new attribute for job subsampling, improving scheduling flexibility.

- **Bug Fixes**
	- Improved error handling and logging for resource evaluation and parsing, providing clearer guidance for users.
	- Enhanced job selection process with a subsampling mechanism to optimize scheduling efficiency.

- **Refactor**
	- Enhanced structure and organization of job scheduling logic for better integration with existing functionality.

### DIFF
--- a/snakemake/scheduler.py
+++ b/snakemake/scheduler.py
@@ -62,6 +62,7 @@ class JobScheduler(JobSchedulerExecutorInterface):
         self.failed = set()
         self.finished_jobs = 0
         self.greediness = self.workflow.scheduling_settings.greediness
+        self.subsample = self.workflow.scheduling_settings.subsample
         self._tofinish = []
         self._toerror = []
         self.handle_job_success = True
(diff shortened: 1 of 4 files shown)


snakemake/scheduler.py
```diff
@@ -263,7 +263,23 @@ def schedule(self):
                         job.reset_params_and_resources()
 
                     logger.debug(f"Resources before job selection: {self.resources}")
-                    logger.debug(f"Ready jobs: {len(needrun)}")
+
+                    # Subsample jobs to be run (to speedup solver)
+                    n_total_needrun = len(needrun)
+                    solver_max_jobs = int(
+                        os.environ.get("SNAKEMAKE_SOLVER_MAX_JOBS", sys.maxsize)
```
johanneskoester, line 270: Let us rather have this as a command line parameter. Then people can also set it via profiles.

johanneskoester (changes requested):

johanneskoester (approved):

Issue #2501, opened 2023-11-03
Bug with storage interface no argument `--storage-http-allow-redirects`


vsoch:
I'm submitting to Google Batch using the latest branches (clones) from the interfaces (storage and executor) and snakemake, and I am hitting this missing argument that is generated by snakemake:

```console
__main__.py: error: unrecognized arguments: --storage-http-allow-redirects
```

PR #3116, opened 2024-10-04
fix: remote pre-command not installing all required storage plugins

jjjermiah:
Related Issues: 
#2501 
https://github.com/snakemake/snakemake-executor-plugin-kubernetes/issues/19 

When the job spawner prepares the `precommand` for remote execution, it only registers the `default storage provider` which fails to acknowledge the presence of a storage provider within the snakefile. 

This PR should now register all storage providers that `workflow` has recognized. 

NOTE: unsure how to I could add a test case for this? 

### QC

## Summary by CodeRabbit

- **New Features**
	- Enhanced package installation logic for storage providers, allowing multiple packages to be installed simultaneously when the auto-deploy feature is enabled. 

- **Bug Fixes**
	- Improved handling of duplicate package names in the installation process.

### DIFF
--- a/snakemake/spawn_jobs.py
+++ b/snakemake/spawn_jobs.py
@@ -201,12 +201,13 @@ class SpawnedJobArgsFactory:
             executor_common_settings.auto_deploy_default_storage_provider
             and self.workflow.storage_settings.default_storage_provider is not None
         ):
-            package_name = StoragePluginRegistry().get_plugin_package_name(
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):

johanneskoester:
Good catch! Indeed, a testcase is tricky, but the code is pretty clear and minimal. I think we can for now live with not having one.

PR #3131, opened 2024-10-11
fix: notebook execution for apptainer

laf070810:
At https://github.com/snakemake/snakemake/blob/afa7bad1a72118dc609f1499560c0847edc36413/snakemake/deployment/singularity.py#L118-L125

snakemake searchpaths are only bound into apptainer when `is_python_script` is `True`. For `notebook` directive, `is_python_script` will be `False` and snakemake searchpaths are not bound into apptainer, causing `ModuleNotFoundError: No module named 'snakemake'` error in the preamble. This PR sets `is_python_script` to `True` for `notebook` directive. 

Also, the name, `is_python_script`, could then be somewhat confusing. Changing it to something like `is_script` or `has_preamble` might be better. 

### QC

## Summary by CodeRabbit

- **New Features**
	- Introduced a new parameter to enhance script execution for Python scripts in Jupyter notebooks.
	- Added functionality to determine whether a notebook should be drafted or executed based on user input.

- **Bug Fixes**
	- Improved validation for local notebook paths to ensure editing is only allowed for local files.

### DIFF
--- a/snakemake/notebook.py
+++ b/snakemake/notebook.py
@@ -94,6 +94,7 @@ class JupyterNotebook(ScriptBase):
                 fname_out=fname_out,
                 fname=fname,
                 additional_envvars={"IPYTHONDIR": tmp},
+                is_python_script=True,
             )
 
             if edit:


laf070810:
Also, it may be better to add some tests for notebook execution in apptainer, to prevent trivial bugs like this one here.

johanneskoester (approved):
Good catch! To me the name also sounds correct for a notebook.

johanneskoester:
> Also, it may be better to add some tests for notebook execution in apptainer, to prevent trivial bugs like this one here.

Indeed! I won't be able to do that now, but if you want to do it in this PR please feel free to do so. If not, it would be ok to merge this without or while adding a TODO.

laf070810:
> > Also, it may be better to add some tests for notebook execution in apptainer, to prevent trivial bugs like this one here.
> 
> Indeed! I won't be able to do that now, but if you want to do it in this PR please feel free to do so. If not, it would be ok to merge this without or while adding a TODO.

Thanks for the quick reply! I would recommend merging this first, as this bug makes `notebook` in apptainer unusable. I can open an issue to track on adding tests and do it later.

PR #3199, opened 2024-11-12
fix: clean env vars in apptainer

laf070810:
The apptainer command used in snakemake now does not have `--cleanenv` (or `-e`) option. The envinronment variables in the host (e.g. PYTHONPATH or CONDA_EXE) will be passed to the container. This may affect reproducibility and even cause some errors. This PR proposes adding the `--cleanenv` option to the apptainer commmand to avoid passing host environment variables. 

### QC

## Summary by CodeRabbit

- **New Features**
	- Enhanced Singularity command execution by adding the `--cleanenv` option, ensuring a clean environment for better handling of environment variables during execution.

### DIFF
--- a/snakemake/deployment/singularity.py
+++ b/snakemake/deployment/singularity.py
@@ -127,7 +127,7 @@ def shellcmd(
     if container_workdir:
         args += f" --pwd {repr(container_workdir)}"
 
-    cmd = "{} singularity {} exec --home {} {} {} {} -c '{}'".format(
+    cmd = "{} singularity {} exec --cleanenv --home {} {} {} {} -c '{}'".format(
         envvars,
         "--quiet --silent" if quiet else "",
         repr(os.getcwd()),


johanneskoester (approved):

ronaldlehnigk:
Dear @laf070810, @johanneskoester 

I am having trouble with this change, because I have the scenario that the container needs environment variables from the host. E.g. I am consuming information like a SLURM_JOB_ID and in order to be able to run multi-node Apptainer jobs via Slurm`s PMI2 interface.

Is it really necessary to hardcode this into the command? After all, it is not the default behavior for Apptainer, so why should it be forced by a downstream tool, like Snakemake in this case. Is it not possible to achieve this via `--apptainer--args`? In case it is, I'd be happy if this could be reverted.

Thanks for the feedback!

ronaldlehnigk:
Dear @laf070810,

have you had a chance to test the proposed approach of just adding `--cleanenv` to `--apptainer--args`. I would like to create a PR reverting the change, but don't want to destroy your workflow of course.

Thanks a lot!

ronaldlehnigk:
https://github.com/snakemake/snakemake/pull/3285

laf070810:
> Dear @laf070810,
> 
> have you had a chance to test the proposed approach of just adding `--cleanenv` to `--apptainer--args`. I would like to create a PR reverting the change, but don't want to destroy your workflow of course.
> 
> Thanks a lot!

I did some tests and it seems OK to use just `--apptainer--args` to add `--cleanenv`. Thank you for pointing this out! I think I was not aware of the `--apptainer--args` option when I proposed this MR. Using `--apptainer--args` should be a better approach.

ronaldlehnigk:
Good to know this works for you. Reverting the change in [PR](https://github.com/snakemake/snakemake/pull/3285) together with a comment in the docs.

PR #3200, opened 2024-11-12
feat: add support for conda-lock unified lockfile

laf070810:
Currently, conda environment pinning supported by snakemake uses only `explicit` pinning file. However, this file does not support pip dependencies. [conda-lock](https://github.com/conda/conda-lock) is a project to lock conda environments. It has a [Unified lockfile](https://conda.github.io/conda-lock/output/) format to lock multi-platform conda and pip dependencies. This PR adds support for using this lock file format. 

The change in this PR favors `.pin.yml` file over `explicit` pinning file `.pin.txt`. When a `.pin.yml` file is present, it will use this file and `conda-lock` command to create new environments. 

Note: this PR is just a proposal now. It doesn't include any tests (although tested on my personal use case) nor documentation. If developers are happy with this feature support, we can continue with adding tests and docs. 

### QC

### DIFF
--- a/snakemake/deployment/conda.py
+++ b/snakemake/deployment/conda.py
@@ -124,11 +124,21 @@ class Env:
 
     @lazy_property
     def pin_file(self):
-        return self._get_aux_file(
-            f".{self.conda.platform}.pin.txt",
+        aux_file = self._get_aux_file(
+            f".pin.yml",
             "Omitting search for pin file "
             "(https://snakemake.readthedocs.io/en/stable/snakefiles/"
             "deployment.html#freezing-environments-to-exactly-pinned-packages).",
+        )
+        return (
+            aux_file
+            if aux_file is not None
+            else self._get_aux_file(
+                f".{self.conda.platform}.pin.txt",
+                "Omitting search for pin file "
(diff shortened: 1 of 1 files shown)


johanneskoester:
To be honest, I am not sure yet. Conda support is in flux, since the appearance of pixi and rattler-py, as well as my work on externalizing software deployment into plugins. It is good to have a proove of concept here, but I am not sure yet whether it should be merged in this form.
A general concern I have about conda-lock (I was aware of it even when implementing the pinning support) is that I so far always had the feeling that plain conda package pinning support is sufficient since one can package any pypi package as a conda package if needed, so that the simple pinning files that we currently use are enough.

laf070810:
One can indeed package any pypi package as a conda package if needed. But if the conda version of a package is not released officially by its developers, additional packaging (and releasing) jobs may need to be done by the user. I think this should not be necessary. 

Nevertheless, new tools like pixi and rattler-py, and externalizing software deployment into plugins are indeed good directions, which I didn't take into account. Will there be different plugins for deployment methods like conda, pixi or rattler-py? so conda-lock support maybe should live in the conda plugin? (or not needed because of better deployment methods)

PR #3216, opened 2024-11-20
docs: refer to Merkle trees instead of "blockchain" in caching.rst

veprbl:
This is a more correct terminology, IMO.

### QC

### DIFF
--- a/docs/executing/caching.rst
+++ b/docs/executing/caching.rst
@@ -9,7 +9,7 @@ For example, in bioinformatics, reference genomes or annotations are downloaded,
 Since such steps are independent of the actual data or measurements that are analyzed, but still computationally or timely expensive to conduct, it has been common practice to externalize their computation and assume the presence of the resulting files before execution of a workflow.
 
 From version 5.8.0 on, Snakemake offers a way to keep those steps inside the actual analysis without requiring from redundant computations.
-By hashing all steps, parameters, software stacks (in terms of conda environments or containers), and raw input required up to a certain step in a `blockchain <https://en.wikipedia.org/wiki/Blockchain>`_, Snakemake is able to recognize **before** the computation whether a certain result is already available in a central cache on the same system.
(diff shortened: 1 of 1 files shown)


docs/executing/caching.rst
```diff
@@ -9,7 +9,7 @@ For example, in bioinformatics, reference genomes or annotations are downloaded,
 Since such steps are independent of the actual data or measurements that are analyzed, but still computationally or timely expensive to conduct, it has been common practice to externalize their computation and assume the presence of the resulting files before execution of a workflow.
 
 From version 5.8.0 on, Snakemake offers a way to keep those steps inside the actual analysis without requiring from redundant computations.
-By hashing all steps, parameters, software stacks (in terms of conda environments or containers), and raw input required up to a certain step in a `blockchain <https://en.wikipedia.org/wiki/Blockchain>`_, Snakemake is able to recognize **before** the computation whether a certain result is already available in a central cache on the same system.
+By hashing all steps, parameters, software stacks (in terms of conda environments or containers), and raw input required up to a certain leaf in a `Merkle tree <https://en.wikipedia.org/wiki/Merkle_tree>`_, Snakemake is able to recognize **before** the computation whether a certain result is already available in a central cache on the same system.
```
johanneskoester, line 12:
```suggestion
By hashing all steps, parameters, software stacks (in terms of conda environments or containers), and raw input required up to a certain step in a `Merkle tree <https://en.wikipedia.org/wiki/Merkle_tree>`_, Snakemake is able to recognize **before** the computation whether a certain result is already available in a central cache on the same system.
```

johanneskoester (changes requested):
I agree! Just a minor thing, since the word step in the change below does not refer to an item in the hash tree but rather to a step in the workflow.

docs/executing/caching.rst
veprbl: Maybe I don't understand what "up to a certain step" means here. I assumed that hash corresponding to the currently considered node in the computational graph/tree has to be used. I called it "leaf" here for a more colloquial term. Even at the last step of hashing you consider not just the "step", but also the "parameters" and "software", that's a whole leaf/node, right?

Issue #2982, opened 2024-07-22
Temp files are not deleted for rule before checkpoint that is rerun after checkpoint with data. 


LeeBergstrand:
**Snakemake version**

Snakemake Version: 8.16.0

Bug first occurs at version 8.6.0
The bug does not occur in 8.5.5

**Describe the bug**

I have a rule that polishes a nanopore genome. This polishing rule leads to a series of other rules, which activate a checkpoint. After checkpoint DAG generation, the polishing rule is again run on a subset of contigs in a different directory. The polishing rule uses wild cards for which directory to run: either the original polishing directory (before the checkpoint) or a later circularization directory (after the checkpoint).

Snakemake only deletes the temp files for the first polishing rule invocation after the checkpoint (in the circularization directory) but doesn't delete the temp files from when the file was initially invoked (in the polishing directory). It also does not delete any temp file generated before the checkpoint (this may be my use case), even if they are not needed after the checkpoint. Running snakemake with flag `--delete-temp-output` deletes these files.

Before 8.6.0 snakemake would delete the temp files for both rules, which is the expected behavior for this use case because these temp files are not needed for later steps, even after DAG regeneration.

This bug appears to occur between [8.5.5](https://github.com/snakemake/snakemake/releases/tag/v8.5.5) and [8.6.0](https://github.com/snakemake/snakemake/releases/tag/v8.6.0). Specifically, the release for 8.6.0 mentions https://github.com/snakemake/snakemake/issues/2732 being fixed by https://github.com/snakemake/snakemake/pull/2737 (fix premature deletion of temp files in combination with checkpoints). I believe the bug was introduced during this pull request.

**Logs**

TODO

**Minimal example**

TODO --> The pipeline is pretty complex, so getting a simplified example may take some time.

Here is the full rule. _R1.clean.sam and _R2.clean.sam as well as temp files in rules leading up to this rule are not deleted. 

```snakemake
rule polish_polypolish:
    input:
        contigs="{sample}/{step}/polypolish/input/{sample}_input.fasta",
        mapping_r1 = "{sample}/{step}/polypolish/{sample}_R1.sam",
        mapping_r2 = "{sample}/{step}/polypolish/{sample}_R2.sam"
    output:
        mapping_clean_r1 = temp("{sample}/{step}/polypolish/{sample}_R1.clean.sam"),
        mapping_clean_r2 = temp("{sample}/{step}/polypolish/{sample}_R2.clean.sam"),
[...]

Issue #2732, opened 2024-03-06
Temp files get deleted before time


fgvieira:
**Snakemake version**

`v8.5.3`

**Describe the bug**

Temp files are being deleted when they will still be needed. In the example below, the problem seems to be that temp file `A.txt` is removed after rule `b1`, but it is still needed by rule `a1`.

**Logs**

```
Assuming unrestricted shared filesystem usage for local execution.
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job      count
-----  -------
a            1
all          1
b1           1
b2           1
total        4

Select jobs to execute...
Execute 1 jobs...

[Wed Mar  6 10:04:25 2024]
localrule a:
    output: A.txt
    jobid: 3
    reason: Forced execution
    wildcards: name=A
    resources: tmpdir=/tmp

[Wed Mar  6 10:04:25 2024]
Finished job 3.
1 of 4 steps (25%) done
Select jobs to execute...
Execute 1 jobs...

[Wed Mar  6 10:04:25 2024]
localrule b1:
    input: A.txt
    output: A.b1.out
    jobid: 2
    reason: Forced execution
    resources: tmpdir=/tmp

[Wed Mar  6 10:04:25 2024]
Finished job 2.
2 of 4 steps (50%) done
Removing temporary output A.txt.
Select jobs to execute...
Execute 1 jobs...

[Wed Mar  6 10:04:25 2024]
localcheckpoint b2:
    input: A.b1.out
    output: A.b2.out
    jobid: 1
    reason: Forced execution
    resources: tmpdir=/tmp
DAG of jobs will be updated after completion.

[Wed Mar  6 10:04:25 2024]
Finished job 1.
3 of 4 steps (75%) done
Select jobs to execute...
Execute 1 jobs...

[Wed Mar  6 10:04:25 2024]
localrule a1:
    input: A.txt
    output: A.a1.out
    jobid: 6
    reason: Forced execution
    wildcards: name=A
    resources: tmpdir=/tmp

Waiting at most 5 seconds for missing files.
[...]
    raise IOError(
OSError: Missing files after 5 seconds. This might be due to filesystem latency. If that is the case, consider to increase the wait time with --latency-wait:
A.txt
```

**Minimal example**

```
rule a:
    output:
        temp("{name}.txt"),
    shell:
        "touch {output}"

rule a1:
    input:
[...]

Issue #609, opened 2020-09-10
checkpoint + temp


nick-youngblut:
**Is your feature request related to a problem? Please describe.**

`checkpoint` is great for dealing with dynamically created files, but it would be great if `temp()` worked on those files so that the files would automatically be removed by snakemake. I know that checkpoint is providing input for downstream rules, while `temp()` is for output, but there doesn't seem to be a good alternative for automatically removing the dynamically generated files, except for creating another rule that deletes all of the files, but the user must then make sure to not delete those file prematurely. 

An example from the docs with `temp()` added:

```
# a target rule to define the desired final output
rule all:
    input:
        "aggregated/a.txt",
        "aggregated/b.txt"


# the checkpoint that shall trigger re-evaluation of the DAG
checkpoint somestep:
    input:
        "samples/{sample}.txt"
    output:
        "somestep/{sample}.txt"
    shell:
        # simulate some output value
        "echo {wildcards.sample} > somestep/{wildcards.sample}.txt"


# intermediate rule
rule intermediate:
    input:
        "somestep/{sample}.txt"
    output:
        "post/{sample}.txt"
    shell:
        "touch {output}"


# alternative intermediate rule
rule alt_intermediate:
    input:
        "somestep/{sample}.txt"
    output:
        "alt/{sample}.txt"
    shell:
        "touch {output}"


# input function for the rule aggregate
def aggregate_input(wildcards):
    # decision based on content of output file
    # Important: use the method open() of the returned file!
    # This way, Snakemake is able to automatically download the file if it is generated in
    # a cloud environment without a shared filesystem.
    with checkpoints.somestep.get(sample=wildcards.sample).output[0].open() as f:
        if f.read().strip() == "a":
            return temp("post/{sample}.txt")
        else:
            return temp("alt/{sample}.txt")

rule aggregate:
    input:
        aggregate_input
    output:
        "aggregated/{sample}.txt"
    shell:
        "touch {output}"
```

As far as I can tell, this method of using `temp()` doesn't work. 

The alternative that would be required to delete the files without temp:

```
# a target rule to define the desired final output
rule all:
    input:
        "aggregated/a.txt",
        "aggregated/b.txt",
        "tmp_deleted.done"

# the checkpoint that shall trigger re-evaluation of the DAG
checkpoint somestep:
    input:
        "samples/{sample}.txt"
    output:
        "somestep/{sample}.txt"
    shell:
        # simulate some output value
[...]

PR #3218, opened 2024-11-22
feat: assume checkpoint safe temp files

mhulsman:
Snakemake can only ensure that temp files will not be needed anymore
after all checkpoint jobs have been executed (#609, #2732, fixed in #2737).
However, waiting for all checkpoint jobs to be completed can lead to
excessive storage use (#2982).

In most cases temp file usage is either unaffected by checkpoint jobs, or
can be explicitly managed by including the temp files as input to
checkpoint jobs or their downstream dependencies. This will ensure they
are kept on disk until they (might) be needed.

To address cases where pipelines using checkpoints become infeasible due
to high storage demands, this commit introduces a new command-line
option `--assume-checkpoint-safe-temp-files`. 

This option lets Snakemake know that it can  safely remove temporary files 
not required by current jobs, even if checkpoint jobs are still pending.

### QC

### DIFF
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -1656,6 +1656,17 @@ Further, an output file marked as ``temp`` is deleted after all rules that use i
         shell:
             "somecommand {input} {output}"
 
+
+When using checkpoint rules, Snakemake cannot determine whether temporary files are no longer needed until 
+all checkpoint jobs have been completed. This delay in removal may result in unexpectedly high storage usage.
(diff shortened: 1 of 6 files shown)


cmeesters:
@mhulsman and @johanneskoester I just merged the main into this PR. And as I just have the usecase for this, I will check out this PR next week and will try to apply it in a complex scenario. Thanks for this contribution!

At quick glance, the review is not overwhelming. Can do, if you want me to, Johannes.

PR #3236, opened 2024-12-19
fix: problem with spaces in path

### LINKED ISSUE #3235: "Snakefile not found" when path contains a space, using `run` directive
**Snakemake version**
8.25.5

**Describe the bug**

Snakemake fails with "Snakefile /path/that/includes/a/space" (part after space truncated) not found.

**Logs**

```bash
$ snakemake
```

```
host: ___________
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 16
Job stats:
job        count
-------  -------
default        1
total          1

Select jobs to execute...
Execute 1 jobs...

[Thu Dec 19 11:54:53 2024]
localrule default:
    jobid: 0
    reason: Rules with neither input nor output files are always executed.
    resources: tmpdir=/tmp

Error: Snakefile "/home/kyle/scratch/test-smk-bug/smk" not found.
[Thu Dec 19 11:54:55 2024]
Error in rule default:
    jobid: 0

Complete log: .snakemake/log/2024-12-19T115453.901323.snakemake.log
WorkflowError:
At least one job did not complete successfully.
```

**Minimal example**

```bash
mkdir smk\ space\ bug && cd smk\ space\ bug
echo 'rule default:\n    run:\n        print("hello")' > Snakefile
snakemake
```

**Additional context**

- The same thing happens in a symlinked folder to the folder with spaces in it
- The same thing happens in a subdirectory without spaces. i.e., a space anywhere in the path seems to break things
- It only happens when using the `run` directive, not `shell`
- It doesn't happen with `--dry-run`, but rather in the execution phase

kjohnsen:
Fixes #3235 
Goes together with https://github.com/snakemake/snakemake-interface-executor-plugins/pull/79 
This PR updates tests to catch the bug (should break at least `test_path_with_spaces` and `test_with_parentheses` and  https://github.com/snakemake/snakemake-interface-executor-plugins/pull/79 actually fixes it

## Summary by CodeRabbit

- **New Features**
	- Introduced a new rule that generates an output file with a predefined string.
- **Tests**
	- Added a test case to verify the handling of paths containing spaces.
	- Enhanced temporary directory naming for clarity during test execution.


kjohnsen:
🚨**Important**🚨 the bug only shows up when I run snakemake from the command line; the automated test doesn't fail. i.e., this works:
```bash
pytest pytest tests/tests.py::test_path_with_spaces
```
but this doesn't:
```bash
snakemake --snakefile tests/test_path\ with\ spaces/Snakefile
```

I discovered this because I saw a pre-existing test should have covered my issue: `test (with parenthese's)`. It passes via pytest but fails with `SyntaxError` when running from the terminal.

kjohnsen:
The previous commit adds the test folder name to the temp folder. Now the parentheses and space tests do fail when I run from pytest

johanneskoester (approved):

Issue #2994, opened 2024-08-02
The priority directive should support lambdas


nh13:
My use case is when I _know_ ahead of time that some executions of a particular rule will take longer than others, on a multi-core machine.  I want the "longer-running" rule to run first.  I was thinking I could specify the priority of the rule via a wildcard, but I get:

```console
RuleException in file /private/tmp/snakemake/Snakefile, line 14:
Priority values have to be numeric.
```

Here's the snakefile:

```snakemake
contigs = {
    "chr1": 1000,
    "chr2": 10000,
    "chr3": 100000,
    "chr4": 100,
}

rule all:
    input:
        [f"{c}.txt" for c in contigs]


rule txt:
    output:
        txt = "{c}.txt"
    priority: lambda wcs: contigs[wcs.c]
    shell:
        "touch {output.txt}"
```

TheSkyentist:
I definitely second this. Certain input/outputs might take longer to process with the same rules simply because of larger data volume. This request appears to be a duplicate of #2560 which didn't seem to gain much attention. I wonder how hard this would be to implement, I assume you'd have to ensure that the output of the function is number-like.

wiedenhoeft:
I third this :-) I am trying to avoid using pipes between rules for various reasons. This option would allow one to use 
`priority: lambda wc, input: input.size_mb`
to get rid of the largest temp files first, which is especially critical if multiple successive downstream rules communicate via temp files.

havardtl:
I also agree that this would be a great functionality for my use case!

TheSkyentist:
I've drafted #3253 which should enable this for wildcards, though not yet for inputs or resources (sorry @wiedenhoeft). Hopefully this can get some movement to enable this feature more fully.

wiedenhoeft:
Great to see this gain traction! Out of curiosity, what prevents me from using wildcards to recreate the input string and get the file size programmatically? Is this a question of timing for the scheduler, i.e. would the priority be fixed at the beginning of a Snakemake run?

TheSkyentist:
@wiedenhoeft, inputs and resources are now included in #3253 :-)

wiedenhoeft:
That is awesome, thank you so much!! Workflows and resource management will become so much more customizable this way :-)

PR #3253, opened 2025-01-14
Added lambda functions for priority in rules

TheSkyentist:
(Partially) adding the functionality desired in #2994. Lambda functions can now be used to specify priority based on input wildcards. 

Folks also desired the ability to use other inputs (such as input resources). Wanted to make sure this was more or less on the right track. Happy to receive feedback. 

I've added some tests to cover priorities as well and updated the docs. 

### QC


JohnAchenbach:
Feedback: Very much looking forward to the new version! I want to use this feature to selectively fast-track the analysis of samples from tumor patients who are in a particularly bad shape and may miss their treatment option (=become too weak to be treated and succumb to disease) if the analysis and hence recommendation of a therapy takes too long. I have a dictionary with all sample names and their priorities and wanted to dynamically retrieve the priority for each given sample from the dictionary using lambda wildcards. Great this is already underway!

JohnAchenbach:
I have applied the changes Raphael committed to dag.py and workflow.py month ago across three different versions of snakemake (7, 8, 9) and it works flawlessly in a context like this:

rule mapping_and_sorting:
    priority:
        lambda wildcards: get_priority(wildcards, sample_info_dict)

To us (Institut für Pathologie & Hämatopathologie Hamburg), this is a vitally important update, as it adds the possibility to dynamically prioritize NGS analyses where cancer patients urgently need results to inform their further treatment over those who can wait for another weekend. Overall, we analyse several thousand cancer samples annually. While this branch is not merged with the main branch, each re-install or update of snakemake breaks the functionality of the pipeline I am building until I manually re-add the changes. Once the pipeline goes productive, this is too much of a risk in terms of portability across multiple nodes. Since I see the finish line, I need to either know soon that I can rely on this function in the future or I'll have to recode everything. Your work is essentially done, it lives up to its purpose, I cannot spot any spelling mistakes, documentation update is intelligible and formatting looks good to me. Therefore, I kindly ask you to prioritize merging this branch with the main.

Thanks a lot!

Best regards,
John

TheSkyentist:
I would also love to see this merged! Perhaps we can get eyes on this @johanneskoester? The code change is quite small but already opens up quite a bit of functionality and comes with tests + documentation changes. My use case 

If possible, it would be great to get advice about how to proceed (perhaps in a second pull request) on how to achieve also passing resources into a priority call such that one could prioritize jobs based on their expected. resource usage.

JohnAchenbach:
> I would also love to see this merged! Perhaps we can get eyes on this @johanneskoester? The code change is quite small but already opens up quite a bit of functionality and comes with tests + documentation changes. My use case
> 
> If possible, it would be great to get advice about how to proceed (perhaps in a second pull request) on how to achieve also passing resources into a priority call such that one could prioritize jobs based on their expected. resource usage.

Good idea! With the functionality you built, you could already achieve that - at least in a simple version. All it takes is a python function accepting e.g. config["rulename_memory"] and/or config["rulename_threads"] as arguments and returning an integer. Then make a lambda call of the function with the required arguments in the prio section and its done ;)

TheSkyentist:
I found this to be trickier in practice, since resources can be dynamically assigned, e.g. based on input file size, and therefore calculated quite often whereas the priorities for the DAG are computed less frequently, only on DAG updates. 

The functionality in this PR was straightforward to implement and perhaps we could get a version of what you are suggesting implemented in the future, maybe under simplifying assumptions, but I believe its currently beyond the scope of this PR. But perhaps we can start a draft and experiment once this one is merged? :-)

JohnAchenbach:
> I found this to be trickier in practice, since resources can be dynamically assigned, e.g. based on input file size, and therefore calculated quite often whereas the priorities for the DAG are computed less frequently, only on DAG updates.
> 
> The functionality in this PR was straightforward to implement and perhaps we could get a version of what you are suggesting implemented in the future, maybe under simplifying assumptions, but I believe its currently beyond the scope of this PR. But perhaps we can start a draft and experiment once this one is merged? :-)

Absolutely! I can help testing, but I would not know how to make the priority section "see" what's inside the resources and threads sections. However, if I were to implement a rule with dynamic resource assignment and resource-dependent priority right now, I'd do dynamic rule generation in a for-loop, do the math for calculating resources and priority there and just hand the results to the respective sections (maybe useful also for a unit test for however you think this can be developed).

TheSkyentist:
Rebased to make merging easier.

TheSkyentist:
I have now fully implemented the functionality in #2994. It would be really great to get eyes on this. 

The docs fail to build but for reasons unrelated to the content added. Seems to be a versioning issue with sphinx? I get the same build errors on the main branch.

src/snakemake/dag.py
```diff
@@ -1674,7 +1674,11 @@ def update_priority(self):
             or not self.priorityfiles.isdisjoint(job.output)
         )
         for job in self.needrun_jobs():
-            self._priority[job] = job.rule.priority
+            self._priority[job] = (
+                job.rule.priority(job.wildcards)
+                if callable(job.rule.priority)
+                else job.rule.priority
+            )
```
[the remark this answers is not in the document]
TheSkyentist: Newest version enforces more rigorous try/except error catching.

docs/snakefiles/rules.rst
```diff
@@ -1219,7 +1219,22 @@ Snakemake allows for rules that specify numeric priorities:
       priority: 50
       shell: ...
 
-Per default, each rule has a priority of 0. Any rule that specifies a higher priority, will be preferred by the scheduler over all rules that are ready to execute at the same time without having at least the same priority.
+Per default, each rule has a priority of 0. Any rule that specifies a higher priority, will be preferred by the scheduler over all rules that are ready to execute at the same time without having at least the same priority. Priority may also be specified with a callable that returns an integer. The signature of the callable must be `callable(wildcards)`.
```
johanneskoester:
```suggestion
Per default, each rule has a priority of 0. Any rule that specifies a higher priority, will be preferred by the scheduler over all rules that are ready to execute at the same time without having at least the same priority.
```

docs/snakefiles/rules.rst
```diff
@@ -1219,7 +1219,22 @@ Snakemake allows for rules that specify numeric priorities:
...
+
+Priority may also be specified with a callable. The callable receives ``wildcards``
+as its first positional argument, and may optionally accept ``input``, ``attempt``,
+and ``rulename`` keyword arguments (similar to param functions):
```
johanneskoester:
```suggestion
and ``rulename`` keyword arguments (similar to param functions). It has to return the priority as an integer or float (will be rounded):
```

Issue #3036, opened 2024-08-20
Two checkpoint workflow fails to output all files in latest version 8.18.1


peterch405:
**Snakemake version**

Tested with v8.18.1
Works with v7.30.2

**Describe the bug**

A collect rule that aggregates files from two sequential checkpoints fails to output all files in the latest version of snakemake. This type of workflow works back in v7.30.2 of snakemake. In the latest version only files from the first checkpoint are produced, the second checkpoint is not executed. The `output_j = glob.glob(f"{checkpoints.second.get(**wildcards, i=i).output}/*/")` returns an empty list and does not trigger the second checkpoint to be run. I am unsure why this no longer works in the latest version, but would love some help making it run.

**Minimal example**

checkpoint_test.smk:
```
import glob
import random
from pathlib import Path

ALL_SAMPLES = ["s1", "s2"]


rule all:
    input:
        expand('collect/{sample}/all_done.txt',
               sample = ALL_SAMPLES)


checkpoint first:
    input:
        expand("{sample}", sample=ALL_SAMPLES)
    output:
        directory('first/{sample}')
    run:
        for i in range(1,5):
            Path(f"{output[0]}/{i}").mkdir(parents=True, exist_ok=True)
            Path(f"{output[0]}/{i}/test.txt").touch()


checkpoint second:
    input:
        'first/{sample}/{i}/test.txt'
    output:
        directory('second/{sample}/{i}')
    run:
        for j in range(6,10):
            Path(f"{output[0]}/{j}").mkdir(parents=True, exist_ok=True)
            Path(f"{output[0]}/{j}/test2.txt").touch()


rule copy:
    input:
        'second/{sample}/{i}/{j}/test2.txt'
    output:
        'copy/{sample}/{i}/{j}/test2.txt'
    shell:
        """
        cp -f {input} {output}
        """


def aggregate(wildcards):

    outputs_i = glob.glob(f"{checkpoints.first.get(**wildcards).output}/*/")

    outputs_i = [output.split('/')[-2] for output in outputs_i]

    split_files = []
    for i in outputs_i:
        output_j = glob.glob(f"{checkpoints.second.get(**wildcards, i=i).output}/*/")
        outputs_j = [output.split('/')[-2] for output in output_j]
        for j in outputs_j:
            split_files.extend(expand(f"copy/{{sample}}/{i}/{j}/test2.txt",
                               sample=wildcards.sample))
    return split_files




rule collect:
    input:
        aggregate
    output:
[...]

Issue #2958, opened 2024-07-10
Chained checkpoints are not evaluated correctly


lczech:
**Snakemake version**
8.15.2

**Describe the bug**
After finally upgrading my pipeline from Snakemake 6.0.5 to 8.15.2 (due to dependency hell with all the tools...), part of my checkpoint setup broke.

**Minimal example**
~~~Snakemake
rule all:
    input:
        "merged.vcf"

checkpoint samtools_faidx:
    output:
        touch("genome.fai")

checkpoint contig_groups:
    output:
        touch("contigs.txt")

def contigs_groups_input(wildcards):
    return checkpoints.contig_groups.get().output[0]

def get_contigs():
    contig_group_file = checkpoints.contig_groups.get().output[0]
    with open(contig_group_file, "r") as f:
        f.read() # <-- (X) FileNotFoundError: 'contigs.txt'
    # Return value based on the file, but hardcoded here for minimalicity.
    return "contig_1"

rule call_variants:
    output:
        touch("{contig}.vcf.tmp")

def merge_variants_input(wildcards):
    # (1) Removing the following line makes the error disappear:
    fai = checkpoints.samtools_faidx.get().output[0]
    # (2) Removing the following line and replacing it with
    # the hardcoded value below also makes the error disappear:
    contigs = get_contigs()
    # contigs = "contig_1"
    return contigs + ".vcf.tmp"

rule merge_variants:
    input:
        # (3) Adding the following input also makes the error disappear:
        # contigs_groups_input,
        merge_variants_input
    output:
        touch("merged.vcf")
~~~

**Logs**
```log
Assuming unrestricted shared filesystem usage.
Building DAG of jobs...
[...]/lib/python3.12/site-packages/snakemake/dag.py:1612: RuntimeWarning: coroutine 'DAG.sanitize_local_storage_copies' was never awaited
  self.sanitize_local_storage_copies()
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
Using shell: /usr/bin/bash
Provided cores: 16
Rules claiming more threads will be scaled down.
Job stats:
job               count
--------------  -------
all                   1
merge_variants        1
samtools_faidx        1
total                 3

Select jobs to execute...
Execute 1 jobs...

[Wed Jul 10 16:25:06 2024]
localcheckpoint samtools_faidx:
    output: genome.fai
    jobid: 2
    reason: Missing output files: <TBD>
    resources: tmpdir=/tmp
[...]

Issue #2762, opened 2024-03-17
When checkpoints are nested, the second checkpoint does not seem to run the rule


fgregg:
**Snakemake version**
8.8.0

**Describe the bug**
When there are nested checkpoints, the interior checkpoint does not seem to be running the shell command.

**Minimal example**
```snakemake
import json
import pathlib

def texts(wildcards):

    with checkpoints.needs_text.get(**wildcards).output[0].open() as f:
        file_names = json.load(f)

    return file_names

rule all:
    input: texts 

def aggregate_pieces(wildcards):

    pieces_directory = pathlib.Path(checkpoints.pieces.get(**wildcards).output[0])
    files = list(pieces_directory.glob("*.txt"))
    return files
        
rule concatenate:
    input: aggregate_pieces
    output: "{name}.txt"
    wildcard_constraints:
        name="[a-z]+"
    shell:
        """
        cat {input} > {output}
        """    

checkpoint pieces:
    output: directory("pieces/{name}/")
    shell:
        """
        mkdir {output}
        echo "hello" > {output}/1.txt
        """

checkpoint needs_text:
    output: "needs_text.json"
    shell:
        """
        echo '["b.txt", "a.txt"]' > {output}
        """
```    

When you run the `all` rule, the `{input}` of the `concatenate` rule does not get hydrated

```console
> snakemake --cores 1 -p
Using shell: /opt/homebrew/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job            count
-----------  -------
all                1
concatenate        2
total              3

Select jobs to execute...
Execute 1 jobs...

[Sat Mar 16 23:08:57 2024]
localrule concatenate:
    output: b.txt
    jobid: 3
    reason: Set of input files has changed since last execution
    wildcards: name=b
    resources: tmpdir=/var/folders/p0/59kzf5n91jl8jxbx7793rg440000gn/T


        cat  > b.txt
```

But if you run snakemake with a named target, that succeeds.

```console
> snakemake --cores 1 -p b.txt
Assuming unrestricted shared filesystem usage.
Building DAG of jobs...
Using shell: /opt/homebrew/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job            count
-----------  -------
concatenate        1
total              1

Select jobs to execute...
Execute 1 jobs...

[Sat Mar 16 23:13:42 2024]
[...]

PR #3256, opened 2025-01-18
fix: Fix bug with checkpoint execution

zmbc:
This is a rough start on fixing #2762, #3036, #2958.

It appears that with this change, though I don't completely understand it, the minimal example in #2762 completes successfully.

### QC

## Summary by CodeRabbit

- **Refactor**
	- Improved logic for checking future outputs in the checkpoint tracking mechanism, enhancing clarity and efficiency.


zmbc:
It seems clear that the fix is nowhere near this simple, closing

PR #3262, opened 2025-01-22
feat: add support for validation of polars dataframe and lazyframe

fgvieira:
### QC

### DIFF
--- a/tests/test_validate/Snakefile
+++ b/tests/test_validate/Snakefile
@@ -1,6 +1,7 @@
 shell.executable("bash")
 
 import pandas as pd
+import polars as pl
 from snakemake.utils import validate
 
 configfile: "config.yaml"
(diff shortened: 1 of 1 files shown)


tests/test_validate/Snakefile
```diff
@@ -10,6 +11,11 @@ validate(config, "config.schema.yaml")
 samples = pd.read_table(config["samples"]).set_index("sample", drop=False)
 validate(samples, "samples.schema.yaml")
 
+samples = pl.read_csv(config["samples"])
+validate(samples, "samples.schema.yaml")
```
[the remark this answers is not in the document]
fgvieira: `separator` added

snakemake/utils.py
```diff
@@ -111,33 +111,67 @@ def set_defaults(validator, properties, instance, schema):
     if not isinstance(data, dict):
         try:
             import pandas as pd
+            import polars as pl
```
johanneskoester, line 114: this should work independently for pandas and polars. Otherwise, people will always have to have installed both just to validate one of them.

johanneskoester (changes requested):

snakemake/utils.py
fgvieira: But both are already part of `snakemake` dependencies, no?

johanneskoester (approved):

johanneskoester (changes requested):
The only thing that is missing here is to mention that both polars and pandas work for validate here: https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html#validation

johanneskoester (approved):

PR #3295, opened 2025-02-15
feat: mastodon robot enabled

cmeesters:
This PR enables auto-posting of release messages to the Mastodon announcement bot at https://fediscience.org/@snakemake

### QC

This PR does not alter any code. It is hence not necessary to include a test case or documentation

### DIFF
--- a/.github/workflows/post_to_mastodon.sh
+++ b/.github/workflows/post_to_mastodon.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+# Extract version from PR tag passed as environment variable
+if [ -z "${PR_TITLE}" ]; then
+    echo "Error: 'PR_TITLE' environment variable is not set."
+    exit 1
+fi
+
+# Check if this is a release PR
+if [[ ! "${PR_TITLE}" =~ ^chore\(main\):\ release ]]; then
+    echo "Not a release PR, skipping Mastodon post"
+    exit 0
+fi
+
+# Extract version (everything after "release ")
+if [[ "${PR_TITLE}" =~ [Rr]elease[[:space:]]+([0-9]+\.[0-9]+\.[0-9]+) ]]; then
+    version="${BASH_REMATCH[1]}"
+else
+    echo "Error: Could not extract version number from PR title"
+    exit 1
+fi
+
+# Validate version format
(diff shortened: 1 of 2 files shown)


cmeesters:
@coderabbitai

Apparently, the codespell check detects a typo. Yet, the line it is referring to is a regular expression. Is there a way to tweak the check to ignore this specific "error"?

cmeesters:
@coderabbit it is actually line 16 in that file. The regex is `[Rr]elease` and the flagged word `elease`, because codespell apparently detects a missing `r`.

.github/workflows/codespell.yml
```diff
@@ -22,4 +22,4 @@ jobs:
       - name: Codespell
         uses: codespell-project/actions-codespell@v2
         with:
-          ignore_words_list: Crate,crate
+          ignore_words_list: Crate,crate,elease
```
johanneskoester, line 25: intentional?

johanneskoester (changes requested):
I think, since we want to have this for every snakemake repo, we should turn this into a github composite action rather now, before we duplicate too much code. Example to learn from is the snakemake github action: https://github.com/snakemake/snakemake-github-action/blob/master/action.yml

.github/workflows/codespell.yml
cmeesters: The regex is `[Rr]elease`. Yet, perhaps the whole idea of having a shell script is ludicrous. See the comment below. By pythonizing, that exception will be gone.

cmeesters:
I will take a look for the action design.

And I had a look onto the python module for Mastodon. I will check this in the coming days: It might allow attaching a thumbnail image, some eye-catcher, per repo. E.g. a rulegraph for workflows, a cluster image for the batch system executors, a "cloud" for the cloud executors and a snakish logo for the main repo (just a braindump ;-) ).

If this works as intended, I will update this PR accordingly.

cmeesters:
Honestly, I do not quite understand how the action template differentiates between repos. And we would need different text and images per repo.

johanneskoester:
> Honestly, I do not quite understand how the action template differentiates between repos. And we would need different text and images per repo.

yes, those can be passed as arguments to the action.

cmeesters:
withdrawn - the action https://github.com/snakemake/mastodon-release-post-action - will make publishing organizationwide possible

PR #3303, opened 2025-02-18
fix: robustly escaping quotes in generated bash scripts (#3297)

### LINKED ISSUE #3297: Unescaped quotes in generated bash shell script
**Snakemake version**

8.27.1

**Describe the bug**

When using the `shell` directive (bash script) with the `config` directive, associative arrays are added to a generated shell script. If a string in the config YAML file contains `'` then the generated associative arrays contain unescaped double-quotes which causes a syntax error.

**Logs**

```
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 20
Job stats:
job      count
-----  -------
test         1
total        1

Select jobs to execute...
Execute 1 jobs...

[Mon Feb 17 11:09:53 2025]
localrule test:
    output: test.txt
    jobid: 0
    reason: Forced execution
    resources: tmpdir=/tmp

/home/rosewood/Desktop/snakemake_shell/.snakemake/scripts/tmpz6rzz406.script.sh: line 9: unexpected EOF while looking for matching `''
/home/rosewood/Desktop/snakemake_shell/.snakemake/scripts/tmpz6rzz406.script.sh: line 13: syntax error: unexpected end of file
Not cleaning up /home/rosewood/Desktop/snakemake_shell/.snakemake/scripts/tmpz6rzz406.script.sh
RuleException:
CalledProcessError in file /home/rosewood/Desktop/snakemake_shell/Snakefile, line 9:
Command 'set -euo pipefail;  bash /home/rosewood/Desktop/snakemake_shell/.snakemake/scripts/tmpz6rzz406.script.sh' returned non-zero exit status 1.
[Mon Feb 17 11:09:53 2025]
Error in rule test:
    jobid: 0
    output: test.txt

Complete log: .snakemake/log/2025-02-17T110953.594439.snakemake.log
WorkflowError:
At least one job did not complete successfully.
```

**Minimal example**

*Snakefile*
```snakemake
configfile: 'config.yaml'

rule test:
    output:
        'test.txt'
    params:
        d={'a': "foo'"}
    script:
        "script.sh"
```

*script.sh*
```bash
#!/usr/bin/env bash

echo "hello" > "${snakemake_output[0]}"
```

*config.yaml*
```yaml
foo:
  bar: "let's go"
```

Run command `snakemake test.txt --skip-script-cleanup -f`

**Additional context**

Here is the generated bash script:

```bash
#!/usr/bin/env bash

declare -A snakemake_input=( )
declare -A snakemake_output=( [0]="test.txt" )
declare -A snakemake_params=( [0]="a" [d]="a" )
declare -A snakemake_wildcards=( )
declare -A snakemake_resources=( [0]="1" [_cores]="1" [1]="1" [_nodes]="1" [2]="/tmp" [tmpdir]="/tmp" )
declare -A snakemake_log=( )
declare -A snakemake_config=( [foo]="{'bar': "let's go"}" )
declare -A snakemake=( [_params_types]="{}" [threads]="1" [rule]="test" [bench_iteration]="None" [scriptdir]="/home/rosewood/Desktop/snakemake_shell" )

echo "hello" > "${snakemake_output[0]}"
```

Culprit is `declare -A snakemake_config=( [foo]="{'bar': "let's go"}" )`. Note that the `snakemake_params` is also incorrect.

**Potential patch**
I've tried the following fix, which solves the above MWE but may incorrectly escape double quotes.

```diff
diff --git a/snakemake/script/__init__.py b/snakemake/script/__init__.py
index c802bc15..4a70c41b 100644
--- a/snakemake/script/__init__.py
+++ b/snakemake/script/__init__.py
@@ -468,7 +468,8 @@ class BashEncoder:
         """Converts a dictionary to an associative array"""
         s = "( "
         for k, v in d.items():
-            s += f'[{k}]="{v}" '
+            formatted_v = f"{v}".replace('"', '\\"')
+            s += f'[{k}]="{formatted_v}" '
 
         s += ")"
         return s
```

prs513rosewood:
The fix runs Python's format-generated string through `shlex.quote` to properly interpret in shell scripts generated with the `script:` directive.

Closes #3297 

### QC

## Summary by CodeRabbit

- **New Features**
  - Introduced new classes and methods for enhanced report URL construction and script execution.
  - Added support for GPU resource management in workflows.
  - Enhanced reporting features with toggle functionality for different result views.

- **Bug Fixes**
  - Improved error handling and reporting for job execution results.
  - Enhanced the clarity of error messages related to file handling and storage.

- **Tests**
  - Added comprehensive test suites for new features and existing functionalities, including validation of output index and prefix lookup.
  - Introduced tests for Conda-related functionalities within a Snakemake context.

- **Documentation**
  - Updated documentation to reflect new features and best practices for workflow management.
  - Added new sections for maintainers and resources in the documentation.


tbooth:
I'm taking a look at this today. This PR is a good start, but I can quite easily add further tests that still break things, so I'm seeing this as an opportunity to beef up the tests further and try to make the entire business of passing values to scripts (including R and Julia) as robust as I can.

In the proposed modified test case, we have:

```
foo:
  bar: "let's go"
```

Then since BASH does not support nested dicts, the entire dictionary is quoted as a string `'{'"'"'bar'"'"': "let'"'"'s go"}'` which avoids a syntax error but is not very useful within the BASH script, and also is not consistent with how resources and params are passed. The existing test Snakefile includes this:

```
rule bash:
...
    params:                                                                                         
        integer=123,                                                                                
        string="foo",                                                                               
        d={"a": "foo"}
```

And here the dict `d` is just represented as the string `"a"`. If I add another key "b" then the dict gets rendered as `"a b"` - so just a string listing the keys. I've never considered that *params* could contain dicts, only strings, numbers and lists. And the expected output does not assert what the behaviour should be, so I'm not clear on whether this behaviour is designed/documented or just it-is-what-it-is because the dict has been treated as a list. Probably the latter (I'll read the docs carefully to check), but I'd like to see if we could make this all consistent without breaking any documented behaviour.

I'll also add some further tests for R and Julia.

If this turns out to be a really big job and can't be done in a few days then the PR as it stands will at least avoid a very problematic failure case, and I think should be applied.

tbooth:
I think I have it all working - see https://github.com/EdinburghGenomics/snakemake/tree/issue_3297

@prs513rosewood how would you like to proceed? I can make a fresh PR based on the above branch, which will be easy to do, but then you will not get any credit for your work on the fix because I somehow lost your commits from my Git history.

Or you can pull the above commit then rebase the changes onto your own branch. It's a little bit of faff for you, but you will then get credit in the changelog for your work. It also avoids the need to make a second PR.

Either is fine by me. Hopefully this being the Snakemake hackathon week it's a good time to get a PR accepted!

tbooth:
Note that in my new version, a nested dict within the config gets represented as a string containing just the dict keys. This is consistent with how params are represented, and consistent with the existing documentation regarding list values. I don't think shell-escaping the full Python representation of the dict is very helpful to anyone - if something nested within a dict within the config needs to be extracted and passed to the shell script then it can easily be copied into a param.

prs513rosewood:
To be honest I don't really have the time to worry about it, so it's probably best if you make a seperate PR.

PR #3307, opened 2025-02-21
chore: test dash shell

johanneskoester:
### QC

## Summary by CodeRabbit

- **Tests**
	- Enhanced automated testing to run in an additional shell environment, ensuring compatibility with Python 3.12.
	- Updated the environment setup to include `dash` for extended test coverage.
	- Improved flexibility of the testing configuration by allowing tests to run in both `bash` and `dash` environments, with specific conditions for shell usage.


peterjc:
Thanks - as per our discussion on Mastodon, I've not yet made a minimal test case but did suspect the system default shell of Dash rather than Bash breaking something.

johanneskoester:
Yes. THese tests for now ensure that having dash as default does not cause failures in the main Snakemake code. It can still be that there is a problem with dash in the slurm plugin.

PR #3334, opened 2025-03-05
test: Tidying up the tests

tbooth:
(I'm struggling with the PR mechanism here as there the "remove datrie" commits and changes are showing up, even though they should be made on both branches. Anyway, I'm hoping I can open the PR first then clean it up.)

The purpose of this change is to try and improve the tests suite by:

1) Providing `tests/README.md` to guide new developers how to run the tests
1) Split out the tests that require conda to run into a separate `tests_using_conda.py`
1) Allow the main tests to pass even if Snakemake cannot determine its own version (and thinks it is version 0.0)

Also:

1) Removed dependency on stress by using stress-ng which is available via Bioconda
1) Clean up some test files that should never have been committed to GIT
1) Change some 10 second stress delays to 4 seconds as there seems no reason for them to be that long
1) Fix a test that modified the environment (`shell.prefix`) so later tests could fail depending on the order of tests run

## Summary by CodeRabbit

- **Chores**
  - Revised build workflows by switching to an improved stress testing tool and enhancing version processing.
  - Refined file tracking rules for smoother development, including updates to `.gitignore`.

- **Tests**
  - Expanded and optimized the test suite by removing redundant tests and adding comprehensive checks for environment management and benchmark timing.
  - Introduced a new test suite specifically for validating Conda functionality within Snakemake.

- **Documentation**
  - Added detailed testing instructions to support proper environment setup.
  - Enhanced guidelines for running tests locally, including setup for Conda environments.


tbooth:
I rebased, force pushed, and the diff now looks correct.

johanneskoester:
I have no clue why the win tests suddenly fail.

johanneskoester:
> > I have no clue why the win tests suddenly fail.
> 
> It's because "stress-ng" is not available for Windows via Conda. I may need to revert this part of the patch.
> 
> I think all the tests that relied on "stress" were being skipped on Windows anyway, but it looks like this whole file needs to be installable on all platforms.

Actually not, there is some (admittedly hacky and hidden) code to remove certain packages from the env on windows: https://github.com/snakemake/snakemake/blob/75afa380edd192fd839c9939a7af471e5ff0aa64/.github/workflows/main.yml#L177
You can put stress-ng there.

tests/README.md
```diff
@@ -0,0 +1,94 @@
+# Unit and regression tests for Snakemake
```
johanneskoester, line 1: Sorry, one final thing besides my comment above: can you merge the content of this readme into https://snakemake.readthedocs.io/en/stable/project_info/contributing.html#testing-guidelines, and just put a link to that doc section into this readme? This way, tests are documented in just one place but the readme still helps people who haven't found the section in the online documentation.

johanneskoester (changes requested):

tests/README.md
tbooth: Yes, will do

tbooth: Content was merged, so we no longer have redundant docs.

One question - the docs mention wercker.com. Is this still a thing? I thought we were all running of GitHub actions now?

docs/project_info/contributing.rst
```diff
@@ -97,26 +97,98 @@ Testing Guidelines
...
+Tests in ``tests/test_api.py`` require a working ``git``.
+This is not included in ``test-environment.yml`` as it's assumed you must have GIT installed to be working on the source code, but installing git into the conda environment should work if need be.
+
+Depending on how the Snakemake code was downloaded and installed in the test environment, Snakemake may not be able to determine its own version and may think that it is version 0.
+The existing unit tests should all cope with this, and in general you should avoid writing tests that rely on explicit version checks.
+
+Continuous integration
+======================
+
+To have integration tests run automatically when committing code changes to Github, you need to sign up on wercker.com and register a user.
```
johanneskoester, line 191:
```suggestion
When creating a pull request for https://github.com/snakemake/snakemake, all tests will be automatically executed in a controlled environment by Github Actions.
```

docs/project_info/contributing.rst
```diff
@@ -97,26 +97,98 @@ Testing Guidelines
...
+
+1. Ensure your Conda config (``~/.condarc``) has these options:
+
+.. code-block::
+
+   channels:
+     - conda-forge
+     - bioconda
+   channel_priority: strict
+   solver: libmamba
```
johanneskoester, line 117: The channel setup is not needed for the tests to work. They all come with their own conda envs, don't they? Also the libmamba setting can be avoided by just requiring a minimum conda version (24.7.1) to be used (which uses libmamba by default). I think that is prefereable, since the libmamba integration in earlier conda versions was problematic.

johanneskoester: Hence, I would suggest to just say here that conda 24.7.1 should be present and strict priorities (linking to https://conda-forge.org/docs/user/tipsandtricks/#using-multiple-channels) and suggesting to run the command `conda config --set channel_priority strict` should be sufficient here.

johanneskoester (changes requested):

johanneskoester:
I will merge this and change the things I mentioned above in a follow up-PR 9as I don't have the permissions to push to your fork.

johanneskoester:
Thanks again for the fantastic work!

Issue #3319, opened 2025-02-28
document `snakemake --report report.zip` functionality


dlaehnemann:
Currently, the docs only list `snakemake --report report.html` format, but not the `report.zip` variant that scales to larger reports.

johanneskoester:
And maybe that should even be the new default when one just specifies --report.

dlaehnemann:
So I have addressed the original issue, but I am not sure where any of the embada-ability checks would go and thus no idea where the error message would go.

cmeesters:
this seems to be "done", right? @dlaehnemann

PR #3357, opened 2025-03-10
docs: clearly explain report rendering to ZIP archive

dlaehnemann:
This addresses the first part of #3319, by making the explanations of ZIP archive reports better.

### QC

### DIFF
--- a/docs/snakefiles/reporting.rst
+++ b/docs/snakefiles/reporting.rst
@@ -4,8 +4,16 @@
 Reports
 =======
 
-From Snakemake 5.1 on, it is possible to automatically generate detailed self-contained HTML reports that encompass runtime statistics, provenance information, workflow topology and results.
+From Snakemake 5.1 on, it is possible to automatically generate detailed self-contained HTML reports that you can easily move and share. 
+They encompass runtime statistics, provenance information and workflow topology by default, and workflow developers can also specify and annotate results for inclusion.
+For smaller default reports, these can be a single :ref:`snakefiles-self_contained_html_file`.
+For more complext reports, one can generate a :ref:`snakefiles-self_contained_zip_archive`, with a contained ``report.html`` as the main entry point.
 **As an example, the report of the Snakemake rolling paper can be found** `here <https://snakemake.github.io/resources/report.html>`__.
+
+.. _snakefiles-including_results_in_a_report:
+
+Including results in a report
+-----------------------------
 
 For including results into the report, the Snakefile has to be annotated with additional information.
 Each output file that shall be part of the report has to be marked with the ``report`` flag, which optionally points to a caption in `restructured text format <https://docutils.sourceforge.io/docs/user/rst/quickstart.html>`_ and allows to define a ``category`` for grouping purposes.
@@ -94,7 +102,7 @@ This works as follows:
             """
 
 Defining file labels
---------------------
+^^^^^^^^^^^^^^^^^^^^
 
 In addition to category, and subcategory, it is possible (and highly recommended!) to define a dictionary of labels for each report item.
 By that, the actual filename will be hidden in the report and instead a table with the label keys as columns and the values in the respective row for the file will be displayed.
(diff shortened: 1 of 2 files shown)


docs/snakefiles/reporting.rst
```diff
@@ -246,36 +254,77 @@ Further, using ``url_args()`` you can add URL arguments and using ``anchor()`` y
             file=f,
         )
 
+.. _snakefiles-rendering_reports:
+
 Rendering reports
 -----------------
 
-To create the report simply run
+All the metadata contained in the report (e.g. runtime statistics) is automatically collected during the rendering of the report.
```
cmeesters, line 262: "are automatically"? As metadata are plural?

docs/snakefiles/reporting.rst
```diff
@@ -246,36 +254,77 @@ Further, using ``url_args()`` you can add URL arguments and using ``anchor()`` y
...
 .. code-block:: bash
 
-    snakemake --report report.html
+    snakemake --report
 
-after your workflow has finished.
-All other information contained in the report (e.g. runtime statistics) is automatically collected during creation.
-These statistics are obtained from the metadata that is stored in the ``.snakemake`` directory inside your working directory.
+after your workflow has finished successfully.
+This will implicitly assume a report file name of ``report.html`` and will create this self-contained HTML report.
```
cmeesters, line 277: This will create a report file named ``report.html`` which includes a self-contained HTML report.

The words "implicitly assume" are semantically wrong associated with the output, because, the file name is a default and hence, it will "implicitly create". But then, it is explicitly created ... . Perhaps the shortened version above will do?

docs/snakefiles/reporting.rst
```diff
@@ -246,36 +254,77 @@ Further, using ``url_args()`` you can add URL arguments and using ``anchor()`` y
...
 .. code-block:: bash
 
-    snakemake --report report.html --report-stylesheet custom-stylesheet.css
+    snakemake --report some_report.zip
 
-In particular, this allows you to e.g. set a logo at the top (by using CSS to inject a background for the placeholder ``<div id="brand">``, or overwrite colors.
-For an example custom stylesheet defining the logo, see :download:`here <../../tests/test_report/custom-stylesheet.css>`.
-The report for above example can be found :download:`here <../../tests/test_report/expected-results/report.html>` (with a custom branding for the University of Duisburg-Essen).
-The full example source code can be found `here <https://github.com/snakemake/snakemake/tree/main/tests/test_report/>`__.
+You can move this file wherever you need it and then unpack it.
```
cmeesters, line 299: You can move this file to wherever you want to view it and then unpack it.

??

docs/snakefiles/reporting.rst
```diff
@@ -246,36 +254,77 @@ Further, using ``url_args()`` you can add URL arguments and using ``anchor()`` y
...
+
+The report can be restricted to particular jobs and results by specifying targets at the command line, analog to normal Snakemake execution.
 For example, with
 
 .. code-block:: bash
 
     snakemake fig1.svg --report report-short.html
 
 the report contains only ``fig1.svg``.
+This can for example be useful when a larger workflow has not yet run to completion, but you already want to explore some intermediate outputs in the report.
```
cmeesters, line 314: Erase the qualifier "for example" - no need.

cmeesters (commented):
Just a few minor suggestions - no need to accept them all.

docs/snakefiles/reporting.rst
dlaehnemann: What about:
```suggestion
This will create a self-contained HTML report with the default filename ``report.html``.
```

docs/snakefiles/reporting.rst
dlaehnemann:
```suggestion
This can be useful when a larger workflow has not yet run to completion, but you already want to explore some intermediate outputs in the report.
```

docs/snakefiles/reporting.rst
dlaehnemann: I went back and forth on this. Grammatically correct is the plural, but singular is also acceptable. So let's maybe go with grammatically correct:
```suggestion
All the metadata contained in the report (e.g. runtime statistics) are automatically collected during the rendering of the report.
```

docs/snakefiles/reporting.rst
cmeesters: much better, indeed!

Issue #3338, opened 2025-03-06
Do not overwrite parameters when using module with `with` directive


LKress:
### Is your feature request related to a problem? Please describe.

When a rule is included via the `module` directive and an argument is modified via `with`
(e.g. parameters), then all other parameters are overwritten. It would be 
helpful to keep all parameters from the imported workflow and only overwrite
the ones that are defined in the imported module.

### Describe the solution you'd like

When the parameters of the imported module are stored, only overwrite the
ones defined in the new workflow.

### Describe alternatives you've considered

If there is a reason for completely overwriting all parameters, the behaviour
can remain the same.

### Minimal example

Snakemake version: 8.29.1

Workflow to import (name: `other_workflow`):

```
rule all:
    input:
        "test_1_file.tsv",
        "test_2_file.tsv"

rule test:
    input:
        file = "file_{nr}.txt"
    output:
        file = "test_{nr}_file.tsv"
    params:
        a = 3,
        b = 6
    shell:
        "echo {params.a} && cp {input.file} {output.file}"
```

Workflow that imports the above test rule:

```
rule all:
    input:
        "test_1_file.tsv",
        "test_2_file.tsv"

module other_workflow:
    snakefile:
        # here, plain paths, URLs and the special markers for code hosting providers (see below) are possible.
        "other_workflow"

use rule test from other_workflow as other_test with:
    params:
        b = 9
```

When running other workflow via `snakemake --snakefile other_workflow` as expected, 3 is printed when rule test is executed.
However, if `Snakefile`  is run via `snakemake`, the following error is thrown:

```
RuleException in rule other_test in file /home/luis/Documents/PhD/development/snakemake_test/other_workflow, line 7:
AttributeError: 'Params' object has no attribute 'a', when formatting the following:
echo {params.a} && cp {input.file} {output.file}
```

johanneskoester:
Yes, this would be very neat, and seemingly also more natural, so that I think one would not even have to declare this as a breaking change.

johanneskoester:
Fixed by #3365. Thanks a lot for reporting!

PR #3365, opened 2025-03-10
Fix: Keep parameters when using 'with' directive

LKress:
Keep parameters when using 'with' directive. 
If the included rule contains parameters that are not defined in the given rule, these parameters are inherited and only the newly specified parameter is overwritten.

Related to issue #3338 

### QC


johanneskoester (approved):
Great work!

PR #3367, opened 2025-03-11
fix: group resources

fgvieira:
Resources on groups are gathered but, currently, it only works if they are all numeric. If some resources are human-readable strings (e.g. "10 m" or "5 GiB"), it fails.

### QC


fgvieira:
Closing, as no longer relevant.

PR #3368, opened 2025-03-11
feat: Maximal file size for checksums

JZentgraf:
Add a new parameter `--max-checksum-file-size` to adjust the maximal file size for which a checksum is computed when the DAG is build.
This is related to https://github.com/snakemake/snakemake/issues/2949

### QC

## Summary by CodeRabbit

- **New Features**
  - Introduced a command-line option to specify a maximum file size for checksum processing, defaulting to 1,000,000 bytes.
  - Added a new configuration parameter for managing the maximum allowable size of checksum files.

- **Refactor**
  - Updated file validation and checksum routines to incorporate the new file size limit, enhancing processing efficiency.
  - Made method signatures more explicit by removing default values for threshold parameters in checksum-related methods.


snakemake/cli.py
```diff
@@ -1299,6 +1299,14 @@ def get_argument_parser(profiles=None):
...
         "to individual checks for the rest.",
     )
+    group_behavior.add_argument(
+        "--max-checksum-file-size",
+        type=int,
+        default=1000000,
+        metavar="BYTES",
+        help="Compute the checksum during DAG computation and job postprocessing "
+        "only for files that are smaller than the provided threshold.",
+    )
```
johanneskoester, line 1309: Snakemake depends on the humanfriendly package, which allows to parse sizes like this from strings. I think we should use that here. You can pass humanfriendly.parse_size to the argument so that it gets directly interpreted and translated into bytes when the user specifies somehing like `--max-checksum-file-size 1MB`.

johanneskoester (changes requested):
Thanks a lot! One minor thing below.

snakemake/cli.py
```diff
@@ -1308,6 +1314,14 @@ def get_argument_parser(profiles=None):
         "network file systems. Hence, we do not spend more than a given amount of time and fall back "
         "to individual checks for the rest.",
     )
+    group_behavior.add_argument(
+        "--max-checksum-file-size",
+        default=1000000,
+        metavar="BYTES",
+        parse_func=parse_size_in_bytes,
+        help="Compute the checksum during DAG computation and job postprocessing "
+        "only for files that are smaller than the provided threshold.",
```
johanneskoester, line 1323:
```suggestion
        help="Compute the checksum during DAG computation and job postprocessing "
        "only for files that are smaller than the provided threshold (given in any valid size "
        "unit, e.g. 1MB, which is also the default). ",
```

johanneskoester (approved):

PR #3374, opened 2025-03-11
fix: only skip eval when resource depends on input

### LINKED ISSUE #3271: accessing mem_mb from resources in params
Maybe this is not a bug but a wanted change in behavior in v8.

**Snakemake version**
- 8.27.1

**Describe the bug**
I started to move my pipelines from v7 to v8. In one of my rules I am accessing mem_mb from resources in params to calculate Gb -> Mb (stored in params.mem_gb). This worked so far without problems but now I am running into an error. I will first post my rule and then the error (log).

Any idea how to solve this? Thanks!

**Rule**
```bash
rule make_ref_dict:
    message: "Create a GATK ref dictionary."
    input:
        reference = rules.reference.output.reference
    output:
        dict = REFERENCE_PREFIX + '.dict'
    resources:
        mem_mb =  lambda wildcards, attempt: attempt * 2000,
        cpus_per_task = 1
    params:
        mem_gb = lambda wildcards, resources: int(resources.mem_mb / 1000)
    shell:
        """
           picard -Xmx{params.mem_gb}g CreateSequenceDictionary -R {input.reference} -O {output.dict}
        """
``` 

**Logs**
```bash
Building DAG of jobs...
InputFunctionException in rule variant-qc.make_ref_dict in file snakefile.smk, line 109:
Error:
  TypeError: __int__ returned non-int (type TBDString)
Wildcards:

Traceback:
  File "snakefile.smk", line 124, in <lambda> (rule variant-qc.make_ref_dict, line 153, snakefile.smk)
```

ezherman:
Fixes #3271.

Currently, evaluation of functions in resources are skipped until after input is evaluated. However, this is not necessary when a function does not rely on input. This PR amends the skipping, to only operate on resources that depend on input.

### QC


snakemake/jobs.py
```diff
@@ -481,6 +481,14 @@ def attempt(self, attempt):
         self._resources = None
         self._attempt = attempt
 
+    def _get_resources_to_skip(self):
+        """Return a set of resource names that are callable and depend on input files."""
+        return {
+            name
+            for name, val in self.rule.resources.items()
+            if is_callable(val) and "input" in val.__code__.co_varnames
```
johanneskoester, line 489:
```suggestion
            if is_callable(val) and "input" in get_function_params(val)
```

johanneskoester (changes requested):
The function comes from snakemake.common in case it is not yet imported in this module.

johanneskoester (approved):

PR #3376, opened 2025-03-11
fix: enhance error message formatting for strict DAG-building mode

### LINKED ISSUE #3331: Strict DAG-building mode to help with debugging
I'm frequently frustrated by `MissingInputException`s and other challenges while trying to debug a complex workflow, especially when making extensive use of `Callable`s for deriving input files based on a config.

I believe that this is often due to unintuitive behavior when an input function raises an error (my understanding is that this error is caught and not re-raised as long as a different source for the prerequisite file is found).

Hard-to-reason-about behaviors like these, while they may often result in the correct DAG and fewer errors for the user to sort through, are potentially harmful and allow workflow writers to be less disciplined in their design.

I'd like to see a "strict mode" for DAG building, where errors like these are re-raised instead of being caught. I.e. a `-Werror -Wall` for Snakemake. I believe that in many (all?) cases, workflows can be designed that do not rely on the current lax/dynamic approach to dependency resolution, and that workflows that strive for this ideal are easier to debug.

Ideally, any workflow that runs successfully in strict mode will be resolved identically in the default mode. For some workflow writers, strict mode would serve as a complement to the linter.

**Additional context**
I'm optimistic that increasing the strictness of DAG resolution may enable other optimizations that speed up DAG building or allow for static analyses.

Marco-Masera:
### Description
This PR adds three CLI parameters to enhance debugging capabilities through strict evaluation modes:
* strict-functions-evaluation
* strict-cyclic-graph-evaluation
* strict-periodic-wildcards-evaluation
These flags are stored in types.DAGSettings and when enabled, enforce strict checking of exceptions that would otherwise be ignored if the rules producing them are not strictly required. The affected exceptions are:

* InputFunctionException
* CyclicGraphException
* PeriodicWildcardError

When these flags are not set, appropriate warning messages are now displayed for these exceptions.

Additionally, small refactor to exceptions.py: the code to stringify exceptions is moved from print_exception into a new function named format_exception_to_string, which is now used to build consistent strings for both error reporting and logging purposes.

Closes #3331 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new command-line option to enable strict evaluation of dependency graphs, allowing users to enforce stringent checks.
  - Consolidated multiple configuration flags into a unified setting for more flexible strict evaluation choices.

- **Refactor**
  - Enhanced error messaging and warning prompts for evaluation issues, providing clearer feedback when stricter rules are applied.


johanneskoester (changes requested):
I would suggest to use an enum and a single arg for this instead, e.g.

```
snakemake --strict-dag-evaluation periodic-wildcards cyclic-graph functions
```

The enum should be defined under snakemake.settings.enums. If you use the same base class as the other enums there (SettingsEnumBase), you can also copy the way the enum is used to offer the choices for the argument parser, e.g. copy from --rerun-triggers.

snakemake/cli.py
```diff
@@ -860,6 +860,15 @@ def get_argument_parser(profiles=None):
         ),
     )
 
+    group_exec.add_argument(
+        "--strict-dag-evaluation",
+        nargs="+",
+        choices=StrictDagEvaluation.choices(),
+        default=[],
```
johanneskoester, line 867:
```suggestion
        default=set(),
```

johanneskoester (approved):

PR #3377, opened 2025-03-11
fix: in remote jobs, upload storage in topological order such that modification dates are preserved (e.g. in case of group jobs)

johanneskoester:
### QC

## Summary by CodeRabbit

- **Refactor**
	- Improved the job output storage workflow by changing the data structure for job processing.

- **New Features**
	- Included template files in the package distribution for enhanced functionality.

- **Chores**
	- Restructured dependencies in the test environment to treat certain tools as top-level dependencies, streamlining package management.

### DIFF
--- a/snakemake/dag.py
+++ b/snakemake/dag.py
@@ -410,7 +410,7 @@ class DAG(DAGExecutorInterface, DAGReportInterface):
             logger.info("Storing output in storage.")
             try:
                 for level in self.toposorted(
-                    list(self.needrun_jobs(exclude_finished=False))
+                    set(self.needrun_jobs(exclude_finished=False))
                 ):
                     async with asyncio.TaskGroup() as tg:
                         for job in level:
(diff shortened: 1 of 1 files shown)


snakemake/dag.py
```diff
@@ -409,25 +409,28 @@ async def store_storage_outputs(self):
...
-
-                        async def tostore(f):
-                            return (
-                                f.is_storage
-                                and f
-                                not in self.workflow.storage_settings.unneeded_temp_files
-                                and await f.exists_local()
-                            )
+                for level in self.toposorted(
+                    list(self.needrun_jobs(exclude_finished=False))
```
johanneskoester, line 413:
```suggestion
                    set(self.needrun_jobs(exclude_finished=False))
```

johanneskoester:
Thanks! Changed to a list since toposorted also uses it for contains queries.

Issue #3355, opened 2025-03-10
introduce localtemp (temp files that shall end up on local storage and lead to auto-grouping like pipe output does)


johanneskoester:
https://github.com/snakemake/snakemake/blob/359ae2e17f31f33ba8059d75239c0e064bbd7520/snakemake/dag.py#L1716

https://github.com/snakemake/snakemake/blob/359ae2e17f31f33ba8059d75239c0e064bbd7520/snakemake/dag.py#L678

DrYak:
Starting to work on PR#3378

DrYak:
PR #3378 is now ready for review.

DrYak:
Note: I am not maried to the "`nodelocal`" name, feel free to rename to something else if needed.

DrYak:
> feel free to rename to something else if needed.

Presentation to end-user changed into a "group_jobs" option of "temp()" as suggested by @johanneskoester

PR #3378, opened 2025-03-11
feat: optionally auto-group jobs via temp files in case of remote execution

DrYak:
issue #3355 

creates a new flag to flag files that are only accessible within a group running on a compute node, and note accessible to the main process (somewhat reminiscent as how pipes and services are grouped together but with regular files and directories on paths that are local to that machin, e.g., that machine's /tmp)

### QC

## Summary by CodeRabbit

- **New Features**
	- Introduced support for node-local storage workflows, enabling local handling of temporary and persistent files.
	- Enabled job grouping for tasks that leverage local storage to optimize performance.

- **Documentation**
	- Added a new section detailing job-grouped temporary files to guide efficient usage of local storage.

- **Refactor**
	- Improved error handling and file flag management to prevent conflicts and ensure smoother operations.


DrYak:
@johanneskoester it works, Peter managed to pin-point the trouble source and [help fix](https://github.com/snakemake/snakemake/pull/3378/files#diff-7898111b777083eaffbb7e986fdf700073f388165a2f752a1477b374eff7ee52R1856-R1858)

DrYak:
Rebased ontop of refactor

johanneskoester (commented):
Sorry for the late review, I have thought  Idea: instead of introducing the new marker name (nodelocal), what about instead adding a keyword argument to `temp` that tells it to infer the group like it happens with nodelocal now. Just basically a different interface bbut same functionality.

```
temp("somefile", group_jobs=True)
```

Seems better than a new marker, as it also shows at the same time that the file will not be preserved after the node group job.

DrYak:
Resolving conflict underway...

DrYak:
> Just basically a different interface bbut same functionality.

Interface updated as requested:
```python
temp("…file…", group_jobs=True)
```
is now the way to put the flag "nodelocal" on a file.

(Under-the-hood logic remains the same, docstring of "temp" explains the internal "nodelocal" flag, main documentation updated).

docs/snakefiles/rules.rst
```diff
@@ -1818,6 +1818,30 @@ Further, an output file marked as ``temp`` is deleted after all rules that use i
 
 .. _snakefiles-directory_output:
 
+Job-grouping temp files
```
johanneskoester, line 1821:
```suggestion
Auto-grouping via temp files upon remote execution
```

docs/snakefiles/rules.rst
```diff
@@ -1818,6 +1818,30 @@ Further, an output file marked as ``temp`` is deleted after all rules that use i
...
+~~~~~~~~~~~~~~~~~~~~~~~
```
johanneskoester, line 1822:
```suggestion
Auto-grouping via temp files upon remote execution
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

johanneskoester (approved):

Issue #730, opened 2020-11-06
configfile does not completely overwrite previous config files


fgvieira:
**Snakemake version**
5.27.4

**Describe the bug**
When there is a config file loaded and another one is loaded through `--configfile`, the config `dict` is updated and not overwritten.

**Minimal example**
With a snakefile:
```
import json
configfile: "config.yaml"
print(json.dumps(config, indent="\t"))
```
and a config file:
```
ref:
  n_chromosomes: 17
  species: saccharomyces_cerevisiae
  release: 100
  build: R64-1-1
  url: ensembl.org
```
you get (with some errors that are not relevant for here):
```
$ snakemake -j 1
{
        "ref": {
                "n_chromosomes": 17,
                "species": "saccharomyces_cerevisiae",
                "release": 100,
                "build": "R64-1-1",
                "url": "ensembl.org"
        }
}
```
but, if you also load this config through the command line:
```
ref:
  n_chromosomes: 17
  species: saccharomyces_cerevisiae
  release: 101
  build: R64-1-1
  name: NEW
```
you get:
```
$ snakemake -j 1 --configfile config_NEW.yaml
{
        "ref": {
                "n_chromosomes": 17,
                "species": "saccharomyces_cerevisiae",
                "release": 101,
                "build": "R64-1-1",
                "name": "NEW",
                "url": "ensembl.org"
        }
}
```

Shouldn't we expect the config to be fully overwritten (and the key `url` should not be there)?
If not and snakemake actually merges config files (instead of overwriting them), is there a way to enforce the old one to be overwritten?

thanks,

sroener:
> Shouldn't we expect the config to be fully overwritten (and the key url should not be there)?

No, as the documentation states:

> Finally, you can use the –configfile command line argument to overwrite values from the configfile statement. Note that any values parsed into the config dictionary with any of above mechanisms are merged, i.e., all keys defined via a configfile statement, or the –configfile and –config command line arguments will end up in the final config dictionary, but if two methods define the same key, command line overwrites the configfile statement.

But I think that an option to overwrite a config file completely could be useful in some cases.

johanneskoester:
Exactly. PRs for an additional flag to modify the behavior in this regard are welcome.

schrins:
As clarification: The desired behavior for such an option is that the config file provided on the CLI (and only this one) should override all configs that are specified in the workflow (and possibly other referenced sub-workflows)? All other config files should be evaluated like they are now?

[...]

PR #3381, opened 2025-03-11
feat: adding --replace-workflow-config to fully replace workflow configs (from config: directive) with --configfile, instead of merging them

schrins:
Adds a new flag `--replace-workflow-config` to change the behavior of config file merging: Instead of extended und updating the values in the config dictionary of the workflow, the config dictionary provided via CLI replaces the workflow dictionary entirely. Keys that are not redefined via the CLI dictionary are not present in the final config.

This addresses the request from #730 

### QC

## Summary by CodeRabbit

- **New Features**
  - Introduced a new command‐line option allowing users to fully replace the workflow configuration with a provided configuration file, ensuring only the specified settings are applied.
  - Enhanced logging now clearly indicates whether configurations are being extended or replaced, providing clearer insights during execution.

### DIFF
--- a/src/snakemake/cli.py
+++ b/src/snakemake/cli.py
@@ -1989,7 +1989,7 @@ def args_to_api(args, parser):
                         config=parse_config(args.config),
                         configfiles=args.configfile,
                         config_args=args.config,
-                        replace_workflow=args.replace_workflow_config,
+                        replace_workflow_config=args.replace_workflow_config,
                     ),
                     storage_settings=storage_settings,
                     storage_provider_settings=storage_provider_settings,
--- a/src/snakemake/cli.py
+++ b/src/snakemake/cli.py
@@ -660,6 +660,16 @@ def get_argument_parser(profiles=None):
         ),
     )
     group_exec.add_argument(
+        "--replace-workflow-config",
+        action="store_true",
+        help=(
+            "Config files provided via command line do not update and extend the config "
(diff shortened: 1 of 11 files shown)


snakemake/settings/types.py
```diff
@@ -348,19 +348,27 @@ class ConfigSettings(SettingsBase):
     config: Mapping[str, str] = immutables.Map()
     configfiles: Sequence[Path] = tuple()
     config_args: Optional[str] = None
+    replace_workflow: bool = False
```
johanneskoester, line 351:
```suggestion
    replace_workflow_config: bool = False
```

snakemake/settings/types.py
```diff
@@ -348,19 +348,27 @@ class ConfigSettings(SettingsBase):
...
         if self.configfiles:
             for f in self.configfiles:
                 update_config(overwrite_config, load_configfile(f))
         if self.config:
             update_config(overwrite_config, self.config)
+        if self.replace_workflow and not overwrite_config:
+            logger.warning(
+                "--replace-workflow-config was used but no config entries are provided via "
+                "command line. This flag will be ignored."
+            )
```
johanneskoester, line 371: Better not warn here. People might want to activate this behavior as their default, and then they would always get a warning if they don't specify a --configfile.

johanneskoester (changes requested):
Nice!

johanneskoester (approved):

PR #3385, opened 2025-03-11
feat: files added in rule params with workflow.source_path will be available in used containers 

SrPeter128:
If files are specified under `params` using workflow.source_path, they are also automatically available in the container. This was achieved by mounting the nextflow source cache in the container by default. This achieves a behavior which is rather natural, as the user probably expects that files which are specified in params using workflow.source_path are also available in the container. 

Due to the fact that the source cache is normally located under /home/<user>/.cache/snakemake/snakemake/source-cache and the container thus gets access to the home directory despite the use of the --no-home flag used in this PR, the documentation explicitly points out that only trustworthy containers should be used. 

### QC

## Summary by CodeRabbit

This release improves container workflow execution and documentation, ensuring a smoother and more secure experience for users.

- **New Features**
  - Enhanced container integration now automatically binds local cache directories to improve workflow performance.
  - Added new test scenarios to validate Snakemake execution with Apptainer deployment.
- **Documentation**
  - Updated guidance on file accessibility within container environments, with added security recommendations to use trusted container images.
- **Tests**
  - Introduced additional test functions to cover various Snakemake execution scenarios.
  - New input and expected output files added for comprehensive testing.


snakemake/deployment/singularity.py
```diff
@@ -127,6 +127,13 @@ def shellcmd(
...
         args += f" --pwd {repr(container_workdir)}"
 
+    # mount the snakemake cache into the container per default so that
+    # params included with workflow.source_path are always mounted in the container
+    if len(args) == 0:
+        args += "--bind --no-home " + os.path.join(
+            get_appdirs().user_cache_dir, "snakemake/source-cache"
+        )
+
     cmd = "{} singularity {} exec --home {} {} {} {} -c '{}'".format(
```
johanneskoester, line 137: but here it binds --home doesn't it? How can it be that this does not already include the cache dir?

SrPeter128: Thanks for the feedback! 

We just had a deeper look and it seems that the cwd is bound as "home" and not the home of the user. So we think binding the source-cache is still needed to make the sourced files accessible in the container. "--no-home" flag is then not needed since "home" is mounted as the current cwd which makes sense. 

We also further tested the capability of singularity to write and see files in the higher level folders and this fails as expected.  We think this makes the use of containers more save then expected and we also implemented a test for that.

We will provide the update in the next commit.

johanneskoester (approved):
Awesome!

PR #3396, opened 2025-03-12
fix: edgecases of source deployment in case of remote execution

jakevc:
Ensures sources are not uploaded on dry run, and they are cleaned up if nothing to be done.

### QC

## Summary by CodeRabbit

- **Bug Fixes**
  - Refined deployment processing to ensure that source uploads occur only during active operations, preventing unintended actions during simulation or when there are no tasks to process.


snakemake/workflow.py
```diff
@@ -1209,7 +1209,7 @@ def execute(
                 and self.remote_execution_settings.job_deploy_sources
                 and not executor_plugin.common_settings.can_transfer_local_files
             )
-            if should_deploy_sources:
+            if should_deploy_sources and not self.dryrun:
```
johanneskoester, line 1212: Why can't we do something like this?
```suggestion
            if should_deploy_sources and not self.dryrun and len(self.dag):
```

johanneskoester: Then we won't even have to deploy the sources if the dag does not contain any needrun jobs.

johanneskoester (changes requested):

jakevc:
I tested this version and it is working as expected!

matthewfeickert:
@jakevc Just rebasing this to resolve conflicts from a PR of mine that was merged earlier.

johanneskoester (approved):

src/snakemake/workflow.py
```diff
@@ -1265,7 +1265,7 @@ def execute(
                 and self.remote_execution_settings.job_deploy_sources
                 and not executor_plugin.common_settings.can_transfer_local_files
             )
-            if should_deploy_sources:
+            if should_deploy_sources and not self.dryrun and len(self.dag):
```
johanneskoester, line 1268: as just discussed, the two criteria should be part of the boolean above already.

johanneskoester (changes requested):

johanneskoester (approved):

Issue #641, opened 2020-09-25
Print shell cmds when using wrappers


fgvieira:
**Is your feature request related to a problem? Please describe.**
When rules are run through a wrapper it is difficult to see which command is actually being run, since the commands are not printed to the screen when using option `-p`.

**Describe the solution you'd like**
Print of shell commands (`-p`) when running a rule through a wrapper.

**Describe alternatives you've considered**
Any other way one could see the commands being run through a wrapper.

vsoch:
Have you tried using “—skip-script-cleanup” to see if it keeps wrapper scripts? And of course running in debug mode?

fgvieira:
I've tried that option but, even though the wrapper scripts are not clean up, they still do not show up:
```
[Mon Oct  5 13:13:13 2020]
rule bam_index:
    input: temp/bams/NA12878_Illumina_HiSeq_Exome_Garvan_fastq_0925201.bam
    output: temp/bams/NA12878_Illumina_HiSeq_Exome_Garvan_fastq_0925201.bai
    log: logs/index/bam/temp/bams/NA12878_Illumina_HiSeq_Exome_Garvan_fastq_0925201.log
    jobid: 19
    reason: Missing output files: temp/bams/NA12878_Illumina_HiSeq_Exome_Garvan_fastq_0925201.bai; Input files updated by another job: temp/bams/NA12878_Illumina_HiSeq_Exome_Garvan_fastq_0925201.bam
    wildcards: prefix=temp/bams/NA12878_Illumina_HiSeq_Exome_Garvan_fastq_0925201
```

I'd like to have the command line from the wrapper printed out, just like when you use `--printshellcmds`.

vsoch:
Can you give me a dummy example using a wrapper (that I can run locally) to reproduce your issue, and then I can take a shot at figuring out either how to find the script, or do the print?

fgvieira:
When a rule is run from wrappers with `-p`, the executed command line is printed:
```
$ snakemake -j 1 --use-conda mapped/a.se.bam -p
rule bwa_bam_se:
    input: reads/a.1.fastq, sai/a.1.sai
    output: mapped/a.se.bam
    log: logs/bwa_bam_se/a.log
    jobid: 0
    wildcards: sample=a

[...]

[...]

PR #3397, opened 2025-03-12
fix: Print shell cmds when using wrappers

schrins:
Addresses the issue mentioned in #641 where the `--printshellcmd` flag does not work for shell commands from wrappers. With this PR, any shell invocations that are passed to snakemake via its `shell` function are now printed on request. In the same manner, shell commands from python scripts that are called via the `script` directive now also print the respective shell commands.

### QC


johanneskoester:
Looks like something went wrong here with the rebasing. @matthewfeickert can you help @schrins?

schrins:
The rebase already happened locally but my testcases fail with the logging changes, so I haven't pushed the current state yet.

matthewfeickert:
> The rebase already happened locally but my testcases fail with the logging changes, so I haven't pushed the current state yet.

Ah okay, so everything worked with your rebase, @schrins? If not please find me.

schrins:
> Ah okay, so everything worked with your rebase, @schrins? If not please find me.

The rebase was fine. I pushed the "broken" state of this PR anyways so that others can see the code. I asked @cademirch for some help with the logger because it does not seem to behave as expected in some place. Once this is resolved and the tests are stable again, I can finalize this PR.

cademirch:
@schrins I pushed some changes to make this work properly with the logger. 

For posterity, the issue was that because scripts are executed in a subprocess that imports snakemake.logging, the shellcmd log was coming from that subprocess in which the logger was not fully configured. To fix this, I've made sure that the logger exported by snakemake.logging is configured with some reasonable defaults, but will be overwritten when logger_manager.setup is called.

Additionally, this highlights some of the difficulty I've had w/ logging and subprocesses. Since we can only communicate with subprocesses via stdin/stdout there isn't a great way for the subprocess (python script or Snakemake instnace) to communicate back with the main snakemake process. I don't think this is worth addressing quite yet as it would be rather substantial work, but just wanting to leave this note here.

fgvieira:
@schrins can you fix the conflicts? It would be really nice to have this feature in `snakemake`!

cademirch:
> @schrins can you fix the conflicts? It would be really nice to have this feature in `snakemake`!

I can try to take a look at this as well. IIRC, one blocker was that we could not get this work in a dry run (when it would be most useful...)

cademirch:
@fgvieira I've looked at this again, and as I noted above, this PR will not print the shell cmd on dry run - only actual execution. Is this useful to you? 

Ideally, we could have both - but from my brief look over again, the difficulty arrises from having to extract any `shell()` calls from a Python `script`, which is all that a `wrapper` is. The solution implemented here by @schrins goes about this in from the other direction, in the `shell` constructor, checking if the call came from a script. 

The fundamental challenge is that during a dry run, the Python scripts/wrappers aren't actually executed, they're just identified as dependencies. 

Let me know what you think.

cademirch:
Okay, I've got this mostly working. Some caveats though:

1. Subprocess execution mode - It's not entirely clear if Python scripts are ever executed in subprocess mode (seems like no?). If they are, logs below `logging.ERROR` will be suppressed, which would probably hide the shell commands we're trying to log.
2. Remote execution -  Script logs will only be visible in the remote Snakemake process's logs, not in the local process that initiated the workflow. Users may need to check remote logs to see the shell commands.

cademirch:
I'll try to get around to the failing tests. Though one caught my eye "test_script_pre_py39" - 3.9 is end of life in a few days funnily enough.

fgvieira:
> @fgvieira I've looked at this again, and as I noted above, this PR will not print the shell cmd on dry run - only actual execution. Is this useful to you?

I think that getting it to print the shell cmd only on execution is already much better than what it is. Specially if getting it to work on a dry-run is a lot of work.

Unless @johanneskoester has some ideas. :smile:

durr1602:
Hi there! Thank you so much for investing time in this, this would be a greatly appreciated functionality, whether it works on dry run and/or execution :)

PR #3398, opened 2025-03-12
fix: 3342 faster touch runs and warning messages for non-existing files

Marco-Masera:
### Description

This PR tries to optimize the --touch runs but running the *sleep* command only for jobs that actually need to touch files.
The *sleep* command is moved into the async *touch* function and run only if there are files to be created.

Sleeping inside this async should block the entire asyncio event loop, which is the desired behavior here as the order of files creation is preserved.

For jobs with non-existing files, a warning messsage is displayed to the user. A new *log_warning* function is added in the Job class.

## Summary by CodeRabbit

- **New Features**
  - Introduced a configurable pause during file processing to enhance operational flexibility.
  
- **Refactor**
  - Streamlined the process for checking and processing output files, ensuring clearer differentiation between file types.
  - Improved logging to notify when certain expected files are not present.


snakemake/executors/touch.py
```diff
@@ -30,22 +30,48 @@
...
+                            for f in job.output
+                            if f not in touch_storage_and_local_files
+                            and await f.exists_local()
+                        ]
+                    )
+                    non_existing_files = (
+                        set(job.output) - touch_storage_and_local_files - touch_files
+                    )
+
+                    if (len(touch_files) + len(touch_storage_and_local_files)) > 0:
```
johanneskoester, line 64:
```suggestion
                    if touch_files or touch_storage_and_local_files:
```

snakemake/executors/touch.py
```diff
@@ -30,22 +30,48 @@
...
+                        for f in touch_storage_and_local_files:
                             await f.touch_storage_and_local()
-                        elif await f.exists_local():
+                        for f in touch_files:
                             f.touch()
+                    if len(non_existing_files) > 0:
+                        job.log_warning(
+                            msg=f"Missing rule's output files: "
+                            + f"{', '.join(non_existing_files)}. --touch skipped for this job."
+                        )
```
johanneskoester, line 74:
```suggestion
                        logger.warning(
                            f"Output files not touched because they don't exist: {', '.join(non_existing_files)}"
                        )
```

johanneskoester (changes requested):
Thanks a lot! Some suggestions below.

johanneskoester (approved):

Issue #1923, opened 2022-10-20
Allow for dynamic module (name) declaration


nesnoj:
Snakemake allows to load modules via the `module` statement which is very helpful.
As far as I've researched, its name cannot be defined dynamically. Example:
```python
for mod_name in ["module1", "module2", "module3"]:
    module mod_name:
        snakefile: "path/to/module/Snakefile"
    use rule * from mod_name as mod_name_*
```
It does not throw an error but 1) "mod_name" is used as module name and 2) the module is simply redeclared in each iteration. If I try to use anything different from a name, e.g. a variable, an error is raised: [`Expected name after module keyword.`](https://github.com/snakemake/snakemake/blob/9c0f1d50616bb3a6872636ac52a25d1ef49bb091/snakemake/parser.py#L929)

It be great to have a more flexible module declaration which allows for dynamic names (and integration into other control flow constructs such as loops), for example

```python
mod_name = "some_module"
module {mod_name}:
    snakefile: "path/to/module/Snakefile"
use rule * from {mod_name} as {mod_name}_*
```

I'm aware that `include` and `module` are quite different in concept and syntax though.

In my use case, I want to search a directory tree for datasets and create modules during runtime. I used `include` like
```python
snakefiles = search_snakefiles()  # gives e.g. ["/path/to/snakefile1.smk", "/path/to/snakefile2.smk"]
for f in snakefiles:
    include: f
```
which worked but is quite limited and has several drawbacks. I'd like to use additional features from module.
BTW, subworkflows show the same limitations.

Does this feature request make sense to you or am I overlooking some other concept?

(not sure if this is a bug or a feature)

fgvieira:
Have been facing the same issue, and my first try was to just write a temp file with the code and `include` it:
```python
fh = open(Path(workflow.basedir) / "Snakefile_modules", "w")
for mod_name in ["module1", "module2", "module3"]:
    code = f"""module {mod_name}:
    snakefile: "path/to/module/Snakefile"
use rule * from {mod_name} as {mod_name}_*
"""
    fh.write(code)
fh.close()
include: "Snakefile_modules"
```

However, this is definitely not ideal. Looking at the code, I feel it should be possible to parse the code:
https://github.com/snakemake/snakemake/blob/0fe0459107db0cd426093eedd7a01410a86839af/snakemake/workflow.py#L1378-L1383

and compile it:

[...]

PR #3401, opened 2025-03-12
feat: Dynamic module name

johannaschmitz:
Allow for dynamic module names
resolves https://github.com/snakemake/snakemake/issues/1923

### QC

## Summary by CodeRabbit

- **New Features**
  - Enhanced modular workflows enable dynamic and semi-dynamic module configuration with flexible aliasing and output management.
- **Documentation**
  - Updated guidelines provide clear instructions on configuring dynamic module imports and rule aliasing, including a new section on "Dynamic Modules."
- **Tests**
  - Expanded test coverage validates diverse module configurations and expected outcomes, ensuring robust and reliable workflows.


src/snakemake/workflow.py
```diff
@@ -2277,16 +2278,31 @@ def userule(
...
+                                modifier = f"{currentframe().f_back.f_globals[name_modifier[:-1]]}*"
+                            else:
+                                modifier = currentframe().f_back.f_globals[
+                                    name_modifier
+                                ]
+                    else:
+                        raise WorkflowError(
+                            "Module {} has not been registered with 'module' statement before using it in 'use rule' statement.".format(
+                                from_module
+                            )
```
johanneskoester, line 2301: Can you add some explanatory comments here? also what kind of fallback this adds (I know now, but I will forget that within the next 2 months ;-))

src/snakemake/parser.py
```diff
@@ -943,6 +966,14 @@ def block_content(self, token):
                     self.has_snakefile = True
                 if token.string == "meta_wrapper":
                     self.has_meta_wrapper = True
+                if token.string == "name":
+                    if self.has_name:
+                        raise self.error(
+                            "Ambiguous module name. Name given after module and name keyword.",
```
johanneskoester, line 972: after module AND name keyword? How?

src/snakemake/parser.py
```diff
@@ -913,10 +924,19 @@ def __init__(self, snakefile, base_indent=0, dedent=0, root=True):
...
+        self.modulename = None
         self.has_name = False
         self.primary_token = None
 
     def end(self):
+        if self.modulename is not None:
+            yield f"name={self.modulename!r}\n"
+        elif not self.has_name:
+            self.error(
+                "A module needs a name.",
```
johanneskoester, line 936: Maybe a bit more details here?

docs/snakefiles/modularization.rst
```diff
@@ -199,6 +199,28 @@ Note that the second use statement has to use the original rule name, not the on
...
+
+.. code-block:: python
+
+    for module, use_as in zip(['module1', 'module2'], ['module1_', 'module2_'])
+        module:
+            name: module
+            snakefile: f"{module}/Snakefile"
+            config: config[module]
+
+        use rule * from module as use_as*
```
johanneskoester, line 214: I think we should explicitly say here that use rule now has the following logic now:
1. it first tries to interpret the given identifier behind from as a module name. In that case, that name is interpreted as the prefix if used behind `as`.
2. If there is no module with that identifier as name, it will try to interpret the identifier as a variable name. In that case, it interprets any identifier used behind the as keyword as variable name as well, which includes eventually trailing underscores. Therefore one has to use a separate variable for those prefixes.

PR #3402, opened 2025-03-12
feat(docker): optimize Dockerfile for caching, security, and efficiency

Alputer:
- Switched to a non-root user (`snakemake`) for improved security.
- Aggregated multiple `RUN` statements to reduce the final image size from 2GB to 1.3GB.
- Moved `ADD . /tmp/repo` to the end to better utilize Docker layer caching.
- Updated permissions to ensure the new user can access required directories.
- Ensured proper ownership of `/opt/conda/envs` and `/tmp/repo` for `snakemake` user.

These changes improve security, reduce build time, and optimize caching.

### QC


Dockerfile
```diff
@@ -1,26 +1,32 @@
...
+WORKDIR /tmp/repo
+RUN chown -R snakemake:snakemake /tmp/repo
 
-RUN micromamba create -q -y -c bioconda -c conda-forge -n snakemake \
-    snakemake-minimal --only-deps && \
-    eval "$(micromamba shell hook --shell bash)" && \
-    micromamba activate /opt/conda/envs/snakemake && \
-    micromamba install -c conda-forge conda && \
-    micromamba clean --all -y
+USER snakemake
```
johanneskoester, line 30: Can this cause any permission problems when mounting user dirs into the container?

johanneskoester (commented):
Fantastic! Please add your name to the list of container image authors :-).

Dockerfile
Alputer: I noticed that the mounted files had the user `UNKNOWN`. I added the `userid` and `groupid` fields, which resolved the permission issues in the mounted directory. The `UID` and `GID` of the user can also be customized via build arguments if necessary for running the containers in environments with specific `UID` and `GID` requirements.

Also, what do you think about removing additional files from `.dockerignore`  which are not really necessary to run snakemake such as `.github` directory, `.gitpod.yml` and `.test_durations` files and some others?

johanneskoester: Yes, removing those files sounds like a good idea.

Regarding UID and GID, would be better if the did not have to rebuild just to use the container in some env with specific requirements. And the problems with UNKNOWN you have seen, on what setup did they appear? K8s, docker, apptainer?

Alputer: Okay, I added additional items to .dockerignore. Let me know if I added too many.

I observed the UNKNOWN user in a Docker container yesterday before I specified the UID and GID parameters in the user-add command.

I tested only the Dockerfile and Apptainer, not Kubernetes. In Apptainer, it seems like the Snakemake user is the owner of every folder, which is what we do not want. I am not very knowledgeable about Apptainer, but I think this might be due to how I configure the Apptainer image build on my side. However, it probably needs more testing for Apptainer at least.

Regarding the UID and GID arguments, I can hard-code the value 1000 in the user-add command if you think that would be better. Currently, they default to 1000 if the arguments are not specified.

johanneskoester: Mhm, uid=1000 is also problematic, because not every outside user will have that uid. I believe that apptainer by default sets everything via the user that calls the container. Seems like having explicit statements in the dockerfile causes problems for apptainer?

Alputer: I agree. From my understanding, Apptainer and Docker handle users differently. In Docker environments, we often use non-root users for enhanced security. However, in Apptainer, users should either match the `host UID` or use the `--fakeroot` option to avoid permission issues.

For the highest compatibility between Docker and Apptainer environments, it seems like it is better to keep the `root user` in the Dockerfile. It is also possible to maintain separate Docker images for Apptainer and Docker use cases but this would add more complexity for end users.

If you're okay with it, I can proceed with switching to the `root user` and removing the `snakemake user` from the Dockerfile. If you have any other suggestions, let me know!

johanneskoester: Sorry for the delay. Yes, sounds reasonable.

johanneskoester: So let us remove the snakemake user.

johanneskoester: And I guess also the UID and GID statements?

PR #3417, opened 2025-03-13
test: Add pytest markers & skips, update docs to reflect changes

jjjermiah:
- Some tests that that are related to s3 storage keep failing locally. I believe these require the S3 credentials (available as secretes in CI) so I've marked them with pytest as `needs_s3` and added a section to the docs regarding this.
    - I could instead have a marker skip automatically but unsure how this works at the moment and dont want to have it start skipping tests in CI/CD. 

- I've also updated the `.test_durations` files since it hasnt been updated in some time, and there are more tests. 

Side note: the docs havent updated due to the src layout not registering in ReadtheDocs. fixed: #3419 

### QC

## Summary by CodeRabbit

- **Documentation**
  - Introduced a new guide on using test markers, providing clear instructions for running or excluding tests based on environment needs.

- **Tests**
  - Revised and reorganized test configurations with adjusted timings and added numerous new test cases.
  - Enhanced test annotations to indicate dependencies on external resources and platform-specific conditions for improved clarity during execution.
  - Added support for macOS in the testing workflow, expanding the environment options.


pyproject.toml
```diff
@@ -134,6 +135,16 @@ cmd = [
 ]
 description = "Run tests only for tests/tests.py"
 
+[tool.pixi.feature.test.target.osx-arm64.tasks.test-simple]
+cmd = [
+  "pytest",
+  "-m",
+  "not (needs_s3 or needs_envmodules)", # needs_s3 is not supported on osx-arm64
+  "-k",
+  "not (test_conda_global or queue)", # queue is buggy on osx-arm64
```
cademirch, line 144: I pulled this and testing locally the queue input dryrun fails for me too (am on ARM tho). Weirdly when I ran the test snakefile directly with Snakemake with --dry-run, the workflow ran. So something funky here. Will investigate.

cademirch (approved):
The queue test failure is weird but I think is unrelated to this. These changes are great, will make testing way easier, especially with the pixi support you added earlier. Great work.

PR #3425, opened 2025-03-14
feat: Allow for dict queries in storage objects

MattMonk:
This is a slightly experimental change so feedback is very welcome!

In LHCb we have a Python package called [apd](https://pypi.org/project/apd/) which allows users to get a list of remote files (XRootD URLs) by specifying values of tags.

It could be cool to have a storage plugin for this, which I made here: https://github.com/snakemake/snakemake-storage-plugin-apd which is a bit of a "meta-plugin" as it would only accept `dict` queries and then use `snakemake-storage-plugin-xrootd` to provide the actual interface to the remote files.

My idea to implement this is for `snakemake-storage-plugin-apd` to have a `get_files()` method which returns a list of remote files for the given `dict` query and then a `base_provider` attribute which is a string to specify the actual storage plugin to use to handle the files -- in this case `xrootd`.

Then in snakemake, if a `dict` is passed as a query to a storage_object it checks that the provider has the `get_files()` method and `base_provider` attribute which can then be used to return a list of storage_objects.

Here's an example of how it would look (which works with the changes in this PR and what is currently in https://github.com/snakemake/snakemake-storage-plugin-apd):

```snakemake
# explicitly set read-only tokens (this is actually the default behaviour)
# any settings specified in the specific instances below override the ones here
storage:
    provider="apd",
    readonly=True,
    keep_local=True

# an instance for a bnoc analysis
storage b2vv:
    provider = "apd",
    working_group = "bnoc",
    analysis = "bdstovvand4h_run3_checks",

# an instance for a b2oc analysis
storage b2dpi:
    provider = "apd",
    working_group = "b2oc",
    analysis = "bd2dpi_run3",
    keep_local=True


VERSION = "v1r1535"
pols = ["magdown", "magup"]

# By default, snakemake will always try to download the remote file
# so use a basic wrapper to prevent that
# easier to configure with https://github.com/snakemake/snakemake/pull/3410
storage_b2vv = lambda x: storage.b2vv(x, retrieve=False)
storage_b2dpi = lambda x: storage.b2dpi(x, retrieve=False)

# Put the keyword arguments you would use with apd into a dict
rule b2vv_rule:
    input:
        lambda w:
            storage_b2vv({
                "version": w.version,
                "polarity": pols if w.polarity == "magboth" else w.polarity,
                "eventtype": w.et
                })
    output: "b2vv_{version}_{et}_{polarity}.flag"
    shell: "touch {output}"


rule b2dpi_rule:
    input:
        lambda w:
            storage.b2dpi({
                "version": w.version,
                "polarity": pols if w.polarity == "magboth" else w.polarity,
                "eventtype": w.et
                })
    output: "b2dpi_{version}_{et}_{polarity}.flag"
    shell: "touch {output}"
```

Which gives you something like this when you run it:

```
$ snakemake --latency-wait 60 -c1 b2vv_v1r1535_13104003_magboth.flag --force -n
Building DAG of jobs...
Job stats:
job          count
---------  -------
b2vv_rule        1
total            1
[Fri Mar 14 08:40:47 2025]
rule b2vv_rule:
    input: root://eoslhcb.cern.ch:1094//eos/lhcb/grid/prod/lhcb/anaprod/lhcb/MC/Dev/NTUPLE.ROOT/00236378/0000/00236378_00000001_1.ntuple.root (retrieve from storage), root://eoslhcb.cern.ch:1094//eos/lhcb/grid/prod/lhcb/anaprod/lhcb/MC/Dev/NTUPLE.ROOT/00236378/0000/00236378_00000002_1.ntuple.root (retrieve from storage), root://eoslhcb.cern.ch:1094//eos/lhcb/grid/prod/lhcb/anaprod/lhcb/MC/Dev/NTUPLE.ROOT/00236385/0000/00236385_00000001_1.ntuple.root (retrieve from storage), root://eoslhcb.cern.ch:1094//eos/lhcb/grid/prod/lhcb/anaprod/lhcb/MC/Dev/NTUPLE.ROOT/00236385/0000/00236385_00000002_1.ntuple.root (retrieve from storage)
    output: b2vv_v1r1535_13104003_magboth.flag
    jobid: 0
    reason: Forced execution
    wildcards: version=v1r1535, et=13104003, polarity=magboth
    resources: tmpdir=<TBD>
Job stats:
job          count
---------  -------
b2vv_rule        1
total            1
Reasons:
    (check individual jobs above for details)
    forced:
        b2vv_rule
This was a dry-run (flag -n). The order of jobs does not reflect the order of execution.
```

### QC

* [x] The PR contains a test case for the changes or the changes are already covered by an existing test case. [I'm not sure it is possible to provide a test case as it requires LHCb credentials to perform a query]
* [x] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake). [will be in the plugin catalog docs]


johanneskoester:
Sorry for the delay. In principle, allowing more complex queries is a reasonable idea. Two thoughts:
1. I think having meta-plugins is not necessary for that.
2. We can offer dataclass-style queries via the plugin interface instead. This way, storage plugins can define their own query dataclasses. They might also allow handling of wildcards and automatic representation as string(s), thereby also representing sets of individual queries.

johanneskoester:
See draft PR for further thoughts: https://github.com/snakemake/snakemake-interface-storage-plugins/pull/76

PR #3427, opened 2025-03-14
feat: initial save write cli command/profile and config info

cademirch:
### QC


ezherman:
Thought: we could add to the issue template a request for the settings file

ezherman:
Note to self: for large config, simply printing the dict to the output file is messy. Find a cleaner way to present the data.

PR #3432, opened 2025-03-14
enable default default prefix for source archive upload

jakevc:
### QC

### DIFF
--- a/src/snakemake/settings/types.py
+++ b/src/snakemake/settings/types.py
@@ -225,6 +225,7 @@ class DAGSettings(SettingsBase):
 class StorageSettings(SettingsBase, StorageSettingsExecutorInterface):
     default_storage_provider: Optional[str] = None
     default_storage_prefix: Optional[str] = None
+    default_storage_protocol: Optional[str] = None
     shared_fs_usage: AnySet[SharedFSUsage] = SharedFSUsage.all()
     keep_storage_local: bool = False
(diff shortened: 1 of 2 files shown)


src/snakemake/settings/types.py
```diff
@@ -241,6 +242,9 @@ def __post_init__(self):
         if self.remote_job_local_storage_prefix is None:
             self.remote_job_local_storage_prefix = self.local_storage_prefix
 
+        if self.default_storage_provider:
+            self.default_storage_protocol = f"{self.default_storage_provider}://"
```
johanneskoester, line 246: This is dangerous. The name of the provider is not always a valid schema name. We should rather extend the storage provider base class to return the desired schema name via a new abstract method.

src/snakemake/workflow.py
```diff
@@ -292,10 +292,11 @@ def upload_sources(self):
                 checksum = hashlib.file_digest(f, "sha256").hexdigest()
 
             prefix = self.storage_settings.default_storage_prefix
-            if prefix:
+            if not prefix or prefix == self.storage_settings.default_storage_protocol:
+                prefix = "s3://snakemake-workflow-sources/"
```
johanneskoester, line 296: s3:// as a default? Is that intentional?

johanneskoester (changes requested):

jakevc:
yes I will mark as draft and continue to think about a more robust implementation

PR #3435, opened 2025-03-14
fix: display group jobs on dryrun

### LINKED ISSUE #3422: Consistent dryrun output for group jobs
When running group jobs, the output with `--dryrun` is different than the one we get when doing a live run - the dry run does not contain information about job grouping. The outputs should match.

johanneskoester:
### Description

Closes #3422 

### QC

## Summary by CodeRabbit

- **New Features**
	- Introduced a new property to evaluate non-local execution and dry run scenarios.
  
- **Bug Fixes**
	- Refined task grouping behavior to ensure accurate assignment based on the updated execution context.

These updates aim to provide more consistent and predictable workflow behavior across varied execution modes.


tadasbar (approved):

tadasbar:
> It appears that if we
> 
>     * supply the `-j` flag,
> 
>     * don't have `-c` supplied and
> 
>     * have job groups defined
> 
> 
> dryrun fails with the following message:
> 
> ```
> WorkflowError:
> Job needs threads=4 but only threads=1 are available. This is likely because two jobs are connected via a pipe or a service output and have to run simultaneously. Consider providing more resources (e.g. via --cores).
> ```
> 
> (adding the `-c1` option fixes the error, contrary to what the error message states)
> 
> In all other cases it works, and is an improvement on the previous dryrun output.

I misunderstood the error message. Supplying the `-c` option fixes the error, as is recommended in the error message.

PR #3439, opened 2025-03-14
Add a check for nodelocal when selecting which files to store

tadasbar:
This change ensures that `temp(, group_jobs=True)` outputs are not uploaded to cloud storage.

### QC

* [ x ] The PR contains a test case for the changes or the changes are already covered by an existing test case.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -451,6 +451,7 @@ class DAG(DAGExecutorInterface, DAGReportInterface):
                                     and f
                                     not in self.workflow.storage_settings.unneeded_temp_files
                                     and await f.exists_local()
+                                    and not is_flagged(f, "nodelocal")
                                 )
 
                             if self.finished(job):


tadasbar:
We get a warning

```
The flag 'temp' used in rule <rulename> is only valid for outputs, not inputs.
```

if we set `temp()` for the same file in section `input:`, not just `output:`.

However, doing it "the right way", and setting `temp()` for a file when it's an output, but not when it's an input, causes the following.

The function `get_missing` in `wait_for_files` in `src/snakemake/io.py` gets called on the same file multiple times (I assume for every appearance in sections `input:` and `output:`), and for some calls `is_flagged(f, "temp")` returns `True`, for others - `False`. This makes the handling of the file based on the flag unreliable.

tadasbar:
Has to rerun tests many times, as pixi kept randomly (?) failing with

```
/home/runner/.pixi/bin/pixi install -e docs --manifest-path pyproject.toml --color always
  thread 'tokio-runtime-worker' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.43.0/src/runtime/task/core.rs:378:22:
  JoinHandle polled after completion
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  Error:   × failed to solve the pypi requirements of 'py312' 'osx-64'
    ├─▶ failed to resolve pypi dependencies
    ╰─▶ Because you require snakemake-interface-storage-plugins>=3.5.0,<4.0 and
        snakemake-interface-storage-plugins==3.4.0, we can conclude that your
        requirements are unsatisfiable.
```

but upon rerun the error would no longer appear.

tadasbar:
Note to self:

Local execution fails when we have `temp(, group_job=True)` - for files in `input:` section we don't have the nodelocal flag, and the `if` statement I've added to `wait_for_files` does not result in an early exit with `None` returned.

Example (aout2.txt is a nodelocal file):

```
Waiting at most 5 seconds for missing files:
gs://tadasbar_scratch/output/snakemake8_test/aout1.txt (missing in storage)
WorkflowError in rule ruleb in file "<...>/Snakefile", line 17:
OSError: Missing files after 5 seconds. This might be due to filesystem latency. If that is the case, consider to increase the wait time with --latency-wait:
gs://tadasbar_scratch/output/snakemake8_test/aout1.txt (missing in storage)
```

Adding the flag `--keep-local-storage-copies` fixes this (took this from the remotely executed command by the executor, as that one completes without the error).

PR #3440, opened 2025-03-14
feature: persist the names of the parameters

DrYak:
names of the params is also stored in the metadata in the same order as the params

### QC


DrYak:
Well, I fixed the [quotes in my test on Windows](https://github.com/snakemake/snakemake/actions/runs/13859945249/job/38785704340?pr=3440#step:9:593), but now a bunch of tests seem to be failing in upstream snakemake v9.0.0 #3449 :sweat_smile:

DrYak:
@johanneskoester: any comment on this implementation, does it look okay?

src/snakemake/rules.py
```diff
@@ -830,6 +830,10 @@ def _apply_wildcards(
                     newitems.append(concrete)
                     if not is_derived and non_derived_items is not None:
                         non_derived_items.append(concrete)
+                        if name is not None:
+                            non_derived_items._add_name(
+                                name
+                            )  # HACK multiple names might be reused
```
johanneskoester, line 836: Why is this needed? Is it related in any way to the changes in persistence.py?

johanneskoester (commented):
Thanks! Question: the params names stored in the metadata are not used so far though apart from being stored, right? What is your plan with them?

src/snakemake/rules.py
johanneskoester: @DrYak?

PR #3441, opened 2025-03-14
feat: enable more flexible use of checkpoint activity without marking…

Hocnonsense:
… rule as checkpoint.

With this feature, output from any rule can just be used as checkpoint.
The disadvantage is that every file in the workflow should be checked for whether it is a checkpoint rule or not, and these additional tests may slow down snakemake.

### QC
* [x] The PR contains a test case for the changes or the changes.
* [x] The documentation (`docs/`) is updated to reflect the changes.

## Summary by CodeRabbit

- **Documentation**
	- Updated the docs to clarify flexible checkpoint usage using a new, simplified syntax.
- **New Features**
	- Enhanced checkpoint handling for more dynamic workflow configurations.
	- Introduced new rules and functions in a sample workflow for processing samples.
- **Tests**
	- Added a new test function to validate the flexible checkpoint functionality.
- **Style**
	- Refined documentation formatting for improved readability.


johanneskoester:
I think this won't happen. The reason is clarity in the code. People should be able to see what is a checkpoint from the workflow definition. Hence, I would rather prefer to keep it like currently is. But nevertheless thanks a lot for bringing up the idea!

PR #3455, opened 2025-03-17
test: plugins as pypi-dependencies to not depend on feedstocks

jjjermiah:
errors like:

```
Error:   × failed to solve the pypi requirements of 'py312' 'win-64'
  ├─▶ failed to resolve pypi dependencies
  ╰─▶ Because you require snakemake-interface-storage-plugins>=3.5.0,<4.0 and snakemake-interface-storage-plugins==3.4.0, we can conclude that your requirements are unsatisfiable.
```

seem to possibly originate from the plugin dependencies that were included for the test suite being sourced as conda packages. This could cause issues since conda package feedstocks need to be manually merged:
https://github.com/bioconda/bioconda-recipes/pull/54568

By moving them to be pypi-dependencies, we then only rely on the pypi packages to be considered for solving.

Since we are not tracking the lockfile, this change will have a negligible effect on solving and wouldnt be noticed anyways. 

### QC

* [NA] The PR contains a test case for the changes or the changes are already covered by an existing test case.
* [NA] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake).


dlaehnemann:
That's a good idea. And I guess that with pixi, it doesn't really matter any more, whether things come from PyPI directly, or via the (bio)conda package, right?

On the other hand, I guess that we do want to test it the way that users will install it eventually. As the [installation recommendation](https://snakemake.readthedocs.io/en/stable/getting_started/installation.html) (and user habits) are still `conda/mamba`, we should probably stick with that for testing, for now? Otherwise, users will end up with this hard-to-track incompatibility, so I guess it is better if we catch those in the CI?

johanneskoester (approved):

PR #3472, opened 2025-03-21
chore: Use data: URLs to work around DefinitelyTyped license discrepancies

musicinmybrain:
See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/50580. While https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/LICENSE and the estree license (https://www.npmjs.com/package/@types/estree?activeTab=code) are both `MIT`, they don’t have the same copyright statement, and the MIT license does require reproducing that.

See also: https://github.com/snakemake/snakemake/commit/55202ba53896109e92fec323dfc8a9f5005a26e9

I’m not sure why unpkg no longer seems to be serving contents of some packages, breaking our previous workaround, but my suggestion is to use [`data:` URLs](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data) in `snakemake/assets/__init__.py` to work around the lack of suitable remote URLs for license text for these two packages.

### QC

* [ ] The PR contains a test case for the changes or the changes are already covered by an existing test case. **N/A**
* [ ] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake). **N/A**

## Summary by CodeRabbit

- **New Features**
  - License information is now embedded directly, ensuring that legal details display reliably without needing external links.
  - Reactivated package entries for `@types-estree` and `@types-clone`, now including their respective license paths.
- **Chores**
  - Removed outdated external references to streamline the handling of license data.

### DIFF
--- a/src/snakemake/assets/__init__.py
+++ b/src/snakemake/assets/__init__.py
@@ -7,6 +7,7 @@ import hashlib
 import importlib.resources
 from pathlib import Path
 from typing import Dict, Optional
+import urllib.parse
 import urllib.request
 import urllib.error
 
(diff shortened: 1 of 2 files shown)


johanneskoester:
I would love to not having to have a copy of those licenses in here. Is there no way to get the correct license from somewhere in their codebase? If it is in the package, it also has to be somewhere in their codebase.

musicinmybrain:
> I would love to not having to have a copy of those licenses in here. Is there no way to get the correct license from somewhere in their codebase? If it is in the package, it also has to be somewhere in their codebase.

I agree with this completely, but I have been entirely unable to find where the license text with `Copyright (c) Microsoft Corporation.` comes from. The string does not exist in the `master` branch of https://github.com/DefinitelyTyped/DefinitelyTyped at all.

On the other hand, it looks like the old unpkg URLs (https://unpkg.com/@types/estree@0.0.50/LICENSE, https://unpkg.com/@types/clone@2.1.1/LICENSE) have started working again, so maybe we can just revert 55202ba53896109e92fec323dfc8a9f5005a26e9 and hope that this was a temporary issue with unpkg’s backend.

johanneskoester:
Closing as #3514 was merged. Thanks a lot!

PR #3481, opened 2025-03-26
fix: Removed output from parse_input

fgvieira:
### QC

## Summary by CodeRabbit

- **Documentation**
  - Enhanced clarity by updating explanations and usage examples, emphasizing how extracted values can now be dynamically used as parameters.

- **Refactor**
  - Simplified internal interfaces by streamlining parameter lists to improve the overall consistency and maintainability of input processing.
  - Updated the checksum parameter assignment to introduce separate checksum computations for improved clarity and functionality.

- **Tests**
  - Improved readability of assertions and updated input parameters to include new test cases and enhanced checksum calculations.

### DIFF
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -585,7 +585,7 @@ The function will return the extracted value and this can, for example, be used 
         output:
             "samples.id",
         params:
-            id=lambda w, input: parse_input(input.samples, parser=extract_id)
+            id=parse_input(input.samples, parser=extract_id)
         shell:
             "echo {params.id} > {output}"
 
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -575,8 +575,7 @@ The parse_input function
 
 The ``parse_input`` function allows to parse an input file and return a value.
 It has the signature ``parse_input(input_item, parser, kwargs)``, with ``input_item`` being the key of an input file, ``parser`` being a callable to extract the desired information, and ``kwargs`` extra arguments passed to the parser.
-The function will return the extracted value.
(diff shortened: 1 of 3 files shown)


docs/snakefiles/rules.rst
```diff
@@ -586,7 +585,7 @@ It can for example be used to extract a value from inside an input file.
         output:
             "samples.id",
         params:
-            id=parse_input(input.samples, parser=extract_id)
+            id=lambda w, input: parse_input(input.samples, parser=extract_id)
```
johanneskoester, line 588: The point of the helper and in particular the input.xyz helpers is to not having to write lambda expressions here. Why would it be needed now?

johanneskoester (changes requested):

docs/snakefiles/rules.rst
fgvieira: The tests have a wildcard to specify the filename, and I can only get it to work with a lambda function.

fgvieira: I am getting an error where `infile` is a `<function RuleItemProxy._deferred_get.<locals>.inner at 0x7f2f2ac5a8e0>` and python throws:
```
>       with open(infile, "r") as fh:
E       TypeError: expected str, bytes or os.PathLike object, not function
```

johanneskoester: Then we have a bug there. It should work with the syntax that we had in the test.

johanneskoester: I have added a commit to fix the issue. In the testcase, it would be good to test that as well. Currently, the testcase requires a more complex lambda because the file argument depends on wildcards, which is probably a special case.

PR #3485, opened 2025-03-26
fix: Restore python rules changes triggering reruns. (GH: #3213)

### LINKED ISSUE #3213: Code changes in python rules no longer trigger re-running of workflows
**Snakemake version**
Since 8.23.0

**Describe the bug**
Ever since #3140 (released in 8.23.0), changing the python code inside a rule doesn't trigger snakemake to rerun the rule. I've reverted to 8.22.0 because I heavily rely on that feature. That PR claims serializing python code consistently across python versions is an unsolved problem. Perhaps the approaches used [here](https://github.com/cedricrupb/code_diff) or [here](https://pypi.org/project/astdiff/) could be employed to bring back support for automatic changes detection?

royjacobson:
Trying to fix #3213 by caching the source code of python rules (using `dis` + `linecache` magic).

I would appreciate any help in how to approach writing a test for this.

cc @corneliusroemer 

### QC

## Summary by CodeRabbit

- **New Features**
	- Enhanced rule processing to capture both shell command and Python function definitions when available.
	- Expanded the internal rule configuration to include additional source code information, supporting improved debugging and documentation.
	- Introduced a method to retrieve and store source details for rules automatically.
	- New workflow defined with rules for generating files based on time-stamped content.
- **Bug Fixes**
	- Added tests to validate the behavior of the new rule processing and ensure expected outputs.

### DIFF
--- a/src/snakemake/persistence.py
+++ b/src/snakemake/persistence.py
@@ -553,11 +553,13 @@ class Persistence(PersistenceExecutorInterface):
 
     @lru_cache()
     def _code(self, rule):
-        # We only consider shell commands for now.
-        # Plain python code rules are hard to capture because the pickling of the code
-        # can change with different python versions.
         # Scripts and notebooks are triggered by changes in the script mtime.
-        return rule.shellcmd if rule.shellcmd is not None else None
+        # Changes to python and shell rules are triggered by changes in the plain text.
+        if rule.shellcmd is not None:
+            return rule.shellcmd
+        if rule.run_func_src is not None:
+            return rule.run_func_src
+        return None
 
     @lru_cache()
     def _conda_env(self, job):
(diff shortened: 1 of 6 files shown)


johanneskoester:
I guess having a testcase with a rule that has a run directive, which first runs the workflow with `cleanup=False` (an argument of the test suites run function), and then a second run that tries to run again should be a start. There one would have to use python's MagickMock to hook into the dag in order to check that no rerun is triggered. Then, in a second rerun, one could first alter the run directive code trivially via some replace command and then check via the same way if a rerun is triggered.

Also, should be assured that the new approach really captures the function code and not lines above or below the actual function.

royjacobson:
@johanneskoester friendly ping :)

royjacobson:
@johanneskoester ?

royjacobson:
Hi @johanneskoester, is there anything I could help with to move this PR forward?
Thanks!

johanneskoester:
Sorry for the delay, there is a huge backlog to catch up with!

johanneskoester:
Best is always to ping me via discord.

PR #3492, opened 2025-03-28
fix: rerun-trigger regression

### LINKED ISSUE #3491: Regression in version 9: Changes to "shell" directive content don't trigger rerun anymore
**Snakemake version**
9.1.1

**Describe the bug**
To my surprise, changes to shell directive content don't seem to trigger reruns anymore.

**Minimal example**

1. Set up Snakefile like this:
```
rule a:
    output:
        "a.txt",
    shell:
        "echo a > {output}"
```
2. Run `snakemake`
3. Edit the content of `shell` directive to `echo b > {output}`
4. Run `snakemake` again

Expected result: Rule reruns
Actual result: Nothing to be done

I installed snakemake v8.30.0 and that one shows expected behavior. So it seems to be a regression in v9 (or if it's intended behavior then it's a breaking change of v9 that isn't documented as far as I can tell)

cademirch:
fixes #3491

Edit: This is reverting 4430e23, which changed the default for rerun-trigger for the docs. Presumably because `RerunTrigger.all()` gives the actual enum values which don't render nicely. 

Will need to find a better solution. Any thoughts @fgvieira?

### QC

## Summary by CodeRabbit

- **New Features**
  - Updated the CLI default so that all available triggers for job reruns are enabled by default, ensuring a more comprehensive and consistent behavior that aligns with workflow configurations.

### DIFF
--- a/src/snakemake/cli.py
+++ b/src/snakemake/cli.py
@@ -721,7 +721,7 @@ def get_argument_parser(profiles=None):
         "--rerun-triggers",
         nargs="+",
         choices=RerunTrigger.choices(),
-        default=RerunTrigger.choices(),
+        default=RerunTrigger.all(),
         parse_func=RerunTrigger.parse_choices_set,
         help="Define what triggers the rerunning of a job. By default, "
         "all triggers are used, which guarantees that results are "


fgvieira:
But why doesn't `RerunTrigger.choices()` trigger the re-runs? It should be the same as `RerunTrigger.all()`, no?

cademirch:
> But why doesn't `RerunTrigger.choices()` trigger the re-runs? It should be the same as `RerunTrigger.all()`, no?

.choices() returns list[str] whereas .all() returns list[RerunTrigger]. 

See this comment https://github.com/snakemake/snakemake/issues/3491#issuecomment-2762522276

fgvieira:
But that is a string when set by the user, so a list[str] should also work as `choices`, no? Unless it is parsed by

https://github.com/snakemake/snakemake/blob/17f0d9b1edd321a8470d1552ead314532f2b3910/src/snakemake/cli.py#L725

and the defaults are not...

cademirch:
That’s a good question. User supplied strings are parsed into RerunTrigger.
I guess the defaults are not?

johanneskoester (approved):

johanneskoester:
Merging this for now to avoid the regression

PR #3493, opened 2025-03-29
fix: improve xonsh script support

jsquaredosquared:
Improved Xonsh script support so that:
- an error is raised if the script exits with a non-zero exit status
- a traceback is shown if an exception occurs in the script

### QC

## Summary by CodeRabbit

- **New Features**
	- Enhanced script execution with improved error reporting and debugging support.
	- Added support for dynamic output generation using external tool information.
- **Chores**
	- Added a new dependency to support updated script tests.


src/snakemake/script/__init__.py
```diff
@@ -1603,8 +1603,27 @@ def execute_script(self, fname, edit=False):
...
+        fd.write(preamble.encode())
+
+        if self.conda_env:
+            conda_xonsh_preamble = textwrap.dedent(
+                f"""
+                    # Workaround for [this issue](https://github.com/conda/conda/issues/7980)
+                    __xonsh__.execer.exec($("$CONDA_EXE" "shell.xonsh" "hook"))
+                    conda activate {self.conda_env}
+                    """
+            )
```
johanneskoester, line 1616: This I don't get. Why do we need an activation of the env in xonsh? Snakemake would activate the env before calling xonsh already.

johanneskoester (commented):
Thanks a lot! One question below.

src/snakemake/script/__init__.py
johanneskoester: i.e. in the shell command that executes the statement returned by execute_script

jsquaredosquared: You are correct, it is not necessary.

I added this because previously I had an issue where if I ran the Xonsh script on the HPC cluster, xonsh would not recognize the executables in the conda environment if the conda environment was activated outside the script.

But now I am able to get it to work so this part is no longer necessary. I have removed it.

I have also modified the test to use bcftools, to make sure the xonsh script works with the conda environment.

Thanks :D

johanneskoester:
Thanks!

PR #3511, opened 2025-04-03
fix: restrict to pulp <2.8 for now because of API breakage

Pablo-Caballero-Maria:
With pulp>=2.8, snakemake fail, yielding the error: AttributeError: module 'pulp' has no attribute 'list_solvers'
### QC


johanneskoester (approved):

musicinmybrain:
The API was simply renamed from `pulp.list_solvers` to `pulp.listSolvers` in PuLP 2.8. Wouldn’t it make more sense just to patch the call site(s) to support both names?

musicinmybrain:
(On the other hand… where *is* `pulp.list_solvers()` called?)

johanneskoester:
Right, we already call listSolvers.

johanneskoester:
Maybe this can simply be closed? Or do we have to restrict to >2.8?

Issue #3221, opened 2024-11-27
Wildcard values modified in some circumstances


irr-github:
**Snakemake version**
8.25.2

**Describe the bug**
If a wild card in a rule input is directly preceded by a "/" its value will change and a "/" prepended to it. This causes breaks. 

**Logs**
NA

**Minimal example**
```
    rule plot_network:
        input:
            network=  base_results_dir +   "/{heating_demand}_steady_state"
```

**Additional context**
This can of course be solved by proper paths handling (os.paths.join) but as there are no clear warnings, it currently causes pain.

Issue #2981, opened 2024-07-22
Python rules do not support single quotes or spaces in path


Warggr:
**Snakemake version**
8.16.0

**Describe the bug**
When the path contains either single quotes `'` or spaces, Python rules (i.e. `run:`) fail.

**Logs**
The error message is
```
/bin/sh: -c: line 1: unexpected EOF while looking for matching `''
```
when the path contains a single quote and when the path contains a space, e.g. `/tmp/users\ directory`:
```
Error: Snakefile "/tmp/users" not found.
```

**Minimal example**
```sh
mkdir "user's folder" && cd user\'s\ folder
cat > Snakefile <<EOF
rule:
    run:
        print("hello, world")
EOF
snakemake
```
Output:
```
Assuming unrestricted shared filesystem usage.
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 4
Rules claiming more threads will be scaled down.
Job stats:
job      count
-----  -------
1            1
total        1

Select jobs to execute...
Execute 1 jobs...

[Mon Jul 22 10:30:59 2024]
localrule 1:
    jobid: 0
    reason: Rules with neither input nor output files are always executed.
    resources: tmpdir=/tmp

/bin/sh: -c: line 1: unexpected EOF while looking for matching `''
[Mon Jul 22 10:30:59 2024]
Error in rule 1:
    jobid: 0

Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: .snakemake/log/2024-07-22T103059.309900.snakemake.log
WorkflowError:
At least one job did not complete successfully.
```

**Additional context**
- shell rules (i.e. `shell:` instead of `run:`) work - so it's possible to just run `shell: "python -c '...'"`

cmeesters:
Your minimal example works like a charm for me. Are you trying to trigger Snakemake in a more elaborate shell script? Could it be that you are dealing with nested quotes in a shell script?

Warggr:
Thanks for looking into this.

I am not using a shell script, just executing `snakemake` from the terminal.
My default shell is zsh; I thought that could maybe be the problem. But when I try running `snakemake` from Bash, or with `SHELL=bash`, or both, the same problem occurs. In any case, the error message says something about `/bin/sh`, which for me is a symbolic link to `/bin/bash`.

To make sure that the problem was not my environment, I tried running the minimal example in a Docker container, with the following Dockerfile:
```Dockerfile
FROM alpine:3.20.2

[...]

PR #3521, opened 2025-04-08
fix: Refactor of format_job_exec related to issue 2981, 3221

tbooth:
This started off as a simple fix for some quoting issues, but it's turned into a substantial re-write of the whole way Snakemake invokes itself. As it stands, this will break all the executor plugins 😱, so I'm presenting it here looking for feedback before I put any more work in. I'm aware that there are still a few failing tests.

When looking into issues #2981 and #3221 I came to the the conclusion that the whole idea of `executor.format_job_exec(job)` called in `executors.local.spawn_job()` was in need of refactoring. You can see [my comment here](https://github.com/snakemake/snakemake/issues/2981#issuecomment-2718384209).

The central change is not actually in this repo, it's the replacement of several related functions in `snakemake-interface-executor-plugins/snakemake_interface_executor_plugins/utils.py` with a new [ShellRunner](https://github.com/EdinburghGenomics/snakemake-interface-executor-plugins/blob/issue_2981_3221/snakemake_interface_executor_plugins/utils.py) class. The idea is that an instance of this class captures everything needed when Snakemake invokes itself. For local execution, calling `my_shell_runner.check_call()` bypasses the need for Bash entirely and therefore the issues with shell escaping go away. For remote calls, where the command must be written to a script, the commands are quoted by `quote_command()` which puts all the quoting logic in one place and uses `shlex.quote()` consistently. Other parts of the code no longer have to worry about Bash/Dash syntax.

As far as I can see this is mostly working well, and to my mind is both robust and tidier than before. I've eliminated the use of base64 encoding for all arguments as (I think !?) `shlex.quote()`, applied consistently, is sufficient.

I've not tested this on Windows but the new ability of Snakemake to invoke a local instance or itself without involving any sort of shell should be good for that platform.

One thing I changed was the encoding of `--target-jobs` where commas within wildcard values can confuse the parser on how to split up the wildcard list, and we had some funky logic to try and deal with that. My solution here was to use JSON encoding for the wildcards dictionary in all cases. I added some unit tests specifically for this.

I would suggest considering bringing the *snakemake-interface-executor-plugins* code back into this git repository. The code in this module seems so intimately connected to the core of Snakemake that I had to make changes in lock step, and I had to fudge the dependencies in `pyproject.toml` to convince `pixi` to run any tests. However, I've not attempted to move any code around as part of this PR.

Any comments?

### QC

* [ N/A ] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake).


cademirch:
> I would suggest considering bringing the _snakemake-interface-executor-plugins_ code back into this git repository. The code in this module seems so intimately connected to the core of Snakemake that I had to make changes in lock step, and I had to fudge the dependencies in `pyproject.toml` to convince `pixi` to run any tests. However, I've not attempted to move any code around as part of this PR.

Still need to look through the changes here, but I strongly agree with this. Perhaps git submodules is a in between option? Though I'm not familiar.

johanneskoester:
First of all thanks a lot for having a look into this. This is a dramatic change that needs very careful investiation. Unfortunately I don't have the capacity to do that at the moment. I still have to merge some remaining PRs from the hackathon and in parallel I am actually on a longer vacation, _and_ have to finalize several other collaborative projects unrelated to snakemake. Could you contact me again about this in June, ideally via discord?

Regarding the interface package, one could move it into the main repo but it should definitely stay a separate package. I would prefer though to rather improve it to limit the connections to main snakemake further than to move back to a single repo.

tbooth:
Hi @johanneskoester thanks for the comments and yes I'm happy to park this until June and then have a discussion on discord. Enjoy your vacation, assuming you get some actual vacation time between the work!

PR #3522, opened 2025-04-08
fix: Resolve issues with hashlib.md5() on systems in FIPS mode

### LINKED ISSUE #3504: hashlib.md5() is not available on FIPS-compliant systems
**Snakemake version**

All recent versions, including v9.1.5 and the main branch

**Describe the bug**

Snakemake uses `hashlib.md5()` in multiple places.

- https://github.com/snakemake/snakemake/blob/3be8ca9308df72494cbca4d6f572486c8486418a/src/snakemake/deployment/env_modules.py#L25
- https://github.com/snakemake/snakemake/blob/3be8ca9308df72494cbca4d6f572486c8486418a/src/snakemake/persistence.py#L530
- https://github.com/snakemake/snakemake/blob/3be8ca9308df72494cbca4d6f572486c8486418a/src/snakemake/deployment/conda.py#L238
- https://github.com/snakemake/snakemake/blob/3be8ca9308df72494cbca4d6f572486c8486418a/src/snakemake/deployment/singularity.py#L48
- https://github.com/snakemake/snakemake/blob/3be8ca9308df72494cbca4d6f572486c8486418a/tests/common.py#L44

However, the MD5 algorithm is not an algorithm supported by the [Federal Information Processing Standards (FIPS)](https://en.wikipedia.org/wiki/Federal_Information_Processing_Standards); it is [not considered secure](https://en.wikipedia.org/wiki/MD5#Overview_of_security_issues). Snakemake will therefore fail on a system with FIPS mode enabled. I am working with a user who is getting the error message

```
md5hash = hashlib.md5()
          ^^^^^^^^^^^^^

ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
```

from persistence.py on a system with FIPS mode enabled. The Python documentation (see https://docs.python.org/3/library/hashlib.html) says that

> [md5()](https://docs.python.org/3/library/hashlib.html#hashlib.md5) is normally available as well [in addition to other algorithms], though it may be missing or blocked if you are using a rare “FIPS compliant” build of Python.

A solution that will work for any version of Python—including versions built for FIPS compliance—is to use a different algorithm for calculating hashes, such as SHA-256. Alternatively, if the hash algorithm is not used for security features, it's still possible to use it by setting `usedforsecurity=False` when calling `hashlib.md5()`.

**Logs**

```
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
```

will be raised whenever `hashlib.md5()` is called.

**Minimal example**

Attempt to use Snakemake on any system with FIPS mode enabled.

**Additional context**

MD5 is supported in FIPS mode if the `usedforsecurity` keyword argument is set to `False`:

```
>>> import hashlib
>>> hashlib.md5()
  File "<stdin>", line 1, in <module>
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
>>> hashlib.md5(usedforsecurity=False)
<md5 HASH object @ 0x148eafc44170>
```

SHA-256 is always supported:

```
>>> import hashlib
>>> hashlib.sha256()
<sha256 HASH object @ 0x151e25bfb7d8>
```

robbenmigacz:
This change adds `usedforsecurity=False` to calls of `hashlib.md5()` to resolve issues with the availability of the MD5 hash algorithm on systems with (United States) Federal Information Processing Standards (FIPS) mode enabled. Snakemake does not work on systems with this security standard because the MD5 algorithm is considered unsafe. For more information, please see #3504.

Resolves #3504

### QC

## Summary by CodeRabbit

- **Chores**
  - Refined several internal hashing routines to clearly indicate that they are not used for cryptographic security, ensuring more consistent behavior across the system.
- **Tests**
  - Updated test validations to align with the revised internal configuration.

No changes have been made to the public interfaces, so you should experience uninterrupted functionality.

### DIFF
--- a/src/snakemake/deployment/conda.py
+++ b/src/snakemake/deployment/conda.py
@@ -235,7 +235,7 @@ class Env:
         return self._content_hash
 
     def _get_hash(self, include_location: bool, include_container_img: bool) -> str:
-        md5hash = hashlib.md5()
+        md5hash = hashlib.md5(usedforsecurity=False)
         if self.name:
             md5hash.update(self.name.encode())
         elif self.dir:
--- a/src/snakemake/deployment/env_modules.py
+++ b/src/snakemake/deployment/env_modules.py
@@ -22,7 +22,7 @@ class EnvModules:
 
     @property
     def hash(self) -> str:
-        md5hash = hashlib.md5()
+        md5hash = hashlib.md5(usedforsecurity=False)
         for name in self.names:
             md5hash.update(name.encode())
         return md5hash.hexdigest()
--- a/src/snakemake/deployment/singularity.py
+++ b/src/snakemake/deployment/singularity.py
@@ -45,7 +45,7 @@ class Image:
 
     @lazy_property
     def hash(self):
-        md5hash = hashlib.md5()
+        md5hash = hashlib.md5(usedforsecurity=False)
         md5hash.update(self.url.encode())
         return md5hash.hexdigest()
 
--- a/src/snakemake/persistence.py
+++ b/src/snakemake/persistence.py
@@ -531,7 +531,7 @@ class Persistence(PersistenceExecutorInterface):
     def _software_stack_hash(self, job):
         # TODO move code for retrieval into software deployment plugin interface once
         # available
-        md5hash = hashlib.md5()
+        md5hash = hashlib.md5(usedforsecurity=False)
         if (
             DeploymentMethod.CONDA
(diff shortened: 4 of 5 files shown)


tbooth (approved):
This change does not impact the functionality of the code, and the use of md5sums here is not for security purposes, so the use of this flag is reasonable. No need for further tests.

johanneskoester (approved):

PR #3524, opened 2025-04-10
Update basics.rst

paulkorir:
It appears that the CLI `--dag` now takes arguments, which is omitted in the documentation. I've added it in the example.

### QC


johanneskoester (approved):

johanneskoester:
Thanks, good catch!

PR #3531, opened 2025-04-14
fix: restore accidentally changed behavior of --cores. If not specified, Snakemake now complains again, asking for --cores to be specified (of --jobs in case of remote exec). As before, if a default is desired, it can be easily set via a profile

johanneskoester:
### QC

## Summary by CodeRabbit

- **Refactor**
  - Updated the logic for assigning CPU cores, removing the automatic fallback to available CPU count when not explicitly set.
- **Bug Fixes**
  - Improved error handling to avoid raising null exceptions after failed executions.
- **Tests**
  - Modified test commands to explicitly specify single-core usage during runs.

### DIFF
--- a/src/snakemake/cli.py
+++ b/src/snakemake/cli.py
@@ -1964,9 +1964,6 @@ def args_to_api(args, parser):
             args.cores = 1
             args.jobs = None
 
-    if args.cores is None:
-        args.cores = available_cpu_count()
-
     # start profiler if requested
     if args.runtime_profile:
         import yappi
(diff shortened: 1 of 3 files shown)


peterjc:
Just got caught out by this change with the (new to me) error:

`ERROR    snakemake.logging:exceptions.py:181 Error: cores have to be specified for local execution (use --cores N with N being a number >= 1 or 'all')`

Easy to work arround with an explicit `--cores all`, and the change does seem reasonable. I am commenting here mainly to get the error message in your issue tracker to help anyone else.

PR #3538, opened 2025-04-16
docs: rewrite the mpi job execution section

dlaehnemann:
This pull request tries to make the explanations in this section a bit clearer and provides a number of useful linkouts, especially to the Open MPI docs, but also to the snakemake executor plugins. Along the way, it also fixes the syntax of existing links.

### QC

## Summary by CodeRabbit

- **Documentation**
  - Improved and expanded the MPI support documentation for Snakemake workflows, including clearer explanations, updated examples, and guidance on customizing MPI launcher commands.
  - Added references to external MPI and cluster documentation for further assistance.
  - Enhanced instructions for overriding MPI resources and passing additional parameters.

### DIFF
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -3095,26 +3095,10 @@ Importantly, they are also "namespaced" per module, meaning that ``inputflags`` 
 MPI support
 -----------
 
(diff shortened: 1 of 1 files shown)


dlaehnemann:
@coderabbit The commands `mpirun` and `mpiexec` are identical in Open MPI, the mpi implementation we link to in the docs. The recommended synonym to use is `mpirun`, which they also use in their documentation. So this is what we should stick to here and there is no need to mention `mpiexec`.

Issue #3558, opened 2025-05-02
Logging broken yet again


lczech:
**Snakemake version**
9.3.3

**Describe the bug**
This is an extension of my earlier bug report [here](https://github.com/snakemake/snakemake/issues/2974), which also seems to still be unresolved.

**Minimal example**

I've tried to adapt to the new logging setup now, but unfortunately, it seems even more broken than before.

```
import logging
logger_manager.logger.addHandler(logging.FileHandler("my.log"))

print("Message: print")
logger.info("Message: logger.info")
logger.warning("Message: logger.warning")


rule all:
    output:
        touch("result.txt")     
```

**Logs**
This behaves differently depending on whether an additional logger is added or not (first two lines).

Without extra log file: The terminal output only prints the "print" and "warning", but not the "info". The log file in `.snakemake/log` contains none of the three messages.

With the extra log file: The "warning" disappears from the terminal output, and instead appears in `my.log` now. The "print" remains in the terminal, the "info" still appears nowhere.

How can I add custom log messages in the new snakemake 9 that actually appear consistently? I would like for my custom info and warning log messages to show up in the terminal _and_ in the log file. Right now, I'm adding the additional print to everything to make messages appear in the terminal, and I am promoting everything to a warning, so that it might at least show up in the snakmake log as well. Not ideal. 

I'd appreciate any hints on how to either fix this with a workaround, or other suggestions! Thanks in advance :-)

lczech:
It gets a bit weirder: The output of `print` in my production pipeline is not printed to terminal on the first "fresh" run of snakemake - only on subsequent runs. I cannot reproduce this in the minimal example above though, but something weird seems to be going on there.

cademirch:
I'll try to take a look at this this week.

cademirch:
I'm not sure what to think about the printing issue, I don't fully understand it as you've described here so perhaps an example will be helpful.

You observation is in part why the logging is a bit complex (more than I'd like it to be), the logger manager mainly is there for managing the state of the logger. Because Snakemake launches itself in subprocesses, we need to suppress logs coming from those subprocesses using the setting `ExecMode`. This is where adding handlers via the workflow is tricky. I need to look deeper but I'm not sure if ExecMode is set in the WorkflowAPI. We could potentially setup the logger there if it is, but there could be other side effects in doing so... 

For my curiosity, I'm curious what kind of information you're logging in your workflow that helps with debugging? Is it more related to workflow configuration?

[...]

PR #3571, opened 2025-05-13
fix: logging refinements

cademirch:
This PR refines some of the changes from the logging refactor to tackle some issues as well as simplify the code.

Changes:
## `ExecMode` added to `OutputSettings`:
- ExecMode is critical to setting up the logger and controls how logs are written, so from this perspective I think it makes sense that ExecMode fits here
- Moving ExecMode here allows the logger to be setup in SnakemakeAPI, which has multiple positive effects: the logger only needs to be setup once in all of Snakemake's lifecycle, the logger is setup before the workflow, allowing for workflow files to use the logger (#3558), and the logger does not need to manage state anymore.
- This could be seen as a breaking change, though I've set the default for this field to be ExecMode.DEFAULT so  existing callers would be unaffected. 

## Setup logger no longer dependent on `executor_plugin.common_settings.dryrun_exec`
- Another reason the logger had to be setup late in the Snakemake lifecycle was that the stdout setting was dependent on this executor plugin setting
- During refactoring I didn't question this, but now after looking at most of the executor plugins, I haven't seen any that use this, so I'm not sure the rationale behind tying stdout to this. @johanneskoester do you happen to know?

## QueueHandler only used for plugins
- Only spin up queue overhead for logger plugins if they are being used. 

## Added more tests
- Added test for specifiyng logger in Snakefile

### QC

## Summary by CodeRabbit

* **New Features**
  * Added configurable output settings to control log level overrides, plugin handler usage, and file logging.
  * Added comprehensive logging tests and supporting test workflows to validate logging behavior and failure scenarios.

* **Bug Fixes**
  * Centralized and stabilized logger initialization for more consistent logging across runs.

* **Refactor**
  * Simplified logging setup by removing mode-based branching and consolidating output-settings creation and handler management.

* **Documentation**
  * Documented the logging module and centralized logger management.

* **Chores**
  * Removed an outdated logfile test and updated test coverage.


src/snakemake/settings/types.py
```diff
@@ -402,6 +402,7 @@ class OutputSettings(SettingsBase, OutputSettingsLoggerInterface):
     keep_logger: bool = False
     stdout: bool = False
     benchmark_extended: bool = False
+    mode: ExecMode = ExecMode.DEFAULT
```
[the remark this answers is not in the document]
johanneskoester: Another alternative: what if we don't set the execMode here but instead offer settings for the individual logger behavior that so far is triggered by the exec mode. Then, whoever calls the API would have full control, and we don't have that duplication of settings. When the CLI calls the API, it could then set those settings according to exec mode (no need to expose them to command line flags). And their default could be the current ExecMode.DEFAULT behavior.

cademirch:
Thanks for the review @johanneskoester, will try to address this weekend.

johanneskoester:
Is this ready to merge?

johanneskoester (approved):

cademirch:
> Is this ready to merge?

Almost! I'm going to add some more tests and finally write some docs on the logging.

cademirch:
Replaced the boolean log_errors_only flag with settings:

`log_level_override`: Allows setting any log level (not just ERROR)
`skip_plugin_handlers`: Controls whether to use plugin handlers/queue infrastructure (for remote execution)

This gives us better control over logging behavior based on execution mode:

Subprocess mode: Sets log level to ERROR (currently), skips file logging
Remote mode: Skips plugin handlers/queue for simpler logging
Main process: Full logging with plugins and all features

Eventually I'd like to clean up the output when a subprocess job fails as its quite verbose. But that is not very high priority.

src/snakemake/settings/types.py
```diff
@@ -403,12 +403,11 @@ class OutputSettings(SettingsBase, OutputSettingsLoggerInterface):
...
-    log_errors_only: bool = (
-        False  # Only log errors. Sets log level to ERROR. Typically used for subprocces mode.
-    )
-    enable_file_logging: bool = (
-        True  # Create and write to logfile. Generally False for remote/subprocess executors.
+    log_level_override: Optional[int] = (
+        None  # Override log level (e.g., ERROR for subprocess)
     )
+    skip_plugin_handlers: bool = False  # Skip plugins/queue (remote mode)
+    enable_file_logging: bool = True
```
johanneskoester: If you would keep the old log_errors_only as well, then we can avoid a breaking change in the API here.

johanneskoester (changes requested):

src/snakemake/settings/types.py
johanneskoester: Wrong, log_errors_only was only an intermediate new attribute in this PR. No breaking change here.

PR #3597, opened 2025-05-21
fix: expandvars for special profile keys

### LINKED ISSUE #3592: Not all values in profile config support environment variables
[Snakemake's documentation](https://snakemake.readthedocs.io/en/v9.4.0/executing/cli.html) is explicit about supporting the environment variables like `$USER` in profiles:
   > Values in profiles can make use of globally available environment variables, e.g. the \$USER variable. For example, the following would set the default prefix for storing local copies of remote storage files to a user specific directory
   > 
   > ```YAML
   > local-storage-prefix: /local/work/$USER/snakemake-scratch
   > ```
   > 
   > Any such environment variables are automatically expanded when evaluating the profile.

Unfortunately, it does not appear that environment variables are expanded for all profile values. Specifically, `$USER` is not expanded when part of the `tmpdir` resource value.

Not sure whether this should be considered a bug or a documentation issue.

----------------

# Example

Tested with Snakemake version 9.3.3.

workflow/Snakefile

```
rule all:
    input:
        "out.txt",

rule rule1:
    output:
        "out.txt",
    shell:
        """
        echo "USER: $USER" >> "{output}"
        echo "TMPDIR: $TMPDIR" >> "{output}"
        """
```

workflow/profiles/default/config.yaml

```
executor: local # or slurm
jobs: 2
latency-wait: 10
printshellcmds: True
verbose: True
default-resources:
    runtime: 5 # minutes
    mem_mb: 1000
    tmpdir: /central/scratch/$USER
```

Command: `snakemake --snakefile workflow/Snakefile --workflow-profile workflow/profiles/default`

Output (out.txt)

```
USER: bentyeh
TMPDIR: /central/scratch/$USER
```

The failure for $USER to be expanded results in the literal string `'/central/scratch/$USER'` to be used as the value of the `$TMPDIR` environment variable within a rule. This is the result for both `local` and `slurm` executors.

----------------

Is there a recommended way for users to configure the temporary directory used by a workflow? I thought that the profile would be the most appropriate place, but since that does not appear to work (for the moment), I could expose a variable in config/config.yaml for the user to set. But then each rule in the Snakefile would need to set the `tmpdir` resource, since it does not appear that the workflow profile `value` for `default-resources: {tmpdir: <value>}` has access to variable values in the Snakefile.

cademirch:
Fixes #3592 

Need to still add tests.

### QC

## Summary by CodeRabbit

- **Bug Fixes**
  - Environment variables within profile configuration values are now automatically expanded, ensuring correct values are used in configurations.

### DIFF
--- a/src/snakemake/profiles.py
+++ b/src/snakemake/profiles.py
@@ -26,7 +26,7 @@ class ProfileConfigFileParser(YAMLConfigFileParser):
             )
 
         def format_one_level_dict(d):
-            return [f"{key}={val}" for key, val in d.items()]
+            return [f"{key}={os.path.expandvars(str(val))}" for key, val in d.items()]
 
         def format_two_level_dict(d, item: str):
             if not all(isinstance(val, dict) for val in d.values()):
@@ -34,7 +34,7 @@ class ProfileConfigFileParser(YAMLConfigFileParser):
                     f"Invalid {item} format in profile. Expected two-level mapping, got {d}"
                 )
             return [
-                f"{key}:{key2}={val2}"
+                f"{key}:{key2}={os.path.expandvars(str(val2))}"
                 for key, val in d.items()
                 for key2, val2 in val.items()
             ]


johanneskoester (approved):

cademirch:
> Good idea! Maybe the profile docs should also be extended to explain that $-variables will be expanded.

It is already documented, as pointed out in https://github.com/snakemake/snakemake/issues/3592#issue-3078111717. The issue fixed here is to make sure that `os.expandvars` is called on all profile values.

PR #3604, opened 2025-05-27
refactor: Minor Logging Improvements

ChristofferCOASD:
I am using SnakeMake with wildcards for a very large number of files. I observed that executimg `snakemake --quiet rules` would still leave blank lines corresponding to the space where the rules would have been.

image

I am not fully fammiliar with the snakemake logging setup but i narrowed down the problem to the `emit` mehod of `ColorizingTextHandler`. My solution was to check for an empty `formatted_message`.

Our large number of files also use a corresponding number of temp files which results in thousands of print messages about which temporary files would be remove. I purpose adding `LogEvent.JOB_INFO`  to the temporary file messages in order to allow quieting these messages.

### QC

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved logging to prevent empty or "None" messages from appearing in the output.
  - Refined logging behavior to respect quiet mode settings, reducing unnecessary log messages during temporary file removal and job scheduling progress updates.
  - Enhanced control over shell command and progress log visibility based on specific quietness flags for a clearer user experience.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -961,17 +961,15 @@ class DAG(DAGExecutorInterface, DAGReportInterface):
 
         for f in unneeded_files():
             if self.workflow.dryrun:
-                if (
-                    not self.workflow.output_settings.quiet
-                    or Quietness.RULES not in self.workflow.output_settings.quiet
-                ):
-                    logger.info(f"Would remove temporary output {fmt_iofile(f)}")
+                logger.info(
+                    f"Would remove temporary output {fmt_iofile(f)}",
+                    extra={"quietness": Quietness.RULES},
+                )
             else:
-                if (
-                    not self.workflow.output_settings.quiet
-                    or Quietness.RULES not in self.workflow.output_settings.quiet
-                ):
-                    logger.info(f"Removing temporary output {fmt_iofile(f)}.")
+                logger.info(
+                    f"Removing temporary output {fmt_iofile(f)}.",
+                    extra={"quietness": Quietness.RULES},
+                )
                 await f.remove(remove_non_empty_dir=True)
 
     async def handle_log(self, job):
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -30,6 +30,7 @@ from snakemake_interface_executor_plugins.dag import DAGExecutorInterface
 from snakemake_interface_report_plugins.interfaces import DAGReportInterface
 from snakemake_interface_storage_plugins.storage_object import StorageObjectTouch
 from snakemake_interface_logger_plugins.common import LogEvent
+from snakemake.settings.enums import Quietness
 
 from snakemake import workflow as _workflow
 from snakemake.common import (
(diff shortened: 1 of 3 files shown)


cademirch (changes requested):
Thanks for opening this! A couple comments

cademirch:
> It could be possible to move the `printshellcmds` check to where the filter is initialized:
> 
> image
> 
> and then we could add a quietness for shellcmds if the `printshellcmds` (`self.settings.printshellcmds`) flag is not present? And then remove the part where we set the message to empty?

Yes! I think that would be a great solution. We should just filter the message properly. But I don't think we should add a new quietness level since its already confusing as is with the issue you referenced. I think we should just do something like:

```python
class DefaultFilter:
    def __init__(self, quiet, debug_dag, dryrun, printshellcmds) -> None:
        if quiet is None:
            quiet = set()
        self.quiet = quiet
        self.debug_dag = debug_dag
        self.dryrun = dryrun
        self.printshellcmds = printshellcmds

    def filter(self, record):
        from snakemake.settings.enums import Quietness

        event, level = get_event_level(record)
        
        ...

        
        # handle shell cmd
        if event == LogEvent.SHELLCMD and not self.printshellcmds:
            return False
```

cademirch (approved):
Thanks for making these changes. Let's make sure tests pass but I think this is okay to merge.

src/snakemake/dag.py
```diff
@@ -960,9 +961,17 @@ def unneeded_files():
...
+                    or Quietness.RULES not in self.workflow.output_settings.quiet
+                ):
+                    logger.info(f"Would remove temporary output {fmt_iofile(f)}")
             else:
-                logger.info(f"Removing temporary output {fmt_iofile(f)}.")
+                if (
+                    not self.workflow.output_settings.quiet
+                    or Quietness.RULES not in self.workflow.output_settings.quiet
+                ):
+                    logger.info(f"Removing temporary output {fmt_iofile(f)}.")
```
johanneskoester: I don't think quietness should be handled explicitly outside of the logger code, rather via setting the log event type. Do you agree @cademirch?

johanneskoester (changes requested):

src/snakemake/dag.py
cademirch: I'm not sure. In general, I do agree. I'm hesitant (and suggested the above solution)  because I wanted to avoid adding more `LogEvent`s and `Quietness` settings except for "bigger" log records. I didn't want to end up back with the pattern we had before, creating a new log methods for each log type. 

But looking at this again, I think I was being overcautious, and this was the point of the LogEvent when we wrote it - to easily add new ones. 

Here is one other idea I've had: just pass the `Quietness` enum value that a log record should be filtered under in the extra. For example:

```python
logger.info(f"Removing temporary output {fmt_iofile(f)}.", extra={"quietness"=Quietness.RULES})
```

Then we can easily handle that in the filter. This would be pretty clean and more cleanly separate the responsibility of filtering off of LogEvent. 

Anyway, sorry for the rambling - I'm OK with adding another LogEvent for this and can help. Also happy with other solutions.

cademirch: I went ahead and committed the quietness solution I proposed above. I think this is the best way to handle quietness, since its the most explicit, and does not depend on LogEvent (which is mainly for informing formatting and context). If you agree with this approach @johanneskoester, lets merge this and then I can tackle updating all the other logger calls in #3571 or another PR.

johanneskoester (approved):

src/snakemake/dag.py
johanneskoester: Great idea!

Issue #3619, opened 2025-06-09
--wms-monitor does not work with snakemake 9.5.1


AdemSaglamRB:
**Snakemake version**
9.5.1

**Describe the bug**
This command below run on any simple Snakefile does not send any message to the given port (i.e. panoptes does not display any progress)
`snakemake --cores 1 --wms-monitor http://127.0.0.1:5000`

**Additional context**
snakemake version 8.20.3 does send messages and panoptes captures those.

cademirch:
Unfortunately, wms monitor was deprecated as of v9.0.0, with the addition of LoggerPlugins. It seems that its still in the CLI - I will address this.

AdemSaglamRB:
Thank you. I leave it to you to close the issue after the fix then,

PR #3622, opened 2025-06-10
fix: remove wms arg, update logging cli docs

### LINKED ISSUE #3620: It would be great to monitor the PID of a Job with --wms-monitor option!
Hello,
I would like to express my sincere gratitude for the incredible dedication of the Snakemake development team. 

I understand that the progress of Snakemake can be monitored using Panoptes. While it is possible to track the progress, I believe it is not yet linked to actual usage data (please correct me if I am mistaken).

To optimize resource allocation for each rule, I need to monitor the real-time discrepancy between the planned resources (like threads and memory specified in the .smk file) and the actual usage. As far as I know, Snakemake itself does not support this functionality, and even when using virtualization tools like Docker, monitoring actual usage is quite tricky and requires separate tools for each application.

I looked into the wms-monitor option and reviewed logging.py, but after inspecting the request data coming into the server, I found that while information like rule names, inputs, and outputs is available, there is not much related to actual process information. For example, there is no way to find out which PID is executing each rule. This is crucial information when monitoring real-time resource usage.

Therefore, I kindly request that you consider adding the process.pid related to each rule in the info output.

Thank you very much!

cademirch:
Removed wms monitor arg and updated monitoring docs. 

closes #3620 #3619 

### QC

## Summary by CodeRabbit

- **Documentation**
  - Updated monitoring documentation to introduce logger plugins as the new monitoring mechanism, replacing previous references to Panoptes and the `--wms-monitor` flag. Added instructions for using logger plugins and linked to the plugin catalog.

- **Bug Fixes**
  - Removed obsolete `--wms-monitor` and `--wms-monitor-arg` command-line options and related configuration support.


docs/executing/monitoring.rst
```diff
@@ -4,28 +4,26 @@
...
-to interact with the server. This means that for Snakemake 3.30.1 and lower, you should use Panoptes 0.1.1 and lower.
-The documentation here will be updated when a new version of Panoptes with the Monitor Schema is released.
+    $ snakemake --logger <plugin-name> --cores 4
+
+Multiple logger plugins can be used simultaneously by specifying the ``--logger`` option multiple times.
+
+Available logger plugins and their configuration options can be found in the `Snakemake plugin catalog <https://snakemake.github.io/snakemake-plugin-catalog>`_.
+The catalog provides detailed documentation for each plugin, including installation instructions, configuration examples, and usage guidelines.
+
+For information on developing custom logger plugins, see the plugin development documentation in the `plugin catalog <https://snakemake.github.io/snakemake-plugin-catalog>`_.
```
m-jahn: there is currently no information about plugin development on the plugin catalog. Should this link rather go to the snakemake docs?

cademirch: Good point, I guess I was referring to the contributing part of the catalog home page, perhaps I can point to the logger-interface repo? Since that explains how to develop one.

PR #3631, opened 2025-06-17
fix: pass kwargs to StorageRegistry call

jakevc:
### QC

## Summary by CodeRabbit

- **New Features**
  - Enhanced storage handling to support passing additional configuration options when retrieving or registering storage objects, enabling more flexible storage operations.
  - Added a new test validating chained rule execution with storage-backed outputs and local file generation to ensure reliability with S3 storage.
  - Updated documentation to illustrate how to pass custom arguments directly to storage calls within rules for improved configuration.


johanneskoester:
Good idea!

PR #3667, opened 2025-07-09
fix ambiguous rule bug

Maarten-vd-Sande:
I noticed an edge case where unused temporary output of different rules would go to the same file. This can introduce nasty side effects, and also made it so that my rules needed to be re-run all the times. This is now detected.

I'm not sure how this affects performance or other stuff, let me know if sth needs changes or if this is actually wanted behaviour :smile: 

### QC

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -273,8 +273,20 @@ class DAG(DAGExecutorInterface, DAGReportInterface):
         self.check_directory_outputs()
 
         # check if remaining jobs are valid
+        seen_outputs: Dict[str, Job] = {}
         for i, job in enumerate(self.jobs):
             job.is_valid()
+
+            # here we check if no two rules make the same output
+            # this can happen in edge cases where the output of jobs
+            # are not part of the targetfiles
+            for output_file in job.output:
+                if output_file in seen_outputs:
+                    other_job = seen_outputs[output_file]
+                    if not self.workflow.execution_settings.ignore_ambiguity:
+                        raise AmbiguousRuleException(output_file, other_job, job)
+                else:
+                    seen_outputs[output_file] = job
 
     def get_unneeded_temp_files(self, job: AbstractJob) -> Iterable[str]:
         if isinstance(job, GroupJob):
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -19,6 +19,7 @@ from snakemake.utils import min_version  # import so we can patch out if needed
 
 from snakemake.settings.types import Batch
 from snakemake.shell import shell
+from snakemake.exceptions import AmbiguousRuleException
 
 sys.path.insert(0, os.path.dirname(__file__))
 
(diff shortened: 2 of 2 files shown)


tests/test_ambiguousruleexception/Snakefile
```diff
@@ -0,0 +1,30 @@
...
+
+rule B:
+    output:
+        b="B/{sample}.txt",
+        c=temp("C/{sample}.txt}"),
+    shell:
+        """
+        touch {output.b}
+        touch {output.c}
+        """
```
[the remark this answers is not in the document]
Maarten-vd-Sande: it's in tests/test.py

johanneskoester:
No need to update the changelog, that will happen automatically via release-please.

src/snakemake/dag.py
```diff
@@ -273,9 +273,21 @@ async def init(self, progress=False):
...
+            # here we check if no two rules make the same output
+            # this can happen in edge cases where the output of jobs
+            # are not part of the targetfiles
+            for output_file in job.output:
+                if output_file in seen_outputs:
+                    other_job = seen_outputs[output_file]
+                    if not self.workflow.execution_settings.ignore_ambiguity:
+                        raise AmbiguousRuleException(output_file, other_job, job)
+                else:
+                    seen_outputs[output_file] = job
```
johanneskoester: This is a very good catch!
One thing though, I think we should move the check (and also job.is_valid()) into DAG.postprocess().
The reason is that the DAG might be updated after checkpoints, where such ambiguities might newly arise.
In postprocess, there is a call of self.check_needrun_jobs(). Above, we can call a new method self.check_jobs(), which could perform this check here and also call job.is_valid().
Analogously to check_needrun_jobs we can thereby ensure that at least is_valid is not called multiple times if postprocess is repeatedly called.

Maarten-vd-Sande: I moved the is_valid() logic into the `check_needrun_jobs` as you suggested, but also moved the checking for duplicate outputs into the `check_needrun_jobs` (including a "cache").

Not sure if that's okay? I for sure do not oversee all the parts here.

p.s. funnily enough, moving it into `check_needrun_jobs` broke a test I wrote 3 years ago :sweat_smile: (fixed by checking for jobids instead of job object equality).

src/snakemake/dag.py
```diff
@@ -1908,6 +1905,22 @@ async def postprocess(
...
+            for output_file in job.output:
+                if output_file in self._seen_outputs:
+                    other_job = self._seen_outputs[output_file]
+                    if other_job.jobid == job.jobid:
+                        continue
+
+                    if not self.workflow.execution_settings.ignore_ambiguity:
+                        raise AmbiguousRuleException(output_file, other_job, job)
+                else:
+                    self._seen_outputs[output_file] = job
```
johanneskoester, line 1930: I think though that this check should happen on all jobs, not only the needrun jobs. Maybe change the name of the method to check_jobs, and have two iterations. One for the needrun jobs, containing check_protected_output, the other one for all jobs, containing is_valid and the new code for the ambiguity detection.

johanneskoester (changes requested):

Issue #3556, opened 2025-05-01
Mermaid-js output flow connections are incorrect


eeidsness:
**Snakemake version**
9.3.3

**Describe the bug**
The output for rulegraph doesn't match that of mermaid-js, specifically the connection of rules by ids is incorrect.

**Logs**

**Minimal example**
In this example - the part I'm referring specifically to is the end section of sequence (id0 --> id0 and so on)

```snakemake --dag```
Output:
```
digraph snakemake_dag {
    graph[bgcolor=white, margin=0];
    node[shape=box, style=rounded, fontname=sans, fontsize=10, penwidth=2];
    edge[penwidth=2, color=grey];
        0[label = "all", color = "0.00 0.6 0.85", style="rounded"];
        1[label = "Step3", color = "0.10 0.6 0.85", style="rounded"];
        2[label = "Step2", color = "0.48 0.6 0.85", style="rounded"];
        3[label = "Step1b", color = "0.29 0.6 0.85", style="rounded"];
        4[label = "Step1c", color = "0.38 0.6 0.85", style="rounded"];
        5[label = "Step4", color = "0.57 0.6 0.85", style="rounded"];
        6[label = "Step1a", color = "0.19 0.6 0.85", style="rounded"];
        4 -> 0
        1 -> 0
        2 -> 1
        6 -> 2
        3 -> 2
        5 -> 4
        1 -> 4
}
```
With mermaid (thanks btw for this!!)
```snakemake --rulegraph mermaid-js```
Output **(incorrect)**:
```
flowchart TB
        id0[all]
        id1[Step3]
        id2[Step2]
        id3[Step1b]
        id4[Step1c]
        id5[Step4]
        id6[Step1a]
        style id0 fill:#D95757,stroke-width:2px,color:#333333
        style id1 fill:#D9A157,stroke-width:2px,color:#333333
        style id2 fill:#57D9C6,stroke-width:2px,color:#333333
        style id3 fill:#7CD957,stroke-width:2px,color:#333333
        style id4 fill:#57D97C,stroke-width:2px,color:#333333
        style id5 fill:#57A1D9,stroke-width:2px,color:#333333
        style id6 fill:#C6D957,stroke-width:2px,color:#333333
        id0 --> id0
        id0 --> id1
        id1 --> id0
        id2 --> id0
        id2 --> id1
        id4 --> id0
        id4 --> id1
```

**Additional context**

PR #3669, opened 2025-07-10
Fix:dag mermaid-js edges (bug:#3556)

b-brankovics:
I noticed that `DAG._dot()` renders the edges correctly, but `DAG._mermaid_js()` makes a mess out of it.

- I copied the bit from the _dot part to the _mermaid_js part
- modified the f"" stretch to use the correct numbers

This fixed the mermaid output so it matches the dot one.

### QC

Unfortunately, I could not figure out how to add a good test for the issue.
I was trying to add this to `tests/test.py`:

```python
def test_github_issue3556():
    run(dpath("test_github_issue3556"), shellcmd="snakemake --dag mermaid-js >stdout.txt")
```

If we need to add a test, I would appreciate some help in how to capture the DAG output to STDOUT and compare it to the expected output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the direction of edges in the Mermaid.js graph representation, ensuring dependencies are accurately visualized.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -2665,10 +2665,12 @@ class DAG(DAGExecutorInterface, DAGReportInterface):
             f"\tstyle id{index} fill:{rulecolor[str(node)]},stroke-width:2px,color:#333333{node2style(node)}"
             for index, node in enumerate(graph)
         ]
+        # node ids
+        ids = {node: i for i, node in enumerate(graph)}
         edges = [
-            f"\tid{index} --> id{index_dep}"
-            for index, (_, deps) in enumerate(graph.items())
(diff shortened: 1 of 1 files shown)


b-brankovics:
I have managed to get a test that passes with the fix and fails without the fix. I don't know how to integrate it using PRs, but here is the link to my commit: b-brankovics/snakemake@e715f3e

johanneskoester:
@b-brankovics just commit your test to the same branch as that of this PR. Then, it will automatically appear here.

b-brankovics:
Thank you @johanneskoester! Merged the test commit and updated the PR to show that there is a test included.

Issue #3095, opened 2024-09-19
Rerun triggers not causing dependent steps to be rerun


powelleric:
**Snakemake version**
8.11.6

**Describe the bug**
In a 2 step workflow, if I run the workflow to completion, then realize I must re-run step 1, I expect that if I run step 1 with --force that step 2 will be run on the next invocation of the whole workflow.

**Minimal example**
Given this snakefile called `bad.snakefile`:

```
rule runall:
    input: "bye.txt"
    output: "hello.txt"
    shell: "echo hello > {output}"


rule realoutput:
    output: "bye.txt"
    shell: "echo bye > {output}"
```

- First run `snakemake --snakefile bad.snakefile --cores 1`.  It works successfully and runs both steps.  
- Next, rerun step 1 using `snakemake --snakefile bad.snakefile --cores 1 --force realoutput`.  This successfully runs the first step.  
- The bug happens when I then try to run the full workflow with the first command: `snakemake --snakefile bad.snakefile --cores 1`.  This should cause Step 2 to be re-run.  However, it says everything is complete.  But it is clear that the mtime for step 1 output is greater than for step 2 output.  This is not fixed by adding `--rerun-triggers=mtime`.

It works correctly if I do `snakemake --snakefile bad.snakefile --cores 1 --forcerun realoutput`.  However, this forces me to run all the steps at once.  I wish to re-run step 1 and then have a chance to check the outcome.  Then, I want to proceed with the re-run of the dependent steps.

tbooth:
I believe this is the same bug as per #2626 and #2011, and in brief summary:

* The behaviour seen by @powelleric is correct/expected but not well documented (see for example https://bioinformatics.stackexchange.com/questions/19923/modification-time-does-not-trigger-a-snakemake-rule#answer-19927)
* The `--rerun-triggers=mtime` option is deffo broken.

The Snakemake code for deciding what to rerun is pretty knarly and easy to break (see for example the attempts to fix #2015) so I think in terms of development the first task is to make a big regression test suite of corner cases. I have it vaguely in the back of my mind that this work could be spun into some sort of masters project or GSoC project, but I've not put any flesh on the bones of that idea yet.

powelleric:
@tbooth Thank you so much for your prompt response.  I tried to search for others complaining about this but did not find any of the ones you linked to.  I really hope that you can get this issue resolved.  Snakemake has been a wonderful tool for me over the years but this bug really hurts.

tbooth:
[...]

Issue #2011, opened 2022-12-14
Snakemake rule does not regenerate output after input has changed


dariober:
**Snakemake version**: tested on 7.18.2 and 7.19.1

**Describe the bug**

Snakemake doesn't re-run the pipeline even if input file is newer than output. Version 6.15.5 behaves as expected (I haven't tested other versions). This is a simplified version of this question on [SO](https://stackoverflow.com/questions/74795939/snakemake-rule-does-not-regenerate-output-after-input-has-changed).

_Unless I'm missing something this looks like an urgent bug to fix?_

**Minimal example**

Minimal snakefile:

```
rule all:
    input:
        'out.txt'

rule compose:
    input:
        'in.txt',
    output:
        'out.txt',
    shell:
        r"""
        touch {output}
        """
```

**Steps to reproduce**

* Create input file and run pipeline (so far so good):

```
touch in.txt
snakemake -p -j 1

snakemake -p -j 1
Building DAG of jobs...
Using shell: /bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job        count    min threads    max threads
-------  -------  -------------  -------------
all            1              1              1
compose        1              1              1
total          2              1              1
...
```

* Update input, rerun pipeline, and you get `nothing to be done`:

```
rm in.txt 
touch in.txt
snakemake -p -j 1
Building DAG of jobs...
Nothing to be done (all requested files are present and up to date).
Complete log: .snakemake/log/2022-12-14T105436.335608.snakemake.log

# Chekc input newer than output:
stat in.txt out.txt
  File: in.txt
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 802h/2050d	Inode: 3153450     Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1001/   dario)   Gid: ( 1001/   dario)
Access: 2022-12-14 11:05:41.726815101 +0000
Modify: 2022-12-14 11:05:41.726815101 +0000
Change: 2022-12-14 11:05:41.726815101 +0000
 Birth: -
  File: out.txt
  Size: 0         	Blocks: 0          IO Block: 4096   regular empty file
Device: 802h/2050d	Inode: 3177457     Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1001/   dario)   Gid: ( 1001/   dario)
Access: 2022-12-14 11:04:28.938495477 +0000
[...]

Issue #1694, opened 2022-05-31
Announcement: Change in rerun behavior in Snakemake 7.8


johanneskoester:
Snakemake 7.8 changes its rerun behavior. Before, rerunning jobs relied purely on file modification times.
Now, also provenance information is considered, like parameter changes, code changes, software environment changes, and changes in the set of input files of a job.

This is intentional, because it provides a higher degree of safety and reproducibility, such that no changes in the workflow definition are missed and the results on disk always represent the state of the codebase.

However, it also has the downside that you might sometimes experience a job rerun that happens just because of cosmetic changes like the removal of some whitespace or the addition of a comment, which might happen from time to time during development.

If you don't want such reruns, you have the following options, which are also displayed by Snakemake at the end of a dryrun in such cases:

* If you prefer that only modification time is used to determine whether a job shall be executed, use the command line option '--rerun-triggers mtime' (also see --help).
* If you are sure that a change for a certain output file (say, <outfile>) won't change the result (e.g. because you just changed the formatting of a script or environment definition), you can also wipe its metadata to skip such a trigger via `snakemake --cleanup-metadata <outfile>`.

johanneskoester:
At the moment, one has to provide all of the outfiles, but you can use bash wildcards (`*`) for that purpose.

jonas-eschle:
First of all, thanks a lot @johanneskoester for this awesome package! I very much appreciate all the work you put into it.

I am however a bit critical of this change: what exactly will now trigger a rerun? I didn't quite find that, could you maybe point me to some resources? I am also a bit worried that it seems to have been introduced as an opt-out in a minor version release breaking current behavior.

corneliusroemer:
`--touch` doesn't seem to work when there's a code change. I could get rid of `Reason: Params have changed since last execution` by running `--touch` but now I get `reason: code has changed since last execution`.

However, `--rerun-triggers mtime` worked as suggested above. Now the only issue is I have to remember and type that string 👀

johanneskoester:
@corneliusroemer you can always write your own profile to fix `--rerun-triggers mtime` if you prefer that behavior.

johanneskoester:
> `--touch` doesn't seem to work when there's a code change. I could get rid of `Reason: Params have changed since last execution` by running `--touch` but now I get `reason: code has changed since last execution`.

[...]

PR #3690, opened 2025-08-04
Rerun rule if timestamp updated, even if checksum same (rerun-trigger mtime)

ckronlage:
Previously, a rule would not be rerun even with newer input files (according to file timestamps) if the checksums were unchanged. This may be unexpected and problematic especially for empty "touch" files. 
There are at least two open issues (#2011 , #3095 ) on the tracker, so I understand this is not the intended behaviour. Also this issue [was discussed](https://github.com/snakemake/snakemake/issues/1694#issuecomment-1986701139) after rerun changes were implemented.

With the minor changes in this PR, if the mtime rerun-trigger is set, checksums are only checked if the timestamp is not newer. Because the mtime rerun trigger is set by default, this also changes default behaviour.
For the old behaviour, only the other rerun-triggers need to be set ("--rerun-triggers code input params software-env"). I believe this is in line with the current documentation and the issues seem to show that some people expected this behaviour. Otherwise, I would suggest to not set the mtime rerun-trigger by default.

Also added test cases to reproduce the two tracker issues.

### QC

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -1428,7 +1428,9 @@ class DAG(DAGExecutorInterface, DAGReportInterface):
                             if (
                                 await f.exists()
                                 and await f.is_newer(output_mintime_)
-                                and not await is_same_checksum(f, job)
+                                and (RerunTrigger.MTIME in self.workflow.rerun_triggers
+                                     or
+                                     not await is_same_checksum(f, job))
                             ):
                                 yield f
 
--- a/tests/test_github_issue2011/Snakefile
+++ b/tests/test_github_issue2011/Snakefile
@@ -1,0 +1,14 @@
+
+rule all:
+    input:
+        'out.txt'
+
+rule compose:
+    input:
+        'in.txt',
+    output:
+        'out.txt',
+    shell:
+        r"""
+        touch {output}
+        """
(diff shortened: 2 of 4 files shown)


johanneskoester:
Thanks for bringing this up! Actually, not rerunning upon newer mtime but same checksum was intentional. But I will now reconsider this in the light of those issues.

Hocnonsense:
Hello, I'm wondering: would it be possible to support more fine-grained control over --rerun-triggers, perhaps at the **rule level**? For example, something like:

```snakemake
rule example:
    input:
        a = ...
    rerun_trigger:
        "mtime", "code", "checksum"
    shell:
        "..."
```
This would allow workflows to more precisely control rerun behavior based on the nature of each rule.

In addition, could checksum triggers also support more specific variants, 
e.g., "checksum:size", "checksum:md5", "checksum:sha256", etc. 
depending on the desired tradeoff between speed and strictness?

Another idea is whether such triggers could also be combined with ancient(), e.g.:
```python
def ancient(file, trigger_metadata: str | List[str] = "mtime"):
    ...

rule example:
    input:
        a=ancient("afile.txt", trigger_metadata="checksum")
```
Would this be conceptually aligned with Snakemake’s current direction, or are there known limitations or design reasons that would make this infeasible?

Thanks!

fgvieira:
@ckronlage can you format the code with `black`?

PR #3701, opened 2025-08-17
fix: parser.py incorrectly parsing multiline f-strings

### LINKED ISSUE #3700: Multiline F-strings remove some lines
**Snakemake version**
9.9.0
Python 3.12.2

**Describe the bug**
Multiline f-strings strip lines that don't contain variables or relevant syntax. See example.
I would expect the example to produce a file that also contains "def", but it doesn't.

**Minimal example**
```
rule test:
    output:
        "test.txt"
    run:
        with open(output[0], "w") as f:
            f.write(f"""abc
                def
                {1+1}
            """)
```

**Output (test.txt) of example**
```
abc
                2
            
```
Note that the second line (containing "def") is missing.

ChristianRomberg:
Fix multiline F-Strings removing lines
fixes #3700

### QC

## Summary by CodeRabbit

- Bug Fixes
  - Correctly parses multi-line f-strings (Python 3.12+), preventing crashes, truncation, and misaligned spans.
  - Preserves pure-literal middle lines in multi-line f-strings and handles complex/nested cases more reliably.

- Refactor
  - Reworked f-string parsing to aggregate content across full token spans for more accurate line handling without changing public APIs.

- Tests
  - Added regression test for preserving literal middle lines in multi-line f-strings (issue #3700) and updated expectations.


ChristianRomberg:
Not sure if this is the best way, but it fixes my minimal example (see #3700 ). Feel free to edit

Hocnonsense:
> can you open a PR for this and tag me in it?

@cademirch I just push the commits here, fell free to edit~

johanneskoester (approved):

johanneskoester:
Thanks, guys!

Issue #3456, opened 2025-03-17
Code rerun trigger cannot be ignored when using a script file.


roshankern:
**Snakemake version**

8.29.0

**Describe the bug**

Code rerun trigger cannot be ignored when using a script file. This only seems to be the case when using the a script, not when including code directly in the snakemake rule.

**Minimal example**

I have a Snakefile like
```
N = 5


rule touch_files:
    params:
        n=N,
    output:
        expand("output/file_{i}.txt", i=range(N)),
    script:
        "scripts/touch_files.py"
```

Which calls the script
```
# scripts/touch_files.py
from pathlib import Path

# Get parameters from Snakemake
n = snakemake.params.n
output_files = snakemake.output

# Create output directory if it doesn't exist
Path("output").mkdir(exist_ok=True)

# Touch each output file
for file_path in output_files:
    Path(file_path).touch()

print(f"Created {n} test files test")
```

I run this Snakefile to produce all output files. Now if I make any change to `scripts/touch_files.py` and try `snakemake --rerun-triggers mtime -n`, Snakemake wants to rerun the rule because the code has changed.
```
Building DAG of jobs...
Job stats:
job            count
-----------  -------
touch_files        1
total              1


[Mon Mar 17 16:10:06 2025]
rule touch_files:
    output: output/file_0.txt, output/file_1.txt, output/file_2.txt, output/file_3.txt, output/file_4.txt
    jobid: 0
    reason: Code has changed since last execution
    resources: tmpdir=<TBD>

Job stats:
job            count
-----------  -------
touch_files        1
total              1

Reasons:
    (check individual jobs above for details)
    code has changed since last execution:
        touch_files
Some jobs were triggered by provenance information, see 'reason' section in the rule displays above.
If you prefer that only modification time is used to determine whether a job shall be executed, use the command line option '--rerun-triggers mtime' (also see --help).
If you are sure that a change for a certain output file (say, <outfile>) won't change the result (e.g. because you just changed the formatting of a script or environment definition), you can also wipe its metadata to skip such a trigger via 'snakemake --cleanup-metadata <outfile>'. 
Rules with provenance triggered jobs: touch_files
```

Note that `--rerun-triggers mtime` does work to avoid rerunning at least due to parameter changes.

roshankern:
I was also able to reproduce this bug in Snakemake `9.0.1`.

alienzj:
I meet this issue too in `Snakemake 8.27.1`.

[...]

PR #3706, opened 2025-08-25
fix: unable to ignore code change for scripts

jasonho1308:
Fix the issue mentioned in #3456, which code change cannot be ignored when using script. The issue in the original code is the following snippets will return `true` even if `rerun_triggers` do not include `RerunTrigger.CODE`
```py
reason.code_changed = any(
    [
            f
            async for f in job.outputs_older_than_script_or_notebook()
    ]
)
if not self.workflow.persistence.has_metadata(job):
...
```

My approch should be able to handle #3148 as well.
                            
### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of when tasks need rerunning by refining metadata, code-change and output-age checks, reducing unnecessary or missed reruns in edge cases.

* **Style**
  * Clarified internal comments and cleaned up related log/message formatting for clearer, more maintainable behavior.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -1447,17 +1447,18 @@ class DAG(DAGExecutorInterface, DAGReportInterface, DAGSchedulerInterface):
                             # for determining any other changes than file modification dates, as it will
                             # change after evaluating the input function of the job in the second pass.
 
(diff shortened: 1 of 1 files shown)


jasonho1308:
#3707 got a better approach

Issue #3474, opened 2025-03-23
`ancient()` statements are arbitrarily ignored when using a workflow as a `module:`


irm-codebase:
**Snakemake version**
Occurs in v9.1.1 (and older)

**Describe the bug**
`snakemake` seems to arbitrarily ignore the `ancient()` setting in rule inputs when sub-workflows are called using the `module:` functionality.

**Minimal example**

In the case below, rule inputs with `ancient` statement within `module_hydropower` may be re-triggered, even if the file exists.

```python
configfile: workflow.source_path("./test_config.yaml")

module module_hydropower:
    snakefile:
        "../../workflow/Snakefile"
    config:
        config["module_hydropower"]

use rule * from module_hydropower as module_hydropower_*

rule all:
    message:
        "Generate CF timeseries for a small country."
    default_target: True
    input:
        "results/by_shape_id/hydro_dam_cf.parquet",
```

**Additional context**
This issue was identified in this PR https://github.com/calliope-project/module_hydropower/pull/1

Here is a failing example: [module_hydropower.zip](https://github.com/user-attachments/files/19407611/module_hydropower.zip). You should be able to re-run our case by running the test workflow in `tests/integration`

This project uses data that requires a specific API key. To allow for CI tests (and to deter constant downloads) we use `ancient` in rules that would re-download data using the API, and include minimal test datasets in the integration test. Unfortunately, `snakemake` is ignoring these pre-existing files during github's CI testing, leading to the failure.

irm-codebase:
Here are some other issues that might be related to this one:

- https://github.com/snakemake/snakemake/issues/2962
- https://github.com/snakemake/snakemake/issues/3024
- https://github.com/snakemake/snakemake/issues/3254

Seems like `ancient` is too temperamental...

Hocnonsense:
Is it possible to describe your problem more detaily? I downloaded your file and try run like this (snakemake version 9.1.4):
```bash
cd module_hydropower/tests/integration/; snakemake -np
```
and these is the output:
```log
Building DAG of jobs...
Job stats:
job                                               count
----------------------------------------------  -------
all                                                   1
module_hydropower_powerplants_adjust_location         1
module_hydropower_powerplants_get_cf_per_shape        1
module_hydropower_powerplants_get_inflow_m3           1
module_hydropower_powerplants_get_inflow_mwh          1
total                                                 5


[Sun Apr  6 16:56:34 2025]
[...]

PR #3707, opened 2025-08-25
fix: only trigger script with CODE label

Hocnonsense:
This PR is inspired by #3706, resolves issues #3456 and #3474, and addresses the concern raised in https://github.com/snakemake/snakemake/pull/3148#discussion_r1837990293_

In brief, commit e502312 made changes so that the modification time (mtime) of script and notebook files no longer triggers a rerun of the workflow, later, commit e8a0b83 reversed this, making mtime changes always trigger reruns -- even when RerunTrigger.CODE is not specified in `--rerun-triggers`.
This created an undesirable situation: legitimate and common scenarios (e.g., copying scripts or pulling them from a Git repository) could unnecessarily trigger reruns — as reported in #3474.

To make the behavior more intuitive, the `outputs_older_than_script_or_notebook` condition should take precedence over metadata checks. However, this condition should still be governed by the presence of `RerunTrigger.CODE`.

### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Code-change reruns now only run when the CODE rerun trigger is enabled; code-change detection correctly accumulates signals from multiple sources to avoid missed or spurious reruns.

* **Tests**
  * Added an MTIME-triggered rerun test validating that outputs retain timestamps when only input mtimes change.

* **Chores**
  * Fixed a test import path and broadened connectivity error handling to include timeouts.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -1448,7 +1448,7 @@ class DAG(DAGExecutorInterface, DAGReportInterface, DAGSchedulerInterface):
                             # change after evaluating the input function of the job in the second pass.
 
                             if RerunTrigger.CODE in self.workflow.rerun_triggers:
-                                # Prefer script/notebook mtime over metadata when CODE trigger is enabled.
+                                # Always check script/notebook mtime unless CODE trigger is disabled.
                                 # Short-circuit on first older output to avoid building a list.
                                 reason.code_changed = False
                                 async for (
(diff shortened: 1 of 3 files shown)


Hocnonsense:
So sad for many fails when pixi install

jasonho1308:
Your approach is better than mine in #3706 as it handled ignoring code change when the file don't have metadata as well! I am gonna close my pull request.

johanneskoester:
Thanks a lot!

PR #3721, opened 2025-09-03
feat!: federated execution by specifying per-rule executors

johanneskoester:
### QC


cmeesters:
could be `snakemake --default-resources executor="'slurm' if resources.mem_mb > 10000 else 'local'", too?

PR #3744, opened 2025-09-18
Do not release lock when job finished

ChristofferCOASD:
Hi,

I've observed an issue with the rate limiter where the main loop is being rerun by the scheduler before the specified wait time has elapsed.

Initially, I suspected the excessive logging of the message "Job rate limit reached..." was the culprit. However, after further investigation, it appears that the main loop is actually triggered multiple times, not just excessively logged.

Running the following with `--max-jobs-per-timespan "1/1s"` will produce the excesive logging messages. (They can become mucm much worse than this. As a workaound i removed the logging on our HPC)
```
nrs = range(20)
rule testRule:
    input:
        expand("OUTPUT/{i}.txt",i=nrs),

rule onefile:
    output:
        "OUTPUT/{number}.txt"
    shell:
        """
        mkdir -p OUTPUT
        echo "TEST" > {output}
        """
```
image

I believe the issue may stem from the semaphore being released when a job finishes. It's unclear to me why the release occurs at this point, as a completed job doesn't necessarily imply that a new slot in the rate limiter has become available.

In this PR, I've removed the semaphore release to initiate a discussion. I'm happy to explore other solutions or make additional changes — any clarification or feedback would be appreciated.

Thanks,
A happy Snakemake user


ChristofferCOASD:
One idea I had was also to ignore `localrules` from the rate limit, since in our case the rate limit is mostly intended to limit the rate at which jobs are submitted to SLURM. Would this make sense in general?

johanneskoester:
> One idea I had was also to ignore `localrules` from the rate limit, since in our case the rate limit is mostly intended to limit the rate at which jobs are submitted to SLURM. Would this make sense in general?

yes, absolutely!

ChristofferCOASD:
> > One idea I had was also to ignore `localrules` from the rate limit, since in our case the rate limit is mostly intended to limit the rate at which jobs are submitted to SLURM. Would this make sense in general?
> 
> yes, absolutely!

Great, here is my idea I will work on implementing:
- I would like to have remote jobs have the "cores" resource be 0 such that the cores resource limit can be kept in place. This will allow the user still to limit resource usage of local jobs. It could be though of as "local cores".
- Remote jobs should instead of being restricted by job_count be restricted by a new resource limit like `job_slots` which will represent the number of available remote execution slots either limited by the rate limiter or by the user specified max_jobs.
- All jobs will still keep the ´job_count=1` and the `job_count=sys.maxsize` will still stay to keep a hard limit on number of jobs.
- The rate limiter will then be changed away from determining `job_count` to instead determine `job_slots`

From my understanding, this will result in the desired behaviour:

- When executing on a cluster, only remote jobs will be rate limited
- The when specifying `jobs: 100` this limits the number of remote jobs, not the local ones
- Local jobs will respect the "cores" limit even when using rate limiter and on the cluster


**Edits/Notes:**
- I considered if `_nodes` could be used instead of introducing a new `_jobs_slot`, but they are not interchangeable I don't think since you can have jobs requiring multiple nodes but only acting as a single submission in terms of rate limiting.
- There could maybe be found an improvement in `schedule_reevaluation` since local rules would no longer count in the rate limit, there might be several jobs ready even before there is space in the rate limiter.

ChristofferCOASD:
For the sake of completeness i made the following to enable testing things related to remote execusion:

Snakefile:
```
rule all:
    input:
        "output.txt"

rule LOCAL:
    localrule: True
    output:
        "LOCAL/{i}.txt"
    shell:
        """
        echo "==== LOCAL ====="
        touch {output}
        """

rule REMOTE:
    output:
        "REMOTE/{i}.txt"
    shell:
        """
        sleep 10
        touch {output}
        """

rule C:
    input:
        expand("LOCAL/{i}.txt", i=range(1, 4)) +
        expand("REMOTE/{i}.txt", i=range(1, 4))
    output:
        "output.txt"
    shell:
        """
        cat {input} > {output}
        """
```

yaml profile:
```
executor: cluster-generic
jobs: 4
rerun-incomplete: True
default_resources: []
quiet: ["rules"]
max-jobs-per-timespan: "1/1s"
cluster-generic-submit-cmd: "ts -L {rule}-{jobid}"
cluster-generic-status-cmd: "./ts_status.sh"
cluster-generic-cancel-cmd: "ts -r {jobid}"
```

taskspooler status wrapper:
```
#!/usr/bin/env bash
jobid="$1"
state=$(ts -s "$jobid" 2>/dev/null || echo "unknown")

case "$state" in
    running|queued)
        echo "running"
        ;;
    finished)
        echo "success"
        ;;
    error|failed)
        echo "failed"
        ;;
    *)
        # if the job no longer exists, assume finished successfully
        echo "success"
        ;;
esac
```

src/snakemake/scheduling/job_scheduler.py
```diff
@@ -588,26 +582,28 @@ def run_selector(job_selector) -> Sequence[AbstractJob]:
...
             return set()
         else:
-            self.resources["_job_count"] = n_free_jobs
+            self.resources["_job_slot"] = n_free_jobs
             selected = run_selector(self._job_selector)
-            # update job rate limiter
-            self.job_rate_limiter.register_jobs(len(selected))
+            # update job rate limiter with remote jobs only
+            self.job_rate_limiter.register_jobs(
+                len([job for job in selected if not job.is_local])
```
johanneskoester:
```suggestion
                sum(1 for job in selected if not job.is_local)
```

johanneskoester:
Is this ready for a merge?

ChristofferCOASD:
@johanneskoester let me do another test today on our cluster just to make sure.

ChristofferCOASD:
From my testing i still need to do a bit of work here. The local rules are properly excluded from the rate limit, but still there is an issue with how often the rate limitter is being "hit" in such a way it prints "Job rate limit reached, waiting for free slots.

From what i can tell #3421 also might have som overlap with any changes i might make in what resources are assigned to local rules to allow propper job selection. I will wait until it is merged to look at this any further.

PR #3752, opened 2025-09-23
feat: add test case for rule redefinition after module import

schrins:
Adds a test case to check whether rule names are correctly linked and visible when importing all from an external workflow as a module and redefining one of those rules. The test case is intended to fail on v9.11.4 and v9.11.5.

### QC

## Summary by CodeRabbit

- New Features
  - Added an example workflow that imports an external module with namespacing and locally overrides a rule with custom threading.
  - Added comprehensive pipeline configuration and a templated scenario for variant calling, filtering, reporting, and alignment.

- Documentation
  - Added a detailed config README covering samples/units, primer trimming, UMI handling, calling scenarios, and examples.

- Tests
  - Added a dry-run test to validate the workflow integration.

- Refactor
  - Revised core rule-registration and rule-definition behavior in the workflow engine.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -2597,7 +2597,7 @@ def test_github_issue3556():
 
 
 def test_module_redefined_rule():
-    run(dpath("test_module_redefined_rule"), executor="dryrun", cores=1)
+    run(dpath("test_module_redefined_rule"), executor="dryrun", check_results=False)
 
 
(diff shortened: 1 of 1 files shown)


johanneskoester:
Closing in favor of #3753. But thanks a lot for the effort, I am sure this helped @Hocnonsense to identify the issue!

PR #3765, opened 2025-10-01
docs: update for discord and stack overflow

cmeesters:
First proposal to rephrase the pointers for help:

Stack Overflow seems to fall short on user's request: https://stackoverflow.com/search?tab=newest&q=Snakemake&searchOn=3

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -76,8 +76,8 @@ Support
 
 * For releases, see :ref:`Changelog <changelog>`.
 * Check :ref:`frequently asked questions (FAQ) <project_info-faq>`.
-* In case of **questions**, please post on `stack overflow <https://stackoverflow.com/questions/tagged/snakemake>`_.
-* To **discuss** with other Snakemake users, use the `discord server <https://discord.gg/kHvtG6N>`_. **Please do not post questions there. Use stack overflow for questions.**
+* For **questions and help**, we recommend joining our active `discord server <https://discord.gg/kHvtG6N>`_ where the community and maintainers regularly help users with their workflows and issues.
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -76,8 +76,8 @@ Support
 
 * For releases, see :ref:`Changelog <changelog>`.
 * Check :ref:`frequently asked questions (FAQ) <project_info-faq>`.
-* In case of **questions**, please post on `stack overflow <https://stackoverflow.com/questions/tagged/snakemake>`_.
-* To **discuss** with other Snakemake users, use the `discord server <https://discord.gg/kHvtG6N>`_. **Please do not post questions there. Use stack overflow for questions.**
+* For **questions and help**, we recommend joining our active `discord server <https://discord.gg/kHvtG6N>`_ where the community and maintainers regularly help users with their workflows and issues.
+* You can also try `stack overflow <https://stackoverflow.com/questions/tagged/snakemake>`_ for questions, though response may be slower or lacking.
```
cmeesters: Meanwhile, IMO, we should discourage Stack Overflow entirely. What do you think @fgvieira ? It seems essentially dead to me. Particularly regarding Snakemake. Highly voted answers are outdated. It is better to work on public relations and help by other means.

fgvieira: I don't think we need to discourage user from StackOverflow (in the end, we want users to help each other as much as possible), but maybe just not encourage them.
We can say that the main source of support is on Discord and, either not mention StackOverflow at all, or mention and say that "responde may be slower".

cmeesters: hi hi, not linking is for me the same as discourage, quality wise. The current PR says ` though response may be slower or lacking.`  So erase the mentioning of SO from the docs? Or leave the sentence as is?

I leave it to you, Filipe. ;-)

fgvieira (approved):

PR #3778, opened 2025-10-06
fix: set log level accordingly when debug dag is set

### LINKED ISSUE #3776: --debug-dag prints nothing
**Snakemake version**

v9.12.0

**Describe the bug**

Adding `--debug-dag` does not change the output when a `MissingInputException` is thrown.  This is a regression that appears to have been introduced in v9.0.1.

**Logs**

**Minimal example**

```
rule all:                                                                       
  input:
    'file2.txt'                                                
                                                                         
rule file2:                                                                     
  output:                                                                       
    'file2.txt'                                                                 
                                                                                
  input:                                                                        
    'file1.txt'                                                                 
                                                                                
  shell:                                                                        
    'cp {input} {output}'                                                       
```

**Additional context**

cademirch:
closes #3776 

debug dag log events were not being emitted because the log level was not being adjusted when the debug-dag arg was supplied. 

### QC

## Summary by CodeRabbit

* **Features**
  * The debug-DAG option now emits DAG debug messages without enabling full verbose debug output; logging respects a separate verbose flag to control non-DAG debug noise.
  * Logger behavior updated so debug-level output is produced when either verbose or debug-DAG is enabled.

* **Tests**
  * Logging tests updated to reflect the new debug/verbose configuration.

### DIFF
--- a/src/snakemake/logging.py
+++ b/src/snakemake/logging.py
@@ -361,6 +361,8 @@ class DefaultFilter:
         Quietness values to filter out.
     debug_dag
         Whether to allow DEBUG_DAG events.
+    verbose
+        Whether full verbose (DEBUG) output is requested.
     dryrun
     printshellcmds
         Whether to allow SHELLCMD events.
@@ -368,6 +370,7 @@ class DefaultFilter:
 
     quiet: Collection["Quietness"]
     debug_dag: bool
+    verbose: bool
     dryrun: bool
     printshellcmds: bool
 
@@ -375,6 +378,7 @@ class DefaultFilter:
         self,
         quiet: Optional[Collection["Quietness"]],
         debug_dag: bool,
+        verbose: bool,
         dryrun: bool,
         printshellcmds: bool,
     ) -> None:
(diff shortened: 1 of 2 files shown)


src/snakemake/logging.py
```diff
@@ -580,7 +580,9 @@ def setup(
                 self.logger.setLevel(settings.log_level_override)
             else:
                 self.logger.setLevel(
-                    logging.DEBUG if settings.verbose else logging.INFO
+                    logging.DEBUG
+                    if (settings.verbose or settings.debug_dag)
+                    else logging.INFO
```
johanneskoester: But debug dag should not trigger all the other debug output. Only the debug dag messages.

johanneskoester (changes requested):

src/snakemake/logging.py
cademirch: This is true. I've reworked this to just emit debug dag as `logging.INFO` so as to not raise the global logger level. These are not emitted on a normal run since they carry the debug dag event. 

That said, having debug_dag as a separate switch layered on top of DEBUG is bit messy. I'd like follow-up PRs to simplify this rather than keep bolting on special cases. I've been putting that off but plan to pick it up soon.

johanneskoester (approved):

PR #3781, opened 2025-10-08
fix!: remove cpu_usage and calc load from cpu_time

fgvieira:
According to `psutil` [docs](https://psutil.readthedocs.io/en/latest/#psutil.cpu_percent), `cpu_usage` is a percentage. Because of that `mean_load` is also a percentage.
I find it a bit misleading since since `load` is usually a fraction (e.g. in `top`).

Also, it is not clear to me what is the difference between `cpu_usage` and `cpu_time` (apart from that `cpu_usage` is a percentage). Do we need both?

### QC

## Summary by CodeRabbit

- **Bug Fixes**
  - Simplified benchmark CPU metrics: removed cpu_usage from standard and extended outputs; mean_load now computed as cpu_time divided by wall-clock running time.

- **New Features**
  - Benchmark records include an extended metrics block (jobid, rule_name, wildcards, params, threads) for richer reporting.

- **Documentation**
  - Added “Benchmark Rules” section; tutorial updated to describe multiple statistics, use “benchmark file” wording, and change memory units to “MB”.


johanneskoester:
Wouldn't this be a breaking change? In principle, I am with you, but I wonder whether one should not rather document this properly and avoid the breaking change.

fgvieira:
I think it would be better to fix it, i.e.:
- remove `cpu_usage`
- use `cpu_time` to calculate `mean_load`

but, since I agree that it could be seen as a breaking change (and it is a bit overkill for such a minor issue), what if we only merge this PR on the next major `snakemake` release?

johanneskoester:
Ok, sounds good!

Issue #3687, opened 2025-07-30
symlink to directory "has older modification time" than target


thekswenson:
**Snakemake version**

9.9.0

**Describe the bug**

I have a workflow that creates a directory with `make_file`, and then symlinks to that directory with rule `link_to_D`.  Despite the input of `link_to_D` depending on the output directory of `make_file`, and the rules being run in the correct order, snakemake complains that the symlink directory "has older modification time (2025-07-30 18:34:17.469428) than input".

**Logs**

```
=> snakemake -c 1
Assuming unrestricted shared filesystem usage.
host: mymachine
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job          count
---------  -------
all              1
link_to_D        1
make_file        1
total            3

Select jobs to execute...
Execute 1 jobs...

[Thu Jul 31 13:33:09 2025]
localrule make_file:
    output: dir1/D, dir1/D/file.txt
    jobid: 2
    reason: Missing output files: dir1/D
    resources: tmpdir=/tmp
[Thu Jul 31 13:33:11 2025]
Finished jobid: 2 (Rule: make_file)
1 of 3 steps (33%) done
Select jobs to execute...
Execute 1 jobs...

[Thu Jul 31 13:33:11 2025]
localrule link_to_D:
    input: dir1/D
    output: dir2/D
    jobid: 1
    reason: Missing output files: dir2/D; Input files updated by another job: dir1/D
    resources: tmpdir=/tmp
WorkflowError in rule link_to_D in file "/home/thekswenson/home/tmp/snakemake_test/Snakefile", line 19:
Output dir2/D has older modification time (2025-07-31 13:33:11.437842) than input dir1/D (2025-07-31 13:33:11.438434). This could indicate a clock skew problem in your network and would trigger a rerun of this job in the next execution and should therefore be fixed on system level. System time: 2025-07-31 13:33:11.439108 (rule link_to_D, line 33, /home/thekswenson/home/tmp/snakemake_test/Snakefile)
Exiting because a job execution failed. Look below for error messages
WorkflowError:
At least one job did not complete successfully.
```

**Minimal example**

```snakemake
import shutil
from pathlib import Path                                                        
                                                                                
rule all:
  input:                                    
    'dir2/D'                                   
                                                          
                              
rule make_file:              
  """ Touch a file. """  
  output:                         
[...]

PR #3782, opened 2025-10-09
test: symlink to directory "has older modification time" than target

thekswenson:
This is a test case for the problem described in #3687.

### QC

## Summary by CodeRabbit

* **Tests**
  * Added an integration test for workflows that create directories and write files, including validation of symlinked directory outputs.
  * Introduced expected results to verify correct resolution of relative symlink targets.
  * Improves confidence in directory handling and symlink behavior across environments.


thekswenson:
I'm closing this because I realize that I put it on the main branch.  Will create a PR with a new branch now.

PR #3783, opened 2025-10-10
fix: make `--containerize` consider conda env specs from all rules, not just from dry-run DAG jobs

tedil:
Since rules that depend on checkpoints aren't part of the initial (dry-run) DAG, we previously didn't include their conda env specs in the `--containerize` call. Now we do ;)

### QC

## Summary by CodeRabbit

- New Features
  - Containerization now discovers Conda environments from both executed rules and imported modules (including checkpoints), producing a unified, de-duplicated and consistently ordered environment set for image builds.
  - Externally managed environments are explicitly skipped with a warning; dynamic or non-file-based env definitions in non-initial rules are validated and reported with clear errors.

- Tests
  - Added end-to-end test covering checkpoints, module rules, multiple env files, and Dockerfile generation to verify environment discovery and containerization.

### DIFF
--- a/src/snakemake/deployment/containerize.py
+++ b/src/snakemake/deployment/containerize.py
@@ -22,10 +22,11 @@ def containerize(workflow, dag):
 
     # collect envs from jobs from the initial DAG.
     conda_envs = {job.conda_env for job in dag.jobs if job.conda_env is not None}
+    dag_jobs = {job.name for job in dag.jobs if job.conda_env is not None}
 
     # collect envs from rules not in the initial DAG (e.g., for rules past checkpoints)
     for rule in workflow.rules:
-        if not rule.conda_env:
+        if not rule.conda_env or rule.name in dag_jobs:
             continue
 
         env_def = rule.conda_env


src/snakemake/deployment/containerize.py
```diff
@@ -4,36 +4,65 @@
...
+    for rule in workflow.rules:
+        if not rule.conda_env:
+            continue
+
+        env_def = rule.conda_env
+
+        if is_callable(env_def) or contains_wildcard(str(env_def)):
+            raise WorkflowError(
+                "Containerization of conda based workflows is not allowed if any conda env definition is dynamic "
+                "(e.g. contains a wildcard or is a function) and the corresponding rule is not part of the initial DAG. "
```
johanneskoester, line 37: How is the latter checked here? Shouldn't there be an if that skips the iteration if the conda env or the env def is already in the conda_envs set?

tedil: Ah, yes, wanted to do that, then thought: I'm using a set anyway, can probably skip this.

johanneskoester: But wouldnt it run into the error case then for jobs that are part of the dag?

tedil: Skipping rules that have jobs that are part of the dag now

PR #3792, opened 2025-10-13
fix: corner case bug causing key error when module statement is used without config directive; better error message when empty string is defined as input or output file

### LINKED ISSUE #3754: `configfile:` in module's Snakefile raises error if no `config:` in module declaration
**Snakemake version**

9.11.4

**Describe the bug**

`configfile:` in module's Snakefile raises error if no `config:` in module declaration

The [Modules](https://snakemake.readthedocs.io/en/latest/snakefiles/modularization.html#modules) documentation implies that a module should use its own configuration *unless* the module declaration specifies a `config:` argument:
> It is possible to overwrite the global config dictionary for the module, which is usually filled by the `configfile` statement (see [Standard Configuration](https://snakemake.readthedocs.io/en/latest/snakefiles/configuration.html#snakefiles-standard-configuration)):
...
In this case, any `configfile` statements inside the module are ignored.

**Logs**

```shell
snakemake --cores 1
```
```output
KeyError in file ".../../workflow2/Snakefile", line 1:
'config'
```

**Minimal example**

```tree
.
├── config
│   └── config.yml
├── workflow
│   └── Snakefile
├── workflow1
│   ├── config
│   │   └── config.yml
│   └── Snakefile
└── workflow2
    ├── config
    │   └── config.yml
    └── Snakefile
```

```yaml
# workflow/Snakefile

from snakemake.utils import min_version
min_version("6.0")

configfile: "config/config.yml"

module workflow1:
    snakefile: "../workflow1/Snakefile"
    config: config["config1"]

use rule * from workflow1 as workflow1_*

module workflow2:
    snakefile: "../workflow2/Snakefile"

use rule * from workflow2 as workflow2_*
```

```yaml
# workflow1/Snakefile

configfile: "config/config.yml"
```

```yaml
# workflow2/Snakefile

configfile: "config/config.yml"
```

**Additional context**

### LINKED ISSUE #3774: blank input from version 7
Hello,

I am trying to update a pipeline i created in snakemake version 7.24 and I am running into an issue with providing empty strings as input. This is becoming an issue due to the following:

In a subsequent rule, I say that the input to that rule is either the output of previous rules (which start processing user input files which can be empty) or directly provided by the user like so:

```
    input:
        lambda wildcards: "RESULTS/PREPROCESSING/CCS_PACBIO/MERGED/" + PREFIX + "_merged_ccs.fastq" if DO_CCS else config["hifi"]

```

You can see that I have added a conditional term (DO_CCS) for when we have this file type and want to process it. If that's set to False the input file doesnt exist (set to '') and it defaults to the input from the config.

How can I code this now in the new (9+) version?

Hocnonsense:
will close #3774 and close #3754

### QC

example for returning an empty list to indicate no input can be found in https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#the-exists-function

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error message for empty file paths, guiding users to use an empty list to indicate “no file.”
  * Module configuration now defaults to an empty config when none is provided, improving predictability.

* **Tests**
  * Updated module tests to cover multiple configfile declarations and missing configs.
  * Test environments pinned to Python 3.13 to ensure consistent test runs.

### DIFF
--- a/tests/test_jupyter_notebook/env.yaml
+++ b/tests/test_jupyter_notebook/env.yaml
@@ -2,6 +2,6 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - python >=3.5
+  - python=3.13
   - jupyter
   - papermill
--- a/tests/test_jupyter_notebook_draft/env.yaml
+++ b/tests/test_jupyter_notebook_draft/env.yaml
@@ -2,5 +2,5 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - python >=3.5
+  - python=3.13
   - jupyter
--- a/tests/test_jupyter_notebook_nbconvert/env.yaml
+++ b/tests/test_jupyter_notebook_nbconvert/env.yaml
@@ -2,5 +2,5 @@ channels:
   - conda-forge
   - bioconda
 dependencies:
-  - python >=3.5
+  - python=3.13
   - jupyter


src/snakemake/modules.py
```diff
@@ -238,7 +236,8 @@ def __init__(
         self.workflow = workflow
         self.base_snakefile = base_snakefile
 
-        self.skip_configfile = config is not None
+        # according to docs, disable any configfile in modules
+        self.skip_configfile = self.parent_modifier is not None
```
johanneskoester: This needs an explanation. The previous code did just skip the configfile if the user did not use the config directive in the module definition, which was the intended behavior. What is different now?

PR #3793, opened 2025-10-13
feat!: replace #3365 with a unified and robust usewith function for rule parameter updates

Hocnonsense:
This PR withdraws #3365 and introduces a general `usewith` function that provides a consistent and explicit way to update rule definitions across different parameter types.  
It offers a unified, explicit, and more maintainable mechanism for handling parameter updates across all rule definitions, ensuring consistent behavior and reducing ad-hoc logic.

## Motivation
The previous implementation only addressed `params` updates and did not generalize well to other rule components.  
This new function unifies the update behavior for all rule sections that can accept either positional or keyword-style arguments.

## Features
The `usewith` function can now handle updates for:
- `input`, `output`, `log` — supporting both unnamed and named files  
- `params`, `resources`, `wildcard_constraints` — supporting both positional and keyword parameters  
- `config` parsed to modules — allowing easy updates without unpacking  
- `envmodules` and `pathvars` — planned for later integration

## Effect
This change modifies how `params` behave since v9.0, but adapting existing workflows will be straightforward.

If the core idea is acceptable, I’d be happy to discuss further implementation details and update the documentation accordingly.

### QC

## Summary by CodeRabbit

* **New Features**
  * Added rule inheritance "with:" merge syntax to selectively update rule properties (params, input/output/log, resources, wildcard constraints) and a namespace-safe guard for merge markers.

* **Documentation**
  * Expanded examples and clarified limits (positional args are fully overwritten), inheritance behavior, execution-option and pathvar rules.

* **Tests**
  * Updated test Snakefile to reflect "with:" usage and adjusted expected outputs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 ### Documentation
 
-* update installation instuctions ([#4025](https://github.com/snakemake/snakemake/issues/4025)) ([bfb6b59](https://github.com/snakemake/snakemake/commit/bfb6b59f82aa5c86d6c9df3c4006cbe73ac2604a))
+* update installation instructions ([#4025](https://github.com/snakemake/snakemake/issues/4025)) ([bfb6b59](https://github.com/snakemake/snakemake/commit/bfb6b59f82aa5c86d6c9df3c4006cbe73ac2604a))
 
 ## [9.18.2](https://github.com/snakemake/snakemake/compare/v9.18.1...v9.18.2) (2026-03-26)
 
--- a/src/snakemake/utils.py
+++ b/src/snakemake/utils.py
@@ -952,6 +952,7 @@ class Paramspace:
     def instance(self, wildcards):
         """Obtain instance (dataframe row) with the given wildcard values."""
         import pandas as pd
+        from snakemake.io import regex_from_filepattern
 
         def convert_value_dtype(name, value):
             if self.dataframe.dtypes[name] == bool and value == "False":


LKress:
@Hocnonsense Thank you for your work! I stumbled across this when I reused rules and had to redefine the entire input, even though certain paths could have been kept exactly as they were. So I would be very happy with this change. In my opinion, backward compatibility should also be guaranteed, since workflows that import the rules but overwrite all inputs and outputs, for example, work exactly the same way. I cannot assess the implementation, @johanneskoester has to check this, as I am not able to judge whether it could have unintended effects with the rest of the codebase.

Hocnonsense:
@LKress Thanks a lot for your positive comment!
I would also like to clarify your concern about backward-compatibility, please let me know if my understanding of your concern is correct:

> In my opinion, backward compatibility should also be guaranteed, since workflows that import the rules but overwrite all inputs and outputs, for example, work exactly the same way.

Using the example:
```snakemake
rule origin:
    input: a="a", b="b"
    params: p="p"

use rule origin as modified with:
    input: b="b1"
    params: q="q"
```

Before #3365, the entire sections were replaced, not merged:
```snakemake
rule modified:
    input: b="b1"
    params: q="q"
```

After #3365, The same code now yields:
```snakemake
rule modified:
    input: b="b1"
    params: p="p", q="q"
```
Here, params suddenly merges while other sections have historically replaced. This inconsistency can easily lead to confusion.

**This PR** proposed explicit merging with `usewith()`:
```snakemake
use rule origin as modified with:
    input: usewith(b="b1")
    params: usewith(q="q")
```
which yields:
```snakemake
rule modified:
    input: a="a", b="b1"
    params: p="p", q="q"
```
While the previous example without `usewith()` continues to produce:
```snakemake
rule modified:
    input: b="b1"
    params: q="q"
```

This design has clear advantages:
- Default (no `usewith`) -> legacy behavior (full replacement)
- `usewith` -> explicit merging

So although `params` behavior changed since Snakemake 9.0, existing workflows that override entire `input`/`output`/`log`/`resources` remain fully compatible.

------

Notes for @johanneskoester: 
1. Concern about code duplication

I noticed that Snakemake already includes a somewhat related helper [`utils.update_config`](https://github.com/snakemake/snakemake/blob/a3e0939522c472a05b85bf4099c579e8fe052b87/src/snakemake/utils.py#L602).

However:
- `update_config` mutates the input dict in place
- My `usewith` implementation creates a new dict, updating only the keys explicitly provided

This raises two design questions:
- Should these two mechanisms be unified?
- Would you prefer a simpler model based on deepcopy + utils.update_config? (More consistent and less error-prone, though potentially more expensive for large and deep dicts.)

I would appreciate your thoughts on which direction better fits Snakemake’s long-term architecture.

2. Possible future DSL extensions

The current approach uses an explicit function: 
```snakemake
use rule ... with:
    input: usewith(b="b1")
```

But if we envision a more expressive DSL in the future, we might consider syntaxes like:
```snakemake
use rule origin as modified with:
    input with: b="b1"
```
or:
```snakemake
use rule origin as modified with:
    input_with: b="b1"
```
These would move the merging semantics into the DSL grammar rather than relying on a Python function, and would likely be clearer for users to understand. However, such changes would require careful design and broader discussion.

------

Apologies in advance -- I'm currently preparing for my pre-defense, so further updates to this PR may need to wait until the end of the month.
Looking forward to your replies and guidance!

LKress:
Thank you for the clarification, @Hocnonsense! I was not aware that your new implementation also requires an explicit definition of your new usewith function if you want to use predefined values. In this case, I am not so sure if that makes sense, as it breaks with the simple syntax of Snakemake. I like the your idea of the more expressive DSL but I would prefer to implement it the way it currently works for the params keyword as this keeps the rule definition simpler. However, I can't judge whether that makes sense and can be implemented.

Issue #3801, opened 2025-10-17
Breaking change in version 9.11.9: all input files must not be empty even if they are not used


corneliusroemer:
When upgrading to snakemake >=9.11.9 a workflow that worked in the past started failing with: `Empty file path encountered. This is likely unintended.`

We were setting the input file path to only be non-empty when the rule was actually being used. In the past this was fine. But this PR #3769 changed that behavior at a minor version increment.

Our code looked like this (over at nextstrain/mpox):

```
rule add_private_data:
    """
    This rule is conditionally added to the DAG if a config defines 'private_sequences' and 'private_metadata'
    """
    input:
        sequences=build_dir + "/{build_name}/good_sequences.fasta",
        metadata=build_dir + "/{build_name}/good_metadata.tsv",
        private_sequences=config.get("private_sequences", ""),
        private_metadata=config.get("private_metadata", ""),
```

https://github.com/nextstrain/mpox/blame/906b020667d33b6570f950d148bc44da98683425/phylogenetic/rules/prepare_sequences.smk#L66-L75

Hocnonsense:
Apologize for my PR that break your workflow. I'd like to create a new one to reverse this break change.

However, I need some advice from you. Which kind of behavior is proper? A warning every time before building the DAG (like warnings for paths with space), or a WorkflowError when checking file persistent, or somewhere else?

Thanks.

corneliusroemer:
Hi @Hocnonsense no worries! It's often not clear what the repercussions of such a change are!

I'm not sure what the best behavior is - I think it's actually ok to error on this in the future, as long as there's a period of deprecation warning or it's accompanied by a major version bump.

A warning sounds good for now - plus a full error in the next major version?

Hocnonsense:
Sure, thanks a lot for your suggestion!

and I'll keep it in mind.

m-jahn:
should be closed with #3810 @corneliusroemer ?

corneliusroemer:
Indeed, thanks @m-jahn and @Hocnonsense 

Fixed in v9.13.5

PR #3810, opened 2025-10-27
fix: revert breaking change in 9.11.9 disallowing empty input files even when unused

Hocnonsense:
Reverts a breaking change introduced in version 9.11.9 where all input files were required to be non-empty, even if they were not actually used by a rule.

#3801 suggested that it would be acceptable to raise an error for this in the future, as long as there is a deprecation warning period or it comes with a major version bump, and a warning sounds good for now.

Thanks @corneliusroemer for pointing this out, and apologies for my earlier overzealous modification.

### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty file paths by providing informative guidance instead of raising an error. The system now logs a warning message directing users to use an empty list ([]) to indicate no files, making workflows more forgiving and easier to debug.

### DIFF
--- a/src/snakemake/io/__init__.py
+++ b/src/snakemake/io/__init__.py
@@ -467,10 +467,13 @@ class _IOFile(str, AnnotatedStringInterface):
         if callable(self._file):
             return
         if self._file == "":
-            raise WorkflowError(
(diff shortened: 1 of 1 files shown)


johanneskoester:
fair enough, makes sense.

PR #3820, opened 2025-10-31
feat: add semantic helper functions choose_file/choose_folder/choose_tmp and allow all semantic helper functions to be used when parsing resources

fgvieira:
Sometimes, HPC nodes have different options to write `temp` depending on the hardware (e.g. some might have `nvme`).

This function allows the use to specify a list of possible temp folders that, on each job submission, is evaluated. The first valid `dir` is selected.

Moved `dynamic` resources to after `standard` and `default` resources.

### QC

## Summary by CodeRabbit

* **New Features**
  * Adds a helper to select a usable temporary directory and exposes it for use in workflows and resource expressions.
  * Resource expressions can now reference ioutils helpers during evaluation.

* **Tests**
  * Workflow tests consolidated to exercise dynamic tmpdir selection via profile-based runs; one legacy default-only test removed.

* **Documentation**
  * Expanded guidance and examples for standard/default resources, dynamic resources, and tmpdir configuration.

* **Chores**
  * CI test matrix parallelism increased for faster runs.


johanneskoester:
Why not just relying on the tmpdir resource, which can be configured by the admins or the user via the standard tmpdir env variable?

fgvieira:
The issue is when HPC have different hardware configurations. In my case, some have a fast `nvme` scratch space, while others only have a slow `nfs` scratch. Since we never know where the jobs will be submitted to, this decision has to be made at runtime.

The best solution would definitely be to have the admins define the `best` temp folder into a env variable, and use it in tmpdir resource. However, it might not be possible.
As for the user defining the tempdir via the standard tmpdir env variable, how would you do it at runtime?

The idea here is to define a helper function that can be passed to `resources.tmpdir` to select the first valid folder of a list or, if none, `system_tmpdir`. For example:
```
resources:
    tmpdir = choose_tmp(["/fast_nvme", "/home/$USER/scratch", "/scratch/$SLURM_JOB_ID"])
```
This function would be executed at runtime and, depending on the node, it would select the first folder that meets certain criteria to be a tmpdir.

docs/snakefiles/rules.rst
```diff
@@ -786,22 +786,32 @@ The ``subpath`` function can be very handy in combination with :ref:`Snakemake's
...
+
+The choose_tmp function
+"""""""""""""""""""""""
+Depending on the architecture of our system, you might have different servers with different "ideal" temp folders. For example, some might have small ``/tmp`` folders but fast ``nvme`` disks, while in others the only option might be the slow NFS. This function allows the user to specify a list of temp folders that, on each job submission, are evaluated. The first valid path is selected; if no path is valid, Snakemake's internal ``system_tmpdir`` is used.
+
+.. code-block:: yaml
+
+    default_resources:
+      runtime: 10
+      tmpdir: choose_tmp(["/fast_nvme", "/home/$USER/scratch", "/scratch/$SLURM_JOB_ID"])
```
johanneskoester: This should mention that one would put the code into a profile. Also, I would remove the runtime statement. People could be misled to add that as well if they just want the tmpdir specification. Also, I think this is the wrong place, it should be documented in the resources section (where also `input.size_mb` is document).

src/snakemake/ioutils/tmp.py
```diff
@@ -0,0 +1,33 @@
+from typing import List
+from pathlib import Path
+
+
+def choose_tmp(tmpdirs: List) -> str:
```
johanneskoester:
```suggestion
def choose_tmp(tmpdirs: List[Union[Path, str]]) -> str:
```

src/snakemake/ioutils/__init__.py
```diff
@@ -25,5 +26,6 @@ def register_in_globals(_globals):
             "parse_input": parse_input,
             "extract_checksum": extract_checksum,
             "flatten": flatten,
+            "choose_tmp": choose_tmp,
```
johanneskoester, line 34: I think in addition this needs to be present in the eval environment for resource expressions, right?

tests/test_tmpdir/Snakefile
```diff
@@ -1,16 +1,16 @@
...
         "echo {resources.tmpdir} > {output}"
 
 
 rule b:
     output:
-        "test2.txt"
+        "test2.txt",
     resources:
-        tmpdir="foo"
+        tmpdir=choose_tmp(["/non_existent", "foo"]),
```
johanneskoester: I guess in addition to this inline usage, a separate test should test how to specify this via a profile.

johanneskoester (changes requested):

src/snakemake/ioutils/__init__.py
fgvieira: Yes, but I think I have already included it in `src/snakemake/resources.py`, no?

cademirch:
This is a nice addition @fgvieira - are you wanting to wait until the hackathon to get this merged? If its okay I'll see if I can get the tests passing and merge this sooner.

cademirch:
I had a thought on the design: instead of exposing a `choose_tmp` function, what if we natively supported a list for tmpdir? Since this is YAML, users could just write:

```yaml
default-resources:
  tmpdir:
    - "/scratch/nvme"
    - "/home/$USER/scratch"
    - "/scratch/$SLURM_JOB_ID"
```

Snakemake would iterate through and pick the first valid path, falling back to `system_tmpdir` if none work.

fgvieira:
@cademirch I also thought about that, but I wondered if it would be too disruptive. An io_utils function is completely independent from the main code and the users are free to use or not.

But if @johanneskoester thinks it is a better solution, I am all for it!

johanneskoester:
For reference: we discussed to generalize this function into something that can work with any list of directories.

johanneskoester:
@fgvieira what do you think about the generalization towards arbitrary dirs?

fgvieira:
@johanneskoester from what I remember, I think that I have already implemented that with the additions (apart from `choose_tmp`) of two extra generic functions (`choose_file` and `choose_dir`).

Or are you thinking about something else?

johanneskoester:
Ah, right, sorry, I now remember that I had already seen that.

docs/snakefiles/rules.rst
```diff
@@ -833,17 +833,52 @@ The ``subpath`` function can be very handy in combination with :ref:`Snakemake's
...
+""""""""""""""""""""
 When selecting input files, sometimes you might end up with an irregular list of lists. To flatten in, you can use:
 
 .. code-block:: python
 
     flatten([1, "a", [2,"b"], ["c","d",["e", 3]]]) # returns ["1", "a", "2", "b", "c", "d", "e", "3"]
 
+.. _snakefiles-choose-f:
+
+The choose_f(ile/older) functions
```
johanneskoester, line 848: I think this section could need some better explanation. (a) the parameters of `choose_f*`, namely read/write/etc. (b) the assertions in the examples can be confusing to non-python-experts. I would rather avoid the asserts and instead explain what the respective calls of choose_file/folder return (e.g. in python comments.

johanneskoester (changes requested):

src/snakemake/ioutils/__init__.py
johanneskoester: As `ioutils` though, right? Maybe rather the individual functions? This way, it would be symmetric between the snakefile and resource definitions.

johanneskoester (approved):

PR #3839, opened 2025-11-13
fix: fix backup for output marked by `update`

FelixMoelder:
When an output file is marked to be updated snakemake fails if that file does not exist. That might happen when a file needs to be created by the rule before being updated in later workflow executions. This is fixed by validating if the output exists before creating a backup. In addition the behavior of restoring a backup after an error has been thrown needs to be adjusted. Previously an error was raised when no backup file exists for a file marked with `update`. As the newly introduced scenario might not have created a backup a waring is thrown instead of an error.

### QC

## Release Notes

* **Bug Fixes**
  * Enhanced backup restoration to issue warnings instead of errors when backups are unavailable
  * Improved backup efficiency by only backing up files that exist
  * Refined directory backup handling to better preserve backup data after restoration

### DIFF
--- a/src/snakemake/jobs.py
+++ b/src/snakemake/jobs.py
@@ -839,7 +839,7 @@ class Job(
                 os.makedirs(self.resources.tmpdir, exist_ok=True)
 
             for f in self.output:
-                if is_flagged(f, "update"):
+                if is_flagged(f, "update") and await f.exists():
                     self.rule.workflow.persistence.backup_output(Path(f))
                 f.prepare()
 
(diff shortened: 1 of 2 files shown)


FelixMoelder:
I think this PR is not finalized as one case I can think of is not implemented.
Let us assume an output is marked with `update` but the output file does not exist, yet.
In that case no backup will be created and in case of failing there will only be a warning that no backup exists. Still, it might be that some output has already been written before failing. If that is the case snakemake will not remove the output.
Solving this should be straightforward but I am not sure if there is some function to call for removing incomplete output.

PR #3840, opened 2025-11-13
fix: mount local storage prefix into containers

johanneskoester:
### QC

## Summary by CodeRabbit

* **New Features**
  * Workflows expose runtime paths that are propagated to scripts, notebooks, wrappers, CWL jobs, and executors.
  * A configurable local storage prefix is accepted and forwarded across relevant components.
  * Container executions now accept and apply runtime/local storage binds so runtime paths are available inside containers.

### DIFF
--- a/src/snakemake/deployment/singularity.py
+++ b/src/snakemake/deployment/singularity.py
@@ -133,6 +133,11 @@ def shellcmd(
         for b in bind:
             if b.exists():
                 args += f" --bind {str(b)!r}"
+            else:
+                logger.debug(
+                    "Skipping apptainer/singularity bind-mount for "
+                    f"non-existent path {str(b)!r}"
+                )
 
     cmd = "{} singularity {} exec --home {} {} {} {} -c '{}'".format(
         envvars,


GaspardR:
Hi @johanneskoester 

Would it be possible to add the same logic when the CLI arg `--cache` is used? Currently, when `--cache` is used with `--sdm apptainer`, the user has to add the extra CLI args `--apptainer-args "-B $SNAKEMAKE_OUTPUT_CACHE:$SNAKEMAKE_OUTPUT_CACHE"`, else the cache is not mounted to the container at runtime and it fails. This addition would allow for a seamless usage between `--cache` and `--sdm apptainer`.

Let me know if you think this is out of scope for this PR, I can open an new issue and/or PR if you prefer! Thanks in advance!

johanneskoester:
> Hi @johanneskoester
> 
> Would it be possible to add the same logic when the CLI arg `--cache` is used? Currently, when `--cache` is used with `--sdm apptainer`, the user has to add the extra CLI args `--apptainer-args "-B $SNAKEMAKE_OUTPUT_CACHE:$SNAKEMAKE_OUTPUT_CACHE"`, else the cache is not mounted to the container at runtime and it fails. This addition would allow for a seamless usage between `--cache` and `--sdm apptainer`.
> 
> Let me know if you think this is out of scope for this PR, I can open an new issue and/or PR if you prefer! Thanks in advance!

Hi, good catch! Indeed, we should solve both here. I will try to solve it generically and future proof directly in this PR.

GaspardR:
Great, thank you very much!

PR #3851, opened 2025-11-27
refactor: LoggerManager setup and scope

jlumpe:
Currently there is a single global `LoggerManager` instance, with its `setup()` method called from `SnakemakeApi`'s constructor. This is a potential source of issues if the global instance gets used multiple times in the same Python session, as happens when running the test suite. It's also not very necessary, as the global is only used in a couple of places (`api.py` and `workflow.py`), and there's not really any setup or configuration that actually needs to be shared (a couple of exceptions are mentioned below).

This PR removes the global instance and instead creates a new one attached to each `SnakemakeApi` instance. This instance is also now passed explicitly to `Workflow`'s constructor. Because the setup now happens when the `LoggerManager` is created, I removed the `setup()` function and merged its contents into `__init__()`.

I also refactored and simplified some of the code previously in `setup()`, and also moved the initial creation of plugin log handlers from `SnakemakeApi.setup_logger()` to `LoggerManager.__init__()`.

Edit: Also removed the `cleanup_logfile()` method and instead remove all handlers from the logger in `stop()`.

#### No tests

Currently there are no tests for most of this behavior, that is fixed in #3848. I have personally tested that those tests still succeed with these changes, but probably best to merge that PR first.

#### Global logger

There is still a single global `Logger` instance, which gets (mostly) reset and then set up again for each `LoggerManager` instance. Ideally each `LoggerManager` would create its own logger, however this is much more difficult to refactor because the global logger is used extensively throughout the codebase. I'll leave that to a separate PR/issue.

#### `OutputSettings.keep_logger`

Setup and teardown of the global `LoggerManager` was previously skipped if `OutputSettings.keep_logger` was True. As far as I can tell this was never the case - the dataclass field's default value is False, I checked all instances I could find of the constructor being called and the only case that sets the field is in `snakemake/cli.py:create_output_settings()`, where it is always set to False ([link](https://github.com/snakemake/snakemake/blob/12ce6c6cb854867cf90255c4119b73614a1ec58a/src/snakemake/cli.py#L1923)). I removed the attribute from `OutputSettings`. If we want to optionally allow for similar behavior, I could add an argument to `SnakemakeApi`'s constructor that accepts an existing instance.

#### Default stream handler level

In current `LoggerManager.setup()`:

```python3
class LoggerManager:
	def setup(...):
		...
	
        # Skip plugin handlers if requested and return early
        if settings.skip_plugin_handlers or not handlers:
        
            handler = self._default_streamhandler()
            if settings.log_level_override is not None:
                handler.setLevel(settings.log_level_override)
            self.logger.addHandler(handler)
			
			...
			
			return
```

I removed this if statement because most of the code was duplicated below (some of it is conditional on no plugin handlers outputting to stream, which is always the case if there are no handlers or plugins are skipped). The only thing that is different is that the default stream handler's level is set based on `log_level_override`, which doesn't happen otherwise. I'm pretty confident this has no effect because the `Logger` the handler is attached to is set to the same level.

## Summary by CodeRabbit

* **Refactor**
  * Centralized logging into a manager with plugin-capable initialization, unified lifecycle and cleanup; logging context is now passed into workflow execution and logfile handling is managed centrally.
* **Chores**
  * Marked the previous keep_logger setting as deprecated and removed its use during output-settings creation to simplify logging configuration.

### DIFF
--- a/src/snakemake/api.py
+++ b/src/snakemake/api.py
@@ -169,7 +169,18 @@ class SnakemakeApi(ApiBase):
 
     def _cleanup(self):
         """Cleanup the workflow."""
-        self.logger_manager.stop()
+        if self.output_settings.keep_logger:
+            import warnings
+
+            warnings.warn(
+                "OutputSettings.keep_logger is deprecated and will be removed in v10.0. "
+                "If you rely on this behavior, please open an issue describing your use case.",
+                FutureWarning,
+                stacklevel=2,
+            )
+        else:
+            self.logger_manager.stop()
+
         if self._workflow_api is not None:
             self._workflow_api._workdir_handler.change_back()
             if self._workflow_api._workflow_store is not None:
--- a/src/snakemake/settings/types.py
+++ b/src/snakemake/settings/types.py
@@ -416,6 +416,8 @@ class OutputSettings(SettingsBase, OutputSettingsLoggerInterface):
     skip_plugin_handlers
         Skip plugins/queue (remote mode).
     enable_file_logging
+    keep_logger
+        Deprecated, no longer functional. Will be removed in v10.0.
     """
 
     dryrun: bool = False
@@ -431,6 +433,7 @@ class OutputSettings(SettingsBase, OutputSettingsLoggerInterface):
     log_level_override: Optional[int] = None
     skip_plugin_handlers: bool = False
     enable_file_logging: bool = True
+    keep_logger: bool = False
 
 
 @dataclass


jlumpe:
@cademirch sorry got another one for you

cademirch:
Will review asap - this looks great at first glance. Thank you @jlumpe.

A few thoughts:
> #### Global logger
> There is still a single global `Logger` instance, which gets (mostly) reset and then set up again for each `LoggerManager` instance. Ideally each `LoggerManager` would create its own logger, however this is much more difficult to refactor because the global logger is used extensively throughout the codebase. I'll leave that to a separate PR/issue.

I agree that this would be ideal, but difficult to do. Lets talk about this more though. 
 
> #### `OutputSettings.keep_logger`
> Setup and teardown of the global `LoggerManager` was previously skipped if `OutputSettings.keep_logger` was True. 

I'm still not sure what this attribute is/was for. Perhaps @johanneskoester can chime in.

jlumpe:
Rebased onto current master, including #3848.

Bioconda appears to be down, which is causing the failing tests.

jlumpe:
I also removed the `cleanup_logfile()` method:

```python
    def cleanup_logfile(self) -> None:
        if self.settings.enable_file_logging:
            for handler in self.logfile_handlers.keys():
                self.logger.removeHandler(handler)
                handler.close()
            self.logfile_handlers.clear()
```

This was only called in `SnakemakeApi._cleanup()`, right before `LoggerManager.stop()`. Instead I remove and stop all handlers in the `stop()` function, I think this should more or less reset the global logger instance (other than it's level). It's not necessary to clear `self.logfile_handlers` as the manger isn't reused.

PR #3857, opened 2025-12-02
fix: fix issues with cyclic dependencies when using the update and before_update flag

FelixMoelder:
### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Cycle detection now ignores cases involving before_update flags to avoid false-cycle reports.
  * Flag precedence corrected so update jobs take precedence over before_update when both apply.

* **Tests**
  * Added tests and fixtures covering cyclic dependency scenarios, including a forced-rerun case, to validate the updated behavior.


johanneskoester:
Great work with the two testcases. They made it super easy to find the issue and will prevent it in the future.

PR #3889, opened 2025-12-17
fix: Prevent broken report_href links by using deterministic report IDs with fixed prefix length

### LINKED ISSUE #3867: `report_href` generates broken links because it doesn't match Snakemake's shortened report paths
**Snakemake version 9.14.1**

**Describe the bug**

I am trying to generate a report and add a Datavzrd table to it. To reference files included in the report from a table column, I add their paths to the data `.tsv` file using `report_href`. The `report_href` function generates a path in the format `<long-hash>/file`.

However, in the final report folder, Snakemake stores these files using shortened paths (since b24d971). The `<long-hash>` is replaced by a `<short-hash>`, and the unused suffix from the original hash is simply dropped. Consequently, when the Datavzrd table tries to reference a file using the original `<long-hash>`, the link is broken because the actual file on disk uses the `<short-hash>`.

Currently, it seems there is no built-in method to resolve the correct shortened path. The only workaround appears to be manually determining the length of the `<short-hash>` and using JavaScript within the HTML page to dynamically reconstruct the correct path from the `<long-hash>` provided in the table data.

**Minimal example**

For example, the documented use case with a file like `test.html` does not work, because in the final report, `test.html` is stored under a shortened hash path, not the long hash path provided by `report_href`.
```python
rule a:
    input:
        report("test.html"),
        report(
            "subdir",
            patterns=["{name}.html"],
        )
    output:
        report(
            "test2.html",
        )
    script:
        "test_script.py"
```
```python
import textwrap

with open(snakemake.output[0], "w") as f:
    print(
        textwrap.dedent(f"""
        <html>
            <head>
                <title>Report</title>
            </head>
            <body>
                <a href={snakemake.report_href("test.html")}>Link to test.html</a>
            </body>
        </html>
        """
        ),
        file=f,
    )
```
**Additional context**

fxwiegand:
This PR makes snakemake shorten report hashes to 16 characters and also makes `report_href()` use the 16 character long prefix which was not done before. Collisions are detected when shortening IDs and an error is raised directing users to open an issue to request a longer hash length. Fixes #3867.

## Summary by CodeRabbit

* **New Features**
  * Reports and generated links now use 16-character shortened file IDs for cleaner, more readable paths.

* **Bug Fixes**
  * Added collision detection for shortened IDs; non-unique shortened IDs now produce a clear error with guidance to request longer hashes.
  * Removed prior silent-warning behavior in favor of explicit error handling.

* **Style**
  * Minor formatting cleanup in generated report preambles (reduced extra blank lines).

### DIFF
--- a/src/snakemake/report/__init__.py
+++ b/src/snakemake/report/__init__.py
@@ -45,7 +45,7 @@ from snakemake.io import (
     contains_wildcard,
 )
 from snakemake.exceptions import InputFunctionException, WorkflowError
-from snakemake.script import Snakemake
+from snakemake.script import Snakemake, FILE_HASH_PREFIX_LEN
 from snakemake.common import (
     get_input_function_aux_params,
 )
@@ -463,7 +463,7 @@ def shorten_ids(results: Mapping[Category, Mapping[Category, List[FileRecord]]])
         for res in catresults
     ]
     full_ids = [rec.id for rec in file_records]
-    shortened_ids = [rec.id[:16] for rec in file_records]
+    shortened_ids = [rec.id[:FILE_HASH_PREFIX_LEN] for rec in file_records]
     # We only need to check for collisions that appear because of the shortening
     if len(set(shortened_ids)) != len(set(full_ids)):
         raise WorkflowError(
--- a/src/snakemake/script/__init__.py
+++ b/src/snakemake/script/__init__.py
@@ -54,6 +54,8 @@ snakemake: "Snakemake"
 # For compatibility with Python <3.11 where typing.Self is not available.
 ReportHrefType = TypeVar("ReportHrefType", bound="ReportHref")
 
+FILE_HASH_PREFIX_LEN = 16
+
 
 class ReportHref:
     def __init__(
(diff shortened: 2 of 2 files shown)


src/snakemake/report/__init__.py
```diff
@@ -462,14 +462,17 @@ def shorten_ids(results: Mapping[Category, Mapping[Category, List[FileRecord]]])
...
-    for id_len in range(8, 65):
-        if len({rec.id[:id_len] for rec in file_records}) == len(file_records):
-            for rec in file_records:
-                rec.id = rec.id[:id_len]
-            return
-    logger.warning(
-        "Obtained result IDs are non-unique. Certain results will be not accessible."
-    )
+    full_ids = [rec.id for rec in file_records]
+    shortened_ids = [rec.id[:16] for rec in file_records]
```
johanneskoester: Please define the 16 as a global constant here, and refer to that constant also below in the ReportHref class.

johanneskoester (changes requested):

src/snakemake/report/__init__.py
fxwiegand: `FILE_HASH_PREFIX_LEN` now holds the 16.

johanneskoester (approved):

PR #3900, opened 2026-01-02
fix: make ilp solver enumeration lazy

coroa:
Hi everyone,

The enumeration of available solvers with `pulp.listSolvers(onlyAvailable=True)` typically acquires license slots of all installed commercial solvers, which becomes problematic, if only a limited number of shared slots are available. pulp correctly releases the licenses immediately again, but it still repeatedly leads to friction.

This PR avoids the pulp solver enumeration by using `pulp.getSolver(...).available()` to check for whether the chosen solver is available and only on failure queries the full list.

The default solver choice from #3736 was updated to rely on `pulp.apis.LpSolverDefault`, which is compatible with "PULP_CBC_CMD" (if it is available).

I also moved the pulp specific checking out of api.py entirely, so that all the solver choice is now handled by the new structure.

With these changes, even if pulp is not installed, snakemake gracefully falls back to the greedy scheduler.

### QC

## Summary by CodeRabbit

* **Improvements**
  * Lazy, more robust solver discovery with improved default selection for ILP scheduling, reducing startup overhead and surprises.
  * Scheduler now detects when the configured solver is unavailable and automatically falls back to the greedy scheduler.

* **New Features**
  * Exposes a solver-availability indicator for easier diagnostics and decisions.

* **Bug Fixes**
  * Clearer warning when falling back, indicating missing solver support and required external solver dependencies.

### DIFF
--- a/src/snakemake/scheduling/milp.py
+++ b/src/snakemake/scheduling/milp.py
@@ -76,7 +76,7 @@ class SchedulerSettings(SchedulerSettingsBase):
     )
 
     @property
-    def lp_solver_available(self):
+    def lp_solver_available(self) -> bool:
         """Check if the configured solver is available via pulp.getSolver().available()."""
         return self.solver in lp_solvers
 


coroa:
@johanneskoester Happy to add docstrings to the methods, if you think that improves legibility. I am doubtful.

coroa:
@coderabbitai Add docstrings, please

coroa:
@coderabbitai generate docstrings

coroa:
I am confident i addressed all open reservations of the bot. Ready for review!

src/snakemake/scheduling/milp.py
```diff
@@ -40,6 +75,11 @@ class SchedulerSettings(SchedulerSettingsBase):
         metadata={"help": "Set the PATH to search for scheduler solver binaries."},
     )
 
+    @property
+    def lp_solver_available(self):
```
johanneskoester:
```suggestion
    def lp_solver_available(self) -> bool:
```

PR #3903, opened 2026-01-04
ci: bump pixi version in actions

cademirch:
Pixi install github action is failing with "failed to parse pypi name mapping" errors likely due to rate limiting when 30+ jobs are kicked off nearly simultaneously

I tested this fix on #3820 since the tests kept failing due to the `pixi` install action failing. After committing this change, [the actions ran successfully](https://github.com/snakemake/snakemake/actions/runs/20684893321).

In [this failing run's](https://github.com/snakemake/snakemake/actions/runs/20682879051/job/59383597583#step:3:3261) debug logs we see:
```
pixi install -e py311
[...]
   WARN resolve_conda{group=py313 platform=win-64}: reqwest_retry::middleware: Retry attempt #1. Sleeping 1.225245051s before the next attempt
  Error:   × failed to parse pypi name mapping
    ├─▶ error decoding response body
    ╰─▶ expected value at line 1 column 1
```
This warning is repeated many times until finally pixi stops retrying - this is what suggested to me that some sort of rate limit was the issue. 

One downside is that this does make the CI take a bit longer to run. We could consider using the `cache` feature of the pixi action. And turning up the max-parallel, or reducing the number of test-groups

### QC

## Summary by CodeRabbit

* **Chores**
  * Updated development toolchain dependencies for improved build and test infrastructure.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,6 @@ jobs:
     if: github.event.pull_request.merged != true || github.ref != 'refs/heads/main'
     strategy:
       fail-fast: false
-      max-parallel: 10
       matrix:
         test_group: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
         # see pyprojec.toml: [tool.pixi.feature.test] for available test types


melund:
@cademirch Did you try just updating the pixi version by it self? 

I guess the guys at prefix may want to know about this. I have mentioned the issue on their discord..

cademirch:
> @cademirch Did you try just updating the pixi version by it self? 
> 
> 
> 
> I guess the guys at prefix may want to know about this. I have mentioned the issue on their discord? 

Yeah good question - I'll test that

cademirch:
Seems like bumping the version worked by itself. If we see more failures like in my initial comment we reevaluate.

PR #3911, opened 2026-01-08
fix: reuse async runner to prevent port exhaustion

coroa:
Interestingly i debugged the same issue as reported in #3909 last week, and came to a slightly different implementation for reusing the asynchronuous loop. 

The benefit is that it is more explicit, than the implicit reuse. The drawback is that you still have multiple async runners, since the `async_run` in [`InputFiles._predicated_size_files`](https://github.com/coroa/snakemake/blob/401c86946e72f4e2112ae393ae745dcccc2ebe91/src/snakemake/io/__init__.py#L1927-L1937) cannot access the workflow object, when it is being used through [`eval_resource_expression`](https://github.com/coroa/snakemake/blob/401c86946e72f4e2112ae393ae745dcccc2ebe91/src/snakemake/resources.py#L526-L571).

### QC

## Summary by CodeRabbit

* **Refactor**
  * Centralized asynchronous dispatch into per-workflow runners with orderly creation and teardown, replacing scattered external async calls.

* **New Features**
  * Workflow-managed async execution now drives scheduling, execution, caching, postprocessing and touch operations.
  * Resource/input/benchmark evaluation callables and input helpers can accept and propagate an async execution handle.

* **Bug Fixes**
  * Fixed resource iteration extraction and simplified an IO-file error message for clearer output.

### DIFF
--- a/src/snakemake/workflow.py
+++ b/src/snakemake/workflow.py
@@ -134,6 +134,7 @@ from snakemake.caching.storage import OutputFileCache as StorageOutputFileCache
 from snakemake.modules import ModuleInfo, WorkflowModifier, get_name_modifier_func
 from snakemake.ruleinfo import InOutput, RuleInfo
 from snakemake.sourcecache import (
+    HostingProviderFile,
     LocalSourceFile,
     SourceCache,
     SourceFile,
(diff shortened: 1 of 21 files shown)


coroa:
@coderabbitai review

coroa:
@coderabbitai review

src/snakemake/executors/local.py
```diff
@@ -235,19 +235,22 @@ def spawn_job(self, job: SingleJobExecutorInterface):
             raise SpawnedJobError()
 
     def cached_or_run(self, job: SingleJobExecutorInterface, run_func, *args):
+        async_run(self.acached_or_run(job, run_func, *args))
```
johanneskoester: 1. Why not using self.workflow._async_runner.run here as well?
2. I would rather expose it via a method Workflow.async_run()

coroa: 1. you do not want to share the same async runner between different threads and was unsure whether the executors aren't called from the worker threads.
2. If you mean classmethod or staticmethod on Workflow, then i like that idea a lot.

coroa:
Ok, i now understood your `async_run` workflow method suggestion, with this it was also possible to make it thread-safe, and after some poking i understood also how to hand the method into the `eval_resource_expression`.

In my mind this fixes everything that was to dislike with this approach, we now hold an async_runner per thread and close down all event loops in the workflow teardown.

src/snakemake/executors/local.py
coroa: A yes, actually why not an instance method that just runs a coro. Done.

This also gives a good entry place to make everything thread-safe.

src/snakemake/workflow.py
```diff
@@ -269,6 +284,9 @@ def tear_down(self):
         if self._workdir_handler is not None:
             self._workdir_handler.change_back()
         self._snakemake_tmp_dir.cleanup()
+        with self._async_lock:
+            for runner in self._async_runners.values():
+                runner.close()
```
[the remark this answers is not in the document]
coroa: Fixed

coroa:
@johanneskoester This one would be ready again.

* I included your suggestion to have an `workflow.async_run` method,
* solved the thread-safety issue and 
* added a shared thread pool executor so that there only is a limited number of threads.

johanneskoester:
Thanks a lot @melund and @coroa! Fantastic work!

PR #3912, opened 2026-01-08
fix: thread bloat and hanging/freeze on windows

melund:
This extends the fix in #3909 to also include linux/mac. 

I've applied the fix globally (Windows & Linux) and added `max_workers=1` to the reused event loop.

1. It reuses a thread-local event loop on all platforms.
2. It limits the default executor of that loop to 1 thread.

This solves both the Port Exhaustion on Windows (by reusing the loop/socket) and the Thread Bloat on Linux (by preventing the creation of hundreds of default ThreadPools).

I have no way to really test this on Linux, so I would think this PR should only be merged once someone has used it for some real-word applications. I am simply not familiar enough with event-loops and asyncio to understand if there are any problems

### QC

## Summary by CodeRabbit

* **Refactor**
  * Enhanced async operation handling to reduce resource consumption and improve system performance on Windows and Linux platforms.

### DIFF
--- a/src/snakemake/common/__init__.py
+++ b/src/snakemake/common/__init__.py
@@ -18,6 +18,8 @@ import uuid
 import os
 import asyncio
 import collections
+import threading
+import concurrent.futures
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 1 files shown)


johanneskoester:
Closing in favor of #3911. But thanks a lot for exploring all these different solutions! That helped a lot to understand the issue and getting a feeling for the best path.

PR #3949, opened 2026-01-29
feat: Make scheduler work gracefully with hash collisions of cached jobs

coroa:
The provenance hash for job caching may lead to (intended) cache collisions, like when different configuration scenarios actually do not lead to any parameter differences in some sub-branch of a larger DAG.

This PR makes the scheduler already compute the provenance hashes and mark everything it is scheduling as active so that only a single instance of the redundant is being run at the same time and the remaining instances can then gracefully reuse the finished results from the cache as soon as they are available.

The PR sits ontop of #3948 , since i would otherwise have had to regenerate the cache format twice. I'd suggest to merge this one after it.

### QC

## Summary by CodeRabbit

* **New Features**
  * Between-workflow caching no longer enforces special multi-output declarations.
  * Concurrency control added to prevent redundant concurrent execution of jobs targeting the same cache entry.

* **Bug Fixes**
  * Improved handling and cleanup of cache marks on selection, completion, and failure to avoid deadlocks or missed releases.
  * Cache path naming adjusted to use quoted output identifiers.

* **Documentation**
  * Clarified caching is intended for public resources and marked the implementation as experimental.

* **Tests**
  * Added tests validating provenance-hash collision behavior and updated existing cache tests.


output_2.txt
```diff
@@ -0,0 +1 @@
+/home/coroa/work/snakemake/cache/a479746dbd17f4089193762007ecd7247a274d8419d9c183eecd8291224a7357_output_{i}.txt
```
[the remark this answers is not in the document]
coroa: Thanks for the hint. Fixed.

coroa:
Ready for review

src/snakemake/scheduling/job_scheduler.py
```diff
@@ -311,11 +311,33 @@ def schedule(self):
...
+                        for job in list(needrun):
+                            if (
+                                cache_mode := self.workflow.get_cache_mode(job.rule)
+                            ) and not self.workflow.output_file_cache.mark_if_schedulable(
+                                job, cache_mode
+                            ):
+                                logger.debug(
+                                    f"Cached job {job} not schedulable, due to a provenance hash collision with another job"
+                                )
+                                needrun.discard(job)
```
coroa, line 326: I don't fully understand the throttling logic in this loop, that is used for example in:
```
                # continue if no new job needs to be executed
                if not needrun:
                    if self.workflow.dag.has_unfinished_queue_input_jobs():
                        logger.info("Waiting for queue input...")
                        # schedule a reevaluation in 10 seconds
                        self._schedule_reevalutation(
                            self.workflow.execution_settings.queue_input_wait_time
                        )
                    continue
```

and i am wondering if we empty out `needrun` here completely, whether we should not also do the same to avoid a busy loop?

coroa: I did a small test by adding a sleep 60 into the Snakefile and then watched the verbose output. The scheduler loop is iterating about once per second (without finding anything to schedule), but only with negligible cpu time, so i think this is fine.

johanneskoester (changes requested):
Could you maybe add a section to the caching docs that explains the use case of this behavior and its implications?

Further, I think this now has to be rebased on the named-output caching changes as your other PR has been merged.

Issue #3915, opened 2026-01-09
feat: support pixi envs


cademirch:
Creating this issue to consolidate discussion and track progress towards a pixi software deployment plugin, as I plan to work on this at (or before) the Hackathon this year. 

## Related Issues
[#3565](https://github.com/snakemake/snakemake/issues/3565)
[#2880](https://github.com/snakemake/snakemake/issues/2880)

## SDM Plugins
[PR #3339](https://github.com/snakemake/snakemake/pull/3339) will add software deployment plugin support (the new software: directive)

It will support alternatives `or` and encapsulation `within`:

```python
rule a:
    software:
        conda(envfile="envs/samtools.yaml") or container("quay.io/biocontainers/samtools:1.12")

rule b:
    software:
        conda(envfile="envs/tool.yaml", within=container("quay.io/conda-forge/rattler:1.0"))
```

## Pixi plugin implementation
Would appreciate thoughts and discussion here, I'm still fairly new to pixi and want to make sure we land on something that makes sense for how people actually use it.

Pixi is workspace/project-oriented rather than environment-oriented like conda. So far, I'm thinking the plugin would let you point to a workspace and specify which environment to use. You could organize workspaces however makes sense:  one per rule, group related tools together, or just have a single workflow-level pixi.toml:

```python
rule align:
    software:
        pixi(workspace="envs/biotools/", env="alignment") # has a pixi.toml and pixi.lock
    shell: "bwa mem ..."
```

```toml
# pixi.toml
[feature.alignment.dependencies]
bwa = "0.7.18"
samtools = "1.21"

[feature.calling.dependencies]  
bcftools = "1.21"
samtools = "1.21"

[environments]
alignment = { features = ["alignment"], solve-group = "default" }
calling = { features = ["calling"], solve-group = "default" }
```
Note: pixi can import environment.yaml files, so we could support those too, but I don't think it's worth it, that's what the conda plugin is for. However if people feel strongly about this being supported, can reconsider.

irm-codebase:
Hello @cademirch! 

Really glad to see that this is being considered. Hopefully this leads to less headaches when trying to make workflows support multiple operating systems.

#### Suggestions

I've been using `pixi` quite a bit since I opened #2880. Here are some details based on what I've experienced:

[...]

PR #3952, opened 2026-01-30
fix: allow relative imports from modules

### LINKED ISSUE #3951: Local imports incompatible with submodules
**Snakemake version**
Snakemake 9.15.0 and main branch.

**Describe the bug**

In principle local imports in the Snakefile are allowed and encouraged, ie. there is even a test-case for it: https://github.com/snakemake/snakemake/blob/main/tests/test_local_import/Snakefile

Those are realized by manipulating `sys.path` on every [workflow.include(...)](https://github.com/snakemake/snakemake/blob/8d0650924d70190578920da24dd1fcb4c78ee13c/src/snakemake/workflow.py#L1668-L1677).

This becomes problematic when there are name clashes between what different modules want to import, since python keeps around the loaded modules in `sys.modules[...]`, so only the first include will get the expected module, all subsequent imports from other workflows will receive the same one.

**Minimal example**

Consider the following example:
```
├── bar.py
├── module-test
│   ├── bar.py
│   ├── foo.py
│   └── Snakefile
├── results
│   ├── inner.txt
│   └── outer.txt
└── Snakefile
```

where the outer `Snakefile` contains
```
import bar

module test:
    snakefile:
        "module-test/Snakefile"

rule all:
    input:
        "results/inner.txt",
        "results/outer.txt"

use rule inner from test

rule outer:
    output:
        "results/outer.txt"
    shell:
        f"echo bar={bar.data} > {{output}}"
```

and the inner `module-test/Snakefile` has the same imports and rules:
```
import foo
import bar

rule inner:
    output:
        "results/inner.txt",
    shell:
        f"echo bar={bar.data} foo={foo.data} > {{output}}"
```

Then `results/outer.txt` will be produced as (as expected):
```
bar=outer-bar
```

And `results/inner.txt` will produce:
```
bar=outer-bar foo=inner-foo
```
ie. while importing seemingly worked in both workflows the name-clash between the different `bar` modules leads to it being also the top-level bar module in `module-test` rather than from `module-test/bar.py`.

**Additional context**

coroa:
Fixes #3951 .

This PR turns each of the different snakemake modules into packages, so that relative imports in the Snakefile or *.smk files are always resolved relative to the snakefile of the current module.

This means that the name-clashes reported in #3951 , can be resolved using relative imports, `from . import foo`, instead, which resolve correctly in each module.

### Notes

1. The imports are relative to the snakefile of the current module, ie.
    ```
    ├── rules
    │   └── a.smk
    ├── scripts
    │   └── _helpers.py
    ├── Snakefile
    ```

    One should use
    ```python
    from .scripts._helpers import y
    ```
    from within `a.smk` as from within `Snakefile`.

2. I left the sys.path.insert statement in `workflow.include(...)` intact so that this change is backwards-compatible; but i added a note to remove it with the next major release; this will then make absolute imports not resolve locally anymore.
 
3. I am currently only setting the paths correctly for Snakefile's which are identified by `LocalSourceFile` and `str`, i guess some of the other `SourceFile` variants should also have local paths, but i did not know which. Help appreciated.

4. I think the `WorkflowModifier` needs a refactor to disentangle path modifications from module encapsulation; i did not dare do that at the same time.
### QC

## Release Notes

* **New Features**
  * Snakemake modules and included files now support importing local Python modules from their directory, enabling better code organization and modularity in workflows.

* **Tests**
  * Added test coverage for local module imports within Snakemake workflows.

### DIFF
--- a/src/snakemake/modules.py
+++ b/src/snakemake/modules.py
@@ -3,13 +3,17 @@ __copyright__ = "Copyright 2022, Johannes Köster"
 __email__ = "johannes.koester@uni-due.de"
 __license__ = "MIT"
 
+import os
 from pathlib import Path
+import sys
 import types
 import re
+from importlib.machinery import ModuleSpec
 from typing import List, Optional, Set, Dict, Callable
 from snakemake.common import Rules
 
 from snakemake.exceptions import CreateRuleException, WorkflowError
+from snakemake.sourcecache import LocalSourceFile
 from snakemake.io.flags import DefaultFlags
 from snakemake.path_modifier import PathModifier
 from snakemake import wrapper
@@ -207,17 +211,44 @@ class WorkflowModifier:
                 globals = dict(workflow.vanilla_globals)
                 self.parent_modifier = workflow.modifier
                 allow_rule_overwrite |= self.parent_modifier.allow_rule_overwrite
+
+                name = f"{self.parent_modifier.module_type.__name__}.{namespace}"
             else:
                 # the first module modifier of workflow
                 self.parent_modifier = None
-            self.globals = globals
-            self.globals["__name__"] = namespace
-            self.globals["rules"] = self.rule_proxies = rule_proxies or Rules()
-            self.globals["checkpoints"] = self.globals[
-                "checkpoints"
-            ].spawn_new_namespace()
+                name = f"__workflow_{id(workflow)}__"
 
-            self.globals["config"] = config if config is not None else {}
+            module_type = types.ModuleType(name)
+            module_type.__dict__.update(globals)
+            self.module_type = module_type
+            self.globals = module_type.__dict__
+
+            spec = ModuleSpec(name, None, is_package=True)
+            module_type.__spec__ = spec
+
+            # Set __path__
+            if isinstance(base_snakefile, str):
+                basedir = os.path.dirname(base_snakefile)
+            elif isinstance(base_snakefile, LocalSourceFile):
+                basedir = base_snakefile.get_basedir().get_path_or_uri(
+                    secret_free=False
+                )
+            else:
+                basedir = None
+            if basedir:
+                module_type.__path__ = [basedir]
+                spec.submodule_search_locations = [basedir]
+
+            # register package
+            sys.modules[name] = module_type
+            if self.parent_modifier:
+                setattr(self.parent_modifier.module_type, namespace, module_type)
+
+            module_type.__name__ = name
+            module_type.__package__ = name
+            module_type.rules = self.rule_proxies = rule_proxies or Rules()
+            module_type.checkpoints = globals["checkpoints"].spawn_new_namespace()
+            module_type.config = config if config is not None else {}
             self.wildcard_constraints: dict = dict()
 
             assert (
@@ -230,6 +261,7 @@ class WorkflowModifier:
         else:
             # use rule (from same include) as ... with: init with values from parent modifier
             self.parent_modifier = parent_modifier = workflow.modifier
+            self.module_type = parent_modifier.module_type
             self.globals = parent_modifier.globals
             self.wildcard_constraints = parent_modifier.wildcard_constraints
             self.pathvars = parent_modifier.pathvars
@@ -302,7 +334,3 @@ class WorkflowModifier:
     def __exit__(self, type, value, traceback):
         # remove this modifier from the stack
         self.workflow.modifier_stack.pop()
-        if self.namespace:
-            namespace = types.ModuleType(self.namespace)
-            namespace.__dict__.update(self.globals)
-            self.workflow.globals[self.namespace] = namespace
--- a/src/snakemake/workflow.py
+++ b/src/snakemake/workflow.py
@@ -1644,6 +1644,11 @@ class Workflow(WorkflowExecutorInterface):
 
         snakefile = infer_source_file(snakefile, basedir)
 
+        if basedir is None:
+            path = snakefile.get_basedir().get_path_or_uri(secret_free=False)
+            self.modifier.module_type.__path__ = [path]
+            self.modifier.module_type.__spec__.submodule_search_locations = [path]
+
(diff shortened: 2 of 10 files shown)


johanneskoester:
Interesting idea, thanks a lot! One issue is though, since modules can come from remote locations, that Python won't see that there are additional files next to the module Snakefile in such cases. Or am I missing something?

johanneskoester:
This is why one usually does not use python imports for local code, but instead the snakemake include or module directive.

johanneskoester:
In theory it might be possible to hook into the local imports and derive the correct source files from the remote for these.

coroa:
> Interesting idea, thanks a lot! One issue is though, since modules can come from remote locations, that Python won't see that there are additional files next to the module Snakefile in such cases. Or am I missing something?

No, you are right, this currently only works if there is a local directory in which the to be imported python modules are actually present, that relates to my note number 3, above:
> I am currently only setting the paths correctly for Snakefile's which are identified by LocalSourceFile and str, i guess some of the other SourceFile variants should also have local paths, but i did not know which. Help appreciated.

And i agree, one can add more import machinery to also resolve remote files correctly, but i did not try this yet. My uses of snakemake have been strictly all local so far.

coroa:
> This is why one usually does not use python imports for local code, but instead the snakemake include or module directive.

In [pypsa-eur](https://github.com/PyPSA/pypsa-eur) we have been breaking that rule quite early on, and i frequently run into situations where i need to use some code used in some of the scripts also already during dag generation. The only safe solution at the moment is to put this code into an external package that needs to be installed in the outer environment and the inner environments.

irm-codebase:
@coroa for what I've seen, in modules it's generally safer to always rely on a `conda` environments for code execution.

If you want to import local python code within your module without it failing due to relative imports, an alternative is to ensure your module calls your local scripts using `workflow.source_path` (see [here](https://github.com/calliope-project/module_euro_gas_grid/blob/ec992d909c08d9b297ff3d4d5d94feedf6377b19/workflow/Snakefile#L29)).

I think this will fit what `pypsa` does rather well... with the exception that you'd still have to move away from importing within the `.smk` file.

coroa:
> @coroa for what I've seen, in modules it's generally safer to always rely on a `conda` environments for code execution.

That is incompatible with pixi. And does not solve the problem here that is happening during DAG generation.

> If you want to import local python code within your module without it failing due to relative imports, an alternative is to ensure your module calls your local scripts using `workflow.source_path` (see [here](https://github.com/calliope-project/module_euro_gas_grid/blob/ec992d909c08d9b297ff3d4d5d94feedf6377b19/workflow/Snakefile#L29)).

I guess this solves some of the local/remote workflow conundrum. We are only running on already checked out local repositories, and there the wrapping in source_path is basically just the identity function, isn't it.
It does not solve the import.

I'd still argue the fixes here make importing saner than it is at the moment.

irm-codebase:
@coroa that's fair. There might be some confusion in what 'module' means here, though.
I am talking about the snakemake `module` [command](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html), not python modules.

`snakemake` does not have `pixi` support (for now https://github.com/snakemake/snakemake/issues/3915), so you need to work around its current limitations. I fear that your suggestion will lead to issues in remote module execution due to the ambiguity Johannes pointed out above.

If your project is fully local and you are *not expecting to ever enable your project for snakemake module execution*, you can get pixi and snakemake to behave by adapting your approach. `pypsa` currently uses `pixi.toml`. Might I suggest to move things to `pyproject.toml` and package your helpers in `src/`? That way you can always just import them as any regular python tool, both within .smk and actual scripts called by snakemake. 

As you said above, you'll need to remove `conda:` from all of your rules in this setup.

Below is an example of how I'm doing it for a similar project:

```yaml
.
├── config
│   ├── config.yaml   # Local config.
│   └── modules        # config. for remote snakemake modules
│       ├── co2stop_cdr.yaml
│       ├── euro_gas_grid.yaml
│       └── geo_boundaries.yaml
├── pixi.lock
├── pyproject.toml  # Has pixi configuration
├── src
│   └── your_things  # should be importable everywhere in your project
│       ├── geo.py
│       ├── __init__.py
│       ├── parser.py
│       ├── plots.py
│       └── utils.py
└── workflow
    ├── envs
    │   └── shell.yaml
    ├── rules
    │   ├── electricity_grid.smk
    │   ├── euro_gas_grid.smk
    │   ├── geo_boundaries.smk
    │   ├── prepare.smk
    │   └── template.smk 
    ├── scripts
    │   ├── cleanup_electricity_grid.py
    │   ├── cleanup_euro_gas_grid.py
    │   ├── cleanup_shapes.py
    │   └── template.py
    └── Snakefile
```

irm-codebase:
Keep in mind that this is not a perfect solution, mostly because you tie your entire workflow to one increasingly complex environment, which is pretty brittle in the long run. Ideally, each of your rules should rely in a minimal environment that only has what that rule needs to maximise stability.

coroa:
Hi @irm-codebase ,

i am not looking for alternative solutions. I am unhappy about a current snakemake feature that is actively misused, that i would either ban or improve.

Without snakemake "modules" it works with using absolute python imports like `from scripts._helpers import ...`, because before every `include` directive snakemake adds the top-level directory to sys.path (there is even a test that confirms this works: https://github.com/snakemake/snakemake/blob/main/tests/test_local_import/Snakefile ). This depends on local execution.

As soon as you bring snakemake modules into this now leads to active mixing up of different absolute imports.

Here i am suggesting that this sys.path mangling should be removed on the long run and in a transitional period the absolute imports should at least be modified to relative imports with the changes here, since this would avoid mixing up different snakemake modules, while still depending on locality.

johanneskoester (changes requested):
May I ask you to add a section in the docs explaining how Python imports are resolved now with both modules and includes, and also informing about where this won't work (remote inclusions)?

Issue #3958, opened 2026-02-04
Rules fail when both using singularity/apptainer and storage plugins


BEFH:
**Snakemake version**

9.16.0, however, the bug is still present in the code on main.

**Describe the bug**

Rules fail when both using singularity/apptainer and storage plugins because the storage path is relative and apptainer does not allow relative binds. This is fixable by changing `self.storage_settings.local_storage_prefix` to `self.storage_settings.local_storage_prefix.absolute()` in https://github.com/snakemake/snakemake/blob/main/src/snakemake/workflow.py#L2116.

This has been tested with apptainer 1.2.5 and 1.3.6.

**Logs**

Here is the output of the MWE with the current Snakemake code:

```
$ snakemake -j1 --use-apptainer out.txt
Assuming unrestricted shared filesystem usage.
host: lc06e04
Building DAG of jobs...
Pulling singularity image docker://alpine:latest.
Using shell: /usr/bin/sh
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job stats:
job                  count
-----------------  -------
http_in_apptainer        1
total                    1

Select jobs to execute...
Execute 1 jobs...
Retrieving from storage: https://raw.githubusercontent.com/snakemake/snakemake/main/README.md
Finished retrieval.

[Wed Feb  4 13:01:39 2026]
localrule http_in_apptainer:
    input: https://raw.githubusercontent.com/snakemake/snakemake/main/README.md (retrieve from storage)
    output: out.txt
    jobid: 0
    reason: Missing output files: out.txt
    resources: tmpdir=/tmp
Activating singularity image /sc/arion/projects/load/users/fultob01/projects/gnomix_adsp95_eurwithsas/sandbox/apptainer_storage/.snakemake/singularity/e6a892a0fa64dbf96a3217013cbad801.simg
INFO:    Cleanup error: while unmounting /hpc/packages/minerva-rocky9/apptainer/1.3.6/x86_64/var/lib/apptainer/mnt/session/final directory: no such file or directory, while unmounting /hpc/packages/minerva-rocky9/apptainer/1.3.6/x86_64/var/lib/apptainer/mnt/session/rootfs directory: no such file or directory
FATAL:   container creation failed: unable to add /sc/arion/projects/load/users/fultob01/projects/gnomix_adsp95_eurwithsas/sandbox/apptainer_storage/.snakemake/storage to mount list: destination must be an absolute path
RuleException:
[...]

PR #3962, opened 2026-02-04
fix: issue 3958 relative path for storage prefix breaks apptainer

BEFH:
This PR addresses https://github.com/snakemake/snakemake/issues/3958 and includes a test. No docs required.

### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Runtime storage path resolution now returns absolute paths for runtime resources.

* **Tests**
  * Added an integration test covering container-based retrieval and execution of storage-backed inputs, plus the expected output resource to validate the workflow end-to-end.


BEFH:
@johanneskoester, this is a pretty simple PR that fixes a big issue. I wonder if it can be merged into the next version

BEFH:
@johanneskoester, I know that an upcoming refactoring of the code is going to make this fix unnecessary, but could we merge it in the meantime as a temporary patch? The test should also be useful when the plugin-based containers are implemented.

johanneskoester (approved):
Thanks! Sorry for missing this.

src/snakemake/workflow.py
```diff
@@ -2081,7 +2081,7 @@ def decorate(ruleinfo):
     def runtime_paths(self) -> List[Path]:
         assert self.storage_settings is not None
         return [
-            self.storage_settings.local_storage_prefix,
+            self.storage_settings.local_storage_prefix.absolute(),
```
johanneskoester: Instead of doing it here, I will rather merge PR #4142 because only apptainer/singularity needs it to be absolute I think.
```suggestion
            self.storage_settings.local_storage_prefix,
```

johanneskoester (approved):

johanneskoester:
Waiting for the merge of #4142 now, and then we can afterwards merge this one if the test passes.

johanneskoester:
Thanks a lot, and sorry for the delay.

PR #3974, opened 2026-02-17
chore: prepare for internal Rust modules by migrating to maturin as build backend

seneschall:
- change build-backend from setuptools to maturin to allow for future usage of Rust crates in Snakemake
- change versioning from dynamic to static, since the version number is currently determined by setuptools-scm, which is incompatible with maturin
- create a simple `lib.rs` and a test for importing one of the functions from `lib.rs` to verify import of Rust functions works

### QC

## Summary by CodeRabbit

* **New Features**
  * Added a native Python extension exposing performant math and greeting utilities for direct use in applications.

* **Tests**
  * Added test coverage to validate the new extension’s behavior.

* **Chores**
  * Updated container build to include additional system build tools.
  * Updated repository ignore rules to exclude Rust build output directories.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ENV LANG C.UTF-8
 ENV SHELL /bin/bash
 USER root 
 
-ENV APT_PKGS bzip2 ca-certificates curl wget gnupg2 squashfs-tools git
+ENV APT_PKGS bzip2 ca-certificates curl wget gnupg2 squashfs-tools git build-essential
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends ${APT_PKGS} \


pyproject.toml
```diff
@@ -296,3 +299,6 @@ build = { cmd = "python -m build", description = "Build the package into the dis
 build-check = { cmd = "python -m twine check --strict dist/*", depends-on = [
   "build",
 ], description = "Check that the package can be uploaded" }
+
+[tool.pixi.dependencies]
+maturin = ">=1.11.5,<2"
```
johanneskoester: Sure that this is needed? The build system section above also defines the dependency.

pyproject.toml
```diff
@@ -62,6 +65,11 @@ dependencies = [
   "conda-inject>=1.3.1,<2.0",
 ]
 
+[dependency-groups]
+dev = [
+    "snakemake",
+]
+
```
johanneskoester:
```suggestion
```

I think this is not needed.

pyproject.toml
```diff
@@ -22,6 +22,9 @@ classifiers = [
   "Environment :: Console",
   "Intended Audience :: Science/Research",
   "Natural Language :: English",
+  "Programming Language :: Rust",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
```
johanneskoester: Not sure whether we really support pypy. This was never tested.

tests/tests.py
```diff
@@ -22,6 +22,7 @@
 from snakemake.settings.types import Batch
 from snakemake.shell import shell
 from snakemake.exceptions import AmbiguousRuleException, WorkflowError
+from snakemake.snakemake import sum_as_string
```
johanneskoester: This `snakemake.snakemake` submodule is a bit confusing. I wonder if we can rather call the central entrypoint for the rust codebase `snakemake.core`.

pyproject.toml
```diff
@@ -1,10 +1,10 @@
 [build-system]
-requires = ["setuptools>=64", "setuptools-scm>=8"]
-build-backend = "setuptools.build_meta"
+requires = ["maturin>=1.11,<2.0"]
```
johanneskoester, line 2: Set a rust version here that matches the rust edition in the cargo.toml.

johanneskoester (changes requested):

tests/tests.py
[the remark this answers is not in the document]
johanneskoester: Move this import into the test case.

rust/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "snakemake"
+version = "0.1.0"
+edition = "2024"
```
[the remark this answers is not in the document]
johanneskoester: See my comment here: https://github.com/snakemake/snakemake/pull/3974#discussion_r2815941070

musicinmybrain:
Looks like some compiled extensions were accidentally committed here, `src/snakemake/core.cpython-312-x86_64-linux-gnu.so` and `src/snakemake/snakemake.cpython-312-x86_64-linux-gnu.so`.

fgvieira:
@seneschall can you check the error:
```
error: linker `cc` not found
```

PR #3975, opened 2026-02-17
fix: Use intended executor for workflow validation in 'dryrun/touch' modes

### LINKED ISSUE #3973: The `--dry-run` flag validates workflow against `dryrun` executor instead of intended executor
**Snakemake version**
Snakemake version v9.16.3

**Describe the bug**
I searched the open issues but couldn't find anything that matched the bug I'm experiencing.

I'm opening this issue after some investigation into the source of [another issue](https://github.com/htcondor/snakemake-executor-plugin-htcondor/issues/17) reported for the HTCondor Snakemake Executor I've been working on.

The issue is that Snakemake is incorrectly rejecting a workflow when run with `--dry-run` even though the executor supports the configuration.

This command successfully executes the workflow as intended:
```
snakemake --jobs 10 --executor htcondor --htcondor-jobdir logs --shared-fs-usage none --verbose
```

but invoking the command in `--dry-run` mode produces an error:
```
snakemake --dry-run --jobs 10 --executor htcondor --htcondor-jobdir logs --shared-fs-usage non
e --verbose
  File "/home/jhiemstra/snakemake-executor-dev/snakemake/src/snakemake/cli.py", line 2194, in args_to_api
    dag_api.execute_workflow(
  File "/home/jhiemstra/snakemake-executor-dev/snakemake/src/snakemake/api.py", line 535, in execute_workflow
    raise ApiError(
snakemake_interface_common.exceptions.ApiError: If no shared filesystem is assumed for input and output files, a default storage provider (--default-storage-provider) and default storage prefix (--default-storage-prefix) has to be set. See https://snakemake.github.io/snakemake-plugin-catalog for possible storage provider plugins.

Error: If no shared filesystem is assumed for input and output files, a default storage provider (--default-storage-provider) and default storage prefix (--default-storage-prefix) has to be set. See https://snakemake.github.io/snakemake-plugin-catalog for possible storage provider plugins.
```

In particular, the HTCondor executor sets `executor_plugin.common_settings.can_transfer_local_files` to `True`, indicating that the executor can transfer files to the remote execution point as needed on Snakemake's behalf. This was added to let similar executors run without a shared filesystem without declaring a default storage provider/prefix.

However, [this check](https://github.com/snakemake/snakemake/blob/f08fae8b13816e6b57b97ccf71e014fa0259f9f4/src/snakemake/api.py#L533) is failing with the HTCondor executor when run in `--dry-run` mode despite that setting.

It looks like the underlying problem is that `cli.py`'s `args_to_api()` [unconditionally overwrites the configured executor](https://github.com/snakemake/snakemake/blob/f08fae8b13816e6b57b97ccf71e014fa0259f9f4/src/snakemake/cli.py#L1950-L1951) when `--dry-run` is used, meaning `DAGApi.execute_workflow()` validates against the dryrun executor's `CommonSettings` rather than the intended executor's. The `touch` executor has the same issue.

If I'm reading through the code correctly, this affects any executor capability expressed through `CommonSettings` that differs between the intended executor and the `dryrun`/`touch` stand in.

I'm currently working on a fix for this that validates against the intended executor but can against the `dryrun` executor, but with the added stress of maintaining open source software in the age of AI, I wanted to give a heads up that a human is behind this.

jhiemstrawisc:
# Description
Previously, the `--dry-run` option was failing with the HTCondor executor when running with `--shared-fs-usage none` and no default storage provider, despite this being supported by the executor. See https://github.com/htcondor/snakemake-executor-plugin-htcondor/issues/17

It turned out the problem was that the 'CommonSettings' used for argument validation in dryrun mode belonged to the dryrun executor, not the condor executor, causing validation to fail.

This diff splits these two concepts -- one executor may be used for validation while the other is used for actual execution. In the aforementioned case, the condor executor is used for validation while the dryrun executor is used to generate the expected output from running with that flag.

# Evidence of Intended Behavior Change
Here is the error message before the diff:
```
$ snakemake --jobs 10 --executor htcondor --htcondor-jobdir logs --shared-fs-usage none --verbose --dry-run
  File "/home/jhiemstra/snakemake-executor-dev/snakemake/src/snakemake/cli.py", line 2194, in args_to_api
    dag_api.execute_workflow(
  File "/home/jhiemstra/snakemake-executor-dev/snakemake/src/snakemake/api.py", line 535, in execute_workflow
    raise ApiError(
snakemake_interface_common.exceptions.ApiError: If no shared filesystem is assumed for input and output files, a default storage provider (--default-storage-provider) and default storage prefix (--default-storage-prefix) has to be set. See https://snakemake.github.io/snakemake-plugin-catalog for possible storage provider plugins.

Error: If no shared filesystem is assumed for input and output files, a default storage provider (--default-storage-provider) and default storage prefix (--default-storage-prefix) has to be set. See https://snakemake.github.io/snakemake-plugin-catalog for possible storage provider plugins.
```

And here is the successful invocation after the diff is applied:
```
$ snakemake --jobs 10 --executor htcondor --htcondor-jobdir logs --shared-fs-usage none --verbose --dry-run
host: ap2002
Building DAG of jobs...
Need to rerun job make_report because of missing output required by all.
Need to rerun job qc_quality_check because of missing output required by all.
Need to rerun job qc_validation_validate_format because of missing output required by all.
Need to rerun job grouped_step3 because of missing output required by all.
Need to rerun job parallel_aggregate because of missing output required by all.
Need to rerun job process_data because of missing output required by make_report.
Need to rerun job all because job make_report has to be rerun.
Need to rerun job qc_validation_validate_format because job qc_quality_check has to be rerun.
Need to rerun job all because job qc_quality_check has to be rerun.
Need to rerun job all because job qc_validation_validate_format has to be rerun.
Need to rerun job grouped_step2 because of missing output required by grouped_step3.
Need to rerun job all because job grouped_step3 has to be rerun.
Need to rerun job parallel_worker_a because of missing output required by parallel_aggregate.
Need to rerun job parallel_worker_b because of missing output required by parallel_aggregate.
Need to rerun job parallel_worker_c because of missing output required by parallel_aggregate.
Need to rerun job all because job parallel_aggregate has to be rerun.
<rest of output truncated for brevity>
```

**Disclosure**: I used generative AI while developing/debugging this patch, but I take responsibility for the code. I've tried to make sure the entire diff is sound and solves my problem without being too short sighted or adding regressions of my own.

### QC

* [ X] The PR contains a test case for the changes or the changes are already covered by an existing test case.
* [ X] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake).

## Summary by CodeRabbit

* **New Features**
  * Dry-run and touch now support a separate execution override, decoupled from validation to avoid mismatched behavior.
  * Execution-mode scheduler activation now follows the selected execution mode.
  * CLI now derives the proper execution override for dry-run/touch while keeping the chosen executor for validation.

* **Bug Fixes**
  * Refined cores/jobs defaults for dry-run/touch (cores forced to 1) and ensured consistent validation rules when executors differ.

* **Tests**
  * Added regression tests covering validation/execution swap scenarios and rejection of invalid dry-run/touch overrides, plus a new Snakefile.

Closes #3973


jhiemstrawisc:
I see the tests are failing because they don't have access to the htcondor executor. I'll work to make them more generic in the coming day(s).

jhiemstrawisc:
@cademirch, I think this is ready for a preliminary review, at your convenience. Although two steps in the CI are still failing, I don't see how my changes could have caused them. Thoughts?

lukavom:
This would solve #4034

src/snakemake/api.py
```diff
@@ -464,6 +464,7 @@ def __post_init__(self):
     def execute_workflow(
         self,
         executor: str = "local",
+        execution_executor: Optional[str] = None,
```
johanneskoester: I wonder a bit about the naming. What about

```suggestion
        pseudo_executor: Optional[str] = None,
```
instead?

johanneskoester: And allowing it to only take the value None, dryrun or touch?

jhiemstrawisc: Addressed in my latest commits.

jhiemstrawisc:
@johanneskoester, apologies for the long delay getting back to this, but it should be ready for another review now.

src/snakemake/api.py
```diff
@@ -509,6 +524,14 @@ def execute_workflow(
                 "immediate_submit has to be combined with notemp (it does not support temp file handling)"
             )
 
+        # Resolve the executor used for actual (non-)execution. A pseudo-executor
+        # ("dryrun"/"touch") overrides the intended executor; otherwise execution
+        # uses the intended executor itself. Falling back to 'executor' ensures
+        # that direct API calls using executor="dryrun" or executor="touch"
+        # (without a pseudo_executor) still trigger the greedy-scheduler
+        # optimisation below.
+        execution_executor = pseudo_executor or executor
```
johanneskoester:
```suggestion
        actual_executor = pseudo_executor or executor
```

src/snakemake/api.py
```diff
@@ -600,6 +623,14 @@ def execute_workflow(
             if execution_settings.debug:
                 raise ApiError("debug mode cannot be used with non-local execution")
 
+        # Note: use_threads is derived from the validation executor's
+        # CommonSettings (executor_plugin), not from the execution executor
+        # (run_executor_plugin, resolved later below). When executor is a
+        # remote plugin (local_exec=False) but execution_executor is "dryrun"
```
johanneskoester:
```suggestion
        # remote plugin (local_exec=False) but actual_executor is "dryrun"
```

src/snakemake/api.py
```diff
@@ -612,8 +643,7 @@ def execute_workflow(
             greedy_scheduler_settings = GreedySchedulerSettings()
 
         if (
-            executor == "touch"
-            or executor == "dryrun"
+            execution_executor in ("touch", "dryrun")
```
johanneskoester:
```suggestion
            actual_executor in ("touch", "dryrun")
```

src/snakemake/api.py
```diff
@@ -642,9 +672,23 @@ def execute_workflow(
         workflow.remote_execution_settings = remote_execution_settings
         workflow.scheduling_settings = scheduling_settings
         workflow.group_settings = group_settings
+        # If the execution executor differs from the validation executor (e.g.
+        # the caller passed executor="htcondor" + pseudo_executor="dryrun"),
+        # swap to the execution plugin for actual execution. All validation
+        # above was performed against the intended executor so that dry-run
+        # and touch accurately reflect the real execution environment.
+        if execution_executor != executor:
```
johanneskoester:
```suggestion
        if actual_executor != executor:
```

src/snakemake/api.py
```diff
@@ -642,9 +672,23 @@ def execute_workflow(
...
+            run_executor_plugin = executor_plugin_registry.get_plugin(
+                execution_executor
```
johanneskoester:
```suggestion
                actual_executor
```

src/snakemake/api.py
```diff
@@ -600,6 +623,14 @@ def execute_workflow(
             if execution_settings.debug:
                 raise ApiError("debug mode cannot be used with non-local execution")
 
+        # Note: use_threads is derived from the validation executor's
+        # CommonSettings (executor_plugin), not from the execution executor
```
johanneskoester:
```suggestion
        # CommonSettings (executor_plugin), not from the actual executor
```

johanneskoester (approved):

johanneskoester (approved):

johanneskoester:
Thank you so much!

PR #3978, opened 2026-02-19
fix: reporting remote nodes number

cmeesters:
When passing `-j unlimited`, workflow.py reports `Provided remote nodes: 9223372036854775807` (this number happens to be my `sys.maxsize`. This is rather confusing. Hence, this PR which in this case just reports `The number of remote nodes is set to be unlimited.`. Hope that is fine. There are no additional tests or documentation setting necessary.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logging clarity for remote execution by displaying "unlimited" when node allocation is unrestricted, and specific numeric values otherwise.

### DIFF
--- a/src/snakemake/workflow.py
+++ b/src/snakemake/workflow.py
@@ -1381,10 +1381,16 @@ class Workflow(WorkflowExecutorInterface):
                     if shell_exec is not None:
                         logger.info(f"Using shell: {shell_exec}")
                     if not self.local_exec:
-                        logger.info(
-                            f"Provided remote nodes: {self.nodes}",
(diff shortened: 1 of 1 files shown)


johanneskoester (changes requested):
Can't we just say "Provided remote nodes: unlimited"? A bit easier to read since it has the same structure as in the other case.

cmeesters:
@johanneskoester yes. Sure. Then again ... the number of nodes is neither finite nor unlimited. It is rather a number of concurrent jobs, we set or allow with `-j`.

johanneskoester (approved):

Issue #3981, opened 2026-02-20
Too many calls to input functions when using checkpoints and wildcards


tkchouaki:
I have a rather complex workflow where I make extensive use of checkpoints with wildcards. I noticed when running the workflow on a cluster environment (slurm) that a lot of time is being spent updating checkpoint dependencies and running input functions over and over and in some cases it takes more than two hours, all the while no job is running on the cluster. 

I tried to look at the code and at a first glance it seemed to me that when a checkpoint job is finished, it is implemented such as only the jobs depending on it are updated on so the input function should be called only for those jobs. 

I wanted to experiment and verify it so I implemented the simple workflow shown below

<details>
<summary> Click to expand </summary>

```python
# Simple checkpoint rule
checkpoint checkpoint_test:
    output:
        "output/test_{branch}_{n}.txt"
    shell:
        '''
        shuf -i 1-100 -n 1 > {output[0]}
        '''

if "width" not in config:
    config["width"] = 10
else:
    config["width"] = int(config["width"])

if "height" not in config:
    config["height"] = 10
else:
    config["height"] = int(config["height"])

# Just evaluates the checkpoint ten times and retrieves the best value
def aggregate_input(wildcards):
    print(wildcards["branch"])
    max_value = None
    selected_file = None
    for i in range(config["height"]):
        check = checkpoints.checkpoint_test.get(branch=wildcards["branch"], n=i)
        curr_file = check.output[0]
        with curr_file.open() as f:
            n = int(f.read())
        if max_value is None or n > max_value:
            max_value = n
            selected_file = curr_file
    return selected_file

rule aggregate:
    input:
        aggregate_input
    output:
        "output/selection_{branch}.txt"
    shell:
        '''
        cp {input[0]} {output[0]}
        '''

rule all:
    default_target: True
    input:
        expand("output/selection_{branch}.txt", branch=list(range(config["width"])))
```
</details>

The `test_checkpoint` rule takes two wildcards, `branch` and `n`. The `aggregate` rule takes only the `branch` wildcard and as an input, the file resulting from the previous rule on the same `branch` while varying `n` where the value of the file is maximized.  This is done in the `aggregate_input` function which contains a print that will allow to track the number of times the function is called.

Then, I use the following script to measure the number of calls while varying the `width` (number of branches) and the `height`.  5 measures are taken for each combination to check if it is deterministic.

<details>
<summary> Click to expand </summary>

```shell
#!/bin/bash

set -e

[...]

PR #3982, opened 2026-02-20
perf: avoiding useless input expand during checkpoint update

tkchouaki:
Attemps to solve #3981. 
I noticed that the calls to the input functions of rules that are not supposed to be updated happen inside `dag.postprocess` when it is called from `dag.postprocess_after_update` at the end of `update_checkpoint_dependencies`. 
I noticed that the method has a `update_incomplete_input_expand_jobs` argument with a default value of `True`. So this PR just makes the call from `postprocess_after_update` by setting the argument to `False`.

It solves the case described in the issue but I'm not sure whether this modification has bad side effects.

Below the same plots as the ones in the issue updated with the results of this PR:

image

image

image
 

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable handling of checkpoints and job expansion to avoid redundant re-expansion, ensure incomplete-input jobs are properly expanded, and unify completion/readiness bookkeeping for grouped and non-grouped jobs.

* **Tests**
  * Added an integration test for checkpoints with dynamic grouping and conditional aggregation, expected-result fixtures, and a small helper execution script for the test.


johanneskoester:
Thanks a lot @tkchouaki!
@Hocnonsense can you have a look at this and make a review? You are probably the single most experienced checkpoint developer we have.

Hocnonsense:
Thanks @tkchouaki for proposing this execllent change!

The PR's fix is mostly correct, and the reasoning can be made more precise.

## What `update_incomplete_input_expand_jobs` flag does

It is the only entry point of `self.update_incomplete_input_expand_jobs()`, which scans all jobs with `incomplete_input_expand == True` and calls `job.updated()` + `replace_job()` to re-expand them.
If any job was updated, a second full `postprocess()` pass is triggered to recompute groups using the now-correct groupid.

## Two independent cases that cause `incomplete_input_expand=True`

1. groupid is unknown at initial DAG construction.
    If an input function references groupid, it cannot be fully expanded when the job is first constructed, because `groupid=None` at that point.
    This is handled by `update_incomplete_input_expand_jobs` in the initial postprocess call, after `.update_groups()` has determined the correct groupid.
    This case is **unrelated** to checkpoints.
2. checkpoint output is unknown at initial DAG construction.
    If an input function references checkpoint output, it cannot be fully expanded until the checkpoint finishes.
    This is handled at runtime by `update_checkpoint_dependencies`, which also calls `job.updated()` + `replace_job()` on all direct downstream jobs once the checkpoint completes.
    After re-expansion, these jobs have `incomplete_input_expand=False` since checkpoint output is now known.
    However, the re-expanded jobs may introduce new jobs into the DAG whose `groupid` is not yet recorded in `DAG._group`, leaving their input functions potentially unexpanded (`incomplete_input_expand=True`).

## Why `update_incomplete_input_expand_jobs` mostly redundant in `postprocess_after_update`

When `update_checkpoint_dependencies` calls `job.updated()` inside the `while job_queue` loop, `DAG._group` has already been correctly populated by the previous `postprocess` run (the one that executed before the checkpoint started running).
Since `job.group` is statically defined in the rule and does not depend on checkpoint output, `update_groups()` in that previous `postprocess` correctly assigned groups to all pre-existing downstream jobs.
Therefore, `get_job_group()` returns the correct `groupid` at the time `job.updated()` is called, and all pre-existing jobs with checkpoint dependencies finished are re-expanded with `incomplete_input_expand=False`.

At this stage, for all pre-existing jobs in the DAG, the `incomplete_input_expand` issue has already been resolved:
- either finished during the previous `postprocess` (Case 1: `groupid` resolution)
- or it is a pending checkpoint dependency that should not be re-expanded until the corresponding upstream checkpoint completes (Case 2: checkpoint output resolution).

Importantly, the re-expansion may introduce entirely new jobs into the DAG that did not exist before the checkpoint ran.
These new jobs are not yet recorded in `DAG._group`, so if their input functions reference `groupid`, they will have `incomplete_input_expand=True` and should be updated.
Therefore, it is both necessary and sufficient to call `update_groups()` followed by a targeted re-expansion exclusively on these newly introduced jobs, before the final `postprocess`.

## The correct fix
The fix should:

```python
    async def postprocess_after_update(self, new_jobs: Set[Job]):
        if new_jobs:
            # update groups of new jobs ibkt
            self.update_groups()
            for job in new_jobs:
                if job.incomplete_input_expand:
                    newjob = await job.updated()
                    await self.replace_job(job, newjob, recursive=False)
        await self.postprocess(update_incomplete_input_expand_jobs=False)
        self._derived_targetfiles = None
```

1. Collect exactly the new jobs added to the DAG during `update_checkpoint_dependencies`
2. Call `update_groups()` to resolve their group assignments
3. Re-expand only those jobs where `incomplete_input_expand=True`
4. Run a single `postprocess()` pass with `update_incomplete_input_expand_jobs=False`, since all jobs requiring re-expansion have already been handled precisely above. 

(Note that `update_incomplete_input_expand_jobs=True` is only needed in the initial `postprocess` call at workflow startup, where no prior group information is available.)

To work with the new function:
- `update_checkpoint_dependencies` should collect and pass the set of newly introduced jobs.
- `update_queue_input_jobs` should pass an empty set, since it does not introduce any new jobs into the DAG — it only updates input file lists of already-existing jobs whose groups are already correctly recorded in `DAG._group`.

src/snakemake/dag.py
```diff
@@ -2233,7 +2233,7 @@ async def is_output_present(job):
         return updated
 
     async def postprocess_after_update(self):
-        await self.postprocess()
+        await self.postprocess(update_incomplete_input_expand_jobs=False)
```
Hocnonsense, line 2336:
```python
    async def postprocess_after_update(self, new_jobs: Set[Job]):
        if new_jobs:
            # update groups of new jobs ibkt
            self.update_groups()
            for job in new_jobs:
                if job.incomplete_input_expand:
                    newjob = await job.updated()
                    await self.replace_job(job, newjob, recursive=False)
        await self.postprocess(update_incomplete_input_expand_jobs=False)
        self._derived_targetfiles = None
```

src/snakemake/dag.py
Hocnonsense: To preciously pick the newly-introduced jobs, I think `updated_jobs.add(updated_job)` in `for job in job_queue.keys():` loop is OK.

Hocnonsense (changes requested):
Thanks @tkchouaki for proposing this change!

The PR's fix is mostly correct, and the changes I propose are listed here.

johanneskoester:
@Hocnonsense do you think you might be able to finalize this with your suggestions?

Hocnonsense:
Sure, I'll take this forward based on my earlier suggestions, aim to have something ready by next week.

Hocnonsense (approved):

Hocnonsense:
@johanneskoester @tkchouaki I've apply changes with a related test. Would you mind to review it?

PR #3998, opened 2026-03-08
feat: add shell completion for cli

jsquaredosquared:
## Problem
- [Prior to Snakemake 8](https://github.com/snakemake/snakemake/blob/main/docs/getting_started/migration.rst#migration-between-snakemake-versions), it was possible to generate bash completions using `snakemake --bash-completion`, but this is no longer the case.
- Given the long list of CLI options provided by Snakemake, completions would be helpful when typing Snakemake commands interactively.

## Solution
This PR uses the `shtab` library to generate completions by running:
```
snakemake --print-shell-completion {bash|zsh|tcsh}
```

This does add a new dependency on [`shtab`](https://github.com/iterative/shtab). An advantage of using `shtab` is that it generates completions for all shells supported by it, not just bash. It may also support more shells in the future.[^1]

### QC

[^1]:  #3997 was a previous attempt using `pycomplete`, but `shtab` is probably better as it provides more complete completions.

## Release Notes

* **New Features**
  * Added shell completion support for the CLI, including an option to output shell completion configuration. This improves command-line usability, speeds common workflows, and makes tab-completion available across supported shells without changing existing command behavior.

### DIFF
--- a/src/snakemake/cli.py
+++ b/src/snakemake/cli.py
@@ -427,6 +427,8 @@ def get_argument_parser(profiles=None):
         config_file_parser_class=ProfileConfigFileParser,
     )
 
+    shtab.add_argument_to(parser, ["--print-shell-completion"])
+
     group_exec = parser.add_argument_group("EXECUTION")
 
     group_exec.add_argument(
@@ -1815,7 +1817,6 @@ def generate_parser_metadata(parser, args):
 
 def parse_args(argv):
     parser = get_argument_parser()
-    shtab.add_argument_to(parser, ["--print-shell-completion"])
     args = parser.parse_args(argv)
 
     snakefile = resolve_snakefile(args.snakefile, allow_missing=True)


jsquaredosquared:
No problem! 

However, I see some hackers are working on an overhaul of how CLI options are handled (#4008, #4020). If `click` is used, the [built-in click shell completion](https://click.palletsprojects.com/en/stable/shell-completion/) could be used.

I'm not sure how long the refactor would take. Do you think we should wait, or proceed with this PR for now?

johanneskoester:
Good point indeed. We will not merge that rewrite this week, but probably within the next one or two months. Thus, adding a new dependency that is then removed again later on is probably not worth it indeed. Although I highly appreciate your work, it is thus probably better to close this PR in favor of #4020.

johanneskoester:
Nevertheless, thanks a lot, I will leave a note in the other PR to ensure that shell completion is added along with it!

PR #4012, opened 2026-03-09
feat: Allow storing snakemake metadata in files or databases

### LINKED ISSUE #2969: Database instead of file-based logging and book keeping? Performance issues on HPC with many Snakemake users and millions of meta files

tedil:
This PR allows choosing whether metadata stored in .snakemake/metadata is stored as files or in a database.
To this end, we expose `--persistence-backend [file|db]` and `--persistence-backend-db-url URL]` cli options to choose between the two. 
There is NO conversion (between files and db backend) logic implemented.
There is also NO db schema migration logic implemented, yet (will leave that as future work ;)).

Resolves #2969 (well, at least I hope so)

## Summary by CodeRabbit

* **New Features**
  * CLI options to select persistence backend (file or db) and to supply a DB URL.

* **Refactor**
  * Persistence redesigned into a pluggable, version-aware backend system with file- and SQLite-backed implementations, migration, locking, backup and cache support; workflow now selects backend dynamically.

* **Tests**
  * Tests updated and parameterized to exercise both persistence backends.

### DIFF
--- a/tests/test_persistence_backend_db_run_directive_concurrency/Snakefile
+++ b/tests/test_persistence_backend_db_run_directive_concurrency/Snakefile
@@ -1,0 +1,17 @@
+import time
+import random
+
+
+rule all:
+    input:
+        expand("out/{i}.txt", i=range(100)),
+
+
+rule foo:
+    output:
+        "out/{i}.txt",
+    run:
+        # just to get some random arrivals
+        time.sleep(random.uniform(0.01, 0.25))
+        with open(output[0], "w") as f:
+            f.write("done")
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -396,6 +396,43 @@ def test_params_outdated_metadata(mocker, tmp_path, backend, PersistenceClass):
     assert spy.spy_return == True
 
 
+def test_persistence_backend_db_run_directive_concurrency(tmp_path):
+    workdir = tmp_path / "workdir"
+    shutil.copytree(
+        dpath("test_persistence_backend_db_run_directive_concurrency"), workdir
+    )
+
+    # generate expected output dynamically
+    expected_out_dir = workdir / "expected-results" / "out"
+    expected_out_dir.mkdir(parents=True)
+    for i in range(100):
+        (expected_out_dir / f"{i}.txt").write_text("done")
+
+    db_file = tmp_path / "metadata.db"
+    db_url = f"sqlite:///{db_file}"
+
+    run(
+        path=workdir,
+        tmpdir=workdir,
+        cleanup=False,
+        cores=20,  # not sure this will work in CI, probably gets downscaled?
+        persistence_backend=PersistenceBackend.DB,
+        persistence_backend_db_url=db_url,
+        no_tmpdir=True,
+    )
+
+    from sqlmodel import Session, select, create_engine
+    from snakemake.persistence.db import MetadataRecordORM
+
+    engine = create_engine(db_url)
+    with Session(engine) as session:
+        stmt = select(MetadataRecordORM)
+        records = session.scalars(stmt).all()
+        assert (
+            len(records) == 100
+        ), f"Expected 100 DB records, but found {len(records)}!"
+
+
 def test_same_wildcard():
     run(dpath("test_same_wildcard"))
 


src/snakemake/persistence_interface.py
```diff
@@ -0,0 +1,495 @@
...
+    def cleanup_backup(self, path: Path) -> None:
+        backup_path = self._get_backup_path(path)
+        if backup_path.exists():
+            if backup_path.is_dir():
+                shutil.rmtree(backup_path)
+            else:
+                backup_path.unlink()
+
+
+class AbstractPersistence(
```
cademirch: Would composition be possible here? 

Something like:
```python
class AbstractPersistence(PersistenceExecutorInterface):
    def __init__(
        self,
        env_maintenance: EnvironmentMaintenance,
        file_backup: FileBackup,
        dag: Any,
    ):
        self.env = env_maintenance
        self.backup = file_backup
        self.dag = dag
```

tedil: Hm, could be possible. I too am not a big fan of multiple inheritance, but seemed reasonable (read "less work") here ;) Will have to think about how to properly do this using composition

tedil: I think using composition here would make the potential plugin implementation a bit more unwieldy here, unless we always set the env and backup management stuff by default, but then we'd have to have corresponding delegate methods for all of their functionality in the interface, which kind of defeats the purpose. Or perhaps there _is_ a nice way to do this, but I think for now multiple inheritance is fine here.

tedil:
I wonder if we could abstract this even further, given that the methods usually do almost exactly the same, but the db backend just queries the db first and converts to/from representations…

tedil:
> I wonder if we could abstract this even further, given that the methods usually do almost exactly the same, but the db backend just queries the db first and converts to/from representations…

Yup, so the "backends" basically just do their specific record reading/writing now, so mostly io related stuff; everything else is handled in the base class.

Also added some autogenerated docstrings for some of the abstract methods to make it a bit clearer on how to implement them.

tedil:
Ah, and right now, I just enabled the db backend by default to see if the CI succeeds with that or not.
(there's no option yet to toggle that from the cli/config; technically, may not be needed if we make it a plugin, not sure about that, needs more discussion)

src/snakemake/persistence/db.py
```diff
@@ -0,0 +1,197 @@
...
+    record_format_version: Mapped[int] = mapped_column(Integer, default=0)
+    conda_env: Mapped[Optional[str]] = mapped_column(String, nullable=True)
+    container_img_url: Mapped[Optional[str]] = mapped_column(String, nullable=True)
+    software_stack_hash: Mapped[Optional[str]] = mapped_column(String, nullable=True)
+    input: Mapped[Optional[List[str]]] = mapped_column(JSON, nullable=True)
+    log: Mapped[Optional[List[str]]] = mapped_column(JSON, nullable=True)
+    params: Mapped[Optional[List[Any]]] = mapped_column(JSON, nullable=True)
+    input_checksums: Mapped[Optional[Dict[str, Any]]] = mapped_column(
+        JSON, default=dict
+    )
```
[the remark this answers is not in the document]
tedil: But `workdir` changes the directory the .snakemake folder and its children are placed in, including .snakemake/metadata(.db).

src/snakemake/persistence/file.py
```diff
@@ -0,0 +1,261 @@
...
+                return
+
+    def _delete_locks(self) -> None:
+        for lockfile in self._lockfile.values():
+            try:
+                os.remove(lockfile)
+            except OSError as e:
+                if e.errno != 2:
+                    raise e
+        self._lockfile.clear()
```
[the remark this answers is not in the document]
tedil: `_write_locks` is invoked only exactly once each for "input" and "output" (in `lock`), so unlikely for that to happen

tedil:
> One immediate request without yet having checked the code: we should include here a test case that ensures that when running with a remote executor (or when having a spawned local job with a run directive), only the main process writes to persistence. Otherwise, we risk unnecessary race conditions. I think such additional non main process writes are completely unnecessary, but it could be that they still happen (as a bug). So ensuring this via a test seems like a good idea.
> 
> Below is a list of tests that simulate remote execution (for inspiration or reuse):
> 
> ```
> tests/test_checkpoint_allowed_rules/qsub   tests/test_groups_out_of_jobs/qsub  tests/test_issue930/qsub                     tests/test_singularity/qsub
> tests/test_default_storage_local_job/qsub  tests/test_group_temp/qsub          tests/test_multicomp_group_jobs/qsub         tests/test_spaces_in_fnames/qsub
> tests/test_groupid_expand_cluster/qsub     tests/test_group_with_pipe/qsub     tests/test_nodelocal/qsub                    tests/test_storage_cleanup_local/qsub
> tests/test_group_job_fail/qsub             tests/test_issue1037/qsub           tests/test_resource_quoting/qsub             tests/test_storage_localrule/qsub
> tests/test_group_jobs_attempts/qsub        tests/test_issue850/qsub            tests/test_restartable_job_cmd_exit_1/qsub   tests/test_wrapper/qsub
> tests/test_group_jobs/qsub                 tests/test_issue860/qsub            tests/test_restartable_job_qsub_exit_1/qsub
> tests/test_group_jobs_resources/qsub       tests/test_issue912/qsub            tests/test_shadow_prefix/qsub
> ```

Yes, agreed.

I have gone ahead and addressed the comments by the rabbit, which seemed reasonable, should now be ready to review ;)

tedil:
> One immediate request without yet having checked the code: we should include here a test case that ensures that when running with a remote executor (or when having a spawned local job with a run directive), only the main process writes to persistence.

I tried running two of the suggested tests (and just looked at os.getpid and threading.get_ident at the time when _write_record is called) and they all report the same pid and tid, so not sure that's helpful? Or I didn't run them correctly…

A quick test to just spawn a couple hundred workers with 1000 jobs each seems to be unproblematic though, sqlite doesn't run into any issues with that, though may not be the case on network filesystems, which we cannot reasonably test.

src/snakemake/persistence/db.py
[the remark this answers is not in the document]
tedil: you're right, in case users are using a central database, we'd have to be able to differentiate them somehow; if they have their own db user, that would already help, but then again they may have multiple workflows (potentially the same on multiple times, or just in different workdirs). so perhaps having some more discriminant, probably the workdir, would indeed make sense!

tedil:
additionally included a `namespace` (which is just the absolute path of the `.snakemake` dir) so if anyone ever uses a central DB to store the metadata, they don't run into issues if any of their workflows produce the same output file(-name)s.

src/snakemake/settings/types.py
```diff
@@ -135,6 +135,8 @@ class WorkflowSettings(SettingsBase):
         default_factory=dict
     )
     runtime_source_cache_path: Optional[Path] = None
+    persistence_backend: str = "file"
```
johanneskoester: Can you put an enum for this in settings/enums.py?

src/snakemake/cli.py
```diff
@@ -1571,6 +1571,21 @@ def get_argument_parser(profiles=None):
         action="store_true",
         help="Write extended benchmarking metrics.",
     )
+    group_behavior.add_argument(
+        "--persistence-backend",
+        choices=["file", "db"],
```
johanneskoester: The enum can deliver the choices here.

src/snakemake/persistence/db.py
```diff
@@ -0,0 +1,259 @@
...
+
+from snakemake.persistence import MetadataRecord, PersistenceBase
+import snakemake.exceptions
+
+
+class Base(DeclarativeBase):
+    pass
+
+
+class MetadataRecordORM(Base):
```
johanneskoester: This is quite redundant? Why not let `persistence.MetadataRecord` inherit from SQLModel instead?

tedil: (initially because I wanted to keep number of additional dependencies low)
now using sqlmodel ;)

src/snakemake/settings/types.py
tedil: Done, good idea

cademirch:
> > One immediate request without yet having checked the code: we should include here a test case that ensures that when running with a remote executor (or when having a spawned local job with a run directive), only the main process writes to persistence.
> 
> I tried running two of the suggested tests (and just looked at os.getpid and threading.get_ident at the time when _write_record is called) and they all report the same pid and tid, so not sure that's helpful? Or I didn't run them correctly…
> 
> A quick test to just spawn a couple hundred workers with 1000 jobs each seems to be unproblematic though, sqlite doesn't run into any issues with that, though may not be the case on network filesystems, which we cannot reasonably test.

To add, in my old testing I ran into issues with subprocesses writing to persistence. It shouldn't occur now, but any workflow with `run` directive will use subprocesses so worth running those tests as well

tedil:
I'll be adding one more test, sorry ^^

tedil:
@cademirch added a test that uses `run:`, not sure it works well in CI (cores=20 will probably get scaled down, right?), but at least covers potential issues a tiny little bit.
@johanneskoester please review most recent changes (which also fixed a nasty "whoops" where I forgot to update the cli default to use the enum… ^^)

PR #4020, opened 2026-03-10
feat: cli refactor and add subcommands

cademirch:
early wip for migrating the snakemake cli from argparse to click. this will be a gradual process spread across multiple PRs — the goal is to keep the current argparse cli fully functional while we evaluate the soundness of this implementation. 

 **motivation:**
 - the current cli is a single ~2000-line main() function in cli.py that builds one massive argparse parser. every flag shares one namespace, which makes it hard to reason about which options apply to which
 operation.
 - subcommands are a natural fit — snakemake lint, snakemake unlock, snakemake run etc. each get their own focused set of options instead of everything being top-level flags.
 - the argparse parser has grown organically and is tightly coupled to the execution logic. separating "parse args" from "do the thing" makes both sides easier to maintain and test.

 **why [click](https://click.palletsprojects.com/en/stable/):**
 - click has mature support for nested subcommand groups, which is exactly what we need. argparse can do subcommands but it can be a bit awkward.
 - i considered [typer](https://typer.tiangolo.com/). but we don't need the pydantic-style type inference here. typer is built on click anyway.
 - click's group/command model lets us incrementally register subcommands without touching the legacy parser. the `invoke_without_command` + fallthrough pattern means we can migrate one command at a time with hopefully minimal (if not _none_) breakage.
 - profile/config file loading maps cleanly to click's `default_map` context mechanism.
 
**approach:**
 - a ProfileAwareGroup click group sits in front of the existing argparse main(). any arg that isn't a registered click subcommand falls through to the legacy parser, so nothing breaks.
 - new subcommands call the snakemake API directly (no argparse roundtrip)
 - `cli.py` gets refactored to extract reusable run_*() functions that both the old and new paths can share fc32676c0b4b3082dd8fa351c2174ccca5e133d7
 
**what works right now:**
 - `snakemake lint` — runs linting via click
 - `snakemake lint -s path/to/Snakefile` — lint a specific file
 - `snakemake unlock` — removes the workflow lock via click
 - `snakemake unlock -d /some/workdir` — unlock with a custom workdir
 - `snakemake --lint`, `snakemake --unlock` — still work through the legacy fallthrough
 - `snakemake --help` — still shows the full argparse help

Overall, the plan is to keep merging small incremental pieces so we don't end up with a massive migration PR (like the logging refactor PR 😅 ). the legacy cli stays as the default until we have enough coverage and confidence to migrate. 

Would appreciate comments, thoughts, and feedback - especially in terms of how to go about this big of a change in a responsible and reasonable way. Testing would be helpful too. 

### TODOs

* [ ] ensure that shell completion works with the new implemention (cf. #3998 by @jsquaredosquared)


schnea:
I think it would probably make sense to create a folder `cli/` where each subcommand gets its own file? 
If we could already create a skeleton structure for that, it would also enable people to work on different subcommands and merge things in bit by bit (<- distracted myself into really wanting to write some sort of  `snakemake new` to scaffold a new project)

cademirch:
> I think it would probably make sense to create a folder `cli/` where each subcommand gets its own file? If we could already create a skeleton structure for that, it would also enable people to work on different subcommands and merge things in bit by bit (<- distracted myself into really wanting to write some sort of `snakemake new` to scaffold a new project)

working on this!

cademirch:
**update: package restructure + tests (a3f9773)**

  this commit delivers the `cli/` folder structure that was discussed — each subcommand now gets its own file, making it easy for people to work on different subcommands in parallel.

  **what changed:**

  - **`cli.py` → `cli/legacy.py`**: the existing argparse cli moved into the package. `cli/__init__.py` re-exports everything via `from snakemake.cli.legacy import *`, so existing imports like `from snakemake.cli
  import parse_batch` should continue to work
  - **`cli_click.py` → `cli/click.py`**: the click entry point moved into the package. helper functions (`_extract_flag`, `_find_workflow_profile`, `_load_profile_defaults`) stay here since they're specific to the
  click group, not shared. subcommand definitions moved out — they're now registered with `cli.add_command()`.
  - **`cli/commands/lint.py` and `cli/commands/unlock.py`**: each subcommand in its own file. uses the new `SnakemakeContext` instead of manually managing `SnakemakeApi` lifecycle.
  - **`cli/common.py`**: new `SnakemakeContext` class that bridges click options → snakemake API. it accumulates settings and provides `.workflow()` / `.dag()` context managers, so subcommands don't need to repeat
  the boilerplate of constructing `SnakemakeApi`, `OutputSettings`, `ResourceSettings`, etc. this is where shared options and settings logic will live as we add more subcommands.
  - **`cli/operations.py`**: the `run_*()` functions extracted from the old `cli.py`. these are thin wrappers around the snakemake API — no CLI concerns, just "do the thing". both the legacy argparse path and new
  click subcommands import from here, so there's a single source of truth for each operation. added type hints and docstrings.
  - **`pyproject.toml` + `__main__.py`**: import paths updated to `snakemake.cli.click:cli`.

  **new: `tests/test_cli.py`**:
  - *legacy fallthrough*: verifies that bare `snakemake`, unknown flags (`--dry-run`), positional targets (`all`), and legacy-style flags (`--lint`, `--unlock`) all fall through to the argparse `legacy_main`
  - *subcommand routing*: confirms `snakemake lint` routes to click (not legacy), while `snakemake --lint` still falls through
  - *lint subcommand*: tests default invocation, `-s` and `-d` options, and exit code 1 on lint failure
  - *unlock subcommand*: tests default invocation, `-s` and `-d` options, and verifies the DAG API gets called
  - mocks `SnakemakeApi` so tests run fast without needing a real snakefile

  **file layout now looks like:**
```
  src/snakemake/cli/
  ├── init.py          # re-exports from legacy for back compat
  ├── click.py             # ProfileAwareGroup + cli entry point
  ├── common.py            # SnakemakeContext (shared settings bridge)
  ├── commands/
  │   ├── init.py
  │   ├── lint.py          # snakemake lint
  │   └── unlock.py        # snakemake unlock
  ├── legacy.py            # existing argparse main()
  └── operations.py        # run_lint(), run_unlock(), etc.
```

  **adding a new subcommand is now:**
  1. create `cli/commands/mycommand.py` with a `@click.command()` function
  2. use `SnakemakeContext` for API access
  3. call the corresponding `run_*()` from `operations.py`
  4. register it in `cli/click.py` with `cli.add_command(mycommand)`

this structure should make it straightforward for multiple people to contribute subcommands without merge conflicts. re: `snakemake new` — this would slot in nicely as `cli/commands/new.py` now 😄

still much todo! but this should be fairly safe to merge soon.

cademirch:
more progress here! mapped out the remaining operations from the legacy cli into logical subcommand groups. each group gets its own file in commands/, same pattern as lint and unlock.

here's what was added:
- `snakemake info` — workflow inspection stuff. nested subcommands for rules, target-rules, summary (with --detailed), changes (takes code/input/params as an argument), untracked, and compilation. the ones that just inspect rules/compilation only need a workflow_api, the rest go through dag_api.
- `snakemake dagviz` — prints the job DAG by default. --rulegraph, --filegraph, and --d3 flags switch the output format.
 - `snakemake clean` — cleanup operations. --all and --temp for deleting output files (with --dry-run support), --shadow for stale shadow dirs, --metadata FILE for removing tracking metadata. requires at least one flag.
- `snakemake software` — nested group for managing deployment environments. snakemake software conda list/cleanup/create for conda envs, snakemake software container cleanup for container images. felt cleaner than cramming these all into flags on a single command.
- `snakemake utils` — grab bag for things that don't fit elsewhere. containerize, generate-unit-tests (takes an optional path argument, defaults to .tests/unit), and archive (takes a required file argument).

all commands follow the same pattern — construct a SnakemakeContext, open the appropriate context manager, call the operation from operations.py. tests cover each command, flag, and argument, plus help text verification for every group and subcommand.

skipped report for now since it needs work on plugin settings passthrough — left a TODO in the file.
the full command tree now looks like:

```bash
snakemake lint
snakemake unlock
snakemake info rules|target-rules|summary|changes|untracked|compilation
snakemake dag [--rulegraph|--filegraph|--d3]
snakemake clean [--all|--temp|--shadow|--metadata FILE]
snakemake software conda list|cleanup|create
snakemake software container cleanup
snakemake utils containerize|generate-unit-tests|archive
```

edit: future cli help could look like this:
```
Usage: snakemake [OPTIONS] COMMAND [ARGS]...

  Snakemake workflow engine.

Commands:
  clean   Remove output files, shadow dirs, or metadata.
  dagviz  Print the workflow DAG for visualization.
  help    Show available subcommands.
  info    Inspect workflow rules, changes, and output files.
  lint    Check the workflow for common issues.
  unlock  Remove the workflow lock.
  utils   Miscellaneous workflow utilities.
  ```

**Comment by schnea** [comment]
The subcommands all run properly, but Whenever I try to run a workflow with `pixi run -e dev snakemake --cores 1`, I get this error: 
```python
  File "/home/aal/snakemake/.pixi/envs/dev/bin/snakemake", line 10, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/aal/snakemake/.pixi/envs/dev/lib/python3.12/site-packages/click/core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aal/snakemake/.pixi/envs/dev/lib/python3.12/site-packages/click/core.py", line 1405, in main
    with self.make_context(prog_name, args, **extra) as ctx:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aal/snakemake/src/snakemake/cli/click.py", line 80, in make_context
    return super().make_context(info_name, args, parent=parent, **extra)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aal/snakemake/.pixi/envs/dev/lib/python3.12/site-packages/click/core.py", line 1216, in make_context
    self.parse_args(ctx, args)
  File "/home/aal/snakemake/src/snakemake/cli/click.py", line 93, in parse_args
    ctx.protected_args = []
    ^^^^^^^^^^^^^^^^^^
AttributeError: property 'protected_args' of 'Context' object has no setter```

**Comment by cademirch** [comment]
> The subcommands all run properly, but Whenever I try to run a workflow with `pixi run -e dev snakemake --cores 1`, I get this error:
> 
> ```python
> Traceback (most recent call last):
>   File "/home/aal/snakemake/.pixi/envs/dev/bin/snakemake", line 10, in <module>
>     sys.exit(cli())
>              ^^^^^
>   File "/home/aal/snakemake/.pixi/envs/dev/lib/python3.12/site-packages/click/core.py", line 1485, in __call__
>     return self.main(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/aal/snakemake/.pixi/envs/dev/lib/python3.12/site-packages/click/core.py", line 1405, in main
>     with self.make_context(prog_name, args, **extra) as ctx:
>          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/aal/snakemake/src/snakemake/cli/click.py", line 80, in make_context
>     return super().make_context(info_name, args, parent=parent, **extra)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/aal/snakemake/.pixi/envs/dev/lib/python3.12/site-packages/click/core.py", line 1216, in make_context
>     self.parse_args(ctx, args)
>   File "/home/aal/snakemake/src/snakemake/cli/click.py", line 93, in parse_args
>     ctx.protected_args = []
>     ^^^^^^^^^^^^^^^^^^
> AttributeError: property 'protected_args' of 'Context' object has no setter```
> ```

can you do `pixi shell -e dev` and then `pip show click`?

edit: i figured it out. i had `click==8.1.8` locally, and `protected_args` is gone in `8.2`. conda seems to be pinning click to 8.1.8 for some platforms unfortunately.

**Comment by cademirch** [comment]
big progress since the initial POC. the subcommand tree:

```
snakemake run [TARGETS] [OPTIONS]
snakemake lint [-s FILE] [-d DIR]
snakemake unlock [-s FILE] [-d DIR]
snakemake clean [--all|--temp|--shadow|--metadata]
snakemake dagviz [--rulegraph|--filegraph|--d3]
snakemake info rules|target-rules|summary|changes|untracked|compilation
snakemake report [--reporter PLUGIN]
snakemake software conda list|cleanup|create
snakemake software container cleanup
snakemake utils containerize|generate-unit-tests|archive
snakemake plugins list [-t TYPE] [-v]
snakemake profile list [-v]
```

legacy fallthrough is fully intact — `snakemake --cores 4` etc. still works through argparse.

## run command

option groups are decorators in `common.py`, each mapping 1:1 to a settings dataclass (`resource_options` → `ResourceSettings`, etc.). the command body parses raw CLI values using the same parse functions from legacy, constructs the settings objects, and calls the API. `--executor`/`--scheduler`/`--logger`/`--reporter` choices are populated from plugin registries.

profiles are scoped to `run` only via `ProfileAwareCommand`, which pre-scans for `--profile` and uses click's `default_map` for precedence. the group itself is now a plain `FallthroughGroup` with no profile logic. i'll make sure to stay updated with #4087

## plugin adapter

`plugin_adapter.py` converts plugin settings dataclasses → click options using the existing `dataclass_field_to_argument_args` from `snakemake-interface-common`. zero changes to the interface package or any plugin. `plugin_settings_from_kwargs` reconstructs settings via `SimpleNamespace` so `PluginBase.get_settings()` works unchanged.

## new subcommands

- `plugins list` — shows installed plugins by type, install paths, and optionally their CLI options

## design decisions

- **decorators collect raw values, `run` constructs settings**: we considered having decorators produce settings objects directly, but cross-cutting concerns (dryrun affects executor choice AND output settings) break the pattern. explicit parsing in `run` is more verbose but fully traceable.
- **profiles on `run` only**: profiles are an execution concern. `snakemake lint` doesn't need  profile.


## pruning deprecated/superseded flags

since `run` is a fresh command, we have a clean slate. flags that are deprecate are notcarried over to `run` they remain available through the legacy fallthrough only. for example:

- `--use-conda`, `--use-apptainer`, `--use-envmodules` was superseded by `--sdm conda|apptainer|env-modules`

I'm also thinking of things like:
--edit-notebook, --draft-notebook, --notebook-listen these could be a subcommand
--export-cwl
--target-files-omit-workdir-adjustment
--restart-times — alias for --retries, just confusing to have both
--output-wait — alias for --latency-wait
--dryrun (no hyphen) — alias for --dry-run, inconsistent style
--configfiles — alias for --configfile, plural is misleading since multiple=True handles it

the legacy fallthrough keeps all of these working. users get clear migration paths 

open question: are there others we should prune? 

## snakedeploy?
now that we have space for subcommands, i think it could be worthwhile to expose snakedeploy's functionality via the snakemake cli. for example `snakemake plugins scaffold <plugin_type>`, or `snakemake deploy ...`. snakedeploy can still exist separetly and become a dependncy. i think this would be great for discoverability and makes sense that the snakemake cli does this. 

## TODOs

- refactor existing subcommands to use `@workflow_options`
- tests for `run`, `report`, `plugins list`
- shell completion (cf. #3998)
- move parse functions from `legacy.py` to `cli/parsing.py`

pdimens:
I'm only seeing this now and think this is a really good and sensible CLI refactoring for both user- and dev-facing sides of snakemake. Since the next version is 10.0 and breaking, it would be ideal to revisit this and consider it for inclusion in 10.0

cademirch:
> I'm only seeing this now and think this is a really good and sensible CLI refactoring for both user- and dev-facing sides of snakemake. Since the next version is 10.0 and breaking, it would be ideal to revisit this and consider it for inclusion in 10.0

Thanks for reminding me of this! I do agree it would be nice to get this in with 10.0. My time is a little fragmented these days but I'll see what I can do! 

@johanneskoester iirc we last talked about this shortly after the hackathon. We had discussed using ddargparse to try and more directly use the settings data classes for building the cli (somewhat like clap derive). I'll have to go dig up the branch and refresh myself on it, but from what I can remember, I preferred the click version because of its more intuitive handling of sub commands

PR #4022, opened 2026-03-10
test: add test that  #3913 is fixed by #4026

### LINKED ISSUE #3913: Greedy scheduler uses more cores than provided
**Snakemake version**

9.14.5

**Describe the bug**

The greedy scheduler sometimes schedules jobs in parallel such that the total number of threads exceeds the provided number of cores. In my case, I ran snakemake with `-c 8`, and two jobs using 8 cores each were scheduled, using 16 cores in total.

**Minimal example**

Create a Conda environment with `conda create -n tmp snakemake-minimal=9.14.5`.

Create this Snakefile:
```python
rule final:
  input: "file1", "file2"

rule prepare:
  output: "common"
  threads: 8
  shell:
    "echo Using {threads} threads; touch {output}"

rule do_something:
  output: "file{number}"
  input: "common",
  threads: 8
  shell:
    "echo Using {threads} threads; sleep 5; touch {output}"
```

Run
```
snakemake -c 8 --scheduler=greedy
```

The log output shows that both `do_something` jobs are started in parallel, consuming 16 threads.

**Additional context**

* If the `common` input file already exists (that is, if the `prepare` job does not need to run), then the two `do_something` jobs are run serially as expected.
* I have not tried with the ILP solver. I use the greedy scheduler because I get `Failed to solve scheduling problem with ILP solver, falling back to greedy scheduler.`
* The first Snakemake version with the problem is 9.10.0. The preceding release 9.9.0 behaves as expected. Looking at the changelog for 9.10.0, this likely has to do with #3676.
* The machine I am running this on has 16 cores including hyperthreading. If I write `threads: 2` in the `do_something` rule and add "file3" up to "file8" to the input of the "all" rule, I can get the scheduler to run 8 jobs at the same time.

edbennett:
Before this change, in some cases jobs were started without calling `update_available_resources` to subtract the resources used from the resource pool. Regardless, when the job completed, the resources were re-added to the resource pool.
This change ensures that the resource pool is always updated, and slightly rearranges the logic to avoid duplicating the call to `job_scheduler_greedy`.

The PR also includes a test that failed before the change, and now passes; it tests the behaviour described in the original issue (resources over allocated), rather than the specific bug fixed above.

fixes #3913

### QC

## Summary by CodeRabbit

* **Tests**
  * Added test suite validating thread allocation and resource constraints during parallel workflow execution
  * Test verifies thread usage remains within configured limits across concurrent tasks
  * Validates job scheduling behavior and completion tracking with multi-threaded rule execution

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -2962,6 +2962,8 @@ def test_github_issue3913():
         cleanup=False,
     )
 
+    started_job_count = 0
+    completed_job_count = 0
     active_job_thread_counts = {}
     with open(next((tmpdir / ".snakemake/log").glob("*.log"))) as logfile:
         for line in logfile:
@@ -2972,10 +2974,12 @@ def test_github_issue3913():
                 thread_count = int(line.split()[-1])
             elif line.startswith("    resources:"):
                 active_job_thread_counts[current_jobid] = thread_count
+                started_job_count += 1
             elif line.startswith("Finished jobid:"):
                 finished_jobid = line.split()[2]
                 assert finished_jobid in active_job_thread_counts
                 del active_job_thread_counts[finished_jobid]
+                completed_job_count += 1
 
             if not active_job_thread_counts:
                 # No active jobs trivially passes the test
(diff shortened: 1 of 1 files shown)


johanneskoester:
@edbennet, I just merged a PR by @fgvieira (#4026) that fixes the same issue. We decided that the advantage of not using the complex scheduler (LP) for just a single job is so marginal that it is not worth the additional code. However, your test case is still worth adding! Can you maybe resolve the conflict and then ping me via discord?

johanneskoester:
Thank you so much!

PR #4033, opened 2026-03-10
fix: reduce number of tests leaving temporary files behind

edbennett:
Currently, many tests in the default test suite fail to clean up after themselves. On CI this doesn't really matter as the VM will be destroyed once the run completes, but it's mildly annoying for local development.

This PR currently cleans up the low-hanging fruit: either removing unneeded `cleanup=False` options presumably left over from debugging, or adding explicit cleanup.

A few cases are harder to deal with: the following tests cause the call to `run()` to raise an exception:

- `test_github_issue1158`
- `test_ambiguousruleexception`
- `test_wildcard_annotatedstrings`

In this case, `run()` never reaches the cleanup code, and the calling function doesn't (and can't) know the temporary path used to be able to clean up.

I can see a few options on how to work around this:

1. Ugly hack: make an extra call to `run()` but requesting a no-op target to generate the temporary directory, then pass this in to the current call to `run()` to perform the test; after generating the error, then remove the temporary directory.
2. Move the logic in `run()` that creates the temporary directory to a separate helper function, and call this explicitly from the error-producing tests, then proceed as in (1) above.
3. Refactor the temporary directory functionality to be a class that manages the temporary directory, and deletes it when the instance is destroyed—this could be just directly using `tempfile.TemporaryDirectory`, or reimplementing similar functionality.

@johanneskoester Do you have a preference (one of these three, or something else entirely)?

### QC

In principle one could add a test case that tests do not leave litter, but having tests for your tests seems a little excessive.

## Summary by CodeRabbit

* **Tests**
  * Added a reusable temp-directory helper to support both in-place and isolated test modes.
  * Standardized temp-dir resolution and result-path handling across tests, with clearer behavior when running without a temp dir.
  * Restored and unified cleanup behavior (including platform-aware handling), ensuring temporary artifacts are removed on success and failure.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -2036,7 +2036,7 @@ def test_strict_mode():
 def test_github_issue1158():
     path = dpath("test_github_issue1158")
     with prepare_tmpdir(path) as tmpdir:
-        run(path, cluster="./qsub.py", tmpdir=tmpdir)
+        run(path, cluster="./qsub.py", tmpdir=tmpdir, cleanup=False)
 
 
 def test_ancient_dag():


johanneskoester:
> I can see a few options on how to work around this:
> 
>     1. Ugly hack: make an extra call to `run()` but requesting a no-op target to generate the temporary directory, then pass this in to the current call to `run()` to perform the test; after generating the error, then remove the temporary directory.
> 
>     2. Move the logic in `run()` that creates the temporary directory to a separate helper function, and call this explicitly from the error-producing tests, then proceed as in (1) above.
> 
>     3. Refactor the temporary directory functionality to be a class that manages the temporary directory, and deletes it when the instance is destroyed—this could be just directly using `tempfile.TemporaryDirectory`, or reimplementing similar functionality.

I think best is to create the tempdir outside of run in these cases, using the tempfile.TemporaryDirectory context manager, and pass it to run via `run(..., tmpdir=...)`.

tests/common.py
```diff
@@ -141,6 +141,25 @@ def get_expected_files(results_dir: StrPath) -> list[str]:
...
+    if not dest_path:
+        # If we need to further check results, we won't cleanup tmpdir
+        tmpdir = next(tempfile._get_candidate_names())
+
+        if snakefile_parent_dirname is None:
+            snakefile_parent_dirname = source_path.name
+        dest_path = os.path.join(
+            tempfile.gettempdir(), f"snakemake-{snakefile_parent_dirname}-{tmpdir}"
+        )
+        os.mkdir(dest_path)
```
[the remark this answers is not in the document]
edbennett: This is mirroring the preexisting implementation; I'm not touching it.

tests/tests.py
```diff
@@ -1782,12 +1782,14 @@ def test_parsing_terminal_comment_following_statement():
...
 def test_github_issue640():
-    run(
+    tmpdir = run(
         dpath("test_github_issue640"),
         targets=["Output/FileWithRights"],
         executor="dryrun",
         cleanup=False,
     )
+    shell(f"chmod -R u+rwX {tmpdir}")
+    shutil.rmtree(tmpdir)
```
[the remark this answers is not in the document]
edbennett: If line 1791 fails, line 1792 will also fail because the files won't be `chmod`ded to be able to be deleted.

tests/common.py
```diff
@@ -141,6 +141,25 @@ def get_expected_files(results_dir: StrPath) -> list[str]:
     ]
 
 
+def prepare_tmpdir(source_path, snakefile_parent_dirname=None, dest_path=None):
```
johanneskoester: The exposed version of this function, couldn't it be a context manager, so that not every test has to do the cleanup with rmtree?
https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

edbennett: It could! I had discounted the option because we can't use the `TemporaryDirectory` context manager directly (without either breaking compatibility with Python < 3.12, or with use cases where developers want a directory to stay intact after the test exits), but thinking about it more, that would make sense. I'll work on that now.

edbennett: I've pushed changes that implement this for the exposed version. I'd hoped that `run()` could also be a context manager to simplify the code in `tests.py` and `tests_using_conda.py`, but my attempt at this broke a bunch of the tests.

PR #4035, opened 2026-03-10
feat: add support for typed output in checkpoints and rules

### LINKED ISSUE #3482: Add ability to use typed outputs
```python
@dataclass
class MyType:
    some_threshold: float
    files: Mapping[str, Path]


# generate typed output 
checkpoint a:
    output:
        files=directory("results/something/{dataset}"),
        meta=typed("results/metadata/something/{dataset}.json", MyType)
    run:
        output.meta(some_threshold=0.3, files={...}).write()


# access typed output directly
rule b:
    input:
        rules.a.output.files,
        meta=rule.a.output.meta,
    output:
        "results/b.txt"
    run:
        ...
        input.meta.some_threshold


# process individual items
rule c:
    input:
        "results/something/{dataset}/{item}.txt"
    output:
        "results/processed/{dataset}/{item}.txt"
    shell:
        ...


# scatter over items in the typed output
def get_processed(wildcards):
    meta = checkpoints.get(**wildcards).output.meta
    for file in meta.files:
        yield file


rule d:
    input:
        get_processed
    output:
        "results/reduced/{dataset}.txt"
    shell:
        ...

```

Hocnonsense:
will close #3482

### QC

* [x] The documentation (`docs/`) is updated to reflect the changes.

## Summary by CodeRabbit

* **New Features**
  * Added `typed()` function to serialize and deserialize checkpoint outputs with automatic format detection
  * Support for multiple formats: JSON, YAML, CSV, Pickle, NumPy, and more
  * Compression support (gzip, bzip2, xz) for serialized outputs
  * Custom loader/dumper capabilities for advanced use cases

* **Breaking Changes**
  * Removed `.open()` method on checkpoint output files; use Python's `open()` instead

* **Documentation**
  * Added comprehensive guide for using the `typed()` function with checkpoints

### DIFF
--- a/src/snakemake/io/resolve_file_format.py
+++ b/src/snakemake/io/resolve_file_format.py
@@ -80,7 +80,13 @@ def resolve_file_format(
                 )
         case "pkl" | "pickle":
             import pickle
+            from snakemake.logging import logger
 
+            logger.warning(
+                "Never unpickle data that could have come from an untrusted source, "
+                "or that could have been tampered with. "
+                "Consider using json/yaml/toml instead."
+            )
             return (
                 lambda f: pickle.load(open_(f, "rb")),  # type: ignore[arg-type]
                 lambda obj, f: pickle.dump(obj, open_(f, "wb")),  # type: ignore


btraven00:
very nice feature! As discussed with @Hocnonsense, it'd be very interesting to use this to type-check different segments of a reusable pipeline for correctness (before materializing the files even, just by matching inputs and outputs). From this implementation, maybe it'd be good to declare what's the expected contract in the type class that is passed as an annotation of the output.

docs/tutorial/additional_features.rst
```diff
@@ -223,4 +223,62 @@ This mechanism helps to solve two kinds of ambiguity.
...
+::::::::::::::::::::::::::::
+
+The ``typed`` function wraps an output file path with a specific type, enabling structured serialization and deserialization of typed objects directly within Snakemake rules.
+
+Declare a output data structure and reference it in a rule (or checkpoint rule)::
+
+    @dataclass
+    class MyType:
+        some_threshold: float
+        files: Mapping[str, str]
```
johanneskoester: Maybe, in this example, use something else than file paths for MyType content. I don't want people to become tempted to add additional output files nested into those objects instead of reporting them toplevel in the output section of the rule.

docs/tutorial/additional_features.rst
```diff
@@ -223,4 +223,62 @@ This mechanism helps to solve two kinds of ambiguity.
...
+            meta=typed("metadata/{dataset}.json", MyType)
+        run:
+            output.meta.dump(some_threshold=0.3, files={"a": "path/to/a"})
+
+The ``dump`` method constructs an instance of ``MyType`` from the given keyword arguments and serializes it to the declared file path as JSON.
+The type passed to ``typed`` can be a ``dataclass``, a ``NamedTuple``, or any class that implements an ``.asdict() -> dict[str, Any]`` method.
+
+.. note::
+   The structured data is serialized as JSON, so field values should be JSON-native types (``str``, ``int``, ``float``, ``bool``, ``list``, ``dict``), or types with built-in coercion support such as ``Path``.
+   Runtime type checking is not enforced; use a validated type like ``pydantic.BaseModel`` if strict validation is required.
```
johanneskoester: Add a linkout to pydantics BaseModel here.

docs/tutorial/additional_features.rst
```diff
@@ -223,4 +223,62 @@ This mechanism helps to solve two kinds of ambiguity.
...
+              shell("ls {input.meta}")
+
+Scattering over typed output
+.............................
+
+When combined with checkpoints, ``typed`` enables structured access to deserialized objects in input functions.
+This pattern avoids parsing raw text files and maintains a typed interface between the checkpoint and its consumers::
+
+    def get_processed(wildcards):
+        meta = checkpoints.a.get(**wildcards).output.meta.value
```
johanneskoester: I think here it would be preferable to have the same syntax as within the run directive above. I.e.


```suggestion
        meta = checkpoints.a.get(**wildcards).output.meta.load()
```

src/snakemake/io/__init__.py
```diff
@@ -1794,18 +1821,10 @@ def __init__(
             if custom_map is not None:
                 self.extend(map(custom_map, toclone))
             elif plainstr:
+                # After this, file will be converted to plain string.
+                # Exception: if the file is typed, should be converted to instance of assigned class
```
johanneskoester: I think we should keep interactions with typed IO consistent and always require to use load() to access the typed value. Thus plainstr should actually not convert the typed IO into the typed value but rather return an str inheriting object that exposes one additional method load() to get the typed value. It should be a special object (not IOFile, as that exposes too many unnecessary internals).

docs/tutorial/additional_features.rst
Hocnonsense: Sure, updated.

docs/tutorial/additional_features.rst
Hocnonsense: Sure, updated in docs.

The current implementation has:
- `value` property: shorthand for `.load()`
- `__getattr__`: transparently proxies attribute access to the deserialized object,
  allowing `input.meta.some_threshold` without an explicit `.load()` call

should these two features also removed from `TypedFile`?

src/snakemake/io/__init__.py
Hocnonsense: Sure, `TypedFile` now inherits from `str`, so it works as a plain path string in all contexts while exposing `load()` and `dump()` as additional methods.

One limitation: since `_plainstrings()` is called uniformly for `input`, `output`, and `log`, there is no clean place to distinguish which side a `TypedFile` comes from. As a result, both `load()` and `dump()` are available on all `TypedFile` instances regardless of context. In practice `dump()` is only meaningful on output/log files, but this cannot be enforced at the type level without a more invasive split of the plainstrings path. Happy to explore that direction if you think the stricter separation is worth the added complexity.

johanneskoester: This is fine!

docs/tutorial/additional_features.rst
johanneskoester: Yes, I would say remove the two. Rather only support one way. This also makes it explicit that load requires computation, so that people are guided towards storing the loaded object in a variable instead of loading multiple times.

docs/tutorial/additional_features.rst
Hocnonsense: Sure. I've changed the example to use `filestems: list[str]` instead of `files: Mapping[str, Path]`, so the typed object only carries stems rather than full paths.

docs/tutorial/additional_features.rst
```diff
@@ -223,4 +223,64 @@ This mechanism helps to solve two kinds of ambiguity.
 
 When dealing with ambiguous rules, it is best practice to first try to solve the ambiguity by using a proper file structure, for example, by separating the output files of different steps in different directories.
 
-As follow-up to this tutorial, we recommend to have a look at the :ref:`interaction, visualization and reporting tutorial <interaction_visualization_reporting_tutorial>`, which focuses on Snakemake's ability to cover the last mile of data analysis, i.e., the generation of publication ready reports and interactive visualizations.
\ No newline at end of file
+As follow-up to this tutorial, we recommend to have a look at the :ref:`interaction, visualization and reporting tutorial <interaction_visualization_reporting_tutorial>`, which focuses on Snakemake's ability to cover the last mile of data analysis, i.e., the generation of publication ready reports and interactive visualizations.
+
+
+.. _tutorial-typed:
```
johanneskoester: I suggest to move the documentation of typed into rules.rst. For the tutorial, this is probably a bit too much. Further, non-beginners never look into the tutorial.

src/snakemake/io/typed.py
```diff
@@ -0,0 +1,172 @@
...
+            try:
+                _loader, typed_file._dumper = resolve_file_format(file)
+            except NotImplementedError:
+                if isinstance(loader, str):
+                    typed_file._dumper = _dumper
+                else:
+                    raise
+            if not loader_set:
+                typed_file._loader = _loader
+        return typed_file
```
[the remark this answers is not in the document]
Hocnonsense: it is designed to bypass the initiation of the declared class, enable sth like pd.read_csv and post_init modification, and fully controlled by the function.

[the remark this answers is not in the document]
Hocnonsense: for those who want to set dumper but keep loader unset, i think they already understand what they are doing. they should use loader to set both if they want to also set up dumper according to implicit agreement

Issue #1805, opened 2022-08-11
using `source_path` in params always triggers unneeded rerun


li-yq:
**Snakemake version**

Both `7.12.0` and `7.6.1`

**Describe the bug**

The returned path from `workflow.source_path` changes cross different executions of snakemake, even though the file does not change.
It causes param changes and triggers rerun of rules.

**Logs**

Using the [source_path test](https://github.com/snakemake/snakemake/blob/bfa0adaa90842f20e74273eb53cd75ec5c54d299/tests/test_source_path/workflow/Snakefile) as an example.

The first run:

```console
$ snakemake --version
7.12.0

$ snakemake -j -p
...
rule a:
    output: test.out
    jobid: 0
    reason: Missing output files: test.out
    resources: tmpdir=/tmp

cat /home/user/.cache/snakemake/snakemake/source-cache/runtime-cache/tmp_kcizqm8/6d573129aa3a1579b69b535c1b44b8db4c81ffdc5b20340dccb22d0576dcd362 > test.out
```

The second run:

```console
$ snakemake -j -p
...
rule a:
    output: test.out
    jobid: 0
    reason: Params have changed since last execution
    resources: tmpdir=/tmp

cat /home/user/.cache/snakemake/snakemake/source-cache/runtime-cache/tmp6tmmr3c9/6d573129aa3a1579b69b535c1b44b8db4c81ffdc5b20340dccb22d0576dcd362 > test.out
```

Note that the temp dir `~/.cache/snakemake/snakemake/source-cache/runtime-cache/tmpXXXXXXXX` has changed.

**Minimal example**

See the link above.

**Additional context**

simonwm:
I also stumbled over this bug which is causing massive reruns in my pipeline using snakemake 7.19.1
IMO this renders `workflow.source_path` unusable in practice, as it should affect its use in basically all directives.

The bug seems to be an interaction of https://github.com/snakemake/snakemake/pull/997 and https://github.com/snakemake/snakemake/issues/1694 : params changes as the caching path is always random and this triggers a re-evaluation of everything downstream.

Possible routes to fix it:
1) Have `workflow.source_path` return the path to the original source_file and not to the cached copy
2) Remove source caching completely
3) Defer the source path translation from original to cached to after the evaluation of which rules need to be rerun due to params changes (and probably all other reasons).

The least feature breaking option would be 3., but it is only implementable with deep insight into the package...

@johanneskoester What do you think?

[...]

Issue #2331, opened 2023-06-28
workflow.source_path does not work as expected


chrarnold:
**Snakemake version**
7.24

**Describe the bug**
In params, I have the following line to specify that the config file for multiqc is in the following location (which is specified relative to where the Snakefile is located):
`config    = workflow.source_path("tools/multiqc_config.yaml")`

However, workflow.source_path does not work as expected, an error is thrown  (see below)

**Logs**
`Error: Invalid value for '-c' / '--config': Path '/home/bla/.cache/snakemake/snakemake/source-cache/runtime-cache/tmpoe2bb5nk/file/{full_path}/tools/multiqc_config.yaml' does not exist.`

Here, I replaced the beginning of the full path with {full_path}.

Does it work differently than what I expected and understood from the documentation?

ftabaro:
I think there should be a switch to tell Snakemake whether to cache or not the given path.

ftabaro:
Here is a possible workaround: https://github.com/snakemake/snakemake/issues/1805#issuecomment-1851506675

Redmar-van-den-Berg:
This bug still occurs in Snakemake 8.24, and the workaround does not work for Snakemake modules, since using `workflow.basedir` then points to the Snakefile that imported the module. This leads to incorrect paths.

tsibley:
@Redmar-van-den-Berg Try using `workflow.current_basedir`, which reflects the basedir of the currently-being-evaluated Snakemake file (e.g. the module's `Snakefile` or some `*.smk` file) instead of the entrypoint Snakemake file like `workflow.basedir` does.

tsibley:
Oh, nevermind. :facepalm: While it's possible to build correct module-relative URLs with `workflow.current_basedir`, most places you'd want to use it don't accept a URL, only a file path. I had misremembered and thought Snakemake accepted URLs and local file paths interchangeably in most places. The URL→local file resolution is the second half of what `workflow.source_path()` does:

> This can be achieved by accessing their path via the `workflow.source_path`, which (a) computes the correct path relative to the current Snakefile such that the file can be accessed from any working directory, and (b) downloads remote files to a local cache:

So I guess `source_path` just needs fixing.

deto:
[...]

PR #4036, opened 2026-03-10
docs: Correct workflow.source_path() description in documentation

LKress:
This PR is intended to clarify the workflow.source_path() documentation and additionally ensures the intended functionality with two new tests. It thereby addresses the following issues: #1805, #2331.

In the documentation, the usage of workflow.source_path() is clarified by specifying what that when workflow.source_path() is used in `input` no rerun is triggered while the use in `params` causes a rerun of the given rule upon every start of the pipeline.

The newly introduced tests check:

* That workflow.source_path() correctly caches files that are at arbitrary positions in the filesystem
* That workflow.source_path() does not result in rule reruns if it is used in `input` but it does if it is used in `params`

### QC

## Summary by CodeRabbit

* **Documentation**
  * Clarified how files are mounted and available in workflows/containers; recommend using input with workflow.source_path to avoid automatic reruns caused by params
  * Updated examples and guidance to prefer input-based access over params for auxiliary source files, and noted source-path inputs aren’t mapped to default storage providers
* **Tests**
  * Added new tests and example workflows covering source_path usage, nested cases, and rerun behavior verification

### DIFF
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -3158,6 +3158,7 @@ This can be achieved by accessing their path via the ``workflow.source_path``, w
 .. note::
     Note that if such source paths are specified as input files, they are automatically considered to be non-storage files.
     This means that Snakemake will not try to map them to an eventually specified default storage provider (see :ref:`storage-support`).
+    Further, note that ``workflow.source_path`` should not be used from ``params:`` but only from ``input:``. The reason is that it returns a cached path that may change between Snakemake runs, thereby triggering spurious reruns if referred via ``params:`` (since Snakemake would think that the parameter has changed.
 
 .. _snakefiles-template-integration:
 


johanneskoester:
closes #1805, #2331.

docs/snakefiles/deployment.rst
```diff
@@ -459,7 +459,11 @@ Note that the isolation of jobs running in containers depends on the container e
 For example, Docker does not pass any host environment variables to the container, whereas Apptainer/Singularity passes everything.
 To override the default behaviour, consider using ``--apptainer-args`` or ``--singularity-args``, e.g. to pass ``--cleanenv``.
 
-Files that are mounted using `params` using `workflow.source_path` are also automatically available in the container. This is realized by mounting the snakemake cache in the container (/home/<user>/.cache/snakemake/snakemake/source-cache) where the sourced files will be cached. 
+Files that are mounted in ``input`` using ``workflow.source_path`` are also automatically available in the container. This is realized by mounting the snakemake cache in the container (/home/<user>/.cache/snakemake/snakemake/source-cache) where the sourced files will be cached. 
+
+.. note::
+    Files that are mounted in ``params`` automatically cause the rule to be rerun on each start, as the path to the cached file changes on each start. To avoid this, consider using ``workflow.source_path`` in ``input`` instead of ``params``.
```
johanneskoester:
```suggestion
    Files that are referred by this from ``params`` automatically cause the rule to be rerun on each start, as the path to the cached file can change between runs. To avoid this, use ``workflow.source_path`` in ``input`` instead of ``params``.
```

docs/snakefiles/rules.rst
```diff
@@ -3148,16 +3148,16 @@ This can be achieved by accessing their path via the ``workflow.source_path``, w
...
         shell:
-            "somecommand {params.json} > {output}"
-
-Note that if such source paths are specified as input files, they are automatically considered to be non-storage files.
-This means that Snakemake will not try to map them to an eventually specified default storage provider (see :ref:`storage-support`).
+            "somecommand {input.json} > {output}"
 
+.. note::
+    Note that if such source paths are specified as input files, they are automatically considered to be non-storage files.
+    This means that Snakemake will not try to map them to an eventually specified default storage provider (see :ref:`storage-support`).
```
johanneskoester, line 3160:
```suggestion
    This means that Snakemake will not try to map them to an eventually specified default storage provider (see :ref:`storage-support`).
    Further, note that ``workflow.source_path`` should not be used from ``params:`` but only from ``input:``. The reason is that it returns a cached path that may change between Snakemake runs, thereby triggering spurious reruns if referred via ``params:`` (since Snakemake would think that the parameter has changed.
```

PR #4044, opened 2026-03-11
fix: cast path to str in _replace_prefix (#3991)

### LINKED ISSUE #3991: `replace_prefix` with module breaks on `PosixPath`s
**Snakemake version**

This is with Snakemkake 9.5.1
It's an edge case that I don't have time to look into right now as I have been kept from upgrading due to an unrelated regression.
I just want to document this, feel free if you feel it is too minor to follow up on.
If I have time later, I can dig in and likely even submit a PR, but no promises as to when.

**Describe the bug**
If loading a worflow as a module without using `replace_prefix`, everything works fine.
However, when adding `replace_prefix` (no other changes), `snakemake` fails with (ultimately)

```
AttributeError: 'PosixPath' object has no attribute 'startswith'
```

**Logs**
No as of now.

**Minimal example**
Didn't have time to isolate the issue as for now I can do without `replace_prefix`.
Might get back to this later if required.

**Additional context**
My guess is that some implementation detail of `replace_prefix` assumes the `output` or `log` or whatever paths to be `str`, while the workflow is using Pathlib. If that's the case, the issue could probably be fixed by a (few) strategic `str(...)` call(s).

Sorry for not being able to provide a proper bug report but as this is a corner case I figured it's better to point it out before I forget about it myself.
Feel free to ping me if I never get back to this.

ccondon894:
_replace_prefix was failing because path wasn't guaranteed to be a string. This adds an explicit str() cast at the top of the method. I did not include a test case but happy to add one if needed.
Fixes #3991
### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path handling so path-like inputs are normalized before processing, preventing type-related issues and improving reliability when working with different path representations. This reduces unexpected failures and improves compatibility with downstream features that expect string paths.

### DIFF
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -2884,8 +2884,20 @@ To illustrate the possibilities of this mechanism, consider the following comple
   # a target rule to define the desired final output
   rule all:
       input:
-          "aggregated/a.txt",
-          "aggregated/b.txt"
+          "aggregated/sample1.txt",
+          "aggregated/sample2.txt"
+
+
+  # generate per-sample input files; filenames are sample IDs,
+  # while file content ("a" or "b") controls downstream branching
+  rule generate_sample_input:
+      output:
+          "samples/{sample}.txt"
+      run:
+          import random
+
+          with open(output[0], "w") as f:
+              f.write(random.choice(["a", "b"]))
 
 
   # the checkpoint that shall trigger re-evaluation of the DAG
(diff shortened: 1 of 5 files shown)


src/snakemake/path_modifier.py
```diff
@@ -75,6 +75,7 @@ def replace_prefix(self, path, property=None):
         return path
 
     def _replace_prefix(self, path, property):
+        path = str(path)
```
johanneskoester:
```suggestion
        if isinstance(path, Path):
            path = str(path)
```

PR #4046, opened 2026-03-11
fix: Ensure pixi tasks may be run as advertised

jonasfreimuth:
Add a constraint to the Pixi workspace for `pixi>=0.63` and pin Pixi to `v0.63.2` on the CI to ensure the `default-environment` field is available for tasks and set it for all tasks. This is to ensure that there are no changes to the formula `pixi run task`, when currently pixi either requires user input or an additional `-e environment` specification when the environment for a task is ambiguous. This helps to ensure existing workflows depending on the simple formula above don't break.

### QC

Existing tests should cover this. Also this PR ensures current documentation is accurate, rather than updating docs to reflect a changed state.

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD workflows to remove explicit Pixi version specifications, allowing workflows to use the default Pixi version instead of previously pinned versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -300,7 +300,7 @@ python-build = ">=1.2.2,<2"
 packaging = ">=24.0,<26"
 
 [tool.pixi.feature.publish.tasks]
-build = { cmd = "python -m build", description = "Build the package into the dist/ directory" }
+build = { cmd = "python -m build", description = "Build the package into the dist/ directory", default-environment = "publish" }
 build-check = { cmd = "python -m twine check --strict dist/*", depends-on = [
   "build",
-], description = "Check that the package can be uploaded" }
+], description = "Check that the package can be uploaded", default-environment = "publish" }


jonasfreimuth:
Closes #4046.

.github/workflows/docs.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: prefix-dev/setup-pixi@v0.9.3
         with:
           environments: "docs"
-          pixi-version: v0.42.1
```
johanneskoester, line 25: I would say let us remove the pixi-version definition entirely. Using the latest version should be fine.

jonasfreimuth: The action docs are a little unclear whether that's possible. Apparently, not specifying a version will attempt to use a locally installed one, as describes for [self-hosted runners](https://github.com/prefix-dev/setup-pixi?tab=readme-ov-file#self-hosted-runners), but I don't know whether Pixi is installed on GitHub runners. I'll try though.

jonasfreimuth: This appears to work.

PR #4047, opened 2026-03-11
feat: new graphs

thomasmulvaney:
This draft pull request introduces two graphing functions for producing GraphViz graphs from Snakemake DAGs.  The functionality extends what is currently present in the `--filegraph`, `--rulegraph` and `--dag` functions at the same time condensing much of the logic into a few core functions. The ultimate goal is to have subcommands (maybe `snakmake info rulegraph` and `snakemake info jobgraph`??) which would have their own sets of flags for customizing graph output.

Given, the visual nature of this feature, attached are some images showing the different outputs which are possible

## The `graphviz.rule_graph`
The plain output looks like this:

new_rg

### `graphviz.rule_graph(dag, inputs=True, outputs=True)`
By toggling input and output flags these fields can be shown on rules in the rulegraph.

new_rg_io

### `graphviz.rule_graph(dag, inputs=True, outputs=True, file_edges=True)`
Edges can be shown between the connecting input and output files rather than the nodes themselves.

new_rg_io_fe

### `graphviz.rule_graph(dag, edge_labels=True)`
A rule graph condenses all the jobs of the DAG into interconnecting rules. Information about the number of connecting jobs can be shown by using the `edge_labels` flag. This also works when the edges are files.

new_rg_el

## The `graphviz.job_graph` function
The job graph is DAG of all jobs. By default it looks like this.

new_jg

### `graphviz.job_graph(dag, wildcards=True)`
In addition, one can show the new wildcards introduced during DAG evaluation.

new_jg_wc

### `graphviz.job_graph(dag, inputs=True, outputs=True,file_edges=True`
The files can be shown using the `inputs` and `outputs` flag. Additionally the edges can be between the files of jobs by setting `file_edges=True`.

new_jg_io_fe

### `graphviz.job_graph(dag,legend=True)`
A picture says a thousand words, but still for someone just looking at a workflow for the first time, some extra details about what each rule does can be helpful. The `legend` option pulls out docstrings for each rule. This option is a bit of an experiment.

new_jg_legend

### QC

## Release Notes

* **New Features**
  * Added Graphviz-based visualization for Snakemake DAGs
  * Supports color-coded rule and job workflow diagrams
  * Optional display of inputs, outputs, and wildcards on nodes
  * Edge labeling and legend support for enhanced workflow documentation

### DIFF
--- a/src/snakemake/graphviz.py
+++ b/src/snakemake/graphviz.py
@@ -1,5 +1,5 @@
-__author__ = "Thomas Mulvaney"
-__copyright__ = "Copyright 2026, Thomas Mulvaney"
+__author__ = "Thomas Mulvaney, Johannes Köster"
+__copyright__ = "Copyright 2026, Thomas Mulvaney, Johannes Köster"
 __email__ = "mulvaney@mailbox.org"
 __license__ = "MIT"
 from collections import defaultdict
@@ -20,7 +20,7 @@ def resolve_functions(input_files):
     return files
 
 
-def enumerate_input_files(job):
+def enumerate_files(job_files, rule_files):
     """Given a job, enumerates all outputs as tuples in the form:
 
         (id, job_file, rule_file)
@@ -28,19 +28,8 @@ def enumerate_input_files(job):
     where job_file is a real path and rule_file is the input string as defined in the rule
     """
     for i, (job_file, rule_file) in enumerate(
-        zip(job._input, resolve_functions(job.rule._input))
+        zip(job_files, resolve_functions(rule_files))
     ):
-        yield (i, job_file, rule_file)
-
-
-def enumerate_output_files(job):
-    """Given a job, enumerates all inputs as tuples:
-
-         (id, job_file, rule_file)
-
-    where job_file is a real path and rule_file is the output as defined in the rule
-    """
-    for i, (job_file, rule_file) in enumerate(zip(job._output, job.rule._output)):
         yield (i, job_file, rule_file)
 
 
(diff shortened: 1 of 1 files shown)


johanneskoester:
Very neat! This is a bit related to what @m-jahn is currently working on: tube maps for Snakemake, which I would see as one additional way of visualizing the DAG. Maybe you two should also have a chat in case there are any synergies or potential unifications across the different types of visualizations.

johanneskoester:
Further thought: does it make sense to additionally allow to set the direction in which the DAG flows (horizontal vs vertical)?

thomasmulvaney:
> Further thought: does it make sense to additionally allow to set the direction in which the DAG flows (horizontal vs vertical)?

Yes, the direction could be made as a flag. Indeed, a few GraphViz parameters would be nice to tweak such as `ranksep` for example.

m-jahn:
Really cool! @thomasmulvaney I'm just trying to get this tool [Snakevision](https://github.com/OpenOmics/snakevision) to render tube maps from the rulegraph `.dot` output in the Snakemake workflow catalog. Not directly related to the snakemake source code though. Your work seems really promising so far.

thomasmulvaney:
I think it is also worth thinking about the rule graph layout. In the examples above, there is a rule `gmx` which is used in multiple steps, upstream and downstream of one another. This leads to a rule graph which has cycles and is maybe less obvious to inspect. I believe this is also the case for the current `--rulegraph`. 

Instead, we could unroll the graph further by identifying a rule by its full set of dependencies rather than its name.  This should be trivial and would aid in visualisation using the tubemap which appears to need a dag and would be useful for @m-jahn for example.

Also, we should probably add an option to write the graph to a file rather than stdout,  as stdout sometimes gets polluted by executing the dag.

cmeesters:
@johanneskoester @thomasmulvaney is this a stale PR by now?

m-jahn:
@cmeesters I don't think it's stale. I thought I'd look into this more as soon as I find time, and Thomas probably has the same issue right now.

thomasmulvaney:
Hi @cmeesters, not stale. Given the flexibility of the graphing output, I was holding out for the new subcommands API, as it would make sense to introduce this then as these plots have additional bells and whistles.

I have infact been playing around with trying to incorporate the `--group` options into the output so that job groups can be visualised as subclusters.

cmeesters:
@thomasmulvaney splendid! I will add a sentence to the preprint citing linking this PR. @m-jahn noticed, there is no reference in this chapter. Which is the reason for me asking.

PR #4053, opened 2026-03-11
fix: CLI/profile runtime override is interpreted as minutes not as seconds

### LINKED ISSUE #4039: runtime values from CLI/profile are silently interpreted as seconds instead of minutes since 9.16.3
**Snakemake version**

9.16.3 (regression from 9.16.2, introduced by PR #3421)

**Describe the bug**

In Snakemake 9.16.3, `runtime` values specified via the command line (`--set-resources`, `--default-resources`) or in profile YAML files are silently reinterpreted as **seconds** rather than **minutes**, resulting in drastically reduced runtime values.

This is caused by the new `Resource.parse_human_friendly` method introduced in PR #3421 ("standardize handling of resources"). When a `runtime` value arrives as a string from the CLI or a profile (as all CLI/profile values do), it is passed through `humanfriendly.parse_timespan()`, which treats bare numbers (without a unit suffix) as **seconds**. The result is then divided by 60 to convert to minutes. So a value like `runtime=60`, which previously meant 60 minutes, is now interpreted as 60 seconds → **1 minute**.

Values specified directly in a Snakefile's `resources:` block (e.g., `resources: runtime=60`) are **not affected**, because they are parsed as Python integers and `parse_human_friendly` returns non-string values without modification. The bug is specific to the CLI and profile YAML path where values are initially strings.

This is a silent, breaking change — no warning or error is emitted. Workflows relying on `runtime` values from profiles or the command line will have their runtime limits silently reduced, potentially causing jobs to be killed for exceeding time limits that are now ~60x smaller than intended.

**Logs**

With the minimal Snakefile below, setting `runtime` from the command line:

```
# Snakemake 9.16.2 (expected behavior)
$ snakemake -c1 --set-resources echo_runtime:runtime=60
# Output: Runtime resource: 60

# Snakemake 9.16.3 (broken behavior)
$ snakemake -c1 --set-resources echo_runtime:runtime=60
# Output: Runtime resource: 1
```

The value 60 is interpreted as 60 seconds → 1 minute instead of 60 minutes.

Similarly, when specifying `runtime` with a unit suffix:
```
# Snakemake 9.16.3
$ snakemake -c1 --set-resources "echo_runtime:runtime=60 minutes"
# Output: Runtime resource: 60
```
This works correctly because `humanfriendly.parse_timespan("60 minutes")` returns 3600 seconds, which is then divided by 60 to yield 60 minutes. But bare integers, which are the common case and the convention used in 9.16.2, are broken.

The same issue applies to profile YAML files:
```yaml
# profile/config.yaml
set-resources:
    echo_runtime:
        runtime: 120
```
In 9.16.2 this would set `runtime=120` (minutes). In 9.16.3 this is parsed as 120 seconds → 2 minutes.

**Minimal example**

Snakefile:
```python
rule echo_runtime:
    resources:
        runtime=60
    shell:
        """
        echo "Runtime resource: {resources.runtime}"
        """
```

Run with an overridden runtime from the command line:
```bash
snakemake -c1 --set-resources echo_runtime:runtime=120
```

Expected output (9.16.2 behavior): `Runtime resource: 120`
Actual output (9.16.3 behavior): `Runtime resource: 2`

**Additional context**

The root cause is in `Resource.parse_human_friendly` in `src/snakemake/resources.py`:

```python
elif name in TimeResources:
    try:
        return max(int(round(parse_timespan(stripped) / 60)), 1)
    except InvalidTimespan as err:
        raise WorkflowError(err_msg.format(unit="minutes")) from err
```

`humanfriendly.parse_timespan("120")` returns `120.0` (treating the bare number as seconds), which is then divided by 60 to yield `2`. The historical convention in Snakemake is that `runtime` values are specified in **minutes** as bare integers. The fix should either:

1. Treat bare numeric strings for `runtime` as minutes (matching the pre-9.16.3 convention) by checking if the string is a plain number and skipping the `parse_timespan` call.
2. At minimum, emit a deprecation warning if a bare number is provided for `runtime` via the CLI/profile, informing users of the new seconds-based interpretation.

Note that `runtime` specified directly in a Snakefile's `resources:` directive is unaffected because the value is already an `int` and bypasses `parse_human_friendly`.

SrPeter128:
Restore pre-9.16.3 behavior: treat numeric runtime values from CLI/profile as minutes (not seconds).

- Fix Resource.parse_human_friendly in resources.py to treat plain numeric strings for time resources as minutes.
- Add test: --set-resources echo_runtime:runtime=120 produces expected output.
- Add test: Setting ```runtime: 120``` in profile  produces expected output.

The parse_human_friendly() function in the Resources class previously divided the specified number by 60, interpreting the specified minutes as seconds.

Fix #4039

### QC

* [ ] The documentation (`docs/`) is updated to reflect the changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake). --> not needed since default behavior is restored

## Summary by CodeRabbit

* **Bug Fixes**
  * Runtime parsing refined: pure numeric strings are no longer treated as human-friendly timespans; non-numeric time strings are parsed with adjusted rounding/division so minute values are interpreted consistently.

* **Tests**
  * Added regression tests and example workflows covering runtime handling via CLI, profile, and default runs; expected outputs updated accordingly.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -3359,6 +3359,18 @@ def test_github_issue_4039_runtime_cli():
 
 
 @skip_on_windows
+def test_github_issue_4039_mem_cli():
+    """Test that mem values from CLI are correctly interpreted as MB, not bytes.
+    Test for https://github.com/snakemake/snakemake/issues/4039"""
+    tmpdir = run(
+        dpath("test_github_issue_4039_mem_cli"),
+        shellcmd="snakemake -c1 --set-resources 'echo_mem:mem=1024'",
+        cleanup=False,
+    )
+    shutil.rmtree(tmpdir)
+
+
+@skip_on_windows
 def test_github_issue_4039_runtime_profile():
     """Test that runtime values from a profile are correctly interpreted as minutes, not seconds.
     Test for https://github.com/snakemake/snakemake/issues/4039"""
--- a/tests/test_github_issue_4039_mem_cli/Snakefile
+++ b/tests/test_github_issue_4039_mem_cli/Snakefile
@@ -1,0 +1,9 @@
+rule echo_mem:
+    output:
+        "output.txt"
+    resources:
+        mem_mb=512
+    shell:
+        """
+        echo "Mem resource: {resources.mem_mb}" > {output}
+        """
--- a/tests/test_github_issue_4039_mem_cli/expected-results/output.txt
+++ b/tests/test_github_issue_4039_mem_cli/expected-results/output.txt
@@ -1,0 +1,1 @@
+Mem resource: 1024


tests/tests.py
```diff
@@ -2934,3 +2934,59 @@ def test_cyclic_dependency_single():
...
     # forcing it.
     run(dpath("test_cyclic_dependency_single"), forceall=True)
+
+
+def test_github_issue_4039_runtime_cli():
+    """Test that runtime values from CLI are correctly interpreted as minutes, not seconds.
+    Regression test for https://github.com/snakemake/snakemake/issues/4039"""
+    tmpdir = run(
+        dpath("test_github_issue_4039_runtime_cli"),
+        shellcmd="snakemake -c1 --set-resources 'echo_runtime:runtime=120'",
```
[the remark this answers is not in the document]
johanneskoester: This does not make sense to me. The string is passed to the arg parser, how windows interprets quotes does not matter in this case I think.

tests/tests.py
```diff
@@ -3013,101 +2936,57 @@ def test_cyclic_dependency_single():
...
+        got = f.read().strip()
+    with open(expected_output, "r") as f:
+        expected = f.read().strip()
 
-@skip_on_windows
-def test_github_issue3913():
-    core_count_limit = 8
+    assert got == expected, (
+        f"Generated output does not match expected.\nExpected: {expected!r}\nGot:      {got!r}"
+    )
```
johanneskoester: There is no need to do comparisons like this manually. The `run()` method already does it internally.

src/snakemake/resources.py
```diff
@@ -859,7 +857,7 @@ def parse_human_friendly(name: str, value: ValidResource) -> ValidResource:
                     raise WorkflowError(err_msg.format(unit="size in MB")) from err
             elif name in TimeResources:
                 try:
-                    return max(int(round(parse_timespan(stripped) / 60)), 1)
+                    return max(int(round(parse_timespan(stripped))), 1)
```
johanneskoester: Maybe add a comment here, mentioning that snakemake expects time resources to be given as minutes and this is what parse_timespan returns?

tests/tests.py
SrPeter128: Oh, right! Learned something new again. Thanks!
But deleting lines 3068 to 3070 was a typo, I think?

src/snakemake/resources.py
cndewey: This logic is incorrect.  `parse_timespan` [returns seconds](https://humanfriendly.readthedocs.io/en/latest/api.html#humanfriendly.parse_timespan).  The issue here is that when a bare number is provided for the runtime via the CLI or profile it still has type "str" and is then parsed using `parse_timespan`.  However, when given a bare number string, `parse_timespan` interprets that number as seconds, not minutes.

```python
>>> from humanfriendly import parse_timespan
>>> parse_timespan("1")
1.0
>>> parse_timespan("1m")
60.0
```

SrPeter128: Thank you for pointing that out. I also just noticed that parse_timespan behaves differently than I thought. Perhaps it would make more sense here to handle the bare number given over the CLI input differently.

cndewey: One possible solution here is to first try to convert to an integer with `int`, which will succeed for any bare integer numbers.  If that doesn't succeed, continue to try to use parse_timespan.  Below is the general idea but this could be refactored to be cleaner.

```python
try:
    return max(int(stripped), 1)
except ValueError:
    pass
try:
    return max(int(round(parse_timespan(stripped) / 60)), 1)
...
```

SrPeter128: Ah, good idea! Thanks for the suggestion! How about a regex? Then it would fit in one line
```python
 elif name in TimeResources and not re.fullmatch(r"\d+", stripped):
                try:
                    return max(int(round(parse_timespan(stripped))/60), 1)
```

cndewey: Further complicating this issue is the following behavior for specification of runtime within a Snakefile, without any CLI or profile values:

```python
rule echo_runtime:
    resources:
        runtime=60
    shell:
        """
        echo "Runtime resource: {resources.runtime}"
        """
```
outputs:
```
Runtime resource: 60
```

Changing 60 to "60"

```python
rule echo_runtime:
    resources:
        runtime="60"
    shell:
        """
        echo "Runtime resource: {resources.runtime}"
        """
```
outputs:
```
Runtime resource: 1
```

This behavior is the same between 9.16.2 and 9.16.3.

It is unclear what the desired behavior is, but it would seem most intuitive to me if a bare number string is interpreted as minutes and any string with units is parsed using parse_timespan.

SrPeter128: Good point! It should also be solved with the regex. I'll implement it and test it! Thank you very much for the many examples!

johanneskoester: I would vote for not allowing `"60"`.

johanneskoester: The docs are pretty clear about that. Either a number, which is minutes, or timespan with a proper suffix.

johanneskoester:
@pvandyken may I ask you to review this, since the new resource class is your baby?
 @SrPeter128 is currently finalizing his suggestion, but we already had one question, namely where the intended validation should be placed. Overall, the goal is (a) to ensure that runtime is always correctly interpreted and propagated as minutes, and (b) that, as already defined in the docs, that it may be either an integer or a timespan given as integer + h/d/etc.

src/snakemake/resources.py
```diff
@@ -857,9 +857,9 @@ def parse_human_friendly(name: str, value: ValidResource) -> ValidResource:
                     return max(int(math.ceil(parse_size(stripped) / 1e6)), 1)
                 except InvalidSize as err:
                     raise WorkflowError(err_msg.format(unit="size in MB")) from err
-            elif name in TimeResources:
+            elif name in TimeResources and not re.fullmatch(r"\d+", stripped):
```
pvandyken: It looks like we'll actually have the same problem in SizedResources. I suggest a single check before the SizedResources block checking if `stripped.isdecimal()`. If so, call `int(stripped)` and return.

pvandyken: Note that `isdecimal() == True` guarentees `int()` will work.

src/snakemake/resources.py
pvandyken: This is a gray area I think... `mem` and `disk` both allow simple numbers in string. `runtime` doesn't have a suffixed version, so it must work with strings. If you wish to prohibit simple numbers in strings, I think it should be for `mem`, `disk`, and `runtime`, but that's adding considerable additional behaviour to this PR

src/snakemake/resources.py
```diff
@@ -1090,10 +1090,22 @@ def from_mapping(cls, mapping: Mapping[str, ValidResource] | Self) -> Self:
...
         if isinstance(mapping, cls):
             return mapping
 
+        for key, val in mapping.items():
+            if (
+                key in TimeResources
+                and isinstance(val, str)
+                and re.fullmatch(r"\d+", val)
+            ):
+                raise WorkflowError(
```
pvandyken: As mentioned above, I'm not a fan of this check.  Just to make the point again:

> `mem` and `disk` both allow simple numbers in string. runtime doesn't have a suffixed version, so it must work with strings. If you wish to prohibit simple numbers in strings, I think it should be for `mem`, `disk`, and `runtime`, but that's adding considerable additional behaviour to this PR

pvandyken: With the current implementation, if a user gives `runtime` a function that returns a decimal string, validation will not occur. In other words:

```py
resources:
  runtime: "60" # disallowed by this check
  runtime: lambda *args: "60" # allowed by this check
```

Note that the validation logic cannot be moved into the `Resource` constructor because all resource values from the CLI are passed as `str`. So decimal strings cannot be excluded as allowable values for `runtime`.

The correct way, I _think_, to do this would be to follow the approach in `Resource.standardize_size`, which uses `Resource._convert_units(converter)` to do conversion and validation. In this case, `converter` should be a function returns the input if an int. If the input is a string, it errors if `str.isdecimal()`, otherwise does the human formatting.

I'm happy to do this myself in a second PR, but truthfully, I think it makes everything unnecessarily complex. I don't think the meaning of `runtime: "60"` is ambiguous, even if it's not explicitly "supported" by the docs.

pvandyken:
@johanneskoester @SrPeter128, thanks for handling this issue, it was a bad miss on my part. From a testing perspective, it occurred because I never tested actual command line use, so the existing tests are only seeing int.

I left my comments above!

src/snakemake/resources.py
johanneskoester: Ok, I see your points. Then, let us still allow it for now at least.

SrPeter128:
Hi @pvandyken!
Thank you very much for the detailed explanations and ideas! I hope I got all your points correctly. I removed the check in the ```from_mappting()``` function so that "60" is still allowed. I also check in the  ```parse_human_friendly()``` for both, the TimeResources and the SizedResources strings, if they are decimal. If so they will be returned directly, if not they will be parsed human friendly. 
Thank you and best regards!

src/snakemake/resources.py
SrPeter128: Thanks for the hint! Is done in the latest commit!

tests/tests.py
```diff
@@ -3261,3 +3261,41 @@ def test_module_onerror():
...
+
+# On Windows this test output is emitted with
+# quotes around the output string what cause these tests to fail
+@skip_on_windows
+def test_github_issue_4039_runtime_cli():
+    """Test that runtime values from CLI are correctly interpreted as minutes, not seconds.
+    Test for https://github.com/snakemake/snakemake/issues/4039"""
+    tmpdir = run(
+        dpath("test_github_issue_4039_runtime_cli"),
+        shellcmd="snakemake -c1 --set-resources 'echo_runtime:runtime=120'",
```
pvandyken, line 3355: Might be worth having a test of mem or disk

cademirch:
pushed a fix to get `test_profile_parse` passing. the `isdecimal()` early return in `parse_human_friendly` was sitting above the `SizedResources`/`TimeResources` checks, so numeric strings like `slurm_partition: "42"` were getting cast to int. i think this is the right decision here, but @pvandyken can you weigh in before I merge?

edit: windows test failing currently is unrelated - seems like a transient pixi error

PR #4059, opened 2026-03-11
Flag optional() added for input to allow partial result

### LINKED ISSUE #4016: Allow optional input files so workflows can continue and produce partial results
Sometimes user would be satisfied with the partial result and does not want to stop the whole analysis because one or several output files in a certain rule were not received.

Right now, if a downstream rule expects a full list of input files, and one file is missing, the workflow stops. For some analyses, especially large ones, this is too strict. The user may still want Snakemake to continue and return a partial result instead of failing the whole analysis.

I am thinking about cases where the user could mark some output files as optional.

I would like to have a way to flag some files as optional, so that Snakemake can continue the workflow even if those files are missing, and allow downstream rules to proceed with the available inputs.

The general idea:

- Snakemake should return partial result of analysis if the expected input was flagged as optional.
- Missing optional files should not stop the workflow.

Cases to consider:
- code output 0, file not found -> continue without interruption
- code output 1, file not found -> error, stop
(the user can adjust logging and avoid erroring in the shell/script if they really want continuation)
- code output 0, file found -> continue without interruption
- code output 1, file found -> error, stop

Possible implementation direction I had in mind:
- implement flag `optional` in src/snakemake/io
- in snakemake.dag, add a check for flag optional and allow the workflow to continue in certain cases(during postprocessing function?)
- remove that missing optional file from downstream dependencies, so following rules can proceed with the remaining available inputs

```python
rule NAME:
    input:
        optional("path/to/noncritical_file")
    output:
        "path/to/outputfile"
    shell:
        "somecommand {input} {output}"
```

-//-

**Additional context**
This would be especially useful for large analyses, where one or a few missing outputs should not necessarily block producing a partial merged result.

hooloobooroodkoo:
**Goal**: implement flag `optional` to enable receiving of a partial result of a workflow. If input is marked as optional, when the workflow can be executed even if optional files are missing. 

Closes #4016 

**Example**:

```python
# assume that previous step should have been produced
counts = ["counts_000.json", "counts_001.json", "counts_002.json"]
# but "counts_002.json" wasn't created

#before
rule merge_counts:
    input: counts                   # workflow breaks with no result
    output: "letter_counts.json"
    script: "scripts/merge_counts.py"

# after
rule merge_counts:
    input: optional(counts)         # returns merged "counts_000.json" + "counts_001.json"
    output: "letter_counts.json"
    script: "scripts/merge_counts.py"
```

**Main conceptual changes:**

- the output check is less strict
- the input list is filtered for the downstream job before it runs and for all rules in the DAG, removing optional files that were not found

**Changes made:**

- `dag.py`
    - modified `DAG.update_()` , added condition if optional, then don’t add to missing_input list and do not raise the `MissingInputException`, but log it
    - modified `DAG.check_and_touch_output()` , exclude optional files from `wait_for_files` not to raise `MissingInpitException` (basically, if optional)
        - for this function `DAG._get_optional_only_outputs(Job)` was added. Where, firstly, the `Job.output` is being iterated through, and every output file is being checked for being dependent on in the job graph. Secondly, other rules are checked for needing the current rule’s output in their inputs(if it’s optional somewhere AND not required anywhere, then the exception is not triggered).
- `jobs.py`
    - modified `Job.prepare()` : changes an input `Job._input`, makes sure optional files are absent from the shell command and from the wait_for_files call below
        - for this `Job._filter_optional_inputs()` implemented. It takes `Job.input`, removes any files that are `optional()` AND don't exist on disk, and returns updated input (that can require recreating the InputFiles())
- `tests.py` and tests folder updated with new tests
- misc/vim/syntax/snakemake.vim `optional` added to be highlighted

If you want to run all the tests for this feature: `pixi run pytest tests/tests.py  -k "optional" -v`

### QC

!!! Documentation is not added yet but I will add it when we agree on this pull request)
* [X] I, as a human being, have checked each line of code in this pull request

 I used AI assistance for:

* [X] Code generation (e.g., when writing an implementation or fixing a bug)
* [X] Research and understanding

## Summary by CodeRabbit

* **New Features**
  * added optional() input flag and is_optional IO property to mark optional inputs.
* **Bug Fixes / Behavior**
  * missing optional inputs now log warnings and are ignored during waiting/validation/touching; upstream job failures still fail the run.
* **Documentation**
  * Added "Optional input" docs with examples.
* **Tests**
  * Added integration tests and fixtures covering multiple optional-input scenarios.
* **Style**
  * Vim syntax updated to recognize the "optional" keyword.


mrmattuschka:
Hi Yana, nice work! This is a feature that I've been looking for.

For my own understanding:
- For a given upstream job, you will check which of its outputs are considered ``optional`` in all downstream jobs using it as an input
  - If so, these outputs missing after execution will not raise an ``MissingOutputException`` - instead just log it 
  - This will **not** allow the job itself to fail - files may only be missing after terminating with exit code 0. So if the user wants a failed shell command to not raise an exception, they will have to add something like ``<command> | exit 0`` or so.
- For any downstream job, remove the optional files from the inputs as part of ``job.prepare()``, right before running said job.

Am I getting everything correctly so far?

I'm wondering how one should actually deal with missing inputs on the side of the downstream rule  
Currently, optional files that are missing are completely hidden from the job, and if they're referenced anywhere using ``input.<missing file>`` it will raise an AttributeError.
Would it maybe not be more useful to preserve this missing (but optional) file in the InputFiles object and leave the logic to handle missing files to the ``script``/``shell``/``run``?
This would allow the downstream rule to react to a missing output in a more sophisticated way, e.g. a reporting rule could then conclude "sample XYZ failed" rather than not being aware of its existence to begin with.

hooloobooroodkoo:
Hi @mrmattuschka!
Yes, you are right. You got everything correct and I can see your point. Thank you! I'm right now looking into `Namedlist` to understand whether I can somehow preserve the existing logic and modify it)

CameronMcClymont:
@hooloobooroodkoo @mrmattuschka My PR #4086 extends this one with some additional functionality (optional `tolerance` and `impatient` parameters). WRT the `AttributeError` being raised when trying to access those optional inputs which were 'removed' from the downstream job, I handle this in my PR by using an empty `NamedList` like `my_input._set_name(name, len(my_input), end=len(my_input))`

dkoppstein:
Hi, this is a really nice idea, sorry for not looking at it before during the hackathon! I think PIs around the world will be happy that this feature exists :) Catching up on it, I was wondering if we can ensure that it interacts with the --retries flag, so that only after it fails the given number of times to produce some intermediate file does it try to produce the end result? Maybe a test case could explicitly deal with this?

CameronMcClymont:
@dkoppstein thanks for raising this, it's a good point. I'm not totally sure how this PR interacts with the `--retries` flag, but in my PR #4086 (which extends this one), retries take priority. When an upstream job fails, Snakemake retries it up to `N` times before considering it definitively failed. The `optional` logic is only evaluated after all retries are exhausted.

CameronMcClymont:
I have also just added a test case for the `--retries` flag interaction in my PR #4086

src/snakemake/jobs.py
```diff
@@ -798,6 +800,70 @@ async def remove_existing_output(self):
...
+                new_start = None
+                new_end = None
+                for old_i in range(start, end):
+                    if old_i not in to_remove:
+                        ni = index_map[old_i]
+                        if new_start is None:
+                            new_start = ni
+                        new_end = ni + 1
+                if new_start is not None:
+                    new_input._set_name(name, new_start, end=new_end)
```
johanneskoester, line 863: * Please check the coderabbit comment https://github.com/snakemake/snakemake/pull/4059/changes#r2919638154, I think it is valid.
* In addition, @hooloobooroodkoo can you please move the code that reduces the input files here into the Namedlist class, e.g. as `Namedlist.remove_values()`?

src/snakemake/jobs.py
```diff
@@ -834,6 +899,10 @@ async def prepare(self):
             if self.benchmark:
                 self.benchmark.prepare()
 
+        # Filter out missing optional input files before waiting / expanding {input}.
+        # This ensures they are absent from the shell command and from the wait_for_files call below.
+        self._input = await self._filter_optional_inputs()
+
```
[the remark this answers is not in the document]
johanneskoester: True, this has to be fixed as well.

johanneskoester (changes requested):

docs/snakefiles/rules.rst
```diff
@@ -287,6 +287,38 @@ Finally, argument values passed to ``expand`` can also be functions or lists of
...
 If that is the case, ``expand`` returns a function again, the evaluation of which is deferred to the point in time when the wildcards of the respective job are known.
 
+.. _snakefiles_optional:
+
+Optional input
+~~~~~~~~~~~~~~~~~~~
+
+In some workflows, it can be desirable to proceed even if one or several expected input files are missing and a small number of missing intermediate results should not necessarily prevent the creation of a partial final result.
+
+By default, Snakemake requires all declared input files of a rule to be present before execution. If any expected input file is missing, the workflow stops with an error. To relax this behaviour for selected inputs, Snakemake provides the ``optional()`` input flag:
```
johanneskoester, line 297: Doesn't this lack documentation for how to mark outputs as optional?

johanneskoester (changes requested):

PR #4062, opened 2026-03-11
feat: add `snakemake new` subcommand

schnea:
Adds a `snakemake new` subcommand that copies the scaffolding from https://github.com/snakemake-workflows/snakemake-workflow-template (or any other URL) into a new folder using [copier](https://copier.readthedocs.io/en/stable/)

## Usage
`pixi run -e dev snakemake new`
`pixi run -e dev snakemake new --from <template_url>`

By default, 

## Notes
Mainly opening this PR to get some thoughts on how to proceed:
- We could get very fancy with copier, but it looks like the template structure would need to be adapted for that. I've discussed with @cademirch  and @dlaehnemann that something more barebones than the current template might be nice
- Aside from the project name, what are things that people would reasonably like to fill out in an initial dialog? What would be good to have as a cli flag as well?

### QC

### DIFF
--- a/src/snakemake/cli/commands/new.py
+++ b/src/snakemake/cli/commands/new.py
@@ -1,3 +1,5 @@
+import os
+
 import click
 from copier import run_copy
 from plumbum.commands.processes import ProcessExecutionError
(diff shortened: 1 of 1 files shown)


schnea:
I've created a template in https://github.com/schnea/snakemake-workflow-template that makes use of some of the copier templating functionality to pre-fill some values and set up the simplest example workflow i could think of. 
You can check it out with 
`pixi run -e dev snakemake new --from https://github.com/schnea/snakemake-workflow-template`

johanneskoester:
This is actually a good idea, I would however suggest to move the implementation into snakedeploy, which is also used for plugin templating and workflow deployment.

johanneskoester:
Closing in favor of a PR on snakedeploy.

cademirch:
> This is actually a good idea, I would however suggest to move the implementation into snakedeploy, which is also used for plugin templating and workflow deployment.

I agree the implementation fits for the nature of snakedeploy, but I think having a `new/init` subcommand in the main snakemake cli is important. Much like `cargo init` or `pixi init` its very intuitive and does not require the user to install another package. Of course we could just depend on snakedeploy here and with our cli refactor provide a thin wrapper on top of snakedeploy. Thoughts @johanneskoester?

PR #4065, opened 2026-03-12
feat: add a helper function that translate Python script names into corresponding module names

### LINKED ISSUE #4003: Helper function to be able to call a Python script by its module name (`python -m`)
There are a number of ways of calling Python code:
  
 - `python path/to/file.py` in a `shell:` block
 - `python -m path.to.file` in a `shell:` block
 - `path/to/file.py` in a `script:` block

For mature libraries that can be called as command-line tools, using `python -m` makes sense. For simple scripts, `python path/to/file.py` or `path/to/file.py` is fine.

For slightly more complex script setups, where hundreds to thousands of lines are written for a specific analysis, one wants to be able to split the code into separate modules, with relative imports allowed. However, for this to work correctly (i.e. without hacking `sys.path`), the code must be invoked with `python -m path.to.file`.

However, since the code is being developed bespoke for the workflow, ideally there would _also_ be an explicit dependence on the script as an `input:` file.

Currently there is no easy way to have a `script` dependence that is then referred to as a Python module in a `shell:` block.

For example, consider the setup:

`scripts/main.py`:

```
from . import dependency

dependency.success()
```

`scripts/dependency.py`:

```
def success():
    print("It works!")
```

`scripts/__init__.py`:
[empty file]

Running `python -m scripts.main` works, but how to express that in a Snakefile?

`workflow/Snakefile`:

```
rule python_dependencies:
    input:
        script="scripts/main.py",
    output:
        "dummy.out"
    shell:
        "python {input.script} > {output}"
```

This fails:

```
ImportError: attempted relative import with no known parent package
```

`workflow/Snakefile`:

```
rule python_dependencies:
    input:
        script="scripts/main.py",
    output:
        "dummy.out"
    shell:
        "python -m scripts.main > {output}"
```

This works, but I have to duplicate the script name in slightly different formats; ideally this would be DRY.

```
def fix_name(filename):
    if not filename.endswith(".py"):
        raise ValueError("Doesn't look like a Python script")
    return filename[:-3].replace("/", ".")


rule step_1:
    input:
        script="scripts/main.py",
    output:
        "dummy.out"
    shell:
        "python -m {fix_name(input.script)} > {output}"
```

This fails because I can't put any function call into `shell:` blocks' curly braces.

A helper function that takes a script filename and returns it reformatted as a module name.

E.g.

```
rule step_1:
    input:
        module=python_module("scripts/main.py"),
    output:
        "dummy.out"
    shell:
        "python -m {input.module} > {output}"
```

The current workaround I use is to add an extra parameter to handle the module:

```
rule step_1:
    params:
        module=lambda wildcards, input: input.script.replace("/", ".")[:-3],
    input:
        script="scripts/main.py",
    output:
        "dummy.out"
    shell:
        "python -m {params.module} > {output}"
```

This is annoying boilerplate to have to add to every rule. A compromise would be if I could use a single named function, e.g.

```
def get_module(wildcards, input):
    return input.script.replace("/", ".")[:-3]


rule step_1:
    params:
        module=get_module,
    input:
        script="scripts/main.py",
    output:
        "dummy.out"
    shell:
        "python -m {params.module} > {output}"
```

Currently this isn't possible, as it doesn't seem to be possible to use named functions in `params:` blocks:

```
NameError in file "…/workflow/Snakefile", line 18:
name 'get_module' is not defined
  File "…/workflow/Snakefile", line 18, in <module>
```

edbennett:
Implements the functionality requested in #4003. This may not be mergeable, but I thought it was easier to discuss pros and cons of a concrete implementation proposal rather than a more abstract description of the idea.

Where workflow-specific Python code is under active development, it may be desirable to have an explicit dependency in a rule on the script implementing it. However,
any non-trivial Python code is likely to sit within a package structure, relying on relative imports
(e.g. `from . import another_module` or `from ..helpers import some_helper`). The correct way to call such code is via the module name, i.e. `python -m package.subpackage.module_name`.

Doing this while depending on the script file requires you either to specify the filename twice in slightly different formats (e.g.
```
input:
    script="package/subpackage/module_name.py",
shell:
    "python -m package.subpackage.module_name"
```

or to use a helper function, e.g.:

```
params:
    module=lambda wildcards, input: input.script.replace("/", ".")[:-3],
input:
    script="package/subpackage/module_name.py",
shell:
    "python -m {params.module}"
```

This PR provides a global helper function that may be used to achieve the latter, rather than every workflow needing to reinvent of this wheel.

### QC

## Summary by CodeRabbit

* **New Features**
  * Added a utility to translate script filenames into validated Python module names and expose it via the utilities namespace.

* **Documentation**
  * Added documentation with examples showing default and custom input-name usage for module invocation.

* **Tests**
  * Added tests and sample packages exercising module-name translation, module-mode invocation, and expected failure paths.


johanneskoester:
Not sure I understand the benefit of this over directly using the script directive. Can you explain that? (also in person if you like)

edbennett:
The `script:` directive still ends up calling `python {script_name}`, so has the same issue. I've pushed a test illustrating this.

(Using the `script` directive also means that if your code is designed to be run as a standalone tool as well as being integrated into Snakemake, then as I understand it you need to add separate code paths to handle the input parameters, which can be frustrating.)

tests/test_github_issue4003/Snakefile
```diff
@@ -0,0 +1,49 @@
...
+        script="some_package/some_subpackage/some_module.py",
+    output:
+        "default_name.out"
+    shell:
+        "python -m {params.module} > {output}"
+
+
+rule test_with_non_default_name:
+    params:
+        module=python_module("script_name"),
```
johanneskoester: The usage should rather resemble other ioutils functions. You can take the subpath function for illustration. Basically, the function should accept both a path and a function, thereby allowing to use the [rule item access helpers](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#rule-item-access-helpers), e.g.

```python
rule:
    input:
        script="some_package/some_subpackage/some_module.py"
    params:
        module=as_py_module(input.script)
```

johanneskoester (changes requested):
Now that I understood the use case, I like this a lot! Below, there is one request about the syntax though.

edbennett:
Thanks for the suggestion; now done.

johanneskoester (approved):

johanneskoester:
Thanks a lot!

Issue #4021, opened 2026-03-10
RFC more "tolerant" scatter-gather


tiborsimko:
### About

This RFC aims at proposing a more tolerant scatter-gather construct that would allow the DAG to proceed with the workflow execution even if a small percentage of scatter jobs are failing for a reason or another.

### Motivation

In experimental particle physics data analyses, it is usually necessary to process large simulated datasets consisting of 50k files, such as <https://opendata.cern.ch/record/8884> containing independent simulated collision events. When a small proportion of input files is experiencing data access troubles (e.g. corrupted files), it is often desired to keep the analysis running, at the price of having lower statistics in the outputs. For example, it can already give a valuable insight to the analyst if one can process say 95% of input files, whilst the ultra exact measurement with 100% of input files can always be carried out later.

One important note is that this is something that is DAG rule specific, e.g. an analyst may want to do this when processing simulated datasets, but not when processing collision datasets. (So we cannot act on a Snakemake global execution option, rather on the Snakemake workflow definition itself)

Moreover, when in a quick testing mood, i.e. when an analyst would satisfy herself with say 20% of input files processed for some rule, she would ideally not like to wait for any "laggard" jobs when the sufficient amount of scatter jobs finished, and may like to kill the laggard jobs and to simply proceed quickly with the rest of the analysis.

### Current approach

If one would like to process 100% of input files but still get some results when some files fail, it is currently possible to amend Snakefile to never fail, and produce some partial results, and rerun Snakemake several times to iterate towards the good number. See for example <https://github.com/tiborsimko/reana-demo-snakemake-continuable-errors> where a small example simulating data access failures is given.

However, this cannot express "do not wait for laggards" concept, hence we would like to propose an extension to the Snakemake scatter-gather-based or wildcard-based job dispatch to the workflow description language.

### Proposal

Consider the following simple example:

```
samples = ["1","2","3","4","5"]

rule scatter:
    input:
        "myinputs{n}.txt"
    output:
        "myresults{n}.txt"
    shell:
      ...

rule gather:
    input:
        expand("myresults{n}.txt", n=samples)
    output:
        "sum.txt"
    shell:
        ...
```

#### Option 1: Introduce per-rule `-keep-going` option

It would be already useful to be able to signal that the scatter rule is allowed to fail, something like a local per-rule `--keep-going` option:

```
rule scatter:
    input:
        "myinputs{n}.txt"
    output:
        "myresults{n}.txt"
    keep_going=True
    shell:
      ...
[...]

PR #4067, opened 2026-03-12
feat(scatter-gather): add support for tolerance and impatient gathering

CameronMcClymont:
Adds support for specifying `tolerance` and `impatient` parameters in the top-level `scattergather` directive:

```
scattergather:
    split={"items": 4, "tolerance": 0.25, "impatient": True}
```

In the example above, `tolerance = 0.25` so up to 25% of the 4 scatter branches are allowed to fail. If 3/4 branches succeed and one fails, then the `gather` rule would collect the results of the succeeding 3 branches and move on (without failing the entire workflow). If multiple branches were to fail, then the workflow would stop.

The example above also sets the `impatient` flag, meaning that the gather rule would run as soon as it knows that the `tolerance` is guaranteed to be satisfied (in this case, as soon as 3/4 branches succeed). It doesn't care whether the 4th branch succeeds or not, it just proceeds without it.

The changes preserve backwards compatibility with the previous API, where only an integer would be given:

```
scattergather:
    split=4
```

Which is equivalent to:

```
scattergather:
    split={"items": 4, "tolerance": 0, "impatient": False}
```

The rationale behind adding support for this feature is described in #4021.

### QC

## Summary by CodeRabbit

* **New Features**
  * Introduced scatter–gather workflow support with configurable per-process items, failure tolerance, and impatient (early-gather) mode; new public primitives to declare and inspect scatter processes and items.

* **Documentation**
  * Expanded docs describing optional failure tolerance, impatient behavior, and example mappings.

* **Tests**
  * Added comprehensive tests and fixtures covering tolerance, impatient execution, external dependencies, fanning intermediates, and complex DAG scenarios.


CameronMcClymont:
I've marked this as a draft again because I'm considering closing this PR since it adds considerable complexity to scatter-gather and I anticipate much lower usage compared to the sibling 'optional inputs' feature in PR #4086. So it may be worth merging #4086 and abandoning this PR in order to avoid introducing 'low value' code into the codebase which would just add complexity. Happy to hear others' thoughts before I close it.

PR #4075, opened 2026-03-12
feat: --interactive mode with a first dryrun, then real run upon user input. fixes #3330

### LINKED ISSUE #3330: allow for directly launching a run, if a `--dryrun` looked good, without restarting DAG building
For large numbers of jobs (mainly due to large datasets with many intermediate files), the DAG building can run for very long times. We have seen multiple hours in some instances. At the same time, it is very important to do a `--dryrun` on such large workflows, to avoid unintended reruns of large parts of a workflow, for example after changing some parameters. However, having to do the DAG building twice, once for the `--dryrun` and once for the actual run, even if the `--dryrun` looks fine, can be a considerable slow-down of data analyses.

I think, the optimal solution would be to turn the `--dryrun` default into asking for confirmation to actually run at the end of every `--dryrun`. However, this is not backwards compatible and would for example break existing continuous integration tests that rely on a `--dryrun` to run through without interaction. Thus, a new `--confirm-with-dryrun` option for this behavior might be the better solution, because it is backwards compatible and the wording of the option is probably a bit clearer.

---

**Additional context**
---

dkoppstein:
This PR adds a new CLI option --interactive (-i for short) which automatically starts a dryrun, calls a new internal function `reset_after_dryrun()` to cleanup the DAG after the dryrun, then prompts the user if they want to do a real run. 

### QC
Documentation is added in `docs/executing/cli.rst`. The `common.run` function has been modified to add a new `user_input` parameter for testing the interactive mode with STDIN input. 

NB: this was PR was largely authored by Copilot with many rounds of human feedback and explanation. 

## Summary by CodeRabbit

* **New Features**
  * Interactive dry-run flow: run a dry-run, receive an in-session prompt, and continue to real execution via the new interactive CLI flag (-i/--interactive).

* **Documentation**
  * Added docs describing interactive dry-run usage, prompt flow, and examples.

* **Tests**
  * Added end-to-end tests, a test workflow, and test harness support for simulated user input.

* **Bug Fixes**
  * Reset execution state after dry-run so a subsequent real run proceeds correctly.


dkoppstein:
This addresses #3330.

dkoppstein:
This is a relatively minimal PR, additional functionality could include e.g. prompting the user to proceed with cleanup metadata, unlock, rerun incomplete, etc. if these pop up. Also things like remote execution still need to be tested in interactive mode.

dkoppstein:
NB: --interactive still preserves proper submission to remote hosts, I tested this on LSF but would be good to test this more generally.

johanneskoester:
@cademirch I believe this should be considered together with the CLI redesign.

Issue #4076, opened 2026-03-12
Support remote snakefile


clelange:
Allow users to specify a snakefile via a remote URL (HTTP/HTTPS) or GitHub/GitLab URL instead of only a local file path. This enables:
- Running workflows directly from HTTP/HTTPS URLs
- Automatic resolution of GitHub/GitLab URLs (including self-hosted GitLab instances)
- Relative include statements are resolved relative to the remote Snakefile location

PR #4077, opened 2026-03-12
feat: Support remote snakefile in CLI and API

clelange:
## Summary

Add support for specifying a remote snakefile via URL (HTTP/HTTPS or GitHub/GitLab) in addition to local file paths.

## Changes

- Allow `snakefile` parameter in API and CLI to accept a URL string instead of only `Path`
- Update `resolve_snakefile()` to handle remote URLs
- Infer hosted git source files from plain URLs (GitHub/GitLab)
- Add tests for running a workflow via HTTP URL and via the Python API
- Added convenience syntax support for GitHub/GitLab shorthand URLs

## Usage

```bash
# HTTP/HTTPS snakefile
snakemake --snakefile https://example.com/Snakefile
# GitHub URL
snakemake --snakefile https://github.com/user/repo/blob/main/Snakefile
# GitHub shorthand syntax
snakemake --snakefile gh:user/repo@branch:path/to/Snakefile
# GitLab (including self-hosted instances)
snakemake --snakefile gl:gitlab.company.com:user/repo@branch:path/to/Snakefile
# Via Python API
api.snakemake(..., snakefile="https://example.com/Snakefile")
```

Relative include statements in remote snakefiles are resolved relative to the remote file location.

The shorthand syntax follows the pattern `gl/gh:<host>:<repo>@<ref>:<path>`, where:

- `<host>` is `github.com` or a GitLab instance hostname (optional)
- `<repo>` is the repository in user/repo format
- `<ref>` is the branch, tag, or commit SHA
- `<path>` is the file path within the repository

Addresses #4076 

CC @lukasheinrich @matthewfeickert 

## Summary by CodeRabbit

* **New Features**
  * Snakefile references may be local paths, HTTP/HTTPS URLs, or shorthand gh:/gl: host:owner/repo@ref:path forms for GitHub/GitLab.

* **Behavior Changes**
  * Relative includes in remote Snakefiles resolve against the remote location.
  * CLI/GUI now preserve non-local Snakefile strings (e.g., URLs); workflow-specific default profile is only considered for local Snakefiles.

* **Documentation**
  * Expanded guidance and examples for remote and hosted Snakefile sources and shorthand syntax.

* **Tests**
  * Added tests covering remote Snakefile fetching and hosting-provider inference.

### DIFF
--- a/src/snakemake/sourcecache.py
+++ b/src/snakemake/sourcecache.py
@@ -411,6 +411,7 @@ class HostingProviderFile(SourceFile):
         commit: Optional[str] = None,
         host: Optional[str] = None,
         cache_path: Optional[Path] = None,
+        _ref_path_candidates: Optional[List[tuple]] = None,
     ):
         if repo is None:
             raise SourceFileError("repo must be given")
@@ -439,6 +440,7 @@ class HostingProviderFile(SourceFile):
         self.host = host
 
         self._cache_path: Optional[Path] = cache_path
+        self._ref_path_candidates: Optional[List[tuple]] = _ref_path_candidates
 
         # Via __post_init__ implementing subclasses can do additional things without
         # replicating the constructor args.
@@ -524,7 +526,31 @@ class HostingProviderFile(SourceFile):
     def get_filename(self):
         return os.path.basename(self.path)
 
+    def _resolve_ref(self):
+        """Resolve ambiguous ref/path candidates using the hosted git repo.
+
+        When a URL contains a branch name with slashes (e.g. perf/autobump),
+        multiple (ref, path) splits are possible. This method tries each
+        candidate and picks the one whose ref exists in the repo.
+        """
+        if not self._ref_path_candidates or len(self._ref_path_candidates) <= 1:
+            return
+
+        for ref_candidate, path_candidate in self._ref_path_candidates:
+            if self.hosted_repo.ref_exists(commit=None, branch_or_tag=ref_candidate):
+                self.branch = ref_candidate
+                self.path = path_candidate
+                self._ref_path_candidates = None
+                return
+
+        # No candidate matched; keep the current (naive) values and clear
+        # candidates so we don't retry.
+        self._ref_path_candidates = None
+
     def fetch_if_required(self) -> Optional[str]:
+        # Resolve ambiguous ref/path before any ref-based operations.
+        self._resolve_ref()
+
         # always fetch if this points to a branch
         if (
             self.branch is not None
@@ -667,6 +693,24 @@ class GitlabFile(HostingProviderFile):
         )
 
 
+def _build_ref_path_candidates(path_parts: list, ref_start: int) -> List[tuple]:
+    """Build all possible (ref, path) splits from the ambiguous path segments.
+
+    Given path_parts and the index where the ref begins, generate candidates
+    by trying progressively longer ref strings.  The list is ordered from
+    shortest to longest ref so that the resolution logic can stop at the first
+    match.
+    """
+    candidates = []
+    # There must be at least one segment for the path, so the ref can span
+    # from ref_start up to (but not including) the last segment.
+    for split in range(ref_start + 1, len(path_parts)):
+        ref = unquote("/".join(path_parts[ref_start:split]))
+        path = unquote("/".join(path_parts[split:]))
+        candidates.append((ref, path))
+    return candidates
+
+
 def _infer_github_file(path_or_uri: str) -> Optional[GithubFile]:
     parsed = urlparse(path_or_uri)
     if parsed.scheme not in {"http", "https"}:
@@ -678,7 +722,13 @@ def _infer_github_file(path_or_uri: str) -> Optional[GithubFile]:
         repo = "/".join(path_parts[:2])
         ref = unquote(path_parts[2])
         path = unquote("/".join(path_parts[3:]))
-        return GithubFile(repo=repo, path=path, branch=ref)
+        candidates = _build_ref_path_candidates(path_parts, ref_start=2)
+        return GithubFile(
+            repo=repo,
+            path=path,
+            branch=ref,
+            _ref_path_candidates=candidates if len(candidates) > 1 else None,
+        )
 
     if (
         parsed.netloc == "github.com"
@@ -688,7 +738,13 @@ def _infer_github_file(path_or_uri: str) -> Optional[GithubFile]:
         repo = "/".join(path_parts[:2])
         ref = unquote(path_parts[3])
         path = unquote("/".join(path_parts[4:]))
-        return GithubFile(repo=repo, path=path, branch=ref)
+        candidates = _build_ref_path_candidates(path_parts, ref_start=3)
+        return GithubFile(
+            repo=repo,
+            path=path,
+            branch=ref,
+            _ref_path_candidates=candidates if len(candidates) > 1 else None,
+        )
 
     return None
 
@@ -733,14 +789,17 @@ def _infer_gitlab_file(path_or_uri: str) -> Optional[GitlabFile]:
             and raw_or_blob_index + 3 < len(path_parts)
             and path_parts[raw_or_blob_index + 1] in {"blob", "raw"}
         ):
+            ref_start = raw_or_blob_index + 2
             repo = unquote("/".join(path_parts[:raw_or_blob_index]))
-            ref = unquote(path_parts[raw_or_blob_index + 2])
-            path = unquote("/".join(path_parts[raw_or_blob_index + 3 :]))
+            ref = unquote(path_parts[ref_start])
+            path = unquote("/".join(path_parts[ref_start + 1 :]))
+            candidates = _build_ref_path_candidates(path_parts, ref_start=ref_start)
             return GitlabFile(
                 repo=repo,
                 path=path,
                 branch=ref,
                 host=parsed.netloc,
+                _ref_path_candidates=candidates if len(candidates) > 1 else None,
             )
 
     return None
--- a/tests/test_sourcecache.py
+++ b/tests/test_sourcecache.py
@@ -1,10 +1,13 @@
 from pathlib import Path
 import tempfile
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock
 from snakemake.sourcecache import (
     GenericSourceFile,
     GithubFile,
     GitlabFile,
+    _build_ref_path_candidates,
+    _infer_github_file,
+    _infer_gitlab_file,
     infer_source_file,
 )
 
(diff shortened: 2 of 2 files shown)


docs/snakefiles/deployment.rst
```diff
@@ -85,6 +85,8 @@ Consider the following example:
 First, we load a local configuration file.
 Next, we define the module ``dna_seq`` to be loaded from the URL ``https://github.com/snakemake-workflows/dna-seq-gatk-variant-calling/raw/v2.0.1/workflow/Snakefile``, while using the contents of the local configuration file.
```
johanneskoester:
```suggestion
Next, we define the module ``dna_seq`` to be loaded from the Github repository ``snakemake-workflows/dna-seq-gatk-variant-calling`` at tag ``v2.0.1`` and path ``workflow/Snakefile``, while using the contents of the local configuration file (see :ref:`snakefile-code-hosting-providers`).
```

docs/snakefiles/deployment.rst
```diff
@@ -85,6 +85,8 @@ Consider the following example:
...
 Note that it is possible to either specify the full URL pointing to the raw Snakefile as a string or to use the github marker as done here.
+GitHub and GitLab URLs are recognized automatically, including GitLab URLs on self-hosted instances.
+As an alternative convenience syntax, the same module source could also be written as ``github.com:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1:workflow/Snakefile``.
```
johanneskoester:
```suggestion
As an alternative convenience syntax, the same module source can also be written as ``github.com:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1:workflow/Snakefile``.
```

docs/executing/cli.rst
```diff
@@ -32,6 +32,11 @@ If called with the number of cores to use, i.e.
 
 Snakemake tries to execute the workflow specified in a file called ``Snakefile`` in the same directory (the Snakefile can be given via the parameter ``-s``).
```
johanneskoester:
```suggestion
Snakemake tries to execute the workflow specified in a file called ``workflow/Snakefile`` or ``Snakefile`` relative to the same directory (alternatively, the snakefile can be given via the parameter ``-s``).
```

docs/executing/cli.rst
```diff
@@ -32,6 +32,11 @@ If called with the number of cores to use, i.e.
...
 
+The ``--snakefile`` or ``-s`` argument can point to either a local file or an HTTP/HTTPS URL.
+In the latter case, relative ``include`` statements are resolved relative to the remote Snakefile location.
+In addition, GitHub and GitLab URLs are supported and resolved automatically.
+As a convenience syntax, hosted source files can also be referenced as ``github.com:owner/repo@ref:path/to/Snakefile`` or ``gitlab.example.org:group/project@ref:path/to/Snakefile``.
```
johanneskoester:
```suggestion
As a convenience syntax, hosted source files can also be referenced as ``github.com:owner/repo@ref:path/to/Snakefile`` or ``gitlab.example.org:group/project@ref:path/to/Snakefile``.

.. note::

    Note that referring to remote workflows via ``-s`` will only retrieve the snakefiles and everything directly referred by them, allowing an ad-hoc run. For workflows that expect configuration files or provide workflow profiles, it is usually more convenient to use `snakedeploy <https://snakedeploy.readthedocs.io/en/stable/workflow_users/workflow_deployment.html>`__ to retrieve a complete and templated instance.
```

src/snakemake/sourcecache.py
```diff
@@ -653,6 +653,117 @@ def get_path_or_uri(self, secret_free: bool) -> str:
...
+            )
+
+    return None
+
+
+def _infer_hosting_provider_file(path_or_uri: str) -> Optional[HostingProviderFile]:
+    return _infer_github_file(path_or_uri) or _infer_gitlab_file(path_or_uri)
+
+
+def _infer_hosting_provider_file_shorthand(
```
johanneskoester, line 808: I would say let us only allow these shorthands via -s (no full URLs).
That should make it simpler and cover 99% of the use cases.

docs/snakefiles/modularization.rst
```diff
@@ -396,6 +396,21 @@ For the latter, it is also possible to specify an alternative host, e.g.
...
+This also applies to self-hosted GitLab instances, as long as they expose the standard GitLab API raw-file URL form.
+
+As a convenience syntax, hosted source files can also be written in the compact form ``host:owner/repo@ref:path/to/Snakefile``.
+For example:
+
+.. code-block:: python
+
+    "github.com:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1:workflow/Snakefile"
+    "gitlab.com:owner/repo@main:workflow/Snakefile"
+    "gitlab.cern.ch:group/project@main:workflow/Snakefile"
```
johanneskoester: We can also support to skip the path in the end e.g.:

```
gh:snakemake-workflow/dna-seq-gatk-variant-calling@v2.0.1
```
If the last field with the path is missing, the infer function should just assume it is under ``workflow/Snakefile``.

src/snakemake/sourcecache.py
```diff
@@ -665,6 +776,9 @@ def infer_source_file(path_or_uri, basedir: Optional[SourceFile] = None) -> Sour
         raise SourceFileError(
             "must be given as Python string or one of the predefined source file marker types (see docs)"
         )
+    hosting_provider_file = _infer_hosting_provider_file_shorthand(path_or_uri)
+    if hosting_provider_file is not None:
+        return hosting_provider_file
```
johanneskoester:
```suggestion
    if hosting_provider_file := _infer_hosting_provider_file_shorthand(path_or_uri)
        return hosting_provider_file
```

src/snakemake/sourcecache.py
```diff
@@ -682,6 +796,9 @@ def infer_source_file(path_or_uri, basedir: Optional[SourceFile] = None) -> Sour
                 f"Failed to read source {path_or_uri} from git repo.", e
             )
         return LocalGitFile(root_path, file_path, ref=ref)
+    hosting_provider_file = _infer_hosting_provider_file(path_or_uri)
+    if hosting_provider_file is not None:
+        return hosting_provider_file
     # something else
```
johanneskoester: We have this above already. Why is a second check needed here?

tests/test_remote_snakefile.py
```diff
@@ -0,0 +1,70 @@
...
+                text=True,
+            )
+
+        assert result.returncode == 0, result.stdout
+
+        for relpath in get_expected_files(expected_results):
+            output = workdir / relpath
+            expected = expected_results / relpath
+            assert output.exists(), f"Missing output {relpath}\n{result.stdout}"
+            assert md5sum(output) == md5sum(expected)
```
johanneskoester: Rather use the run-function from tests/common.py. It already handles comparison of outputs with expected results. Further, this test can be moved into tests.py, no need for a separate module.

tests/test_remote_snakefile.py
```diff
@@ -0,0 +1,70 @@
...
+        ("127.0.0.1", 0), partial(QuietHTTPRequestHandler, directory=str(path))
+    )
+    thread = threading.Thread(target=server.serve_forever, daemon=True)
+    thread.start()
+    try:
+        yield f"http://127.0.0.1:{server.server_port}"
+    finally:
+        server.shutdown()
+        thread.join()
+        server.server_close()
```
johanneskoester: Move this into tests/common.py and avoid double definition in two test modules by importing it from there.

clelange: Implemented in a minimal way by moving the shared HTTP test helper into `common.py` and reusing it from `test_api.py` and `test_remote_snakefile.py`.

docs/snakefiles/modularization.rst
clelange: Done

clelange: Omitted shorthand paths now default to `workflow/Snakefile` in `sourcecache.py`, and the docs were updated in `cli.rst`, `deployment.rst`, and `modularization.rst`.

src/snakemake/sourcecache.py
clelange: Given that the underlying source-loading layer already supported generic HTTP/HTTPS URIs, and explicit `github(...)` / `gitlab(...)` provider markers already existed for module/workflow sources, I'd say that it doesn't hurt to support generic URIs here as well.

src/snakemake/sourcecache.py
clelange: The two checks cover different inputs. The early check handles the shorthand forms (`gh:` / `gl:`), while the later one handles full provider URLs like GitHub/GitLab `https://...` URLs after local and `git+file:` handling.

clelange:
Thanks for the review @johanneskoester !

I hope I caught all points. I'll be away next week, can get back to this afterwards if needed.

docs/snakefiles/deployment.rst
clelange: Done

docs/executing/cli.rst
clelange: Done

src/snakemake/sourcecache.py
clelange: I refactored this given another of your comments

src/snakemake/sourcecache.py
clelange: This is changed and simplified now

tests/test_remote_snakefile.py
clelange: Done

clelange:
I think that with the exception of your comment on what URLs are supported

> I would say let us only allow these shorthands via -s (no full URLs).

If you have a strong opinion on that one, I can remove things.

Besides that, I should now have managed to address all your comments, including getting rid of the additional test module. I have added some more test cases to ensure the parsing works.

docs/executing/cli.rst
```diff
@@ -30,7 +30,15 @@ If called with the number of cores to use, i.e.
...
+
+The ``--snakefile`` or ``-s`` argument can point to either a local file or an HTTP/HTTPS URL.
+In the latter case, relative ``include`` statements are resolved relative to the remote Snakefile location.
+In addition, GitHub and GitLab URLs are supported and resolved automatically.
+As a convenience syntax, hosted source files can also be referenced as ``gh:owner/repo@ref:path/to/Snakefile``, ``gh:github.example.org:owner/repo@ref:path/to/Snakefile``, ``gl:group/project@ref:path/to/Snakefile``, or ``gl:gitlab.example.org:group/project@ref:path/to/Snakefile``. If the final path is omitted, Snakemake assumes ``workflow/Snakefile``.
+
+.. note::
+    Note that referring to remote workflows via ``-s`` will only retrieve the snakefiles and everything directly referred by them, allowing an ad-hoc run. For workflows that expect configuration files or provide workflow profiles, it is usually more convenient to use `snakedeploy <https://snakedeploy.readthedocs.io/en/stable/workflow_users/workflow_deployment.html>`__ to retrieve a complete and templated instance.
```
johanneskoester, line 41:
```suggestion
    Note that referring to remote workflows via ``-s`` will only retrieve the snakefiles and everything directly referred by them, allowing an ad-hoc run. For workflows that expect configuration files or provide workflow profiles, it is usually more convenient to use `snakedeploy <https://snakedeploy.readthedocs.io/en/stable/workflow_users/workflow_deployment.html>`__ to retrieve a complete instance with an editable and pre-filled configuration file.
```

docs/snakefiles/deployment.rst
```diff
@@ -83,9 +83,11 @@ Consider the following example:
...
+Next, we define the module ``dna_seq`` to be loaded from the Github repository ``snakemake-workflows/dna-seq-gatk-variant-calling`` at tag ``v2.0.1`` and path ``workflow/Snakefile``, while using the contents of the local configuration file (see :ref:`snakefile-code-hosting-providers`).
 Note that it is possible to either specify the full URL pointing to the raw Snakefile as a string or to use the github marker as done here.
-With the latter, Snakemake can however cache the used source files persistently (if a tag is given), such that they don't have to be downloaded on each invocation.
+GitHub and GitLab URLs are recognized automatically, including GitLab URLs on self-hosted instances.
+As an alternative convenience syntax, the same module source could also be written as ``gh:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1`` or ``gh:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1:workflow/Snakefile``.
+With the latter, Snakemake can, however, cache the used source files persistently (if a tag is given), such that they don't have to be downloaded on each invocation.
```
johanneskoester, line 93:
```suggestion
Both with the github/gitlab markers and with the convenience syntax, Snakemake can cache the used git repository, such that it does not have to be downloaded on each invocation.
```

docs/snakefiles/modularization.rst
```diff
@@ -396,6 +396,24 @@ For the latter, it is also possible to specify an alternative host, e.g.
...
+.. code-block:: python
+
+    "gh:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1:workflow/Snakefile"
+    "gh:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1"
+    "gh:github.example.org:owner/repo@main:workflow/Snakefile"
+    "gl:owner/repo@main:workflow/Snakefile"
+    "gl:gitlab.cern.ch:group/project@main:workflow/Snakefile"
+
 While specifying a tag is highly encouraged, it is alternatively possible to specify a `commit` or a `branch` via respective keyword arguments.
 Note that only when specifying a tag or a commit, Snakemake is able to persistently cache the source, thereby avoiding to repeatedly query it in case of multiple executions.
```
johanneskoester, line 418:
```suggestion
While specifying a tag is highly encouraged, it is alternatively possible to specify a `commit` or a `branch` via respective keyword arguments.
Note that only when specifying a tag or a commit, Snakemake is able to persistently cache the source, thereby avoiding to repeatedly query it in case of multiple executions.

In many cases, it is also possible to provide plain URLs instead of markers.
Snakemake automatically recognizes GitHub and GitLab URLs and treats them like the corresponding hosting-provider-backed source files.
This also applies to self-hosted GitLab instances, as long as they expose the standard GitLab API raw-file URL form.

As a convenience syntax, hosted source files can also be written with an explicit provider prefix.
Use ``gh:owner/repo@ref:path/to/Snakefile`` for GitHub, optionally inserting a custom host as ``gh:github.example.org:owner/repo@ref:path/to/Snakefile``.
Use ``gl:group/project@ref:path/to/Snakefile`` for GitLab, optionally inserting a custom host as ``gl:gitlab.example.org:group/project@ref:path/to/Snakefile``.
If the trailing path is omitted, Snakemake assumes ``workflow/Snakefile``.
For example:

.. code-block:: python

    "gh:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1:workflow/Snakefile"
    "gh:snakemake-workflows/dna-seq-gatk-variant-calling@v2.0.1"
    "gh:github.example.org:owner/repo@main:workflow/Snakefile"
    "gl:owner/repo@main:workflow/Snakefile"
    "gl:gitlab.cern.ch:group/project@main:workflow/Snakefile"

```

src/snakemake/sourcecache.py
```diff
@@ -653,6 +653,61 @@ def get_path_or_uri(self, secret_free: bool) -> str:
...
+
+    if (
+        parsed.netloc == "github.com"
+        and len(path_parts) >= 5
+        and path_parts[2] == "blob"
+    ):
+        repo = "/".join(path_parts[:2])
+        ref = unquote(path_parts[3])
+        path = unquote("/".join(path_parts[4:]))
+        return GithubFile(repo=repo, path=path, branch=ref)
```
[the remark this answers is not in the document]
johanneskoester: I don't understand how this can ever be properly resolved. COnsider the following:
`https://github.com/snakemake-workflows/dna-seq-varlociraptor/raw/perf/autobump/.editorconfig`

The branch name is here `perf/autobump` and the path `.editorconfig`. But the code currently would take perf as the `ref`. And there is also no way to distinguish and correctly decide. The same holds for the interpretation of urls starting with raw.githubusercontent above.

src/snakemake/sourcecache.py
```diff
@@ -653,6 +659,143 @@ def get_path_or_uri(self, secret_free: bool) -> str:
...
+        query = parse_qs(parsed.query)
+        if any(
+            token_name in query
+            for token_name in ("private_token", "access_token", "job_token")
+        ):
+            return None
+        ref = query.get("ref")
+        if ref and len(ref) == 1:
+            return GitlabFile(
+                repo=unquote(path_parts[3]),
```
johanneskoester, line 773: Same here. Branches with slashes cannot be properly distinguished from paths.

clelange: Yes, indeed, I hadn't considered that well. Now I also understand better why you had concerns parsing full URLs as opposed to only the short handles...

In any case, I think I found a solution. Since the URL is inherently ambiguous, i.e. we cannot tell where the branch ends and path begins without querying git, the approach now is to generate all candidate (ref, path) splits at inference time via `_build_ref_path_candidates`, then resolve at fetch time via `_resolve_ref`, which checks `ref_exists` on the cloned repo to find the correct split.

I've added several test cases to make sure this works. If you think that adds to much overhead/makes things too complicated, I can also remove those additions and we only support the short handles.

PR #4078, opened 2026-03-12
feat: add --share-benchmark flag to send benchmark telemetry to a remote collector

btraven00:
This is an experimental feature that enables submission of benchmark information to a remote telemetry collector, after collective discussion on specs on Mar 11th and feedback from @johanneskoester @cmeesters and @tedil 

The submission protocol, as long as the data model and general implementation concerns is [under active discussion](https://github.com/btraven00/psb/blob/main/docs/spec.md#f-open-questions). An open collector implementation for testing, [written in go](https://github.com/btraven00/psb), is deployed at https://psb.biofx.dev. Examples including workflow, tool and command annotations with the proposed syntax can be found in the [examples folder of the psb repo](https://github.com/btraven00/psb/tree/main/examples).

Below there's a screenshot of the record view of the public collector.

image

### QC

### DIFF
--- a/src/snakemake/cli.py
+++ b/src/snakemake/cli.py
@@ -1571,6 +1571,21 @@ def get_argument_parser(profiles=None):
         action="store_true",
         help="Write extended benchmarking metrics.",
     )
+    group_behavior.add_argument(
+        "--share-benchmark",
+        default=False,
+        action="store_true",
+        help="Opt-in: collect benchmark telemetry for all shell rules and submit "
+        "to a PSB (Planetary Scale Benchmark) collector at workflow end. "
+        "Requires --share-benchmark-collector.",
+    )
+    group_behavior.add_argument(
+        "--share-benchmark-collector",
+        metavar="URL",
+        default=None,
+        help="URL of the PSB telemetry collector endpoint "
+        "(e.g. http://psb.example.com). Used with --share-benchmark.",
+    )
 
     group_cluster = parser.add_argument_group("REMOTE EXECUTION")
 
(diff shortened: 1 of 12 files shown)


tedil:
Haven't checked the commits yet, but here's some general remarks:
- it's quite a lot to read, and perhaps also a little too much functionality, I think starting with less may be a better strategy
- instead of having special _psb marker params, may make more sense to have a telemetry wrapper function (similar to `report(...)`) that allows specifying/overwriting/opting-out of some of the auto-collection, something like:
  ```
    params:
      number_of_parrots=telemetry(42, field="birbs"),
  ```
- I think auto-detection of tools either needs to be more sophisticated or (I think) better not be done at all; I think just dumping the shell/script/run code completely is way more valuable and much less error-prone. As I said before: I'd rather try to keep it simple with as little internal heuristic magic as possible.
- the software env definition should also be included (so conda env.yaml, container reference etc) _as is_
- regarding payload size limit: could buffer and submit once limit is reached or just make some smaller chunks afterwards; but requires proper rate limiting on the server side
- `--share-benchmark` or `--enable-telemetry`?
- a little more trust could be established via e.g., github oauth or similar? but I'd say that's for later
- I'd stick with the extended-benchmark columns for compatibility with what's already available in snakemake, and then just tack on some extra information that you already collect (cpu info etc)
- ...

cademirch:
I'm curious if this could be part of the logger? This seems like a fairly natural fit for a logger plugin, we would just need to send the benchmark info in a log record.

tedil:
> I'm curious if this could be part of the logger? This seems like a fairly natural fit for a logger plugin, we would just need to send the benchmark info in a log record.

that's a good idea, the question is if we can gather all necessary information when designing this as a logger plugin

btraven00:
after discussing with @cademirch and @tedil I will try to re-do this PR as a logger plugin. Additionally: 

- drop the auto-detection for tool and command, since this is fragile.
- pass annotations as a dictionary under a single pbs_telemetry param
- `--enable-telemetry` to explicitly opt-in  to sending (plust `--telemetry-collector`)

cmeesters:
@btraven00 Thank you for this PR! It is awesome! (Even if we need a few rounds to consolidate.)

PR #4079, opened 2026-03-12
feat: add workflow info to log

fgvieira:
### QC

## Summary by CodeRabbit

* **New Features**
  * Startup now logs a richer multi-field header (version, date/time, platform, user/environment, Python/conda info, command context, and config paths) and includes a unique workflow identifier.

* **Bug Fixes / Behavior**
  * The workflow-start event is now emitted and shown in end-user logs (no longer suppressed), replacing the simpler host-only startup message for clearer launch visibility.


src/snakemake/workflow.py
```diff
@@ -266,6 +266,50 @@ def async_run(self, coro):
...
+
+        conda_bin = shutil.which("conda")
+
+        return {
+            "datetime": datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
+            "snakemake_version": __version__,
+            "platform": platform.platform(),
+            "host": platform.node(),
+            "user": getpass.getuser(),
+            "conda_version": (
```
johanneskoester: I would leave the conda specific information out. Sooner or later it will be pixi, and moreover there is starship to show these things.

src/snakemake/logging.py
```diff
@@ -152,19 +152,40 @@ def default_formatter(rd):
...
+  Python version: {python_version}
+  Conda version: {conda_version}
+  Conda env: {conda_env} ({conda_prefix})
+  Command: {cmd}
+  Snakefile: {snakefile_main}
+  Base directory: {basedir}
+  Run directory: {rundir}
+  Working directory: {cwd}
+  Config file(s): {configfiles}
+  Config: {config}
```
johanneskoester: I would not display the config dictionary like this.

johanneskoester: Seems too messy for reasonable configs.

src/snakemake/logging.py
```diff
@@ -152,19 +152,40 @@ def default_formatter(rd):
...
+  Snakemake version: {snakemake_version}
+  Python version: {python_version}
+  Conda version: {conda_version}
+  Conda env: {conda_env} ({conda_prefix})
+  Command: {cmd}
+  Snakefile: {snakefile_main}
+  Base directory: {basedir}
+  Run directory: {rundir}
+  Working directory: {cwd}
+  Config file(s): {configfiles}
```
johanneskoester, line 179: Shouldn't these be joined by a comma or so?

johanneskoester (changes requested):

src/snakemake/logging.py
johanneskoester: I think it might be better to do the formatting in here, instead of joining stuff upstream in the info_header method. This way, logger plugins can decide how they display the list of config files instead of having to parse the string.

johanneskoester (changes requested):

src/snakemake/workflow.py
```diff
@@ -267,6 +267,40 @@ def async_run(self, coro):
...
+            "snakemake_version": __version__,
+            "platform": platform.platform(),
+            "host": platform.node(),
+            "user": getpass.getuser(),
+            "python_version": sys.version,
+            "cmd": " ".join(sys.argv),
+            "basedir": self.basedir,
+            "rundir": self.rundir,
+            "cwd": self.workdir_init,
+            "configfiles": ", ".join([str(cfg) for cfg in self.configfiles]),
```
fgvieira:
```suggestion
            "configfiles": self.configfiles,
```

src/snakemake/logging.py
fgvieira: `info_header` now returns `configfiles` as a `list`. We could join with a comma, but I think this way is a bit cleaner and looks almost the same (apart from the `[]` of the list).

johanneskoester (approved):

PR #4086, opened 2026-03-13
feat: add `tolerance` and `impatient` params to optional input

CameronMcClymont:
Builds on #4059 to also support `tolerance` and `impatient` options when using `optional()` inputs.

- `tolerance` (default: `1`) allows for specifying the fraction of inputs that are allowed to be missing
- `impatient` (default: `False`) can be set to start the rule as soon as the `tolerance` is satisfied, using whatever inputs have been received so far

The use of `tolerance` and `impatient` only makes sense when `optional()` wraps a list-like input. This could be e.g. a literal list of files, a wildcard path, or an `expand()`.

PR #4067 adds the same support, but specifically for scatter-gather.

The rationale behind adding support for this feature is described in https://github.com/snakemake/snakemake/issues/4021.

### Examples

Example 1 - wait for relevant upstream rules to finish and proceed if at least 2 of `["a.json", "b.json", "c.json", "d.json"]` are available:

```
rule merge:
    input:
        optional(["a.json", "b.json", "c.json", "d.json"], tolerance=0.5),
...
```

Example 2 - proceed as soon as 2 of `["a.json", "b.json", "c.json", "d.json"]` are available:

```
rule merge:
    input:
        optional(["a.json", "b.json", "c.json", "d.json"], tolerance=0.5, impatient=True),
...
```

Example 3 - proceed as soon as 90% of expected files are available:

```
rule merge:
    input:
        optional("myDir/{myFile}.json", tolerance=0.1, impatient=True),
...
```

### Implementation

Conceptually, #4059 introduces a second dependency type (optional-hard), and this PR introduces a third.

- Hard dependency (existing behaviour) - the upstream job must finish successfully before the downstream job can start
- Optional-hard (tolerance) - the downstream job still waits for the upstream job to finish, but when the upstream job fails, `_n_until_ready` is decremented anyway which allows the downstream job to start. The downstream job then determines whether the failure is tolerated.
- Optional-soft (impatient) - the downstream job does not wait for the upstream job at all (it’s excluded from `_n_until_ready`). It starts as soon as enough files have been produced to meet its tolerance demands

### QC

When #4059 has docs, I will build on them in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added optional() inputs with per-group tolerance and impatient mode to allow tolerant, early or deferred merging of missing inputs.
  * IO now exposes an is_optional flag for inputs.

* **Bug Fixes**
  * Missing optional-only outputs are no longer treated as hard failures; missing optional inputs are skipped with warnings.

* **Documentation**
  * Docs updated describing optional inputs, tolerance, and impatient behavior.

* **Tests**
  * Extensive tests covering tolerance, impatient behavior, and multi-stage scenarios.


CameronMcClymont:
@coderabbitai review

CameronMcClymont:
@coderabbitai review

CameronMcClymont:
@coderabbitai review

CameronMcClymont:
There has been a suggestion to support the automatic cancellation of upstream jobs left 'dangling' by rules using `impatient` when they start early due to their `tolerance` being satisfied. This would require several more changes in this repo as well as adding a `cancel_jobs(self, jobs: List[JobExecutorInterface])` method in the `AbstractExecutor` of the executor plugin interface repo to support cancelling remote jobs. The downstream executor plugins would need to then be updated to implement this method, so it would likely take some time.

I'm interested to hear thoughts on whether we should try and include this before merging, or open an issue and a new PR to handle the cancellation of dangling upstream jobs.

Issue #3994, opened 2026-03-05
Workflow profile selection broken in 9.16.3


zwets:
**Snakemake version**
9.16.3

**Describe the bug**
Passing `--workflow-profile default` (or any name other than `default`) produces this error:

```
Using workflow specific profile default for setting default command line arguments.
Error: profile given but no config.yaml found. Profile has to be given as either absolute path, relative path or name of a directory available in either /etc/xdg/snakemake or /hpc/users/zwets/.config/snakemake.
```

The error occurs regardless of whether `profiles/default/config.yaml` is present in the working directory, next to the Snakefile, or in `workflow`. These are its proper locations [according to the documentation](https://snakemake.readthedocs.io/en/stable/executing/cli.html#profiles).

Note that the error message reports the **profile** locations as opposed to **workflow-profile** locations. 

When `--workflow-profile` is omitted, the file `profiles/default/config.yaml` _does_ get read (in both working directory and under `workflow`), as can be easily verified by adding a non-existing `borkbork: true` parameter to it.

The same error occurs for any other workflow profile in `profiles`. 

**Minimal example**
Put a file `profiles/default/config.yaml` in any workflow directory, then try run with `--workflow-profile default` and the error will occur.

Issue #4006, opened 2026-03-09
feat: new default profile file name `profiles.yaml`, also allow specifying a yaml file instead of a directory


dlaehnemann:
Currently, users often confuse the profile YAML files that can live in subfolders of `~/.config/snakemake/` or `workflow/profiles/` or be specified via `--workflow-profiles` with the configuration of the workflow itself, that usually lives inside the `config/` folder, as both are named `config.yaml`.

In a brief discussion this morning, including @johanneskoester , we decided that a good way to improve this in a non-breaking way, would be to move to `profiles.yaml` as the default name (also mentioned in the docs), but still allow for the `config.yaml` file name for profiles.

In addition, we want to be able to have the `--workflow-profile` point to a specific YAML file instead of a directory, but also keep the directory as the default way.

PR #4087, opened 2026-03-13
feat: clean up profile handling, switch to `profile.yaml` as default file name, allow specifying yaml file instead of dir

dlaehnemann:
This pull request cleans up the profile handling behavior, adds a number of relevant test cases and improves the respective documentation, especially also clarifying in which order profiles (and command-line arguments) take precedence and that they overwrite each other on the level of top-level YAML keys.

### QC

## Summary by CodeRabbit

* **New Features**
  * Profiles may be specified as file paths; resolver accepts profile.yaml or config.yaml, supports multiple --profile entries, and resolves workflow-profile paths relative to the workflow or current directory.
* **Bug Fixes**
  * Clearer help/error messages for missing profiles; more predictable profile selection and re-resolution when profiles or defaults are provided.
* **Documentation**
  * Major expansion of Profiles docs with usage, lookup rules, precedence, templating, and examples.
* **Tests**
  * Added tests for multiple profiles, env-var profiles, filename handling, and workflow-profile resolution.


dlaehnemann:
This PR addresses #4006. In addition, it should resolve #3994.

cademirch:
great work @dlaehnemann I'll take a closer look at this later today.

docs/executing/cli.rst
```diff
@@ -116,100 +117,377 @@ We advice all workflow developers to inform potential users of the best suited b
...
+1. You give it a standardised filename (for example ``profile.v9+.yaml``, see the section on :ref:`profile-files`) and save it in the default folder hierarchy relative to the ``Snakefile`` or the current working directory, usually either ``profiles/default/profile.yaml`` or ``workflow/profiles/default/profile.yaml``.
+2. You use the command line argument ``--workflow-profile``.
+
+.. note::
+
+  When using modules, the profile will not be propagated to the main workflow importing that module.
+  However, using `snakedeploy deploy-workflow` to deploy a workflow as a module, will also copy any profiles included under the standard location `workflow/profiles` (for more info, see `the snakedeploy documentation for deploying workflows <https://snakedeploy.readthedocs.io/en/stable/workflow_users/workflow_deployment.html>`_).
+  Starting from this import, or starting with a new file, users can create a profile for that main workflow.
+
+When using the command line argument, the ``default/`` location is not searched (unless you explicitly specify ``--workflow-profile default``).
```
cademirch: This is slightly confusing to me still. If I don't use `--workflow-profile`, but there is a profile in `default/` will it be used? I think the answer is yes. 

I think this is unintuitive: a user who never passes `--workflow-profile` might not realize a workflow profile is being loaded, especially if they inherited a project that happens to have a `profiles/default/profile.yaml`

cademirch: I think somewhere in this section we need something like:
>Even without specifying ``--workflow-profile``, Snakemake will automatically
   search for and apply a workflow profile in ``profiles/default/`` (relative to
   the Snakefile or working directory). To prevent any workflow profile from
   being loaded, use ``--workflow-profile none

dlaehnemann:
Hey Cade, many thanks for having a look in detail. I agree that this is a lot; it took me a week to figure out all of the details of the behavior and properly test it.

And I do think, that the documentation could be even better. I tried to somewhat structure it so that you first get a high-level overview which profiles are for which purposes, and then get all the details for each profile type. But maybe a longer main profile section with examples would make it quicker for new users (or when you are just looking it up for the 20th time;).

Also, after we merge this pull request, I would like to move this to its own section under `Executing Workflows`. That makes it more discoverable and the `CLI` section a bit cleaner, especially as the `profiles` section has now considerably grown in lenght. But I didn't want to confuse re-writing with moving, so that the diffs are still somewhat readable...

src/snakemake/cli.py
```diff
@@ -449,13 +460,14 @@ def get_argument_parser(profiles=None):
...
...ontaining the profile YAML or the actual YAML file itself. If a profile name or folder is given,it has to contain a file `profile.yaml` (or a `config.yaml` file, for backwards compatibility). This file can have an optional infix specifying a minimal snakemake version (for example `profile.v9+.yaml`). Cluster-specific profiles can be obtained from https://github.com/snakemake/snakemake-cluster-profiles. The profile can also be set via the environment variable `$SNAKEMAKE_PROFILE`. However, once you provide a profile via the command line argument `--profile`, this environment variable is ignored. And to override this variable without setting another one, provide the value `none` to this argument. Finally, you can specify this argument multiple times. In this case, the profiles get merged with the later `--profile` instances overriding top-level entries in profiles specified earlier. For example, if the last `--profile` specifies the top level `default-resources:` keyword, all entries under that keyword from previous `--profile`s will be ignored. Similarly, also specifying any of the top-level keys from your profile as a command line argument will overwrite this whole top-level key.",
```
johanneskoester:
```suggestion
        help=f"Profile to use for configuring the Snakemake run with settings regarding the compute environment. Every key in this YAML file gets parsed into the respective command line argument, so e.g. `executor: slurm` will get parsed to the argument `--executor slurm` and `default-resources: mem_mb: 16000` is interpreted as `--default-resources mem_mb=16000. You can specify a Snakemake profile as (i) a profile name, (ii) a relative path to a folder or (iii) the relative path to the profile YAML file itself. Snakemake will look for the existence of the relative path or a folder with the profile name in `{dirs.site_config_dir}`, `{dirs.user_config_dir}` and the current working directory. Alternatively, this can be also be an absolute path to a folder containing the profile YAML or the actual YAML file itself. If a profile name or folder is given,it has to contain a file `profile.yaml` (or a `config.yaml` file, for backwards compatibility). This file can have an optional infix specifying a minimal snakemake version (for example `profile.v9+.yaml`). Example profiles for certain infrastructure can be obtained from https://github.com/snakemake/snakemake-cluster-profiles. The profile can also be set via the environment variable `$SNAKEMAKE_PROFILE`. However, once you provide a profile via the command line argument `--profile`, this environment variable is ignored. And to override this variable without setting another one, provide the value `none` to this argument. Finally, you can specify this argument multiple times. In this case, the profiles get merged with the later `--profile` instances overriding top-level entries in profiles specified earlier. For example, if the last `--profile` specifies the top level `default-resources:` keyword, all entries under that keyword from previous `--profile`s will be ignored. Similarly, also specifying any of the top-level keys from your profile as a command line argument will overwrite this whole top-level key.",
```

dlaehnemann: Adressed this in:
https://github.com/snakemake/snakemake/pull/4087/commits/ae64d72f950b19e55d57e2e3537ad3f7aade3a82

dlaehnemann:
Ah yeah, that will make the diffs easier to read in the future! Done.

PR #4090, opened 2026-03-13
feat: add a function to help with prepending arguments to filenames; close #672

### LINKED ISSUE #672: Inline join/expand syntax to support args into multiple command line arguments
I'd love for this syntax to be supported (many GATK/Picard tools require `--input` in front of every argument.
```
rule all:
    input:
        'out.txt'

rule:
    input:
        args = ['1', '2', '3']
    output:
        'out.txt'
    shell:
        '''echo {' '.join('--input ' + v for v in params.args)} > {output}'''
```

Otherwise, I have to write:
```
rule all:
    input:
        'out.txt'

rule:
    input:
        args = ['1', '2', '3']
    params:
        cl_args = ' '.join('--input ' + v for v in ['1', '2', '3'])
    output:
        'out.txt'
    shell:
        '''echo {params.cl_args} > {output}'''
```

This s redundant and gets complicated when `['1', '2', '3']` is a checkpoint function or something complicated

edbennett:
Per issue #672, there are many situations where you want to pass a list of input files, each prepended by the same command-line option. Currently, this requires a monstrosity of a list comprehension in the `shell:` directive, or an awkward list comprehension in `params:`.

This PR provides a helper function`prepend`, which makes it easier to generate the argument list in a `params:` directive.

    params:
        data=prepend("--input", input.data)
    input:
        data=["a.dat", "b.dat", "c.dat"],
    shell:
        "some_tool {params.data}"

will run

    some_tool --input a.dat --input b.dat --input c.dat

### QC

* [X] I, as a human being, have checked each line of code in this pull request

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **New Features**
  * Added a utility to prepend flags/prefixes to input filenames or callable-returned inputs, with configurable spacing.

* **Documentation**
  * Added a guide for the prepend helper with examples showing spaced and no-space usage and multiple input patterns.

* **Tests**
  * Added tests and a sample CLI helper validating behavior for single, multiple, unnamed, and no-space scenarios.


johanneskoester:
TBH, I think I would rather not want parameter syntax like this to live in the input directive. Imagine rule inheritance for example. Anybody replacing the input during that with different paths will have to do this again. However, I definitely see the utility of this and would thus suggest the following syntax instead, in line with other helper functions that we already have (https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#semantic-helpers):

```python
rule a:
    input:
        data=["a.dat", "b.dat", "c.dat"]
    params:
        prepend("--input ", input.data)  # note that we here make use of the rule item access helpers, obviating the need for defining a new function here (https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#rule-item-access-helpers)
    shell:
        "some_tool {input}"
```

edbennett:
Makes sense, thanks!

docs/snakefiles/rules.rst
```diff
@@ -798,6 +798,39 @@ When selecting input files, sometimes you might end up with an irregular list of
     flatten([1, "a", [2,"b"], ["c","d",["e", 3]]]) # returns ["1", "a", "2", "b", "c", "d", "e", "3"]
 
 
+.. _snakefiles-prepend:
+
+prepend
```
johanneskoester: Can you please move the section into the section about semantic helper functions in rules.rst?

docs/snakefiles/rules.rst
```diff
@@ -798,6 +798,39 @@ When selecting input files, sometimes you might end up with an irregular list of
...
+
+Some tools require adding a flag before each filename they are given.
+To add a flag before each input file in a list,
+the ``prepend`` helper can be used.
+For example:
+
+.. code-block:: python
+
+    params:
+        data=prepend("--input", input.data)
```
johanneskoester: To avoid ambiguity, we should give it a more unique name:

```suggestion
        data=prepend_param("--input", input.data)
```

johanneskoester (changes requested):
Nice! Just two minor things.

edbennett:
Done!

johanneskoester (approved):

Issue #3009, opened 2024-08-10
get() method of unexecuted checkpoint not raising exception


jlumpe:
**Snakemake version**
Occurs in `8.16.0` and `8.13.0`, does not occur in `7.32.4`.

**Describe the bug**

I am running into an issue where under certain circumstances, a checkpoint's `get()` method does not raise the expected `IncompleteCheckpointException` when it has not yet been run. I've been able to create minimal example below, The problem happens in the input function of the `problem` rule, which reads the output of the `my_checkpoint` checkpoint rule.

The example runs fine starting from scratch. However, if I create an empty file for the `problem` rule's output file, then execute a rule that takes that file as an input, the call to `checkpoints.my_checkpoint.get()`  succeeds even though the checkpoint has not been run.

Several warnings like the following are generated, which may or may not be related:

```
RuntimeWarning: coroutine 'DAG.sanitize_local_storage_copies' was never awaited
```

**Minimal example**

```python
import os, sys

checkpoint my_checkpoint:
    output: 'checkpoint.txt'
    shell: 'echo 1 2 > {output}'

rule make_problem_input:
    output: 'input-{i}.txt'
    shell: 'echo data-{wildcards[i]} > {output}'

def my_input_func(wc):
    print('!!!!!!!!!! input function called', file=sys.stderr)
    checkpoint_file = checkpoints.my_checkpoint.get().output[0]
    assert os.path.exists(checkpoint_file), f'{checkpoint_file} does not exist!'
    print('!!!!!!!!!! checkpoint file exists', file=sys.stderr)
    data = open(checkpoint_file).read()
    nums = data.strip().split()
    return expand(rules.make_problem_input.output, i=nums)

rule problem:
    input: my_input_func
    output: 'problem.txt'
    shell: 'cat {input} > {output}'

rule final:
    input: 'problem.txt'
    output: 'final.txt'
    shell: 'cat {input} > {output}'
```

**Logs**

With directory empty except for the `Snakefile`:

```bash
snakemake --profile none final.txt
```

```
Assuming unrestricted shared filesystem usage.
Building DAG of jobs...
!!!!!!!!!! input function called
!!!!!!!!!! input function called
/home/jared.lumpe/opt/mambaforge/envs/signature-finder/lib/python3.12/site-packages/snakemake/dag.py:1612: RuntimeWarning: coroutine 'DAG.sanitize_local_storage_copies' was never awaited
  self.sanitize_local_storage_copies()
[...]

PR #4092, opened 2026-03-13
fix: consider checkpoint jobs during update_needrun

### LINKED ISSUE #3879: Deletion of checkpoint outputs is not properly detected
Hello, 
After running a pipeline with checkpoints,and then deleting an output generated by a checkpoint, rerunning the pipeline does not properly detect that the file is missing and that the checkpoint needs to run again.
Below an example pipeline

```python

checkpoint checkpoint_test:
    output:
        "output/test_{n}.txt"
    shell:
        '''
        shuf -i 1-100 -n 1 > {output[0]}
        '''

    max_value = None
    selected_file = None
    for i in range(10):
        check = checkpoints.checkpoint_test.get(n=i)
        curr_file = check.output[0]
        with open(curr_file) as f:
            n = int(f.read())
        if max_value is None or n > max_value:
            max_value = n
            selected_file = curr_file
    return selected_file

rule aggregate:
    input:
        aggregate_input
    output:
        "output/selection.txt"
    shell:
        '''
        cp {input[0]} {output[0]}
        '''

rule all:
    default_target: True
    input: "output/selection.txt"

```
If one of the checkpoint outputs is removed before re-executing the pipeline, checkpoints.get passes and leads to a crash in the following line. 

The shell script below systematically produces the error

```shell
rm -rf output
snakemake --cores 1
rm -rf output/test_4.txt
snakemake --cores 1
```

I tried looking at the code, it seems to me that this function https://github.com/snakemake/snakemake/blob/9598655f9ba99380b13b8a5797110352d41d2831/src/snakemake/dag.py#L1416 doesn't check if output files are missing or have changed. A check is done here https://github.com/snakemake/snakemake/blob/9598655f9ba99380b13b8a5797110352d41d2831/src/snakemake/dag.py#L1598 but only for jobs that have been already considered for rerunning in the aforementioned function. However I am not sure enough of the reasons behind the current implementation to propose a PR.

tedil:
Resolves #3879

Not sure this is the correct solution, though (might have unforeseeable side consequences, though I believe this should be fine, may just cause some more I/O overhead for checking the checkpoint's dependencies).

### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved checkpoint job handling to ensure they are properly considered alongside target jobs during workflow execution evaluation, affecting how missing outputs and updated inputs are handled.

* **Tests**
  * Added test to verify that missing checkpoint outputs are correctly regenerated when workflows are re-run.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -1439,7 +1439,7 @@ class DAG(DAGExecutorInterface, DAGReportInterface, DAGSchedulerInterface):
 
             if is_forced(job):
                 reason.forced = True
-            elif job in self.targetjobs:
+            elif job in self.targetjobs or job.is_checkpoint:
                 if not job.has_products(include_logfiles=False):
                     if job.input:
                         if job.rule.norun:
--- a/tests/test_checkpoint_missing_output/Snakefile
+++ b/tests/test_checkpoint_missing_output/Snakefile
@@ -1,0 +1,34 @@
+checkpoint checkpoint_test:
+    output:
+        "output/test_{n}.txt"
+    shell:
+        '''
+        echo {wildcards.n} > {output[0]}
+        '''
+
+def aggregate_input(wildcards):
+    max_value = None
+    selected_file = None
+    for i in range(10):
+        check = checkpoints.checkpoint_test.get(n=i)
+        curr_file = check.output[0]
+        with open(curr_file) as f:
+            n = int(f.read().strip())
+        if max_value is None or n > max_value:
+            max_value = n
+            selected_file = curr_file
+    return selected_file
+
+rule aggregate:
+    input:
+        aggregate_input
+    output:
+        "output/selection.txt"
+    shell:
+        '''
+        cp {input[0]} {output[0]}
+        '''
+
+rule all:
+    default_target: True
+    input: "output/selection.txt"
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1629,6 +1629,24 @@ def test_checkpoint_rerun():
         assert f.read().strip() == "1"
 
 
+def test_checkpoint_missing_output():
+    """test for issue 3879"""
+    test_dir = dpath("test_checkpoint_missing_output")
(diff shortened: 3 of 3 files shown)


cademirch (dismissed):
~looks like a reasonable solution to me.~

edit: i re-read the referenced issue. i am unsure about the solution here now. how can we know if (and which) outputs of a checkpoint are missing? i guess we assume that checkpoints are determinstic each run? in any case im confused

tedil:
After discussion w/ @johanneskoester this may turn out to be problematic when a checkpoint generates temporary files. While this fix would be the only way to notice this, this would also force re-runs for checkpoints with temp outputs, even though technically no changes that'd change the outcome have occurred.
Since this may be unexpected behaviour for most users, perhaps we should instead just document that behaviour and add an example that demonstrates this?

johanneskoester (changes requested):
First decision after live discussion: we think that it rather should be documented that intermediate jobs between the checkpoint and the input function can not be triggered for reruns if the checkpoint output files are missing. Avoiding that would mean to cause reruns of the checkpoint and everything downstream if the checkpoint output is missing (e.g. because of being marked as temp). Workaround: use -R <checkpoint_name> in such cases.

src/snakemake/dag.py
```diff
@@ -1439,7 +1439,7 @@ async def update_needrun(job):
 
             if is_forced(job):
                 reason.forced = True
-            elif job in self.targetjobs:
+            elif job in self.targetjobs or job.is_checkpoint:
```
Hocnonsense, line 1442: I don't think this is the accurate place to fix this.
Marking every checkpoint as needrun unconditionally here has a broader effect than intended: it forces re-evaluation of checkpoints whose outputs already exist and are up to date, which can cause unnecessary re-runs in incremental workflows.

This is a similar case to #3009, and the fix should be in `update_checkpoint_dependencies`.

johanneskoester:
Closing in favor of the merged #4124. Nevertheless, thanks for bringing this up and helping to find the correct solution!

PR #4097, opened 2026-03-13
fix(checkpoints): Fix importing checkpoints into modules

### LINKED ISSUE #4096: Can't import checkpoints from module
**Snakemake version**

9.16.3

**Describe the bug**

Importing checkpoints from a module via the `use rule` directive does not work.

**Logs**

**Minimal example**

There currently exists a branch with a test case, the PR draft to that will be linked once it's created.

**Additional context**

jonasfreimuth:
As described in https://github.com/snakemake/snakemake/issues/4096, importing checkpoints is currently not possible. This PR currently provides a test case to verify this.

When the actual fix is implemented, this hopefully closes https://github.com/snakemake/snakemake/issues/4096

### QC

This PR currently only provides a test for a feature that, while not explicitly documented, can be expexted to work.

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **New Features**
  * Imported-module checkpoints are now propagated to parent workflows so imported checkpoints can be referenced and used at runtime.

* **Bug Fixes / Behavior**
  * Checkpoint lookup/reporting improved: missing checkpoints produce clearer errors that include available checkpoint names and dynamic resolution of imported checkpoints at runtime.

* **Tests**
  * New tests cover importing, resolving, and using checkpoints across modules and nested module scenarios.


Hocnonsense:
Mark: after modify `test_module_nested`, `test_checkpoint_import` may be duplicated

src/snakemake/checkpoints.py
```diff
@@ -35,17 +35,19 @@ class CheckpointsProxy(Checkpoints):
...
+    def __getattr__(self, name):
+        if name in self._rules:
+            return self._rules[name]
+        raise WorkflowError(
+            f"Checkpoint {name} is not defined in this workflow. "
+            f"Available checkpoints: {', '.join(self._rules)}"
+        )
+
+    def _register(self, rule: "Rule", fallback_name=None):
+        self._rules[fallback_name or rule.name] = Checkpoint(rule, self._parent)
```
johanneskoester: This I don't get. If fallback_name is not None, then this will be the only name that this checkpoint is registered in. Isn't it then kind of the opposite of a fallback_name?

Hocnonsense: Thanks for point this out. Sure, here `fallback_name` will never be `None`, and fall back of `self._rules[rule.name]` is useless because all those in `child_modifier.globals["checkpoints"]` will be added to the outer one with name.
With this insight, I also see `WorkflowModifier.inherit_rule_proxies` can be problemic, only the `avail_rulename`s should be added from the module. @coderabbitai do you have any idea?

Hocnonsense: @coderabbitai I suggess `def _register(self, rule: "Rule", name=None): self._rules[name] = Checkpoint(rule, self._parent)`, since it is only called in Workflow.rule.<decorate> via `self.globals["checkpoints"]._register(rule, fallback_name=orig_name)`, where `orig_name` never be `None`
Wait, Since checkpoint is registered only if the rule used (`if not name: return ruleinfo.func`), the check can skip, and only those in `child_modifier.rule_proxies._rules` need to be checked?

PR #4099, opened 2026-03-13
fix: persistence-backend cli enum parsing

tedil:
what it says on the tin. but doesn't seem to work, even though other arguments to it exactly the same way? What am I missing?

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Chores**
  * Improved parsing and validation for the --persistence-backend command-line argument to accept values via the backend parser while preserving existing behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 * Allow storing snakemake metadata in files or databases ([#4012](https://github.com/snakemake/snakemake/issues/4012)) ([dd75f31](https://github.com/snakemake/snakemake/commit/dd75f31bffecdbed1db5597c58f58fcca7ded2d2))
 * Allow to specify comparison command per-unit test ([#3956](https://github.com/snakemake/snakemake/issues/3956)) ([b88171c](https://github.com/snakemake/snakemake/commit/b88171cea0752b65cb6687906f7778c7f229c93d))
-* job table orderd topological when run is started ([#4018](https://github.com/snakemake/snakemake/issues/4018)) ([75cf506](https://github.com/snakemake/snakemake/commit/75cf506bc367eb0fd9ef75cdff30ebfe87714bc4))
+* job table ordered topological when run is started ([#4018](https://github.com/snakemake/snakemake/issues/4018)) ([75cf506](https://github.com/snakemake/snakemake/commit/75cf506bc367eb0fd9ef75cdff30ebfe87714bc4))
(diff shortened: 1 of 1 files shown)


tedil:
There's at least one more cli option with this issue, which is ExecMode (should use choices instead of all, and parse_func instead of type).

tedil:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> ⚠️ Outside diff range comments (1)
> 
> > src/snakemake/cli.py (1)> `1576-1583`: _⚠️ Potential issue_ | _🔴 Critical_
> > > **Fix `choices` parameter to use `all()` instead of `choices()`.**
> > > The `type=PersistenceBackend.parse_choice` and `choices=PersistenceBackend.choices()` combination is incompatible. argparse applies type conversion before validating against choices. After `parse_choice` converts the string input to an enum member, argparse checks if this enum member exists in `choices`, which contains strings. This comparison fails.
> > > Looking at working patterns in this file:
> > > 
> > > * Line 1164: `--list-changes` uses `choices=ChangeType.all()` with `type=ChangeType.parse_choice` ✓
> > > * Line 1796: `--mode` uses `choices=ExecMode.all()` with `type=ExecMode.parse_choice` ✓
> > > 
> > > Both use `all()` instead of `choices()`. Change line 1577 to match:
> > > 
> > > Suggested fix
> > > ```diff
> > >      group_behavior.add_argument(
> > >          "--persistence-backend",
> > > -        choices=PersistenceBackend.choices(),
> > > +        choices=PersistenceBackend.all(),
> > >          default=PersistenceBackend.FILE,
> > >          type=PersistenceBackend.parse_choice,
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@src/snakemake/cli.py` around lines 1576 - 1583, The argparse configuration
> > > for the "--persistence-backend" option uses PersistenceBackend.choices() while
> > > also supplying type=PersistenceBackend.parse_choice, causing a type/choice
> > > mismatch; update the choices argument to use PersistenceBackend.all() so
> > > argparse compares enum members returned by parse_choice against a list of enum
> > > members (change the call from PersistenceBackend.choices() to
> > > PersistenceBackend.all() in the option definition for "--persistence-backend"
> > > near the references to PersistenceBackend.parse_choice).
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> ℹ️ Review info

Yeah, the enum choices parsing seems a bit inconsistent. Would not be much of an issue with click.
I think choices is correct because it makes the enum names available for proper display on the cli. Not sure about value conversion.

PR #4108, opened 2026-03-16
feat: add temp to default pathvars

fgvieira:
### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **New Features**
  * Added "stats", "reports", and "temp" as new default path variables alongside results, resources, logs, and benchmarks for use in Snakemake workflows.

* **Documentation**
  * Updated docs to list the expanded set of default path variables and their mapping to output paths.

### DIFF
--- a/docs/snakefiles/rules.rst
+++ b/docs/snakefiles/rules.rst
@@ -362,7 +362,7 @@ The values of pathvars can thereby even contain wildcards themselves.
 Pathvar defaults
 """"""""""""""""
 
-By default, Snakemake offers the pathvars ``results``, ``resources``, ``logs``, ``benchmarks``.
+By default, Snakemake offers the pathvars ``results``, ``resources``, ``temp``, ``logs``, ``benchmarks``.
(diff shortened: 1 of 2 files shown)


johanneskoester (commented):
Interesting idea! So, I was so far assuming that such a folder (if a special one is needed) would rather be the system tempdir. Maybe that should be the default instead of the local relative folder `temp`? Unsure though.

fgvieira:
I always have both temp folders:
- system tempdir for rules temp files (those generated by rules - i.e. sort)
- local tempdir for workflow temp files (those generated by the workflow - i.e. temp intermediate output files)

johanneskoester:
> I always have both temp folders:
> 
>     * system tempdir for rules temp files (those generated by rules - i.e. sort)
> 
>     * local tempdir for workflow temp files (those generated by the workflow - i.e. temp intermediate output files)

That makes a lot sense! And by making it a pathvar, people can even choose to set it to a different location (e.g. a fast but small NVMe SSD or something like that).

johanneskoester (approved):

PR #4119, opened 2026-03-19
fix: ensure that strings that purely contain integers or floats (e.g. "42") remain strings when parsing profiles

### LINKED ISSUE #3992: Numeric Slurm Partition Names Summed in Snakemake Group Jobs
**Snakemake version**
v9.16.3 (also checked HEAD of `main` — not fixed)

**Describe the bug**
When a group job has 2+ sibling rules, Snakemake sums *all* integer-typed resources across siblings — including categorical resources like `slurm_partition`. If the partition name happens to be numeric (e.g. `20`), the summed value (`40`) is passed to `sbatch`, which fails.

The root cause is a two-step type-information loss:

1. **Coercion 1** — String serialization in [`profiles.py:29`](https://github.com/snakemake/snakemake/blob/main/src/snakemake/profiles.py#L29): `f"{key}={val}"` flattens all values to strings, so a YAML string `"20"` becomes indistinguishable from the integer `20`.
2. **Coercion 2** — Integer parse in [`resources.py:1041`](https://github.com/snakemake/snakemake/blob/main/src/snakemake/resources.py#L1041): `val = int(val)` coerces any numeric-looking string back to `int`.

Since the value is now an integer, `_is_string_resource()` classifies it as numeric, and the intralayer merge at [`resources.py:351`](https://github.com/snakemake/snakemake/blob/main/src/snakemake/resources.py#L351) sums it: `20 + 20 = 40`.

Quoting in YAML does not help — Coercion 1 destroys the type, and Coercion 2 re-parses it as int regardless.

This affects any resource whose value is a valid integer and is used in a group job with 2+ siblings (e.g. `slurm_partition`, `slurm_account`, `slurm_cluster`, `constraint`, or any custom resource with a numeric-looking value). Non-grouped jobs are unaffected (no summing with a single rule), and non-numeric string values are unaffected (`int()` fails and they stay as strings).

**Logs**
```
sbatch: error: invalid partition specified: 40
```

**Minimal example**

`Snakefile`:
```python
rule a:
    input: "data.txt"
    output: "a.txt"
    group: "mygroup"
    shell: "cp {input} {output}"

rule b:
    input: "a.txt"
    output: "b.txt"
    group: "mygroup"
    shell: "cp {input} {output}"
```

`profile/config.yaml`:
```yaml
executor: slurm
default-resources:
  slurm_partition: 20
  tasks: 1
```

Run with:
```bash
snakemake --profile profile/
```

Expected: `sbatch ... -p 20`
Actual: `sbatch ... -p 40`

**Additional context**
Our slurm cluster uses Ubuntu-version-based partition names (`18`, `20`, `26`), which is how we encountered this.

I appreciate that snakemake services many different backends, so simply fixing `partition` is no good. But perhaps we could consider:
- a) Preserving type information through the serialization round-trip, so only truly numeric values are summed.
- b) Providing a mechanism to declare certain resource keys as non-summable (e.g. via a CLI flag like `--non-summable-resources` or via executor plugin metadata).

I would be happy to explore PRs for either approach if deemed suitable.

**Workaround:** Use non-numeric partition names (e.g. `u20` instead of `20`).

johanneskoester:
fixes #3992

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI-style configuration output now preserves integer/float-like string values by quoting them so they aren’t misinterpreted; non-string values retain prior formatting.
* **Chores**
  * Profile parsing now emits brief debug output during execution.
* **Tests**
  * Added unit tests covering profile parsing and CLI-style serialization to ensure values are preserved and evaluate correctly.

### DIFF
--- a/src/snakemake/profiles.py
+++ b/src/snakemake/profiles.py
@@ -1,12 +1,9 @@
 from collections import OrderedDict
 import os
-import re
 from pathlib import Path
 from configargparse import YAMLConfigFileParser, ConfigFileParserException
 
 from snakemake_interface_common.exceptions import WorkflowError
-
-int_re = re.compile("^-?\d+$")
 
 
 class ProfileConfigFileParser(YAMLConfigFileParser):
@@ -29,12 +26,18 @@ class ProfileConfigFileParser(YAMLConfigFileParser):
             )
 
         def format_val(val):
-            if isinstance(val, str) and int_re.match(val):
-                # when the value is an explicitly string, we have to quote it to prevent
-                # an evaluation as integer in the next pass
-                return repr(val)
-            else:
-                return val
+            def repr_if_numeric(val, numtype):
+                try:
+                    numtype(val)
+                    return repr(val)
+                except ValueError:
+                    return None
+
+            if isinstance(val, str):
+                # when the value is an explicit string enclosing int or float, we have
+                # to quote it to prevent an evaluation as integer in the next pass
+                return repr_if_numeric(val, int) or repr_if_numeric(val, float) or val
+            return val
 
         def format_one_level_dict(d):
             return [f"{key}={format_val(val)}" for key, val in d.items()]


andynu:
| Scenario                  | sbatch -p | Result                        |
|---------------------------|-----------|-------------------------------|
| Serial + Unquoted (20)    | -p 20     | PASS (trivially, 1 job/layer) |
| Parallel + Unquoted (20)  | -p 40     | FAIL - bug reproduced         |
| Serial + Quoted ("20")    | -p 20     | PASS                          |
| Parallel + Quoted ("20")  | -p 20     | PASS - PR fix works           |

This does appear to fix the issue when we explicitly quote the partition name.

johanneskoester:
Thanks for your help @andynu!

PR #4124, opened 2026-03-20
fix: do not rerun when checkpoint job missing but downstream file exists

### LINKED ISSUE #3879: Deletion of checkpoint outputs is not properly detected

Hocnonsense:
will fix #3879. See also the discussion in #4092

### QC

* [x] This is not necessary to update the documentation (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake).

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Checkpoint handling now skips instances whose outputs are missing, preventing failures when checkpoint outputs are deleted and allowing workflows to continue.

* **Tests**
  * Added an integration test that verifies reruns succeed when checkpoint outputs are removed after creation.
  * Removed an obsolete checkpoint-related test.

[Review Change Stack](https://app.coderabbit.ai/change-stack/snakemake/snakemake/pull/4124?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


themavik (commented):
Gating checkpoint enqueue on is_output_present fixes the stale-node case. nit: confirm dropping test_checkpoint_missout does not leave 3009/3879 uncovered.

src/snakemake/dag.py
```diff
@@ -2270,7 +2270,9 @@ async def is_output_present(job):
                 await self.update_needrun()
                 for job, posterior_checkpoint_deps in candidate_job_queue.items():
                     for checkpoint in posterior_checkpoint_deps:
-                        if not self.needrun(checkpoint):
+                        if not self.needrun(checkpoint) and await is_output_present(
```
johanneskoester, line 2275:
```suggestion
                        # the second clause ensures that we only process checkpoints
                        # where the output is present (see test_checkpoint_missing_output)
                        if not self.needrun(checkpoint) and await is_output_present(
```

johanneskoester (approved):

PR #4125, opened 2026-03-20
feat: allow Snakemake to discover all required checkpoints in a single evaluation

### LINKED ISSUE #439: Checkpoints that can be done in parallel are evaluated and executed sequentially
**Snakemake version**
Reproducible in 5.19.2

**Describe the bug**
When a rule depends on multiple checkpoints, i. e. multiple `checkpoints.x.get()` calls in the input function, the checkpoints are evaluated sequentially. Initially snakemake considers that the rule only depends on the first checkpoint, so the second checkpoint isn't queued up until the DAG reevaluation after the completion of the first checkpoint.

**Minimal example**
```
from itertools import product

src_lang = config['src_lang']
trg_lang = config['trg_lang']

rule all:                        
    input: "data/corpus.txt"

checkpoint shard:                    
    output: "data/batches.{lang}"    
    shell: '''                     
        for i in $(seq 0 $(( $RANDOM % 5 + 1))); do
            for j in $(seq 0 $(( $RANDOM % 2 + 1))); do
                mkdir -p data/{wildcards.lang}/$i/$j
                echo 'hello {wildcards.lang}' > data/{wildcards.lang}/$i/$j/text
            done
        done
        ls -d data/{wildcards.lang}/*/* > {output}  
        '''
                                                                
rule combine:                                              
    input:                                                
        l1='data/{src_lang}/{shard}/{src_batch}/text',
        l2='data/{trg_lang}/{shard}/{trg_batch}/text'
    output: 'data/{src_lang}_{trg_lang}/{shard}.{src_batch}_{trg_batch}.combined'
    shell: ''' paste {input.l1} {input.l2} > {output} '''
                                                                                                
def get_batches_pairs(src_lang, trg_lang):                                 
    src_batches = []      
    trg_batches = []                                        
    with checkpoints.shard.get(lang=src_lang).output[0].open() as src_f, \
            checkpoints.shard.get(lang=trg_lang).output[0].open() as trg_f:
        for line in src_f:                      
            src_batches.append(line.strip().split('/')[-2:])                                                                                
        for line in trg_f:
           trg_batches.append(line.strip().split('/')[-2:])
    iterator = product(src_batches, trg_batches)
    return [(src_shard, (src_batch, trg_batch)) for ((src_shard, src_batch), (trg_shard, trg_batch)) in iterator if src_shard == trg_shard]

rule corpus:           
    input: lambda wildcards: [f'data/{src_lang}_{trg_lang}/{shard}.{src_batch}_{trg_batch}.combined' for (shard, (src_batch, trg_batch)) in get_batches_pairs(src_lang, trg_lang)]
    output: 'data/corpus.txt'
    shell: ''' cat {input} > {output} '''
```
When running this example, fist `shard(lang=src_lang)` will be executed, and when it finishes snakemake will reevaluate the DAG and queue up `shard(lang=trg_lang)`

**Additional context**
Related to #16 
A possible workaround would be to add all of the checkpoints outputs to the final target, 
```
rule all:
    input: "data/corpus.txt", f'data/batches.{src_lang}', f'data/batches.{trg_lang}'
```
but that could potentially get messy if there are a lot of outputs.

### LINKED ISSUE #3475: Checkpoint "get" still only resolvable once per aggregate function

Hocnonsense:
will close #439, close #3475, and also see #3562

### Problem
Once a checkpoint output is missing, Snakemake immediately re-evaluates the DAG.
Multiple checkpoints in an input function raise exceptions one by one, leading to unnecessary sequential scheduling.

### Solution
To allow Snakemake to discover all required checkpoints in a single evaluation, here introduce the ``with checkpoints:`` context manager, which defers exception raising until all `.get()` calls in the block have been evaluated.
Inside the block, incomplete checkpoints return a placeholder `CheckpointJob` instead of raising immediately.
All missing targets are collected and raised together at the end of the block, so Snakemake can discover and schedule all missing checkpoints and their upstream dependencies, in a single DAG re-evaluation.

```python
    with checkpoints:
        # multiple checkpoints of the same rule can be accessed in the same block
        series = {checkpoints.step_a.get(**wildcards, a=a).output[0] for a in range(3)}
        # different checkpoint rules can also be accessed together
        b = checkpoints.step_b.get(**wildcards).output[0]
    # both step_a and step_b are guaranteed complete beyond this point
    ...
```

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **New Features**
  * Added a with checkpoints: context manager that batches multiple checkpoint accesses so missing checkpoints are discovered together.

* **Improvements**
  * Defers and aggregates checkpoint-missing errors during the with block, enabling parallel scheduling instead of sequential retries.
  * Consolidates reported missing checkpoint targets for clearer diagnostics.

* **Documentation**
  * Documented context-manager usage and noted file I/O on checkpoint outputs must occur after the with block completes.

* **Tests**
  * Updated tests to precompute checkpoint outputs inside the context to avoid repeated evaluations.


src/snakemake/exceptions.py
```diff
@@ -577,13 +589,22 @@ def __init__(self, rule, targetfile):
...
+        self._targetfile = checkpoint_target(targetfile)
+        self.extra = extra
+
+    @property
+    def targetfile(self):
+        targetfile = [self._targetfile]
+        if self.extra:
+            for e in self.extra:
+                targetfile.extend(e.targetfile)
+        return targetfile
```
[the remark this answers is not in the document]
Hocnonsense: modified `handle_incomplete_checkpoint` of `Rule.expand_params`

themavik (commented):
IncompleteCheckpointException.targetfile is now a property that aggregates multiple paths; any extension or user code that assumed a single string path (or compared with == to one file) could break until updated to handle a list.

src/snakemake/checkpoints.py
```diff
@@ -35,17 +37,40 @@ class CheckpointsProxy(Checkpoints):
...
-        checkpoint = Checkpoint(rule, self)
-        if fallback_name:
-            setattr(self, fallback_name, checkpoint)
-        setattr(self, rule.name, checkpoint)
+        self._rules[fallback_name or rule.name] = Checkpoint(rule, self._parent)
+
+    def __enter__(self):
+        if getattr(self._local, "cache", None) is None:
+            self._local.cache = []
+        # do not lost message with nested checkpoint
```
johanneskoester:
```suggestion
        # do not lose message with nested checkpoint
```

src/snakemake/checkpoints.py
```diff
@@ -35,17 +37,40 @@ class CheckpointsProxy(Checkpoints):
...
+
+    def __exit__(self, exc_type, exc, tb):
+        cache: "list[IncompleteCheckpointException]" = self._local.cache.pop()
+        if exc_type is None and cache:
+            e, *_e = cache
+            for c in self._local.cache:
+                _e.extend(c)
+            e.extra = _e
+            raise e
+        return False
```
johanneskoester, line 102: This is a really smart way of solving this longstanding issue. However, to me the code does not yet make it immediately obvious how this now actually happens. How does this local cache actually capture the exceptions instead of passing them to the outer context? Can you add some comments in the code that explain how this actually happens? Maybe this is some context manager behavior that I am not aware of?

johanneskoester (changes requested):

docs/snakefiles/rules.rst
```diff
@@ -3116,6 +3116,34 @@ Here a new directory will be created for each sample by the checkpoint.
...
+
+.. code-block:: python
+
+    def aggregate_input(wildcards):
+        with checkpoints:
+            # multiple checkpoints of the same rule can be accessed in the same block
+            series = {checkpoints.step_a.get(**wildcards, a=a).output[0] for a in range(3)}
+            # different checkpoint rules can also be accessed together
+            b = checkpoints.step_b.get(**wildcards).output[0]
+        # both step_a and step_b are guaranteed complete beyond this point
```
johanneskoester: To make this semantically easier to read, I suggest to put the context manager under an attribute/property `waitfor`, such that one instead has to write:

```suggestion
        with checkpoints.waitfor:
            # multiple checkpoints of the same rule can be accessed in the same block
            series = {checkpoints.step_a.get(**wildcards, a=a).output[0] for a in range(3)}
            # different checkpoint rules can also be accessed together
            b = checkpoints.step_b.get(**wildcards).output[0]
        # both step_a and step_b are guaranteed complete beyond this point
```

johanneskoester (changes requested):

docs/snakefiles/rules.rst
Hocnonsense: Thanks for the suggestion! waitfor as a property name is much clearer about intent.

One follow-up thought: `with checkpoints.waitfor:` still modifies the shared `CheckpointsProxy` state to activate collection mode, which may be unsafe if two coroutines enter the block concurrently (they share the same `threading.local()` stack since asyncio runs on a single thread).

What if we go one step further with `with checkpoints.waitfor as waitfor:`? The waitfor object would be a children `checkpoints` collector, checkpoint accesses go through `waitfor.foo.get(...)` rather than `checkpoints.foo.get(...)`, so no shared state is mutated at all. Each with block gets its own isolated namespace, which sidesteps the concurrency concern entirely.

Does that seem worth the slightly longer access path, or is concurrent DAG input evaluation guaranteed non-interleaved in practice?

Hocnonsense: Oh I just recalled the reason of using `with checkpoints:` weeks ago: since `CheckpointsProxy` resolves member names to checkpoint rule names via `__getattr__`, and a rule named waitfor may conflict, I'd like to keep `checkpoints.waitfor` as an optional alias at most. `with checkpoints as waitfor:` avoids this since it never go through `__getattr__`.

src/snakemake/checkpoints.py
Hocnonsense: Sure, now the docstring is added to describe the behavior

docs/snakefiles/rules.rst
Hocnonsense: @johanneskoester After further consideration, I see three possible syntaxes for the checkpoint collection block, each with different trade-offs:

1. **`with checkpoints.waitfor:`**: init a `IncompleteCheckpointException` collector and redirects all `checkpoints.foo.get()` calls inside the block to the collector. However, this affects *any* code reachable from the block, including helper functions that call `checkpoints.foo` without knowing they are inside a collection context. This could silently suppress `IncompleteCheckpointException` in unexpected places, so I'd consider this unsafe.

2. **`with checkpoints as waitfor:`**: declare a named collector with its own `__getattr__` function. Only accesses through `waitfor.foo` are collected; `checkpoints.foo` inside the block still raises immediately.
    - To guarantee coroutines safety, the collector is maintained in a module-level `ContextVar` stack in `CheckpointsProxy.__enter__`, and it will be popped and handled in `CheckpointsProxy.__exit__`. The `ContextVar` is purely an internal bridge invisible to users, though it does add some implementation complexity for maintainers.

3. **`with checkpoints.waitfor as waitfor:`** only: the collector is a self-contained context manager (`__enter__` and `__exit__` move from `CheckpointsProxy` to the collector), so `ContextVar` is no longer needed. The only concern is that `waitfor` occupies a name in `CheckpointsProxy`'s attribute namespace, which is otherwise reserved for checkpoint rule names. 
    - Following the `NamedTuple` convention (`_make`, `_asdict`, `_replace`), I'd suggest `_waitfor` to signal it is an internal name that won't conflict with user-defined rule names, though `with checkpoints._waitfor as waitfor:` feels somewhat redundant.

Currently I'm using option 2 because personally I think it is the cleanest user-facing API, with `checkpoints.waitfor` as an alias for entering the collection context.
I'd be happy to go with whichever direction you think fits the project best.

PR #4137, opened 2026-03-28
fix: close plugin handlers after draining QueueListener in LoggerManager.stop()

### LINKED ISSUE #4136: bug: _cleanup() closes logger handlers before draining QueueListener, silently losing events
### Description

In `snakemake/api.py`, the `_cleanup()` method calls `cleanup_logfile()` before `stop()`:

```python
def _cleanup(self):
    if not self.output_settings.keep_logger:
        logger_manager.cleanup_logfile()  # closes handlers first
        logger_manager.stop()             # drains queue second
```

This ordering creates a race condition. `cleanup_logfile()` iterates `logfile_handlers` and calls `handler.close()` on all handlers where `writes_to_file = True`, which includes logger plugin handlers. At this point, the `QueueListener` background thread (set up in `logging.py`) is still running and may still be processing queued events.

The `QueueListener._monitor()` loop only catches `queue.Empty`, so if a handler's `emit()` raises after the handler has been closed (e.g., writing to a closed file), the background thread crashes and any remaining queued events are silently lost.

### Affected Code Paths

- `snakemake/api.py:_cleanup()` (~lines 170-174)
- `snakemake/logging.py:LoggerManager.cleanup_logfile()` (~lines 693-698) — closes plugin handlers
- `snakemake/logging.py:LoggerManager.stop()` (~lines 722-724) — drains QueueListener

### How to Reproduce

This is a race condition, so it manifests as flaky behavior. We discovered it while developing a logger plugin ([snakemake-logger-plugin-snakesee](https://github.com/fulcrumgenomics/snakesee)) that writes events to a JSONL file. Integration tests intermittently fail on Snakemake 9.14.0 because events are lost — the plugin's file writer is closed by `cleanup_logfile()` while the QueueListener thread still has events to process. The race window is small, making this a flaky test failure.

Any logger plugin that writes to a file (i.e., sets `writes_to_file = True`) is affected.

### Expected Behavior

All queued log events should be fully processed and delivered to handlers before those handlers are closed.

### Suggested Fix

Reverse the call order — drain the queue first, then close handlers:

```python
def _cleanup(self):
    if not self.output_settings.keep_logger:
        logger_manager.stop()             # drain queue first
        logger_manager.cleanup_logfile()   # then close handlers
```

### Version Information

- Snakemake: 9.14.0
- Python: 3.12
- OS: macOS / Linux

nh13:
## Summary
- Plugin handlers (managed by QueueListener) were never explicitly closed during `LoggerManager.stop()`, leaking file handles and potentially losing buffered data in logger plugins that write to files
- Adds explicit close loop for plugin handlers after the queue is drained but before the logger's own handlers are cleaned up
- Adds regression test that verifies plugin handlers are closed after `stop()`

Closes #4136

## Test plan
- [x] New unit test `test_stop_closes_plugin_handlers` fails without the fix (RED) and passes with it (GREEN)
- [ ] Existing logging tests pass (`test_logfile`, `test_issue4063`, `test_log_events_dryrun`, `test_log_events`, `test_rule_failure`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin-managed log handlers are now explicitly flushed and closed during logger shutdown to ensure no pending log records remain.

* **Tests**
  * Added a regression test verifying plugin handler flush/close on shutdown and ensuring background logging threads are cleaned up to prevent leakage.


cademirch (approved):
Thanks for this! Looks good to me. Could you just run `pixi run format`?

Also, we've recently updated the PR template, please fill it in for future PRs.

jlumpe:
Beat me to it. I think `flush()` should also be called on all loggers as well? To my knowledge `close()` doesn't necessarily do this. I think the builtin `logging.shutdown()` function flushes all handlers in a `try/except` block before closing them.

PR #4143, opened 2026-04-03
fix: Flush global log handlers in LoggerManager.stop()

jlumpe:
- Call the `flush()` and `close()` methods of all plugin handlers in `LoggerManager.stop()` so they process any remaining records and properly release any resources. This does not otherwise happen because they are attached to the queue listener instead of directly to the global `Logger` instance. Do it after the queue listener is flushed so that it sends any remaining records to them.
- Also adds calls to `flush()` before shutting down handlers directly attached to the logger (which would be the default file and stream handlers).
- Also wraps all calls to `flush()` in `try/except` and ignores errors. This is how the standard library `logging.shudown()` does it, not 100% sure if it's completely necessary but doesn't seem like it can hurt much. Possibly a warning should be generated from any errors to assist in plugin debugging.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logging shutdown so all attached log handlers are flushed and then removed/closed during shutdown, leading to more reliable cleanup of logging output and resources.

[Review Change Stack](https://app.coderabbit.ai/change-stack/snakemake/snakemake/pull/4143?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/snakemake/logging.py
+++ b/src/snakemake/logging.py
@@ -738,13 +738,27 @@ class LoggerManager:
 
     def stop(self) -> None:
         """Shut down logging, removing and closing all log handlers."""
+
         # Stop the queue listener (if it exists) - this finishes processing the remaining records
         # and waits for the thread to exit.
         if self.queue_listener is not None and self.queue_listener._thread is not None:
             self.queue_listener.stop()
 
-        # Remove and close all handlers - this should mostly clean up the global logger instance.
+        # Flush and close plugin handlers (not attached directly to global logger instance)
+        for handler in self.plugin_handlers:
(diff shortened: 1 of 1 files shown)


jlumpe:
@cademirch sorry got another one for you here

cademirch:
Can't take a closer look right now but is this the same as #4137?

jlumpe:
Ah yeah definite duplicate

jlumpe:
Mostly duplicate of #4137, will file separate PR with remaining changes.

PR #4148, opened 2026-04-11
fix: cannot drop ruleinfo of `a` and `use rule a as b` in module if `use b from module`

### LINKED ISSUE #4147: bug: WorkflowError with module and github()
**Snakemake version**
`9.19.0`

**Describe the bug**
In a workflow remotely hosted, I use `input: rules.foo.output.bar` to not rewrite the file path each time. In another workflow, I wanted to reuse one rule from that remote workflow:
```python
module other_workflow:
    snakefile: github("...", path="workflow/Snakefile", branch="main")

use rule bar from other_workflow as bar
```
I tried a dry-run `snakemake -np` and received a `WorflowError`.

**Logs**
This is from the **Minimal example**
```
$ pixi run snakemake -np
Fetching latest changes of github.com/GaspardR/snakemake-github-module-keyerror to /Users/gaspard/Library/Caches/snakemake/snakemake/source-cache/snakemake-git-cache/github.com/GaspardR/snakemake-github-module-keyerror
WorkflowError in file "https://raw.githubusercontent.com/GaspardR/snakemake-github-module-keyerror/main/workflow/Snakefile", line 27:
Rule B is not defined in this workflow. Available rules: all, A
  File "https://raw.githubusercontent.com/GaspardR/snakemake-github-module-keyerror/main/workflow/Snakefile", line 27, in <module>
```

**Minimal example**
I created a public repo for testing: https://github.com/GaspardR/snakemake-github-module-keyerror

Here's that repo's Snakefile:
```python
rule all:
    default_target: True
    input:
        "results/c.txt",


rule A:
    output:
        a="results/a.txt",
    shell:
        "touch {output[0]}"


use rule A as B with:
    output:
        b="results/b.txt"


rule C:
    input:
        a=rules.A.output.a,
        b=rules.B.output.b,
    output:
        c="results/c.txt",
    shell:
        "cat {input.a} {input.b} > {output.c}"
```

And the local Snakefile to reproduce the bug. Note the commented line, I created a branch for testing that replaces `rules.A.output.a` and `rules.B.output.b` with the file paths. The branch `only-paths` works, whereas `main` does not.
```python
module test_github:
    snakefile:
        github("GaspardR/snakemake-github-module-keyerror", path="workflow/Snakefile", branch="main")
        # github("GaspardR/snakemake-github-module-keyerror", path="workflow/Snakefile", branch="only-paths")

use rule A from test_github as A
```

**Additional context**
This is probably related to #3753

### LINKED ISSUE #4201: Question about modular design
**Snakemake version**
9.21.0

**Describe the bug**

Hello developers,

I have a question regarding the modular design.

When importing A.smk into B.smk, I try to modify the rules defined in A.smk. If I only modify the input of either t1 or t2, B.smk runs correctly. However, when I modify both t1 and t2 at the same time, I get the following error:

UnknownRuleException: There is no rule named t2

Does this mean that, by design, this module only supports modifying the input of one rule at a time?

**Logs**

```
host: MacOS
Building DAG of jobs...
UnknownRuleException:
There is no rule named t2.
```
**Minimal example**

```
echo "this is a" > a.txt
echo "this is b" > b.txt
echo "this is c" > c.txt
```
A.smk
```
rule all:
    input:
        "res.txt",


rule t1:
    input:
        file="../a.txt",
    output:
        res="tmp.txt",
    shell:
        "cat {input.file} > {output.res}"


rule t2:
    input:
        file=rules.t1.output.res,
        file2=[],
    output:
        res="res.txt",
    shell:
        "cat {input.file} {input.file2} > {output.res}"
```
B.smk
```
module A:
    snakefile:
        "A.smk"
    prefix:
        "new_results"


use rule * from A as A_*


use rule t1 from A as A_t1 with:
    input:
        file="../b.txt",


use rule t2 from A as A_t2 with:
    input:
        file=rules.A_t1.output.res,
        file2="../c.txt",


rule all:
    input:
        "res.txt",
```
Works as expected:
`snakemake -s A.smk --cores 1 -d results`

Error occurs:
`snakemake -s B.smk --cores 1 -d results`

However, if I comment out either `use rule t1 from A as A_t1 with` or `use rule t2 from A as A_t2 with`, it also works as expected.
**Additional context**

Hocnonsense:
will fix #4147 and fix #4201 and close #4110

## Root Cause

When a snakefile is used as a module and only selected rules are imported (i.e., not `use rule * from`), the module snakefile is re-executed on every `use rule <name> from <module>` statement.

<details>
<summary>
This is because `ruleinfo` objects are discarded immediately after each execution if they are not among the requested rules.
</summary>

https://github.com/snakemake/snakemake/blob/0fc5fd14bf8707812abac14ccfeac0eb8093005d/src/snakemake/workflow.py#L1846-L1891

Rules not in the "whitelist" will not be recognized

https://github.com/snakemake/snakemake/blob/0fc5fd14bf8707812abac14ccfeac0eb8093005d/src/snakemake/modules.py#L269-L279

related: `module` statement only stores initial settings in `ModuleInfo`

https://github.com/snakemake/snakemake/blob/0fc5fd14bf8707812abac14ccfeac0eb8093005d/src/snakemake/workflow.py#L2360-L2387

and Module snakefile is only executed on `use rule ... from`

https://github.com/snakemake/snakemake/blob/0fc5fd14bf8707812abac14ccfeac0eb8093005d/src/snakemake/workflow.py#L2389-L2401

</details>

This design has two consequences:

1. **Sensitive repeat execution:** Every `use rule <name> from <module>` triggers a full re-execution of the module snakefile, since unused `ruleinfo` objects are not cached.
    While this short-circuit avoids permanently storing unused `ruleinfo` objects, it trades memory efficiency for repeated snakefile reloading.
2. **No value override support:** It is impossible to modify values within a module (see PR #4110 and [Discord discussion](https://discord.com/channels/753690260830945390/1483123196277428296)).

Issue #4147 surfaces a failure case that breaks the safety assumption underlying the short-circuit design: namely, that each rule is independent and can be safely discarded if not explicitly requested.
When an outer module requests rule `b`, which internally depends on rule `a` that was *not* explicitly requested, rule `a`'s `ruleinfo` has already been discarded.
To recover under the current lazy-reload model, the only option would be to re-execute the module snakefile once more when `b` is encountered, this time with `a` added to the whitelist — an implicit and fragile workaround.

## Fix Approach

Since each `use rule <name> from` triggered a separate re-execution, a fresh globals dict was created each time, and modifications from the outer Snakefile were not preserved in already-registered rule functions.
This PR caches module execution so specific-rule imports share one globals dict, so computing and memory costs decrease, and consistency is ensured.

### Lazy first-load via `_cached_namespace`

- If the first `use rule` is `* from`, the module executes at that point (since `name_modifier` is not known at the `module ...:` declaration), preserving correctness for inter-rule references and checkpoints.
  All `ruleinfo`s are held in `module.rule_proxies` (accessible as `ModuleInfo._cached_namespace.rules`), so subsequent specific-rule imports of the same module skip execution and read from the cache.
- Otherwise, `use rule <name> from` requires a pre-execution that does not affect `workflow.rules`.
  The first access to `_cached_namespace` triggers a "dry" run that populates `module.rule_proxies` without registering any rules in the main workflow.
  From this point on, the cached `ruleinfo` is processed identically to a local `use rule ... with:` statement via `WorkflowModifier.for_userule`.
- A warning is emitted if `use rule * from` is declared after any prior `use` of the same module (i.e. the module was already loaded), because inter-rule references must be rebuilt from a fresh execution, which invalidates all previously cached ruleinfos.

### Reuse `ruleinfo` from cached `RuleProxy`

`Workflow.rule()` always registers all `Rule` objects in `rule_proxies` unconditionally, so the cached `ruleinfo` reflects the module execution state.
A subsequent `copy.copy()` followed by `apply_modifier()` correctly applies any `with:` overrides without double-modifying paths.

`for_userule` is now shared between local and cross-module `use rule`.
The caller pre-computes `allow_overwrite`: local `use rule A as B with:` always passes `False`.
Cross-module `use rule <name> from` calls `module.check_overwrite(rule_, rulename_modifier)`, which returns `True` if the rule was previously declared via `use rule * from`, allowing the import to overwrite it.

### Use `ModuleType.__dict__` directly as `WorkflowModifier.globals`

Previously, a plain `dict` was copied into a `ModuleType` via `update()`, so direct variable reassignment in the outer workflow (see #4110) was not reflected in rule functions.
By creating the `ModuleType` first and passing its `__dict__` directly into `WorkflowModifier`, the module namespace and execution globals are always the same object.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **New Features**
  * Enhanced module rule importing with flexible aliasing, selective imports, renaming, and output overrides.
  * Preserved rule metadata when importing and reusing rules across modules.

* **Bug Fixes**
  * Improved module composition and rule inheritance for more predictable reuse and overwrite handling.

* **Tests**
  * Added end-to-end coverage for module reuse, rule inheritance, aliasing, and cross-module wiring.


Hocnonsense:
> Outside diff comments:
> In `@src/snakemake/modules.py`:
> - Line 137: Update the namespace initialization in the module modifier to
> populate _cached_namespace from self.parent_modifier.globals instead of
> self.workflow.vanilla_globals, ensuring module rules see bindings declared by
> the outer workflow before use rule.

@coderabbitai NO -- it is intend, to isolate 'module' environment with global namespace.

PR #4150, opened 2026-04-11
fix: suppress duplicate SpawnedJobError tracebacks in local executor

coroa:
## Problem

When a spawned job (e.g. a `run:` block) fails, the parent process printed the `SpawnedJobError` / `concurrent.futures` / `asyncio` call chain — twice. This is pure infrastructure noise with no information useful to the user:

```
During handling of the above exception, another exception occurred:

  File ".../concurrent/futures/thread.py", line 86, in run
    result = ctx.run(self.task)
  File ".../concurrent/futures/thread.py", line 73, in run
    return fn(*args, **kwargs)
  File ".../local.py", line 236, in cached_or_run
    self.workflow.async_run(self.acached_or_run(job, run_func, *args))
  File ".../workflow.py", line 268, in async_run
    return runner.run(coro)
  File ".../asyncio/runners.py", line 127, in run
    return self._loop.run_until_complete(task)
  File ".../asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
  File ".../local.py", line 253, in acached_or_run
    run_func(*args)
  File ".../local.py", line 233, in spawn_job
    raise SpawnedJobError()
snakemake.exceptions.SpawnedJobError
```

## Root cause

snakemake/snakemake#3309 changed `_callback` in `local.py` from:

```python
ex = future.exception()
if ex is not None:
    raise ex  # flows into except chain, SpawnedJobError caught silently
```

to:

```python
ex = future.exception()
if ex is not None:
    print_exception(ex, ...)  # unconditionally prints SpawnedJobError too
    self.report_job_error(job_info)
```

This made the `except SpawnedJobError` handler (which exists precisely to suppress printing) dead code. 

## Fix

- Restore the original `raise ex` pattern so exceptions flow through the `except` chain, allowing `except SpawnedJobError` to suppress the noisy chain as intended.
- In `except SpawnedJobError`, log the wrapped `CalledProcessError` via `log_verbose_traceback` (i.e. at debug level, visible with `--verbose`) so the failed command is still available for diagnostics.
- Remove the `verbose or not is_group` gate from `except Exception` so pipe/group job errors always print — preserving the fix from snakemake/snakemake#3309 without requiring `--verbose`.

## After

Normal output on job failure now only shows what the subprocess printed:

```
RuleException:
NameError in file "Snakefile", line 13:
name 'a' is not defined
  File "Snakefile", line 13, in __rule_1
```

With `--verbose`, the `CalledProcessError` (which command failed and with what exit code) is additionally shown once:

```
  File ".../local.py", line 231, in spawn_job
    subprocess.check_call(cmd, shell=True)
subprocess.CalledProcessError: Command '...' returned non-zero exit status 1.
```

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable and consistent exception reporting for failed jobs, with improved logging of nested/inner exceptions and unconditional display of exception tracebacks for clearer diagnostics.

* **Tests**
  * Updated logging-related test expectations to reflect the adjusted error reporting behavior.

### DIFF
--- a/src/snakemake/executors/local.py
+++ b/src/snakemake/executors/local.py
@@ -264,21 +264,19 @@ class Executor(RealExecutor):
         try:
             ex = future.exception()
             if ex is not None:
-                print_exception(ex, self.workflow.linemaps)
-                self.report_job_error(job_info)
-            else:
-                self.report_job_success(job_info)
+                raise ex
+            self.report_job_success(job_info)
         except _ProcessPoolExceptions:
             self.handle_job_error(job_info.job)
             # no error callback, just silently ignore the interrupt as the main scheduler is also killed
-        except SpawnedJobError:
+        except SpawnedJobError as ex:
             # don't print error message, this is done by the spawned subprocess
+            # but log the inner exception at debug level for diagnostics
+            if ex.__context__ is not None:
+                log_verbose_traceback(ex.__context__)
             self.report_job_error(job_info)
         except Exception as ex:
-            if self.workflow.output_settings.verbose or (
-                not job_info.job.is_group() and not job_info.job.is_shell
-            ):
-                print_exception(ex, self.workflow.linemaps)
+            print_exception(ex, self.workflow.linemaps)
             self.report_job_error(job_info)
 
     @property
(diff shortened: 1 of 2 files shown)


johanneskoester (approved):

johanneskoester:
Thanks a lot!

PR #4151, opened 2026-04-11
feat: enable caching for named multi-output rules

coroa:
Between-workflow caching was previously restricted to rules with a single output file or
with `multiext()` outputs, using cache filenames like:
```
{cache_dir}/{hash}.{ext}          # single output or multiext.
```

This PR extends caching to also support rules where all output files are explicitly named,
using the output key to disambiguate cache entries:
```
{cache_dir}/{hash}_{key}.{ext}
```

For example, given:
```python
rule a:
    output:
        first="1.txt",
        second="2.txt",
    cache: True
    shell: "touch {output}"
```

the cache directory is populated as:
```
cache/
├── d5281ea3...b1_first.txt
└── d5281ea3...b1_second.txt
```

### Design

Three cases are now handled in `get_outputfiles()`:

1. **Single output** — unchanged, uses `.{ext}` suffix
2. **`multiext()` output** — unchanged, uses the suffix from the multiext prefix
3. **Named multi-output** — new, uses `_{key}.{ext}`

Unlike the previous attempt (#3948), this approach is **backward compatible**: existing
cache entries for single-output and multiext rules are unaffected since their filename
format does not change. Only the new named multi-output case uses a different scheme.

Rules with multiple unnamed outputs still fail validation with a clear error message.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Improvements**
  * Rules with multiple outputs are now eligible for between-workflow caching if all outputs are explicitly named, in addition to multiext-style outputs.

* **Documentation**
  * Updated caching eligibility docs to reflect the expanded multi-output cache support.

* **Tests**
  * Updated tests to exercise successful multi-output caching with named outputs.

### DIFF
--- a/docs/executing/caching.rst
+++ b/docs/executing/caching.rst
@@ -48,8 +48,7 @@ For workflows defining cache rules like this, it is enough to invoke Snakemake w
 
 without explicit rulenames listed.
 
-Note that only rules with just a single output file (or directory) or with :ref:`multiext output files <snakefiles-multiext>` are eligible for caching.
-The reason is that for other rules it would be impossible to unambiguously assign the output files to cache entries while being agnostic of the actual file names.
+Note that rules with multiple output files are only eligible for caching if they use :ref:`multiext output files <snakefiles-multiext>` or if all output files are named.
 Also note that the rules need to retrieve all their parameters via the ``params`` directive (except input files).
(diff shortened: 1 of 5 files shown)


johanneskoester (approved):
Fantastic!

PR #4152, opened 2026-04-12
perf: adjust default sqlite PRAGMAs, auto detect network fstype

tedil:
This PR adjusts the default PRAGMAs to better suit networked file systems (and tries to auto-detect them, though we may have to extend the list or just always apply the same PRAGMAs, open to discussion).

### QC

### AI-assistance disclosure

I used AI assistance for:
* [x] Code generation (e.g., when writing an implementation or fixing a bug) (review, coderabbitai)

## Summary by CodeRabbit

* **New Features**
  * Automatic detection when the persistence database is on a network filesystem and application of different SQLite settings to improve reliability.

* **Documentation**
  * Clarified behavior and listed PRAGMA sets used for network FS (journal_mode=PERSIST, synchronous=OFF, temp_store=MEMORY, cache_size=-64000) versus local FS (journal_mode=TRUNCATE, synchronous=NORMAL); reordered caveats and recommendations.


src/snakemake/persistence/db.py
```diff
@@ -224,3 +242,48 @@ def _delete_locks(self) -> None:
...
+    path_obj = Path(path).resolve()
+
+    # track the best mount point match
+    longest_prefix_length = -1
+    fstype = ""
+
+    try:
+        for part in psutil.disk_partitions(all=True):
+            mount_path = Path(part.mountpoint).resolve()
+            if path_obj.is_relative_to(mount_path):
```
cmeesters: My cluster home is `PosixPath('/fshpc/meesters')`. This will raise `AttributeError: 'PosixPath' object has no attribute 'is_relative_to'`. Even though we are dealing with a network file system.

While this is a special case (a NetApp mount with a special name), it still might happen elsewhere. Perhaps we need to consider an AttributeError as being equivalent to `is_relative_to`???

I am sorry, I know this will result in ugly, obfuscated code.

src/snakemake/persistence/db.py
```diff
@@ -224,3 +242,48 @@ def _delete_locks(self) -> None:
...
+                mount_depth = len(mount_path.parts)
+                if mount_depth > longest_prefix_length:
+                    longest_prefix_length = mount_depth
```
cmeesters: for which purpose do we need to store `longest_prefix_length`? If `part.fstype` in `network_fs_types`, can't we return `true` immediately?

src/snakemake/persistence/db.py
```diff
@@ -224,3 +242,48 @@ def _delete_locks(self) -> None:
...
+    # track the best mount point match
+    longest_prefix_length = -1
+    fstype = ""
+
+    try:
+        for part in psutil.disk_partitions(all=True):
+            mount_path = Path(part.mountpoint).resolve()
+            if path_obj.is_relative_to(mount_path):
+                mount_depth = len(mount_path.parts)
+                if mount_depth > longest_prefix_length:
```
cmeesters: do we need `longest_prefix_length` at all?  (see comment about storing it again)

src/snakemake/persistence/db.py
tedil: Because mountpoints can be nested, e.g., `/users/meesters` pointing somewhere else than `/users/meesters/scratch` ; so we have to keep track of the best match.

src/snakemake/persistence/db.py
tedil: what do you mean? `PosixPath("~/projects").is_relative_to("~")` works for me (py3.12, perhaps that didn't exist in earlier versions?) would os.path.commonpath work for you?

cmeesters: No, I meant that I have tested interactively with that snippet and supplied `/fshpc/meesters` which is my home. And that raised the AttributeError. My Python in this env is 3.13.2.

Testing a workflow on this HPC system (with `--directory` to `/lustre/...` works just fine). Testing with toy data all locally within my HOME works fine, too. I concede, there is no issue.

src/snakemake/persistence/db.py
cmeesters: Yes, this part of our work is ugly. 

How about:

```Python
try:
        best_match = max(
            (
                part
                for part in psutil.disk_partitions(all=True)
                if path_obj.is_relative_to(Path(part.mountpoint).resolve())
            ),
            key=lambda part: len(Path(part.mountpoint).resolve().parts),
            default=None,
        )
        fstype = best_match.fstype if best_match else ""
```
I admit: This requires some explanation, but it should be functionally identical and does not work with intermediates. You should pick - in any case, I think the PR is good to merge.

tedil: yeah, that works for me, I'll update accordingly. and rename part to partition or mountpoint or so, because part and .parts is confusing.

cmeesters (approved):

src/snakemake/persistence/db.py
```diff
@@ -224,3 +242,47 @@ def _delete_locks(self) -> None:
...
+    if os.name != "posix":
+        return False
+
+    path_obj = Path(path).resolve()
+
+    try:
+        best_match = max(
+            (
+                mount
+                for mount in psutil.disk_partitions(all=True)
```
johanneskoester, line 317: Querying disk partitions with all=True can apparently be very slow (on our cluster it takes a minute). Thus, we should not do this check each time (maybe only once upon DB initialization)?

src/snakemake/persistence/db.py
```diff
@@ -224,3 +242,47 @@ def _delete_locks(self) -> None:
...
+            "fuse.sshfs",
+            "glusterfs",
+            "gpfs",
+            "lustre",
+            "nfs",
+            "nfs3",
+            "nfs4",
+            "orangefs",
+            "pvfs2",
+            "smbfs",
```
johanneskoester, line 341:
```suggestion
            "smbfs",
            "fuse.juicefs",
```

johanneskoester (changes requested):
Cool! Two things below.

src/snakemake/persistence/db.py
cmeesters: I do not _know_ what it does, yet the docs state:

> Return mounted disk partitions as a list. This is similar to the df command on UNIX. ...

When a `df` does not return (quickly), you are likely dealing with a file system issue. In your case, after the chats we had, I suspect a faulty NFS parameterization. I agree, having defensive approach is to be favoured. If without `all=True` no important bits are missed and the speed increases, we should go for it. (And I would be interested in the diff of with and without this flag ;-) .)

src/snakemake/persistence/db.py
tedil: I sorted them alphabetically, would be nice to keep it that way ;)

src/snakemake/persistence/db.py
johanneskoester: I checked it and network mounts only seem to show up with `all=True`. Yes, I agree that the setup I have access to is likely not optimal, but as you say, better avoid running into issues on such setups. Maybe they are not so rare. So, since disk_partitions() can thus be slow, we should be smarter about those checks.

tedil: alright, adjusted accordingly, should now only do the check once, write that into a record in the db and on subsequent launches check for that record in the db and use its value. I hope the additional (raw) connection doesn't add too much overhead

src/snakemake/persistence/db.py
```diff
@@ -224,3 +296,49 @@ def _delete_locks(self) -> None:
...
+def is_network_filesystem(path: Path | str) -> bool:
+    """
+    Detects if a given path resides on a network filesystem.
+    """
+    if os.name != "posix":
+        return False
+
+    path_obj = Path(path).resolve()
+
+    try:
```
johanneskoester, line 313:
```suggestion
    logger.info("Determining filesystem type for metadata persistence storage...")
    try:
```

johanneskoester: So that people know what takes so long.

johanneskoester (changes requested):

src/snakemake/persistence/db.py
tedil: do you want to name it metadata partition? because I have been using terms such as persistence backend / storage so far

johanneskoester: Called it "metadata persistence storage" now. I think persistence alone can be misleading. Metadata storage alone would be fine though. But if persistence occurs in the cli, let us reflect that here as well.

src/snakemake/persistence/db.py
```diff
@@ -1,21 +1,32 @@
...
     delete,
     event,
 )
+from sqlalchemy.engine.url import make_url, URL
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import DeclarativeBase, Session
 from sqlmodel import SQLModel, Field
+import sqlite3
 
 from snakemake.persistence import MetadataRecord, PersistenceBase
```
johanneskoester, line 20:
```suggestion
from snakemake.persistence import MetadataRecord, PersistenceBase
from snakemake.logging import logger
```

johanneskoester (approved):

PR #4155, opened 2026-04-16
Revert "chore: prepare for internal Rust modules by migrating to maturin as build backend"

johanneskoester:
Reverts snakemake/snakemake#3974

## Summary by CodeRabbit

* **Refactor**
  * Removed Rust extension module and its associated functionality

* **Chores**
  * Simplified Docker image by removing unnecessary build dependencies
  * Updated version control configurations


johanneskoester:
@fgvieira we can't merge this yet. It means that we cannot build snakemake as noarch anymore on bioconda. Thus, to maintain win support, we have to introduce cross compilation in bioconda. And this first has to be implemented, via adding rattler-build as an alternative backend to bioconda-utils.

PR #4161, opened 2026-04-20
feat: support remote snakefile via gh/gl shorthand

clelange:
I had the impression that #4077 was too big to be reviewed and that the preference was to initially only support the `gh`/`gl` shorthand. I've therefore extracted this part out so that it can be reviewed more easily and eventually be merged. I think that almost everything has been reviewed already, but please take another look. 

This PR adds support for using remote Snakefiles via the explicit hosted-source shorthand syntax:

- `gh:owner/repo@ref:path/to/Snakefile`
- `gl:group/project@ref:path/to/Snakefile`

The trailing path remains optional and defaults to `workflow/Snakefile`.

The explicit shorthand form should be unambiguous and avoids the review concerns around parsing full hosted URLs whose refs may contain `/`. Generic remote Snakefile execution via plain HTTP/HTTPS URLs still remains supported; this PR only avoids the provider-specific inference from full GitHub/GitLab URLs.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **New Features**
  * Support for remote Snakefiles via HTTP/HTTPS and hosted-source shorthands (gh:, gl:) with defaulting to workflow/Snakefile when path omitted.

* **Improvements**
  * CLI, API and GUI accept and preserve snakefile URIs/strings; include paths resolve relative to remote Snakefiles; cache keys and hosted-source handling more robust (URL-encoding/ref sensitivity). Remote -s fetches only directly referenced files.

* **Documentation**
  * Expanded docs and examples for shorthand syntax, default path behavior, remote resolution scope.

* **Tests**
  * Added tests and a local HTTP test-server helper for remote Snakefile scenarios.


clelange:
FYI @matthewfeickert @lukasheinrich

src/snakemake/sourcecache.py
```diff
@@ -673,7 +759,14 @@ def infer_source_file(path_or_uri, basedir: Optional[SourceFile] = None) -> Sour
         raise SourceFileError(
             "must be given as Python string or one of the predefined source file marker types (see docs)"
         )
-    if is_local_file(path_or_uri):
+    if hosting_provider_file := _infer_hosting_provider_file(path_or_uri):
+        return hosting_provider_file
+    # Unsupported or malformed URIs should fall through to GenericSourceFile below.
+    try:
+        is_local = is_local_file(path_or_uri)
+    except (NotImplementedError, ImportError, ValueError):
```
johanneskoester: Why would these exceptions occur in is_local_file? It worries me that we have to keep this in mind everytime we use is_local_file (this PR does it above as well). Ideally, is_local_file should not throw exceptions other than stuff caused by bugs, which in turn should not need to be handled.

src/snakemake/api.py
```diff
@@ -91,6 +93,16 @@ def resolve_snakefile(path: Optional[Path], allow_missing: bool = False):
             raise ApiError(
                 f"No Snakefile found, tried {', '.join(map(str, SNAKEFILE_CHOICES))}."
             )
+
+    if isinstance(path, str):
+        try:
+            is_local = is_local_file(path)
+        except (NotImplementedError, ImportError, ValueError):
```
johanneskoester: see my comment on the other try-except around is_local_file below.

src/snakemake/gui.py
```diff
@@ -71,7 +71,15 @@ def log_handler(msg):
 
     run_snakemake(list_resources=True, log_handler=[log_handler])
     app.extensions["resources"] = resources
-    app.extensions["snakefilepath"] = os.path.abspath(args.snakefile)
+    try:
+        is_local = is_local_file(args.snakefile)
+    except (NotImplementedError, ImportError, ValueError):
```
johanneskoester: same here

src/snakemake/sourcecache.py
```diff
@@ -512,18 +523,32 @@ def hosted_repo(self) -> HostedGitRepo:
 
             # Ensure that multiple threads don't concurrently create the same instance
             with self._lock:
-                hosted_repo = HostedGitRepo(
-                    self.repo, self.cache_path, self.auth, self.host
-                )
-                self._hosted_repos[self.repo] = hosted_repo
-            return hosted_repo
+                try:
+                    return self._hosted_repos[cache_key]
```
johanneskoester: Why the second try here to get the object from the cache dict? Didn't we do that before in line 515?

src/snakemake/sourcecache.py
```diff
@@ -661,6 +688,65 @@ def get_path_or_uri(self, secret_free: bool) -> str:
...
+    path = unquote(path)
+
+    if provider == "gh":
+        return GithubFile(repo=repo, path=path, branch=ref, host=host)
+    else:
+        return GitlabFile(repo=repo, path=path, branch=ref, host=host)
+
+
+def _infer_hosting_provider_file(path_or_uri: str) -> Optional[HostingProviderFile]:
+    return _infer_hosting_provider_file_shorthand(path_or_uri)
```
johanneskoester: Why this function if it just delegates?

johanneskoester (changes requested):
Hi Clemens! Thanks a lot, this is very nice. Please find my comments below.

src/snakemake/sourcecache.py
clelange: Agreed. I changed `is_local_file()` so it behaves as a direct predicate for normal path/URI inputs instead of depending on `smart_open.parse_uri()`. Remote or unsupported schemes such as `gh:`, `gl:`, and `s3://` now return `False` without callers needing to catch transport/parser exceptions.

src/snakemake/api.py
clelange: Addressed via the same helper change. `resolve_snakefile()` now calls `is_local_file()` directly and preserves remote strings unchanged.

src/snakemake/gui.py
clelange: Addressed via the same helper change. The GUI path handling now calls `is_local_file()` directly and keeps remote snakefile values unchanged.

src/snakemake/sourcecache.py
clelange: I rewrote this to make the intent clearer. The second lookup is still there, now in the form of `.get()` inside the lock, to handle the case where another thread populates the cache after the initial fast-path miss but before this thread acquires `_lock`. That preserves one `HostedGitRepo` instance per cache key. If you prefer removing the inner check entirely, let me know.

src/snakemake/sourcecache.py
clelange: You're right. I removed the wrapper and now `call _infer_hosting_provider_file_shorthand()` directly.

johanneskoester (approved):

johanneskoester:
Thanks a lot @clelange!

clelange:
Thanks to you, @johanneskoester 🙂

PR #4173, opened 2026-04-29
fix(docs): remove misleading "touch"  example and explain checksum behaviour

### LINKED ISSUE #4172: Docs: touch does not always cause workflow to rerun because of additional checksum checks
**Snakemake version**

v9.19.0

**Describe the bug**

The documentation still describes Snakemake rerun decisions as if a newer input `mtime` is sufficient to trigger a rerun.

Affected documentation:

- [`docs/tutorial/basics.rst` lines 142-144](https://github.com/snakemake/snakemake/blob/main/docs/tutorial/basics.rst?plain=1#L142-L144) says Snakemake reruns jobs when an input file is newer than an output file.
- [`docs/tutorial/basics.rst` lines 206-215](https://github.com/snakemake/snakemake/blob/main/docs/tutorial/basics.rst?plain=1#L206-L215) specifically suggests `touch data/samples/A.fastq` and says this will make Snakemake rerun the job for `mapped_reads/A.bam`.
- [`docs/snakefiles/rules.rst` lines 2109-2113](https://github.com/snakemake/snakemake/blob/main/docs/snakefiles/rules.rst?plain=1#L2109-L2113) describes rerun decisions as timestamp based.
- [`docs/project_info/faq.rst` lines 683-690](https://github.com/snakemake/snakemake/blob/main/docs/project_info/faq.rst?plain=1#L683-L690) describes reruns as based on file modification dates plus metadata, but does not mention input checksum suppression.

This is no longer generally true for checksum-eligible input files. Since PR #1568, Snakemake records input checksums and suppresses reruns when an input is newer than the output but has the same checksum as recorded in metadata.

Relevant implementation:

- [`src/snakemake/dag.py` lines 1492-1501](https://github.com/snakemake/snakemake/blob/main/src/snakemake/dag.py#L1492-L1501) requires an input to be newer than the output and not have the same recorded checksum.
- [`src/snakemake/dag.py` lines 1405-1424](https://github.com/snakemake/snakemake/blob/main/src/snakemake/dag.py#L1405-L1424) checks whether the current input checksum matches the recorded checksum.
- [`src/snakemake/settings/types.py` line 217](https://github.com/snakemake/snakemake/blob/main/src/snakemake/settings/types.py#L217) sets `max_checksum_file_size` to `1000000`.
- [`src/snakemake/cli.py` lines 1378-1385](https://github.com/snakemake/snakemake/blob/main/src/snakemake/cli.py#L1378-L1385) documents the `--max-checksum-file-size` option as defaulting to 1 MB.

The tutorial example uses a pure `touch`, which changes only the file timestamp. For small files under the default checksum threshold of 1 MB, Snakemake can detect that the content did not change and skip the rerun.

I actually tripped over this issue last week when giving a tutorial on Snakemake. I first thought it might be related to the file system being AFS, but then I found the `--max-checksum-file-size` option...

**Suggested fix**

Update the docs to clarify that:

- Snakemake uses `mtime`s for freshness checks, but for checksum-eligible files it may avoid rerunning if the input content checksum is unchanged.
- A pure `touch` is not a reliable way to demonstrate rerun behavior for small files.
- The tutorial should use an actual content-changing edit instead of `touch`, or explicitly explain why `touch` may not trigger a rerun --- my preference would be to get rid of `touch` here since it's not a very useful example anymore.
- Users who intentionally want timestamp-only behavior can adjust `--max-checksum-file-size`, for example setting it to `0`, or use explicit forcing options such as `--force`, `--forceall`, or `--forcerun`.

**Relevant history**

This behavior was introduced by PR #1568

It was first included in Snakemake `v7.7.0`.

**Logs**

N/A

**Minimal example**

See linked docs.

**Additional context**

Linked/added above

I'd be happy to fix this.

clelange:
Given that running `touch` on an input file does not necessarily mean that a workflow rule will be rerun because of the checksum feature (controlled by ``--max-checksum-file-size``) that's been introduced for Snakemake v7, I've tried to explain this in the corresponding places in the docs and removed the `touch` example in the basic tutorial.

Fixes #4172 

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Documentation**
  * Clarified FAQ entry explaining how Snakemake uses content checksums for small input files (default up to 1 MB) instead of relying solely on timestamps, and documented the `--max-checksum-file-size` option for controlling this behavior.
  * Updated freshness checking documentation to reflect checksum comparison logic for small files.
  * Revised tutorial re-execution explanations to emphasize consistency checking between outputs, inputs, and rule definitions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 ### Documentation
 
-* update installation instuctions ([#4025](https://github.com/snakemake/snakemake/issues/4025)) ([bfb6b59](https://github.com/snakemake/snakemake/commit/bfb6b59f82aa5c86d6c9df3c4006cbe73ac2604a))
+* update installation instructions ([#4025](https://github.com/snakemake/snakemake/issues/4025)) ([bfb6b59](https://github.com/snakemake/snakemake/commit/bfb6b59f82aa5c86d6c9df3c4006cbe73ac2604a))
 
 ## [9.18.2](https://github.com/snakemake/snakemake/compare/v9.18.1...v9.18.2) (2026-03-26)
 
--- a/docs/project_info/faq.rst
+++ b/docs/project_info/faq.rst
@@ -686,6 +686,12 @@ Snakemake wants to rerun a rule that has been already executed, what can I do?
 Snakemake tries to ensure consistency between input and output files.
 This is based on file modification dates (input files may not be newer than output files of the same job), as well as execution metadata like the used software stack (e.g. conda env or container image), the non-file parameters, the set of input files, and the code of the rule.
 If Snakemake wants to rerun a rule that has been already executed, it is because one of these criteria has changed and detailed information about the reasoning is given in the job description of Snakemake's output as well as in the final summary at the end of a dry-run.
+However, there is one exception:
+For files that are smaller than a certain size,
+by default 1 MB, as controlled by ``--max-checksum-file-size``,
+Snakemake calculates a checksum of the file content and only reruns the rule if that checksum has changed,
+even if the timestamp of the input file is newer than the output file(s).
+To rely on file modification dates exclusively, set ``--max-checksum-file-size=0``.
 
 If your job is triggered by newer input files, but you are sure that the input files did not change on a semantic level (i.e. won't yield different results), you can mark those input files as ancient via the command line, or (usually better) via a :ref:`workflow specific profile <profiles>`.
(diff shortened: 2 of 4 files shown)


clelange:
Since this only touches the documentation, the CI failures are unrelated

johanneskoester (approved):
Indeed, thanks a lot!

PR #4180, opened 2026-05-03
fix: treat np.nan as null in branch function (closes #4005)

### LINKED ISSUE #4005: np.nan not evaluated as None in branch function
**Snakemake version**
`9.16.3`

**Describe the bug**
When branch function receives `np.nan`, the internal conversion to `None` (https://github.com/snakemake/snakemake/blob/b88171cea0752b65cb6687906f7778c7f229c93d/src/snakemake/ioutils/branch.py#L29-L30) evaluates `np.nan` as a `float` which is `True`, which in turn the `branch` will always return the `then` statement where it should return the `otherwise` value.

**Minimal example**
```
import numpy as np
print(branch(np.nan, then="foo", otherwise="bar"))
```
Returns `foo`

**Additional context**
This typically occurs when `lookup` is used in conjunction with `branch`. Also, this is particularly true when the `lookup` is done in a pandas dataframe and there are missing values, pandas will use `np.nan` instead of `None` to represent missing values.

wali-reheman:
## Summary

`np.nan` is a float and therefore truthy in Python, so before this fix:

incorrectly returned `'foo'` (taking the `then` branch) instead of `'bar'` (the `otherwise` branch).

This is particularly problematic when `branch()` is used after `lookup()` on pandas DataFrames with missing values — pandas uses `np.nan` for NA values, which would cause branch to always take the `then` path.

## Fix

Added `_is_null()` helper that returns `True` for:

- `None`
- `float('nan')` / `np.nan` (via `math.isnan`, guarded by `try/except` for non-numeric types)

The `convert_none` → `handle_callable` refactor now checks _is_null() on the resolved value before returning the falsy `[]` fallback.

## Testing

- Regression test added to `tests/test_io.py::test_branch_np_nan`
- Covers: `np.nan`, `None`, `True`/`False`, `float('nan')`, `[]`, `['a']`
- All existing branch behavior preserved

## Issue

Fixes https://github.com/snakemake/snakemake/issues/4005

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed branch function to correctly preserve valid falsy values (such as 0, empty strings, and False) instead of incorrectly treating them as null. Now only None and NaN are recognized as null values.

### DIFF
--- a/src/snakemake/ioutils/branch.py
+++ b/src/snakemake/ioutils/branch.py
@@ -1,3 +1,4 @@
+import math
 from collections.abc import Mapping, Callable
 from typing import Optional, Union
 
@@ -26,14 +27,23 @@ def branch(
     input function that will be evaluated once the wildcards are known.
     """
 
-    def convert_none(value):
-        return value or []
+    def _is_null(value):
+        """Check if a value is null: None, float NaN, or math.nan."""
+        if value is None:
+            return True
+        try:
+            return math.isnan(value)
+        except (TypeError, ValueError):
+            return False
 
     def handle_callable(value, wildcards):
         if isinstance(value, Callable):
-            return convert_none(value(wildcards))
+            result = value(wildcards)
         else:
-            return convert_none(value)
+            result = value
+        if _is_null(result):
+            return []
+        return result
 
     def do_branch_then_otherwise(wildcards):
         if handle_callable(condition, wildcards):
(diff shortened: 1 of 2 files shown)


GaspardR:
Hello! This seems to be a duplicate, #4031 proposes a fix and is accepted and waiting for #4009 to be merged.

wali-reheman:
Thanks for the heads-up @GaspardR! Confirmed — closing this as duplicate of #4031. Good to see the fix is already accepted and waiting on #4009. Thanks for the review!

wali-reheman:
Closing as duplicate per maintainer feedback — fix is in #4031.

PR #4184, opened 2026-05-05
fix: scoped module container directive

### LINKED ISSUE #3186: Global container image specification overridden by module import
**Snakemake version**

8.23.1

**Describe the bug**

If the Snakefile has a global container image specification `container: ...`, imports an external workflow through `module other_workflow:`, and the module also has a global container image specification, then the container image will be overridden by that of the external workflow. 

I would expect that the global container specification is isolated within each module. 

**Logs**

**Minimal example**

**Additional context**

twillis209:
When using top-level `container:` directives in both a workflow and a second workflow imported by the first using `module`, the container specified for the second workflow will overwrite the container specified for the first, even for rules not imported from the second workflow. This was reported as [issue 3186](https://github.com/snakemake/snakemake/issues/3186).

A toy example of this issue (available for convenience in [this repo](https://github.com/twillis209/container-bug-repro)) where the following workflow imports a second workflow as `mymodule`:
```
container: "docker://ubuntu:22.04"


module mymodule:
    snakefile:
        "module.smk"


use rule * from mymodule


rule all:
    input:
        "result.txt",
    output:
        "final.txt",
    shell:
        "cat {input} > {output}"
```
The second workflow imported as `mymodule`:
```
container: "docker://alpine:3.18"


rule a:
    output:
        "result.txt",
    shell:
        "echo hello > {output}"
```

To try to fix this, I moved the `glboal_container_img` and `global_is_containerized` attributes from `Workflow` to `WorkflowModifier`. Seems to work to fix the bug. This is my first time attempting a PR so please let me know if something is missing.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **New Features**
  * Improved container settings handling for module-based workflows, ensuring global container defaults are applied in the right scope.
  * Module and parent workflows now preserve their own container settings independently.

* **Bug Fixes**
  * Fixed container inheritance so rules use the expected global container image only when appropriate.
  * Added coverage to verify module and master workflows each receive the correct container image.


twillis209:
Seems there was some issue with `pixi` last time the workflows were triggered for this PR, but I ran it on my own fork today and it seems good to go: https://github.com/twillis209/snakemake/pull/1

johanneskoester (changes requested):
Thanks a lot, this looks very good! A guess: I think the test directory test_module_global_container is missing from your commit. This can happen because the whole test subfolders are in gitignore. The latter is an oversimplictic measure to avoid that non-needed test files are leaked into the repo and only stuff that it explicitly enforced with `git add -f` ends up in the repo. We should probably change this approach into something more convenient sooner or later.

twillis209:
Oof, I'm sorry I missed that failing test. There were some which I believed were platform-related and that caused me to miss that one. I think it works now. Thanks for checking.

johanneskoester (approved):

PR #4185, opened 2026-05-06
feat: allow users to define a shell wrapper around all shell directives

LukaP-BB:
In this PR, I propose a way to wrap all shell directives in a way similar to `shell.prefix` and `shell.suffix`, but that is aware of the rule context (wildcards, logs, etc..).
The intended use-case is to factor out all boilerplate shell commands from the rules.
Here is a barebone example of what the user would be able to write : 
```py
def shell_wrapper(shell_script, ctx) : 
  """
  wraps the shell command, 
  redirects all un-captured stderr/stdout to the log file, 
  prints the commands with set -x, 
  prints the inputs and outputs as seen by snakemake in the log file
  """
  return f"""
    (
    set -euo pipefail
    echo [inputs] {ctx['input']}
    echo [outputs] {ctx['output']}
    set -x
    {shell_script}    
    ) > {ctx['log']} 2>&1 
    """

shell.add_wrapper(shell_wrapper)
```
A more real-life implementation that handles some edge-cases is available here : https://github.com/LukaP-BB/snakemake_test_cases/blob/master/logging/Snakefile

### QC

### AI-assistance disclosure
Only organic neurons where used in the process of making this PR.

## Summary by CodeRabbit

* **New Features**
  * Added customizable shell wrapper mechanism to extend shell command execution with custom logic

* **Documentation**
  * Introduced "Shell wrapper" documentation section with implementation examples and best practices


LukaP-BB:
Regarding the test, it actually tests that the wrapper worked since the log file contains the command line that was run (set -x).
Regarding the other points, they are valid, I'll wait on human input to see how to proceed :)

docs/snakefiles/rules.rst
```diff
@@ -974,6 +974,46 @@ Further, it is possible to set global prefixes and suffixes for all shell comman
...
+
+Using this, it is possible to define some usefull generalisations, such as : 
+
+* putting your multi-line shell statements into a sub-shell capturing all stderr/stdout that is not explicitely redirected
+* auto-generating a log path from rule name and wildcards : 
+
+.. code-block:: python
+
+    wildcards_strs = "_".join([f"{k}={v}" for k, v in context['wildcards'].items()])
+    log_path = f"log/{context["rule"]}/{wildcards_strs}.log"
```
cademirch, line 1007: This is cool idea, but is slightly confusing to me as it does not update or set the actual `log` attribute/directive on the given rule.

PR #4189, opened 2026-05-07
fix: storage temp() file cleanup with RemoteProvider 

swrosati:
## fix: `temp()` files not removed from remote storage (S3)

When using `temp()` with a remote storage backend (e.g. S3) and a remote executor (e.g. AWS Batch), temporary files are never cleaned up from storage after they are no longer needed. Two bugs combine to cause this:

**Bug 1 — `is_needed_tempfile` always returns `True` in the main process (`6d7c8823`)**

`is_needed_tempfile` gated cleanup on `storage_settings.unneeded_temp_files`, which is only populated in spawned subprocesses via the `--unneeded-temp-files` CLI arg. In the main orchestrator process (where `handle_temp` runs), this set is always empty, so `is_unneeded_outside` was always `False` and temp files were never passed to `remove()`. The main process has full DAG knowledge and already correctly determines liveness via `is_needed_by_subsequent_job`, so the gate is unnecessary.

**Bug 2 — set intersection loses the `temp` flag (`4bf0afeb`)**

`handle_temp` computes files to remove via `tempfiles & files` — a set intersection between output `_IOFile` objects (carrying the `temp` flag) and input `_IOFile` objects from `_dependencies` (without the flag). Since `_IOFile` inherits from `str`, both sides hash/compare equal on the path string alone. Python's `set.__and__` does not guarantee which side's objects are returned; when it returns the *input* objects, `is_flagged(file, "temp")` in `remove()` evaluates to `False` and `managed_remove()` is skipped.

Replaced with `{f for f in tempfiles if f in files}` to explicitly preserve the output `_IOFile` instances.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined temp-file cleanup logic: subprocess runs now always preserve temp files, while remote runs respect storage settings to determine which temp outputs can be removed; improves accuracy of cleanup and avoids incorrect deletions.

* **Tests**
  * Added tests validating temp-file needed/unneeded decisions across local, remote, and subprocess execution modes and a regression test for temp-file selection during cleanup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/snakemake/snakemake/pull/4189?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


swrosati:
Hi @johanneskoester - I'm just checking to see if this PR needs any additions or edits?  If so, please don't hesitate to let me know!

src/snakemake/dag.py
```diff
@@ -952,16 +952,18 @@ def is_needed_tempfile(self, job, tempfile, outside_of_group_job=None):
...
-            )
-        else:
-            is_unneeded_outside = True
+        # The main process always has full DAG knowledge, so it can rely on
+        # is_needed_by_subsequent_job below to determine if the file is still
+        # needed.  The unneeded_temp_files set is only meaningful in spawned
+        # subprocesses (populated via --unneeded-temp-files CLI arg) and is
+        # always empty in the main process, so gating on it here would prevent
+        # temp files from ever being cleaned up from remote storage.
+        is_unneeded_outside = True
```
johanneskoester: Not sure I understand this. The main process passes to the the remote job temp files that are not needed outside of the remote job, via unneeded_temp_files. If we always assume this to be true in the remote job, this can lead to cleanup of temp files in the remote job that are later needed by another remote job. Only the main process can know this.

johanneskoester: @copilot, what do you think about my concern?

swrosati:
Thank you, @johanneskoester!  Much appreciated!

PR #4199, opened 2026-05-21
fix: wildcard constraints and pathvar expansion

### LINKED ISSUE #4197: Wildcard constraints applied before pathvar expansion
**Snakemake version**

9.21.0

**Describe the bug**

In the example below, rules `numeric_sample` and `alpha_sample` conflict in their wildcard structure, so a wildcard constraint is added to each rule to disambiguate them. This works fine when the output path is `<result>/{sample}.txt`, but it _fails_ when output path is `<sample>.txt` with the `<sample>` pathvar defined accordingly.

In other words, it seems to me that wildcard constraints are somehow applied _before_ expanding the wildcards, leading to them being ignored when they're not visible before expansion? I'm assuming this is not intended behaviour, judging from this sentence in the [docs](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#pathvar-usage):

> Pathvars are resolved when the rule is parsed (before wildcard resolution and DAG construction). The values of pathvars can thereby even contain wildcards themselves.

**Minimal example**

```snakemake
pathvars:
    sample="<results>/{sample}",


rule all:
    input:
        "results/foo.txt",
        "results/42.txt",


rule numeric_sample:
    output:
        "<sample>.txt",  # Fails
        # "<results>/{sample}.txt",  # Works
    wildcard_constraints:
        sample="[0-9]+",
    shell:
        "touch {output}"


rule alpha_sample:
    output:
        "<sample>.txt",  # Fails
        # "<results>/{sample}.txt",  # Works
    wildcard_constraints:
        sample="[a-z]+",
    shell:
        "touch {output}"
```

fgvieira:
Fixes #4197

Currently, pathvars are expanded under `path_modifier`, and these should be done before applying `wildcard_constraints`. In `log`, this is already the case:

https://github.com/snakemake/snakemake/blob/35ab0421813248cea482de3fd796ad48e53d88d7/src/snakemake/rules.py#L619-L626

as well as for `benchmark`:

https://github.com/snakemake/snakemake/blob/35ab0421813248cea482de3fd796ad48e53d88d7/src/snakemake/rules.py#L243-L251

but for `output`, wildcard constraints seem to be applied before expanding pathvars, leading to wildcards inside pathvars not to be evaluated.

##### Test
The added test has two rules with similar output with a pathvar. The pathvar as a wildcard that should be constrained and allow to differentiate both outputs.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of wildcard constraints when multiple outputs are defined, ensuring numeric and alphabetic sample patterns are applied correctly.

* **Tests**
  * Added integration coverage for path-variable behavior with wildcard constraints, including expected outputs for numeric and alphabetic samples.

* **Chores**
  * Updated path-variable configuration so generated text and CSV artifacts use relative output templates across related rules and aliases.


johanneskoester (commented):
Thanks! Can you briefly describe the changes and how they solve the issue? Including the changes to the test? Thanks!

PR #4232, opened 2026-06-17
fix: temp files not deleted when using remote storage plugins

schrins:
When using the fs storage plugin, snakemake did not remove temporary files in the "main" storage after they were not needed anymore. The place, where this is supposed to happen, is the `remove` method in the `_IOFile` class. However, the cleanup via the storage object is only invoked for files, which are not supposed to be retrieved from the storage. This causes said temp files to be removed twice on the local storage (once because they are temp, once because the storage plugin cleans up local fiels behind itself) but not on the main storage. The expectation would be that temp files never permanently remain in the main storage, regardless of how transfers are handled by any storage plugin.

This PR removes the condition that currently prevents the cleanup. In addition, the `only_local` flag is now only set to `True` for callers of the `remove` method, when it actually wants to remove the global file and not just the local one. This might have been another bug that was masked by the excess condition that is now gone. Finally, a new test case ensures that temp files are properly removed now using the fs storage plugin.  

### QC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup behavior for temporary, storage-backed inputs so deletions correctly target the intended local copy when the files are no longer needed.
* **Tests**
  * Added an integration test workflow to verify `temp()` outputs are deleted as expected with the filesystem storage provider.
  * Updated the expected results fixture and extended the test harness to check cleanup outcomes.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -3493,4 +3493,7 @@ def test_delete_temp_fs():
         cleanup=False,
     )
     temp_file = os.path.join(outdir, "a")
-    assert not os.path.exists(temp_file), "temp file was not removed in main working directory"
+    assert not os.path.exists(
+        temp_file
+    ), "temp file was not removed in main working directory"
+    shutil.rmtree(tmpdir)
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -3484,3 +3484,13 @@ def test_github_issue_4039_runtime_no_override():
         cleanup=False,
     )
     shutil.rmtree(tmpdir)
+
+
+def test_delete_temp_fs():
+    outdir = run(
+        dpath("test_delete_temp_fs"),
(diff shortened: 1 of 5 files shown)


johanneskoester (approved):
Thanks a lot!

tests/tests.py
```diff
@@ -3484,3 +3484,16 @@ def test_github_issue_4039_runtime_no_override():
...
+    outdir = run(
+        dpath("test_delete_temp_fs"),
+        shellcmd="snakemake -c1 --default-storage-provider fs",
+        cleanup=False,
+    )
+    temp_file = os.path.join(outdir, "a")
+    assert not os.path.exists(
+        temp_file
+    ), "temp file was not removed in main working directory"
+    shutil.rmtree(tmpdir)
```
johanneskoester:
```suggestion
    shutil.rmtree(outdir)
```

PR #4259, opened 2026-07-25
fix: leak of benchmark monitor threads

bkeepers:
`ScheduledPeriodicTimer._action` re-arms a fresh `DaemonTimer` on every fire without checking the stop flag, so a cancel() that lands between a timer fire and _action is immediately undone — the monitor thread then lives for the rest of the process. This fires on normal job completion, so at scale the leaked threads accumulate until the scheduler starves: on a ~23k-job run throughput decayed monotonically (from ~11 min per 1% early to ~131 min per 1% near the end) and the run stalled, with repeated "Benchmark: unable to collect ... process no longer exists" messages from timers still polling exited PIDs.

`_action` now honors the stop flag; `cancel()` sets the flag before touching the timer (so a racing `_action` observes it) and is None-safe; and `benchmarked()` cancels in a `finally` so a raising job body cannot leak the monitor.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Reworked periodic benchmarking timer handling to prevent cancellation/rescheduling races and avoid unintended re-arming.
  * Improved shutdown behavior so monitoring stops when the process exits and no monitor is left running after failures.
  * Made cancellation more robust by safely handling cases where the underlying timer is not yet available.
* **Tests**
  * Added regression coverage for timer cancellation/re-arming race conditions and for ensuring monitor cleanup even when the benchmark body raises.

### DIFF
--- a/src/snakemake/benchmark.py
+++ b/src/snakemake/benchmark.py
@@ -273,6 +273,11 @@ class ScheduledPeriodicTimer:
 
     def _action(self):
         """Internally, called by timer"""
+        # A cancel() can land between the timer firing and this call. Without this guard
+        # the reschedule below would resurrect a cancelled timer, leaking the monitor
+        # thread for the rest of the process' life.
+        if self._stopped:
+            return
         self.work()
         self._times_called += 1
         if self._times_called > self._interval:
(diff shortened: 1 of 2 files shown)


johanneskoester (approved):
Thanks a lot!!

PR #4273, opened 2026-08-11
fix: cache path existence

cmeesters:
The other day, I was launching a new workflow and defined `SNAKEMAKE_OUTPUT_CACHE`. The workflow crashed and crashed. In the office it was > 30 C, so I failed to notice the error message. What if we ensure the cache directory is created, in case it does not exist?

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Output caching now automatically creates missing cache directories, including nested parent directories.
  * Directory-creation failures are reported using the existing cache write error behavior.

* **Tests**
  * Added coverage confirming cache directories are created correctly when configured through `SNAKEMAKE_OUTPUT_CACHE`.

### DIFF
--- a/src/snakemake/caching/__init__.py
+++ b/src/snakemake/caching/__init__.py
@@ -33,6 +33,12 @@ class AbstractOutputFileCache:
             self.cache_location = os.environ[LOCATION_ENVVAR]
         except KeyError:
             raise MissingOutputFileCachePathException()
+
+        try:
+            Path(self.cache_location).mkdir(parents=True, exist_ok=True)
+        except OSError as e:
+            self.raise_write_error(self.cache_location, e)
+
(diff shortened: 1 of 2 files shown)


johanneskoester:
I see your point, but since SNAKEMAKE_OUTPUT_CACHE is a setting that will usually be shared between several users I think the logic of failing if it is not present is safer. Because it will let the user know if the path is misspelled. If it is just silently created, we risk that several wrong directories are created by different users who just have typos in they respective variables.

cmeesters:
@johanneskoester yes, I failed to see this implication. Perhaps I'll implement a "The folder 'huffelpuff' does not exist. Did you mean 'hufflepuff'?" some time in the future. ;-)

PR #4274, opened 2026-08-12
test: change dependencies

Hocnonsense:
try to use system-level apptainer to bypass hang on of 'test_singularity_conda'.

### QC

* [x] The changes are already covered by an existing test case.
* [x] The changes or this is not necessary (e.g. if the change does neither modify the language nor the behavior or functionalities of Snakemake).

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Tests**
  * Added coverage for running Conda-based workflows through the udocker deployment method.
  * Added a workflow test using a pinned Bioconda/Conda-Forge environment with BWA.
  * Disabled the equivalent Singularity test pending resolution of an external Conda-Forge issue.

### DIFF
--- a/tests/test_udocker_conda/Snakefile
+++ b/tests/test_udocker_conda/Snakefile
@@ -1,0 +1,12 @@
+
+
+singularity: "docker://condaforge/miniforge3:26.3.2-3"
+
+
+rule a:
+    output:
+        "test.out",
+    conda:
+        "test-env.yaml"
+    shell:
+        "bwa 2> {output} || true"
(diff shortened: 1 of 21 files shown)


johanneskoester (commented):
I would prefer to simply use the udocker runtime of the container plugin. Any reason to not do that?

Hocnonsense:
I'm all for it! I also think the new installation method of apptainer is too heavy. Also, I'm really looking forward to seeing an example of using udocker in a Snakefile!

tests/test_udocker_conda/Snakefile
```diff
@@ -0,0 +1,12 @@
+
+
+singularity: "docker://condaforge/miniforge3:26.3.2-3"
```
johanneskoester:
```suggestion
container: "docker://condaforge/miniforge3:26.3.2-3"
```

PR #4285, opened 2026-08-20
fix: Snakemake sometimes checks the wrong copy of an output file's info, causing spurious reruns

Maarten-vd-Sande:
When deciding whether a job's dependency is already done, snakemake currently sometimes checks a copy of that output's file reference instead of the real one the producing rule declared, and that copy can be missing important information (like "this lives in S3, not on local disk"), causing Snakemake to look in the wrong place and wrongly conclude the file doesn't exist.

The fix looks up the producing rule's real output object by matching its path, and uses that instead.

First pushed the tests that break currently (see red check), then pushed the fix.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed workflows with storage-backed outputs that could be incorrectly treated as missing after dependency graph updates.
  * Ensured downstream steps reuse existing producer outputs instead of rerunning completed work.
  * Improved reliability when local storage staging files are removed and outputs must be recreated.

* **Tests**
  * Added integration coverage for S3-backed workflows and dependency edge handling.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -1493,8 +1493,10 @@ class DAG(DAGExecutorInterface, DAGReportInterface, DAGSchedulerInterface):
                             files.update(
                                 f for f in job.products() if f in self.targetfiles
                             )
+                    canonical = {str(f): f for f in job.output}
+                    canonical_files = [canonical.get(str(f), f) for f in files]
                     reason.missing_output.update(
-                        [f async for f in job.missing_output(files)]
+                        [f async for f in job.missing_output(canonical_files)]
                     )
             if not reason:
                 output_mintime_ = output_mintime.get(job)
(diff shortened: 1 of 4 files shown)


src/snakemake/dag.py
```diff
@@ -1636,7 +1638,11 @@ async def updated_input():
 
                 for job_, files in dependencies[job].items():
                     if job_ in candidates_set:
-                        missing_output = [f async for f in job_.missing_output(files)]
+                        canonical = {str(f): f for f in job_.output}
+                        canonical_files = [canonical.get(str(f), f) for f in files]
+                        missing_output = [
+                            f async for f in job_.missing_output(canonical_files)
+                        ]
```
johanneskoester: Can't we move this logic into Job.missing_output? Then, it would only have to be implemented once and not the caller needs to know about this behavior.

johanneskoester (changes requested):

src/snakemake/dag.py
Maarten-vd-Sande: Yes, makes much more sense!

src/snakemake/jobs.py
```diff
@@ -723,7 +723,14 @@ async def get_mtime(f):
         return mintime
 
     async def missing_output(self, requested):
+        """Yield requested outputs that are missing."""
+        # requested may hold path-equal copies of our own output that lost
+        # their flags (e.g. storage).
+        output_by_path = {str(f): f for f in self.output}
```
johanneskoester: I think this should become a lazy_property, such that it is reused and not recreated every time missing_output is called.

johanneskoester (changes requested):

johanneskoester (approved):

johanneskoester:
Thanks a lot, great work!

PR #4286, opened 2026-08-20
fix: preserve flags when formatting _IOFile

Maarten-vd-Sande:
the _IOFile.format() method is inherited from str.format(), which has no concept of flags and silently returns a plain string with the storage_object/temp/etc. flag gone entirely. For example, apply_wildcards() preserves flags across wildcard substitution. This fix mirrors the existing guard already in expand() (which rejects an already-flagged pattern outright) by raising a clear WorkflowError pointing at apply_wildcards() instead, rather than letting the mistake silently corrupt downstream existence/storage semantics.

First pushed the tests that break currently (see red check), then pushed the fix.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of temporary and directory-marked files during formatting.
  * Formatting now preserves applicable file flags.
  * Formatting storage-backed files now reports a clear workflow error and recommends wildcard substitution instead.
  * Standard formatting continues to work as expected when no restricted flags are present.

* **Tests**
  * Added coverage for multiple file flags, path-modifying flags, wildcard substitution, and formatting error handling.

### DIFF
--- a/src/snakemake/io/__init__.py
+++ b/src/snakemake/io/__init__.py
@@ -896,6 +896,25 @@ class _IOFile(str, AnnotatedStringInterface):
             with open(file, "w") as f:
                 pass
 
+    def format(self, *args, **kwargs):
+        """Override str.format() to guard against silently dropping flags."""
+        from snakemake.path_modifier import PATH_MODIFIER_FLAG
+
+        significant_flags = {
+            key: value
+            for key, value in self.flags.items()
+            if key != PATH_MODIFIER_FLAG
+        }
+        if significant_flags:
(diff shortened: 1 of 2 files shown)


johanneskoester (commented):
Good catch! I wonder whether alternatively we could actually reapply the flags after the formatting has happened?

johanneskoester (commented):
Should we do the same for AnnotatedString? Format could become the method of a common mixin?

johanneskoester (approved):
Perfect, thanks!!

PR #4291, opened 2026-08-24
test: Add test for checkpoint output retrieval from S3 storage

fxwiegand:
Adds a new test showing that a checkpoint whose output lives in S3 fails on a second run.

## Summary by CodeRabbit

- **New Features**
  - Added a checkpoint workflow that dynamically discovers generated items and combines them into a final result.

- **Tests**
  - Added integration coverage for checkpoint workflows using filesystem-backed storage.
  - Verified that local copies for running jobs are preserved while concurrently finishing checkpoints are processed.
  - Added coverage for dynamically collecting generated files and producing a combined result.


Maarten-vd-Sande:
@fxwiegand I took a look because I thought maybe my PR would solve this issue (https://github.com/snakemake/snakemake/pull/4285/changes).

Upon further inspection I think the "problem" is in your code, and not snakemake. The correct way to do this is by using snakemake internals:

```
with checkpoints.items.get().output[0].open() as f:
```

It's a bit hidden in the example under: https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#data-dependent-conditional-execution

fxwiegand:
Thanks @Maarten-vd-Sande for having a look and pointing out the issue! Would this be something that `snakemake --lint` could flag maybe? I'd be happy to try and add this.

One more question: when the checkpoint output is a directory how do things work there?

Maarten-vd-Sande:
> Thanks @Maarten-vd-Sande for having a look and pointing out the issue! Would this be something that `snakemake --lint` could flag maybe? I'd be happy to try and add this.
> 
> One more question: when the checkpoint output is a directory how do things work there?

I'm not part of the snakemake dev team, so I'm not sure about the linter. Also not sure about what to do if the output is a directory..... :thinking:

fxwiegand:
> > Thanks @Maarten-vd-Sande for having a look and pointing out the issue! Would this be something that `snakemake --lint` could flag maybe? I'd be happy to try and add this.
> > One more question: when the checkpoint output is a directory how do things work there?
> 
> I'm not part of the snakemake dev team, so I'm not sure about the linter. Also not sure about what to do if the output is a directory..... 🤔

I know who to ask then, again thanks for pointing me in the right direction 🙂

PR #4293, opened 2026-08-26
feat: Print job params in log

### LINKED ISSUE #4290: Include parameter values in dry run output.
I often pass parameters to scripts via the `snakemake` Python/R objects.
Checking my workflow pre-execution, it would be great to be able to verify that the (dynamically generated) parameters passed to each instance of the script match my expectation.

I would like the diagnostic / dry-run output to include `params` alongside `input`, `output`, `log`, `jobid`, `reason`, `wildcards`, and `resources`.

For shell scripts I can see the parameters by adding `--printshellcmds`, but for R/Python scripts or Notebooks I have not found any way to check the parameters without *actually* executing them (at which point I can print them at runtime). This means for long-running pipelines I can see the values passed to downstream jobs only after waiting for their dependencies to (successfully) finish first.

**Additional context**
I did search the issues and documentation for any previous discussion or reason why this might be considered unwanted or hard to implement, but I didn't find anything.

fgvieira:
Closes #4290 

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

- **Enhancements**
  - Job information logs now include configured rule parameters when available.
  - Named and unnamed parameters are displayed in a clearer, more readable format.
  - Parameter values from mappings and structured objects, including tabular data, are supported.

- **Bug Fixes**
  - Improved the completeness and consistency of parameter details in workflow log events.
  - Unsupported parameter values now produce clearer error messages.

### DIFF
--- a/src/snakemake/logging.py
+++ b/src/snakemake/logging.py
@@ -81,7 +81,19 @@ def format_dict(dict_like, omit_keys=None, omit_values=None) -> str:
 
 format_resources = partial(format_dict, omit_keys={"_cores", "_nodes"})
 format_wildcards = format_dict
-format_params = format_dict
+
+
+def format_params(params) -> str:
+    from snakemake.iocontainers import Namedlist
+
+    if isinstance(params, Namedlist):
+        items = params._allitems()
+    elif isinstance(params, Mapping):
+        items = params.items()
+    else:
+        raise ValueError("params neither a dict nor a Namedlist")
+
+    return ", ".join(f"{name}={value}" if name else str(value) for name, value in items)
 
 
 def format_resource_names(resources, omit_resources="_cores _nodes".split()):
--- a/src/snakemake/logging.py
+++ b/src/snakemake/logging.py
@@ -81,6 +81,7 @@ def format_dict(dict_like, omit_keys=None, omit_values=None) -> str:
 
 format_resources = partial(format_dict, omit_keys={"_cores", "_nodes"})
 format_wildcards = format_dict
+format_params = format_dict
 
 
 def format_resource_names(resources, omit_resources="_cores _nodes".split()):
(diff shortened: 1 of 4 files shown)


src/snakemake/logging.py
```diff
@@ -83,6 +83,19 @@ def format_dict(dict_like, omit_keys=None, omit_values=None) -> str:
...
 
+def format_params(params) -> str:
+    from snakemake.iocontainers import Namedlist
+
+    if isinstance(params, Namedlist):
+        items = params._allitems()
+    elif isinstance(params, Mapping):
+        items = params.items()
+    else:
+        raise ValueError("params neither a dict nor a Namedlist")
```
cademirch: I don't think we should raise here. Params can be callables so I'm pretty sure they can be checkpoint dependent.

This might actually be fine since the checkpoint return will be in the namedlist/mapping but we should probably check. 

In general I think raising in formatting should be avoided and we just fallback to whatever the types repr is.

cademirch (changes requested):
Just one comment. Otherwise I think this is good work! Thanks Filipe! 

Am on mobile, excuse my brevity.

src/snakemake/logging.py
fgvieira: I just used the same logic as in `format_dict` where that error is supposed to warn about a bug and, under normal circumstances, should never be triggered.

johanneskoester (commented):
I am worried about params that are non-primitive, lengthy, or even contain newlines or commas.

fgvieira:
Can you give some example of a non-primitive?
I can try adding these cases to the test and see.

johanneskoester:
Imagine a pandas dataframe for example. Or a dataclass with a lot of fields.

fgvieira:
Just added some more example (e.g. dataframe and dataclass) and they work fine but, if they have a lot of fields, it might clutter the screen. But I guess that is the same when we have a longe `string` or `list` as a parameter, no?

PR #4305, opened 2026-09-03
fix: guard is_edit_notebook_job check to prevent crash

tgbrooks:
Fixes a crash when using `--rulegraph` under some conditions. Unfortunately, I have not been able to find a reproducible example of this situation so I cannot easily add a test case. The pipeline I encounter it in is not easily minimized. Another user also encountered this error recently: https://stackoverflow.com/questions/79999901/issue-with-rulegraph-for-v9-20

The error this avoids is:
```
File "[...]/lib/python3.12/site-packages/snakemake/dag.py", line 612, in is_edit_notebook_job
    self.workflow.execution_settings.edit_notebook
AttributeError: 'NoneType' object has no attribute 'edit_notebook''
```

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented errors when notebook editing is checked without execution settings configured.
  * Notebook editing is now treated as unavailable when execution settings are not present.
  * Improved handling of notebook-related workflow checks in configurations without execution settings.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -628,7 +628,8 @@ class DAG(DAGExecutorInterface, DAGReportInterface, DAGSchedulerInterface):
 
     def is_draft_notebook_job(self, job):
         return (
-            self.workflow.execution_settings.edit_notebook
+            self.workflow.execution_settings is not None
+            and self.workflow.execution_settings.edit_notebook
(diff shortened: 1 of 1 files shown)


johanneskoester (approved):
Thanks, good catch!

PR #4306, opened 2026-09-03
fix: improve script file check error messages and performance

tgbrooks:
ScriptProcessor previously read in the contents of every script file and then discarded it. It then tried to check if the file exists. This meant that the error message displayed for a missing script was not the intended one (since it threw an exception while trying to read in the file rather than properly reporting the error after the existence check).

I changed it to now just get the script file path, which is all that it used. This seems more logical and improves the error message as described above.

This also gives a small but noticeable performance boost, since this check is run once per job: `snakemake -c1 --dry-run` goes from 4s to 3.5s for a test case with 1000 jobs. This performance gain may not be typical since I used a cluster with a network file system which could exacerbate the IO cost. It's also arguable that this should be cached since many jobs will use the same script, however, the performance benefit for that change is smaller in my use case.

I added a test, however I don't see a way to capture stdout with the `run()` function, so it doesn't actually check that the error message has improved.

### QC

### AI-assistance disclosure

I used AI assistance for:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved script source-file resolution, including paths using wildcards and parameters.
  * Improved error reporting for workflows that reference unavailable script files.

* **Tests**
  * Added coverage for workflows that reference unavailable script files.

### DIFF
--- a/src/snakemake/dag.py
+++ b/src/snakemake/dag.py
@@ -3516,9 +3516,8 @@ class ScriptProcessor(DAGProcessorBase):
             if isinstance(script_entry, Path):
                 script_entry = str(script_entry)
 
-            script_sourcefile, _, _, _, _ = script.get_source(
+            script_sourcefile = script.get_source_file(
                 script_entry,
-                self.dag.workflow.sourcecache,
                 job.rule.basedir,
                 job.wildcards,
                 job.params,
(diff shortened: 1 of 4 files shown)


tgbrooks:
"⚠️ This pull request shows signs of AI-generated slop (redundant_comments). It has been flagged by CodeRabbit slop detection and should be reviewed carefully."

I literally didn't add any comments! Is it referring to the docstrings? Those were added to satisfy the *other* checks it performs. Can't win 'em all, I guess.

